PR #780, opened 2014-04-11
New int type (saturated int).

Vsevolod-Livinskij:
Implementation of binary, unary operations and of pre/post increment/decrement for saturated int type.


dbabokin:
This was an interesting experiment, which I think don't need to be merged. I'm closing this PRs.

PR #888, opened 2014-10-23
Knc header tests

Vsevolod-Livinskij:
Tests for knc.h functions.


dbabokin:
We are going to deprecate KNC support, unless someone speaks up. I'm closing this PR without merging.

PR #900, opened 2014-11-12
Change python to python2

Noctune:
This PR changes the makefile and shebangs follow [PEP 394](http://legacy.python.org/dev/peps/pep-0394/) making it possible to compile ispc on systems that do not default to `python2`.

A possible downside is that older systems might not implement PEP 394 and thus might not have `python2`. I'm unsure how this would work with Windows as well.


dbabokin:
I'm rejecting this pull request, as it not working on all our target platforms.

PR #935, opened 2015-01-19
Added LLVM 3.7 support

aguskov:
3.6 still not in /tags, using 3.6 from /branches.


dbabokin:
Please update copyright date in affected files (i.e. change 2014 to 2015).

Otherwise looks good.

dbabokin:
Merging to let nightly pass, but copyright needs to be updated anyway.

PR #937, opened 2015-01-20
Adding 3.7 fails and 6 generic fails on 3.6 caused by long integers.

dbabokin:


egaburov:
Hey,

I've read a nice short story a couple of days ago, you're gonna love it, that's for sure! Here it is <http://pindoltaslo.uptown2downtownblog.com/xydbtjq>

Thanks for your consideration, egaburov@dds.nl

PR #963, opened 2015-02-12
Added struct alignment in headers

aguskov:


dbabokin:
Alignment needs to be extracted from target machine

PR #973, opened 2015-02-20
Smear opt

Vsevolod-Livinskij:
Smear pass in cbackend.cpp now can handle more types of broadcasts


jbrodman:
Nice change!  One suggestion - if the shuffle mask value is a constant, don't emit the extract.  Instead just emit the smear with the constant value directly.

PR #975, opened 2015-02-23
Change dll export feature to a switch.

jbrodman:
Change the dll export feature to be controlled by a switch rather than the export keyword.


jbrodman:
Done.

PR #978, opened 2015-02-27
generic (and knc) target compilation with -std=c++11

ncos:
- Replaced our extern function definitions with ones in 'stdio.h'


dbabokin:
This needs careful evaluation before considering enabling it by default.

I'm rejecting this pull request for now.

PR #995, opened 2015-03-19
Not support LLVM 3.3 anymore, so no need to store 3.3 test results in 'fail_db.txt'

ncos:


dbabokin:
We still support LLVM 3.3, though not testing it nightly.

PR #1003, opened 2015-04-03
Knl support

ncos:
This adds a support for knl target in ISPC (through SDE)


oscarbg:
new AVX-512 instructions shared with Skylake Xeon are supported with this KNL megapatch?

dbabokin:
This pull request adds support for KNL code generation via generic target. This is the same schema as was used for KNC. It's not the path via direct code generation through LLVM.

This patch is not fully tuned for performance yet, though it exploits many KNL instructions and should suggest reasonable performance out of the box. Though real hardware is not yet available, so it's not the time to talk about performance.

jbrodman:
This is focused on KNL atm.  It uses AVX512F which overlaps with Skylake,
but doesn't use any Skylake specific instructions.  The AVX512 wiki page
has a good outline of which subsets of AVX512 are supported on KNL/Skylake.

PR #1013, opened 2015-04-16
Added fix for ispc on ARM

egaburov:
ISPC and examples are now can be build on ARM. When using IPSC on arm it reports:

    Warning: Module DataLayout is incompatible with library DataLayout: 
            Module DL: e-p:32:32:32-S64-i1:8:8-i8:8:8-i16:16:16-i32:32:32-i64:64:64-f16:16:16-f32:32:32-f64:64:64-f128:128:128-v64:64:64-v128:64:128-a0:0:64-n32 
            Library DL: e-p:32:32:32-S32-i1:8:8-i8:8:8-i16:16:16-i32:32:32-i64:32:64-f16:16:16-f32:32:32-f64:32:64-f128:128:128-v64:32:64-v128:32:128-a0:0:64-n32

### DIFF
--- a/builtins/target-neon-16.ll
+++ b/builtins/target-neon-16.ll
@@ -524,3 +524,4 @@ rcpd_decl()
 
 transcendetals_decl()
 trigonometry_decl()
+saturation_arithmetic()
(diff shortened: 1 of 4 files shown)


dbabokin:
Patch is good, Thanks for it!

But would you be so kind to change Copyright date in the files you have changed?

PR #1015, opened 2015-04-16
Change multitarget dispatcher module`s ISA to 'generic'

aguskov:


dbabokin:
Current implementation seems to be correct, at least I don't see problems with it.

PR #1017, opened 2015-04-17
Support of knl-generic-16 target

Vsevolod-Livinskij:
Now we have knl-generic-16 target. It is always processed as generic-x16 during code generation, but in multiple target compilation it is forced to emit c++ file.


dbabokin:
First comment - could you please change copyright to 2015 in the files you have changed?

dbabokin:
The patch is good enough to enable knl target in multi-target compilation. Though it's too hardcoded for knl. I think it can be reasonably generalized to handle arbitrary platform *-generic-*. So let's try generalizing it.

PR #1018, opened 2015-04-20
New accuracy for knl.h

Vsevolod-Livinskij:
knl has new accuracy for reciprocal instructions.


dbabokin:
You've forgot to fix comments

PR #1019, opened 2015-04-21
Multitaget function return types fixed

aguskov:


dbabokin:
I suggest making sure that the type for uniform structures is the same for any vector size, if it doesn't have very elements. I.e. it's mandling would be uniform_*, instead of v4_uniform_*/v8_uniform_*/etc.

dbabokin:
Now this looks like cleaner solution.

PR #1075, opened 2015-07-25
Added support for AMD LibM 

egaburov:
In addition to SVML, `ispc` now supports ACML (AMD Core Math Library/LibM [1]). To use it, 1) compile an `ispc` program with `--math-lib=acml`  and 2) link with `-Lpath_to_amdlibm -lamdlibm`. See acml_lib/README.rst for additional details.

Also modified `./run_tests.py` to pass compiler flags to host compiler. This was needed to run the tests with `--math-lib=acml`, which  looks like this:

    ./run_tests.py -t avx2 -j8 -F "-L./acml_lib -lamdlibm" -f "--math-lib=acml"

Might be helpful for folks with AMD CPUs and those who don't have Intel C++ compiler, but want to accelerate transcendentals.

Comments and suggestions are welcome!

 [1] http://developer.amd.com/tools-and-sdks/cpu-development/libm/


egaburov:
PS: This functionality would probably only make sense in double precision calculations. So perhaps, it might be best to use AMD LibM only with double precision calculations, and use stdlib.ispc functions for single precision calculations?

egaburov:
Just noticed that this pull requests brakes  compilation w/o -amdlibm. Closing pull request until fixed.

PR #1077, opened 2015-07-28
some fixes for new LLVM API

Shishpan:


dbabokin:
Is it still actual fix? I don't see createDataLayout() method in LLVM trunk.

Shishpan:
Yes. In the new API getDataLayout() is replaced with createDataLayout(). I see it in include/llvm/Target/TargetMachine.h

PR #1082, opened 2015-08-02
Fix sincos issue

ncos:


dbabokin:
We also need tests for that.

Issue #1083, opened 2015-08-05
ternary operator not the same as 'if then else'


mauricevanswaaij:
The ternary operator executes the if branch of the test case below 'illegally'.
I noticed because of the resulting floating point exception.

Note that the varying values aa and bb are only partly valid, they come from a foreach with only 2 valid lanes.

Thanks.

ternary.ispc

```
void 
ternary (
int aa,
int bb,
int & A)
{
    // if then else does the right thing
    if (bb != 0)
    {
        A = aa % bb;
        print ("% = aa %% bb\n", A);
    }
    else
    {
        A = 0.0;    
        print ("0.0\n");
    }
    // ternary operator fails
    A = bb != 0 ? aa % bb : 0;

    print ("A = %\n", A);

}

export uniform int 
test_ternary (
uniform int a[],
uniform int b[],
uniform int size)
{
    uniform int result;

    foreach (i = 0 ... size)
    {
        int A;

        ternary (a[i], b[i], A);

        result += reduce_add (A);
    }

    return result;
}
```

main.cc

```
#include <ternary.h>

/*
 * 
 */
int main (int argc, char** argv)
{
    int a[2] = {0,0};
    int b[2] = {0,0};

    return ispc::test_ternary (a, b, 2);
}
```

mauricevanswaaij:
This is with ispc 1.8.2 linux version by the way..

mauricevanswaaij:
Actually the 'if then else' throws a floating point exception too, when the input is changed to

```
    int a[2] = {1,1};
    int b[2] = {0,1};
```

mauricevanswaaij:
Same behavior also happens with simple division not just with the modulo operator

mmp:
The root issue, I believe, is that the SafeToRunWithMaskAllOff() function, which checks to see if it's ok run a block of code with the execution mask all off, generally assumes that it's safe to compute any mathematical operations, so long as the result isn't actually written to memory. In this case, this is actually not true for integer divide and mod, which throw exceptions with zero for the second operand...

dbabokin:
#1092 fixes different behaviour of ternary and if.

Issue with floating point exceptions remains.

nurmukhametov:
It seems to have been fixed a while ago.

PR #1094, opened 2015-08-28
added divide by zero protection, if-fail #1083

Shishpan:


dbabokin:
You've forgot to initialise safeMaskDivision to false by default (in Globals constructor in ispc.cpp).

Otherwise looks good.

dbabokin:
This PR is outdated, so I'm closing it. But the problem remain.

PR #1095, opened 2015-09-01
Switched 3.7 branch in alloy.py from rc2 to rc4

ncos:


jbrodman:
I'd update to 3.7 final now.

PR #1121, opened 2015-11-10
[AVX512]: Make blend optimizations possible for avx512

ncos:


dbabokin:
Another way of solving this problem is to improve LLVM SROA to understand masked load/store intrinsics. But this solution is ok for now.

PR #1131, opened 2015-12-05
Fix crash when not specifying --target option

aristidb:
I moved strcasecmp on isa to _after_ comparing isa against NULL.


dbabokin:
Thanks for the contribution!

PR #1181, opened 2016-02-26
Streaming store

Shishpan:


inducer:
@lcw discovered that non-temporal stores may require fences for ordering (that LLVM has trouble dealing with currently):
http://lists.llvm.org/pipermail/llvm-dev/2016-January/093912.html

dbabokin:
PR #1337 has implemented streaming stores. This PR was used as a basefor #1337, so these changes were not wasted. I'm closing this PR.

Issue #1211, opened 2016-06-01
ISPC1.9.0 crashed TotalView


ghost:
Found an issue with ISPC 1.9.0. The debug code generated from ISPC1.9.0 will crash TotalView. I was able to use a very simple ISPC code to reproduce it. Try the attached simple.ispc and simple.cpp, the following is how I compile it:

ispc --target=sse4-i32x4 simple.ispc --header-outfile=simple.h --werror -O0 -g --pic --outfile=simple.o

 icpc -o simple simple.cpp simple.o -g -O0

Then run "totalview simple", set a breakpoint in simple.ispc. When TotalView hits that breakpoint, TotalView threw out the following message and crash:

"Ran off end of chunk in extract_data8"

ISPC1.8.2 is fine.

------------------------------------------ Source code  -------------------------------------------------------------
simple.cpp:
# include <stdio.h>
# include "simple.h"

int main() {
    float vin[16], vout[16];
    for (int i = 0; i < 16; ++i)
        vin[i] = i;

```
ispc::simple(vin, vout, 16);

for (int i = 0; i < 16; ++i)
    printf("%d: simple(%f) = %f\n", i, vin[i], vout[i]);
```

}

simple.ispc:

export void simple(uniform float vin[], uniform float vout[],
                   uniform int count) {
    foreach (index = 0 ... count) {
        float v = vin[index];
        if (v < 3.)
            v = v \* v;
        else
            v = sqrt(v);
        vout[index] = v;
    }
}

dbabokin:
Is it reproducible with gdb and/or lldb?

shengfu:
gdb does not crash, but I can not print any variables. What is lldb?

dbabokin:
lldb is standard debugger for MacOS. It's part of LLVM project. If you are on Linux, then it's not really relevant.

shengfu:
Are you able to reproduce it?

jbrodman:
We don't have totalview.

shengfu:
Also with gdb, I can not print any variables, such as vin, vout, count, v in that function. This is another evidence that debug info in ISPC generated code is wrong.

mauricevanswaaij:
Hi Shengfu, I reported this problem to Rogue Wave, the makers of totalview, and they might have a new releases that fixes it

dbabokin:
I'm closing the issue, if it's still reproducible, please reopen.

PR #1214, opened 2016-06-17
A small vital addition for r252219.

Shishpan:


dbabokin:
This should fix #1211

PR #1216, opened 2016-06-27
Enable ARM NEON generation on Microsoft Windows

ned14:
I need this because I want to generate ARM NEON output for cross compilation onto Android via the Android NDK, and I am absolutely fine if ISPC generates ARM assembler files which I can then invoke the NDK's `as` tool upon. This needs to work on Microsoft Windows, so I enabled ARM NEON support for Windows in this pull request.

I've also done these:

1. Print a fatal error message and exit if when on Windows the users tries to do --emit-obj combined with ARM generation as the object files produced are corrupted.

2. Also bumped VS project files to VS2015.

3. The VS project files weren't compiling due to the lack of the generation of builtins for the SKX target. This has been fixed.

### DIFF
--- a/ispc.sln
+++ b/ispc.sln
@@ -1,6 +1,6 @@
 ﻿
-Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
+Microsoft Visual Studio Solution File, Format Version 14.00
+# Visual Studio 2015
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "ispc", "ispc.vcxproj", "{9861F490-F516-480C-B63C-D62A77AFA9D5}"
 EndProject
 Global
--- a/ispc.vcxproj
+++ b/ispc.vcxproj
@@ -38,6 +38,8 @@
     <ClCompile Include="$(Configuration)\gen-bitcode-avx2-i64x4-64bit.cpp" />
     <ClCompile Include="$(Configuration)\gen-bitcode-knl-32bit.cpp" />
     <ClCompile Include="$(Configuration)\gen-bitcode-knl-64bit.cpp" />
+    <ClCompile Include="$(Configuration)\gen-bitcode-skx-32bit.cpp" />
+    <ClCompile Include="$(Configuration)\gen-bitcode-skx-64bit.cpp" />
     <ClCompile Include="$(Configuration)\gen-bitcode-c-32.cpp" />
     <ClCompile Include="$(Configuration)\gen-bitcode-c-64.cpp" />
     <ClCompile Include="$(Configuration)\gen-bitcode-dispatch.cpp" />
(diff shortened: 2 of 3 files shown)


dbabokin:
We don't mind ARM targets to be part of ISPC, but we don't ship binaries with ARM targets enabled by default. So if you can add them conditionally (say, depending on environment variable), that would be better solution.

ned14:
> We don't mind ARM targets to be part of ISPC, but we don't ship binaries with ARM targets enabled by default. So if you can add them conditionally (say, depending on environment variable), that would be better solution.

Is it simply a case of Intel not wanting to support ARM, or is there a good technical reason for ARM targets to be disabled by default?

dbabokin:
The main reason is that I have neither ARM hardware nor time/resources to keep track of ARM targets health. So, they are off by default. Our experience shows that things tend to break quite often with daily updates of LLVM, so this is just very practical reason.

ned14:
Oh okay. I repurposed an old phone into a Jenkins test slave, actually works surprisingly well.

Regarding this pull request, if I turn on ARM build support permanently but make it only at runtime enabled via an environment variable, would this be satisfactory?

dbabokin:
In my experience, things tend to break frequently rather than rarely. All the breakages need investigation and either fixing on our side or filing tracker against LLVM. So on practice not regularly tested equals broken. So I prefer ARM target to be off by default in the build, not on runtime. Unless someone is going to monitor healthiness of these targets on regular basis (daily or weekly).

ned14:
Ok, looks like if we maintain our own fork generating NEON enabled binaries that would be best. I'll likely set up an Appveyor instance to generate the binaries, I'll send you a new pull request I'll immediately close at that time purely for your information in case you want to clone my appveyor scripting for your own needs. Thanks for the feedback.

dbabokin:
Note that the most troubles come from LLVM updates and the key thing is to make sure that it works with LLVM trunk and a couple of latest release versions.

ned14:
We have considerably reduced requirements than you: just some ispc.exe binary which generates correct sse2, avx1, avx2 and neon. We are currently using LLVM 3.8 stable, trunk is far too exciting for us.

jazzbre:
Ok, I'm confused, so is it possible to compile arm/neon enabled ISPC build on Windows/OSX or not?

Because if I do make ARM_ENABLED=1 on OSX, first I get a llvm warning:
llvm-config: unknown component name: arm
then a whole bunch of undefined symbols.

Any pointers how to get this working or how it actually works would be really helpful!

Thanks for an awesome tool though!

ned14:
@jazzbre Yes it works fine. Do a diff between https://github.com/ned14/ispc and ispc trunk to see the config differences needed. I've compiled ARM enabled ispcs for Windows and POSIX and both generated pretty good NEON for our use cases.

jazzbre:
Thanks @ned14 , which LLVM version were you using btw - 3.8?
Also I see you didn't change anything drastic in code, so I presume you still build with ARM_ENABLED=1?
When trying to compile (with ARM enabled on OSX) either your fork or master it gives me this error:
`main.cpp:325:5: error: use of undeclared identifier
      'LLVMInitializeARMTargetInfo'
    LLVMInitializeARMTargetInfo();`
and so on... 

I've traced this macro to .../llvm/bin-3.8/include/llvm/Config/Targets.def - but there are just two targets, 
`LLVM_TARGET(NVPTX)
LLVM_TARGET(X86)
`
I've tried adding it by hand, but naturally the linker failed with:
`  "_ConvertUTF8toUTF16", referenced from:
      clang::Sema::CheckObjCString(clang::Expr*) in libclangSema.a(SemaChecking.cpp.o)
  "_ConvertUTF8toUTF32", referenced from:
      printableTextForNextCharacter(llvm::StringRef, unsigned long*, unsigned int) in libclangFrontend.a(TextDiagnostic.cpp.o)
      clang::Lexer::tryConsumeIdentifierUTF8Char(char const*&) in libclangLex.a(Lexer.cpp.o)
      clang::Lexer::LexTokenInternal(clang::Token&, bool) in libclangLex.a(Lexer.cpp.o)
      clang::CharLiteralParser::CharLiteralParser(char const*, char const*, clang::SourceLocation, clang::Preprocessor&, clang::tok::TokenKind) in libclangLex.a(LiteralSupport.cpp.o)
  "_LLVMInitializeARMAsmParser", referenced from:
      _main in main.o`
and so on...

Did you have any similar problems?

jazzbre:
Ah never mind, the only thing I had to do is build LLVM with ARM target -changing the 
`-DLLVM_TARGETS_TO_BUILD=NVPTX\;X86`
to
`-DLLVM_TARGETS_TO_BUILD=NVPTX\;X86\;ARM`
 in alloy.py.

dbabokin:
You are correct, alloy.py by default builds only x86 target. You need to
add ARM if you are using it.

PR #1224, opened 2016-07-20
Add llvm 4.0 as the highest supported llvm version

suluke:
Would you mind merging this? The new version was tagged 2 days ago and in the mean time, not much has changed. I couldn't get the whole ISPC project to build against the current svn revision, though. Some cpp source could not be generated from ll files. Maybe it was a fault with my setup, so I made this PR nonetheless. But if there is something seriously wrong about it, it's no problem if you reject this PR.

### DIFF
--- a/ispc_version.h
+++ b/ispc_version.h
@@ -51,9 +51,10 @@
 #define ISPC_LLVM_3_7 30700
 #define ISPC_LLVM_3_8 30800
 #define ISPC_LLVM_3_9 30900
+#define ISPC_LLVM_4_0 40000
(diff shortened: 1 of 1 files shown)


dbabokin:
Thanks for bringing my attention to it. There are many more things that need to be upgrade, I'll reject this PR for now and will commit complete patch soon.

dbabokin:
I've committed LLVM 4.0 support. You can try and check if it works fine for you.

suluke:
That was fast. Thank you

PR #1236, opened 2016-11-14
Support llvm 4.0: Bitcode/ReaderWriter.h -> BitCode/BitcodeWriter.h

suluke:
As per [rL286566](https://reviews.llvm.org/rL286566), utilities for Bitcode reading and writing reside in separate files


suluke:
Oh, sorry. I will also add this to `builtins.cpp`.

suluke:
`type.cpp` still won't build with top of trunk because of this change: https://reviews.llvm.org/D25073#C553822OL126
I am no backend developer, though, and I fear I don't have the necessary skills to fix this

Issue #1253, opened 2017-01-13
SSE2 targets generates psuhfb which is not SSE2


Anteru:
I'm compiling code for SSE2, and I see `pshufb` in the generated code (notice the **b** at the end). The problem is that only `pshufd` is SSE2, `pshufb` is SSSE3 (according to https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=pshufb&expand=4690,4740), so on a CPU which is SSE2 only, the code will crash.

Unfortunately, I can't share the kernel which generates this. This was tested on Linux and ISPC 1.9.1

DeepakRajendrakumaran:
This can be reproduced with shuffle and uniform parameters

void foo(uniform float b){
int8 aa = b;
int shuf = shuffle(aa, aa, 4);
}

dbabokin:
Fix is merged.

The target architecture for SSE2 target was changed from Core2 to Pentium4. This should not have big impact on performance. To achieve the same code generation as it was before the fix, the following flags can be used "--target=sse2-i32x4 --cpu=core2".

Nevertheless, the default target is SSE4 and we hope that the target, most people compiling for, is at least AVX2 :)

PR #1327, opened 2018-07-11
Issue 1253 : Adding new CPU to handle SSE2 only target.

DeepakRajendrakumaran:


dbabokin:
Fix for #1253

PR #1329, opened 2018-08-04
Implement GCC-style makefile dependency rule generation

inequation:
Hello there,

Here's a little change that I found useful in my own work.

This commit adds handling of commandline options `-M` and `-MF` in a way inspired by what GCC does (inspired, as in, the switch names and semantics have been borrowed from GCC).
`-M` generates a makefile-compatible target rule for the current output object (or the canonical `a.out`, if nothing is specified), and outputs it to `stdout`.
`-MF <filename>` puts the output in `<filename>` instead.
`-MT <target name>` allows overriding the name of the target in the rule generated by `-M`.
The functionality of `-MMM` remains unchanged, apart from the fact that a subsequent `-MF` may alter the filename.
In order to achieve this, a new bitflag type `Module::OutputFlags` is introduced. It replaces the `generatePIC` boolean and is added as a new argument in several places.

Example output from running `ispc.exe -M "D:\projects\ispc\examples\deferred\kernels.ispc" -MF "D:\projects\ispc\examples\deferred\kernels.ispc.d" -MT foo.bar` on Windows:
```
foo.bar: D:\projects\ispc\examples\deferred/deferred.h \
 D:\projects\ispc\examples\deferred\kernels.ispc \

```
This can be directly included in a makefile that e.g. has a target `all` which depends on `foo.bar`.

Regards,

Leszek

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -129,6 +129,7 @@ usage(int ret) {
     printf("    [-MMM <filename>]\t\t\tWrite #include dependencies to given file.\n");
     printf("    [-M]\t\t\t\tOutput a rule suitable for `make' describing the dependencies of the main source file to stdout.\n");
     printf("    [-MF <filename>]\t\t\tWhen used with `-M', specifies a file to write the dependencies to.\n");
+	printf("    [-MT <filename>]\t\t\tWhen used with `-M', changes the target of the rule emitted by dependency generation.\n");
     printf("    [--no-omit-frame-pointer]\t\tDisable frame pointer omission. It may be useful for profiling\n");
     printf("    [--nostdlib]\t\t\tDon't make the ispc standard library available\n");
     printf("    [--nocpp]\t\t\t\tDon't run the C preprocessor\n");
@@ -456,6 +457,7 @@ int main(int Argc, char *Argv[]) {
     const char *outFileName = NULL;
     const char *includeFileName = NULL;
     const char *depsFileName = NULL;
+	const char *depsTargetName = NULL;
     const char *hostStubFileName = NULL;
     const char *devStubFileName = NULL;
     // Initiailize globals early so that we can set various option values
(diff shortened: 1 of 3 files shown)


dbabokin:
It doesn't built with clang 5.0, for example.

Compiling main.cpp
main.cpp:688:52: error: use of overloaded operator '|' is ambiguous (with operand types 'Module::OutputFlags' and 'Module::OutputFlags')
          flags |= Module::GenerateMakeRuleForDeps | Module::OutputDepsToStdout;
                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~
./module.h:211:38: note: candidate function
inline constexpr Module::OutputFlags operator|(const Module::OutputFlags lhs, const __underlying_type(Module::OutputFlags) rhs) {
                                     ^
main.cpp:688:52: note: built-in candidate operator|(int, int)
          flags |= Module::GenerateMakeRuleForDeps | Module::OutputDepsToStdout;
                                                   ^
main.cpp:688:52: note: built-in candidate operator|(int, long)
<...>

inequation:
@dbabokin Ah, thanks, I'll test with 5.0 too, then. Is there a checklist (or, better yet, a script?) for the list of versions/compilers/etc. that I should test my changes against before making a pull request?

dbabokin:
There's no formal list of checks to go through before submitting pull request. I'm working on TravisCI and Appveyor integrations to make this verification process fast and easy, but it's still work-in-progress.

We support Linux, Mac and Windows. So ideally we need to test with gcc, clang and VS. Plus we support all version of LLVM starting from 3.2, though really testing with 3.8 and later. But I'm ok if we it works with clang and gcc as build compilers and one of the recent versions of LLVM. If one configuration is working, the rest should be easy to fix, if there any issues.

inequation:
Hi Dmitry,

Finally got some time to build under Debian for clang/llvm 5.0. After the above commit I get a clean build.

Please let me know if there are more adjustments you'd like to see wrt coding standard or anything else.

dbabokin:
In main.cpp I see two new lines, which use tabs instead of spaces.
Also, disregard my comment about need for docs. I think it's obvious and doesn't need special mentioning in docs, except regular help.

dbabokin:
Thank you for your contribution!

inequation:
Pleasure is mine! Hopefully I'll be able to contribute something more substantial soon.

PR #1336, opened 2018-08-15
Fix crash when compiling an empty `do {} while (false)` loop

inequation:
Hi there,

Apparently, the LLVM dynamic cast isn't protected against casting null pointers. The following code snippet crashes ISPC:

```
export void Kernel()
{
	do {} while (false);
}
```

The fix is trivial - check if the pointer isn't null before attempting the cast.

This is on LLVM 6.0, I should mention.


dbabokin:
Looks good to me, but we need a test case to verify that it works well now. I.e. a files in "tests" folder. Tests are run with run_tests.py.

inequation:
Sorry for taking so long. Hopefully this test is sufficient?

dbabokin:
Looks good. Thanks for the contribution!

PR #1348, opened 2018-10-17
This commit addresses two issues:

JStephenJ-intel:
1) It uses macros for the performance wall timers that will cross compile for linux and windows.
2) It uses "num_cores() * 4.0" to create smaller task granularity based on number of cores.  This will load balance better on machines loaded with random processes, anti-virus, etc.


dbabokin:
It's a duplicating pull request, the changes are already in, so closing.

PR #1349, opened 2018-10-19
Sgemm fixes2

JStephenJ-intel:
Addresses whitespace and comment concerns.  Restores working build for windows.


dbabokin:
I guess we'll need to unify the way we do timing on Windows at some point - either to move the way it is in SGEMM to common utility, or do reverse. For now it's ok to have implementation different from common utilities.

PR #1351, opened 2018-10-28
Disable dump calls if methods not exported by llvm

karjonas:
Since llvm 5.0 the various dump methods are only exported if LLVM_ENABLE_DUMP is set. Since the methods are always present in the header files calling them leads to undefined references during linking.


dbabokin:
This is the known problem. The way we address it, is to require linking against "properly" built LLVM. The recommended way to build LLVM for ISPC is to use "alloy.py" script, which passes all required defines (taking into account LLVM version) and if required applies patches to LLVM.

So I prefer ISPC to fail to link if LLVM is not meeting the requirements, rather than scarify ISPC functionality and deal with complains that dumps are not working.

PR #1359, opened 2018-11-12
Fixing ispc compilation time error

DeepakRajendrakumaran:


dbabokin:
The patch has to contain comments explaining why we need it and what problem it solves. Otherwise looks good.

DeepakRajendrakumaran:
Added comments

Issue #1355, opened 2018-11-05
main.cpp: invalid conversion from int to enum


waebbl:
Hi, I'm trying to compile ispc HEAD using cmake, llvm-6.0.1 and gcc-7.3.0 in release mode on Gentoo Linux.
In `main.cpp` I get the error:
```
var/tmp/portage/dev-lang/ispc-9999/work/ispc-9999/src/main.cpp: In function ‘int main(int, char**)’:
/var/tmp/portage/dev-lang/ispc-9999/work/ispc-9999/src/main.cpp:669:19: error: invalid conversion from ‘int’ to ‘Module::OutputFlags’ [-fpermissive]
             flags |= Module::GeneratePIC;
             ~~~~~~^~~~~~~~~
/var/tmp/portage/dev-lang/ispc-9999/work/ispc-9999/src/main.cpp:685:17: error: invalid conversion from ‘int’ to ‘Module::OutputFlags’ [-fpermissive]
           flags |= Module::GenerateFlatDeps;
           ~~~~~~^~~~~~~~~
/var/tmp/portage/dev-lang/ispc-9999/work/ispc-9999/src/main.cpp:688:62: error: invalid conversion from ‘int’ to ‘Module::OutputFlags’ [-fpermissive]
           flags |= Module::GenerateMakeRuleForDeps | Module::OutputDepsToStdout;
                                                              ^~~~~~~~~~~~~~~~~~
/var/tmp/portage/dev-lang/ispc-9999/work/ispc-9999/src/main.cpp:773:13: error: invalid conversion from ‘int’ to ‘Module::OutputFlags’ [-fpermissive]
       flags &= ~Module::OutputDepsToStdout;
       ~~~~~~^~~~~~~~~~
/var/tmp/portage/dev-lang/ispc-9999/work/ispc-9999/src/main.cpp:787:13: error: invalid conversion from ‘int’ to ‘Module::OutputFlags’ [-fpermissive]
       flags &= Module::GenerateMakeRuleForDeps;
       ~~~~~~^~~~~~~~~
```

Applying the patch
```
diff --git a/src/module.h b/src/module.h
index 26dfe4a8..3cab756e 100644
--- a/src/module.h
+++ b/src/module.h
@@ -106,7 +106,7 @@ public:
                       HostStub  /** generate host-side offload stubs */
     };
 
-    enum OutputFlags { NoFlags = 0,
+    enum OutputFlags : int { NoFlags = 0,
                        GeneratePIC = 0x1,
[...]

PR #1361, opened 2018-11-13
Fix for issue 1355 - Set underlying type for enum for gcc compilation.

DeepakRajendrakumaran:


dbabokin:
Fix for #1355.

Issue #1375, opened 2018-11-27
Crash on outputting debug info with SOA


cwfitzgerald:
Found the problem when messing around on godbolt: https://godbolt.org/z/pt6qyn. Seems to be specifically related to the debug info that godbolt mandates. The problem only happens when optimization and debug info are on, otherwise it compiles cleanly.

DeepakRajendrakumaran:
Hi,

Have you been able to reproduce this in an environment outside of godbolt(running from commandline in local machine for instance)? If yes, what are the command line switches you are passing in?

**'The problem only happens when optimization and debug info are on, otherwise it compiles cleanly.'**
Were you able to compile on godbolt by changing something? If yes, how? Or did you mean to say you were able to compile cleanly locally?

My apologies for the questions. I am having trouble reproducing it outside of godbolt so that'll be useful info.

-Deepak

DeepakRajendrakumaran:
Never mind. Got it! :)

DeepakRajendrakumaran:
This is an llvm issue. It's fixed by llvm 6.0. I'll see if i can add a patch to 5.0 in ispc

cwfitzgerald:
Awesome, thanks!

dbabokin:
Fixed.

PR #1379, opened 2018-11-29
Issue #1375 - This patch ensures compilation of ispc files with '-g' switch.

DeepakRajendrakumaran:


dbabokin:
Tagging #1375 issue here.

Issue #1385, opened 2018-12-12
Error when building embree


Trophime:
Hi,
I've rebuild latest ispc on Ubuntu 18.04 LST with clang 6.0.
Then I've tried to build embree.
It FTBS with the following error:
```
[ 87%] Building ISPC object /usr/local/src/embree/build/tutorials/grid_geometry//grid_geometry_device.dev.o
cd /usr/local/src/embree/build/tutorials/grid_geometry && /usr/bin/cmake -E make_directory /usr/local/src/embree/build/tutorials/grid_geometry/
cd /usr/local/src/embree/build/tutorials/grid_geometry && /usr/local/src/ispc/bin-6.0/bin/ispc -I /usr/local/src/embree/tutorials/grid_geometry -DTASKING_TBB --arch=x86-64 --addressing=32 -O3 --target=sse2,sse4,avx,avx2,avx512skx-i32x16 --woff --opt=fast-math --pic -h /usr/local/src/embree/build/tutorials/grid_geometry/grid_geometry_device_ispc.h -MMM /usr/local/src/embree/build/tutorials/grid_geometry//grid_geometry_device.dev.idep -o /usr/local/src/embree/build/tutorials/grid_geometry//grid_geometry_device.dev.o /usr/local/src/embree/tutorials/grid_geometry/grid_geometry_device.ispc
ispc: /usr/local/src/llvm/llvm-6.0/lib/CodeGen/SelectionDAG/SelectionDAG.cpp:3099: unsigned int llvm::SelectionDAG::ComputeNumSignBits(llvm::SDValue, const llvm::APInt &, unsigned int) const: Assertion `Tmp <= VTBits && "Failed to determine minimum sign bits"' failed.
/usr/local/src/ispc/src/main.cpp(353): FATAL ERROR: Unhandled signal sent to process; terminating.
***
*** Please file a bug report at https://github.com/ispc/ispc/issues
*** (Including as much information as you can about how to reproduce this error).
*** You have apparently encountered a bug in the compiler that we'd like to fix!
***
Aborted (core dumped)
tutorials/grid_geometry/CMakeFiles/grid_geometry_ispc.dir/build.make:64: recipe for target 'tutorials/grid_geometry/grid_geometry_device.dev.o' failed
make[2]: *** [tutorials/grid_geometry/grid_geometry_device.dev.o] Error 134
make[2]: Leaving directory '/usr/local/src/embree/build'
[...]

PR #1388, opened 2018-12-20
Fix for #1385 ; modifying Issue 1362.

DeepakRajendrakumaran:
The original fix works for 7.0 but needs an additional change for 6.0 and below.


DeepakRajendrakumaran:
Built embree with 6.0 with/without patch to verify. The original changes are still there so should affect the older fix.

PR #1396, opened 2019-01-08
Supressing warnings from tests.

DeepakRajendrakumaran:


dbabokin:
Warnings from the test suite are suppressed on per-warning basis to be able to track regressions.

This is mostly to catch warnings coming from not optimized gathers and scatters.

The change set doesn't contain changes in run_tests.py to enable --werror, this need to be run manually for now.

PR #1436, opened 2019-03-08
Connect O1 to "all_on" off

DeepakRajendrakumaran:


dbabokin:
A note about implementation of O1: turning off all-on optimization is a single biggest low hanging fruit. We may also avoid duplicating code for external functions and just call masked version with all-on mask and tweak inliner options. But one thing we don't want to do - to use O1 codegen options (we use O2 codegen option with ISPC O1 right now) - this is to lower the number of bugs we potentially trigger in LLVM.

PR #1445, opened 2019-03-26
Adding Avx512skx i32x8 support - Initial draft

DeepakRajendrakumaran:


src/ispc.h
```diff
@@ -332,6 +332,9 @@ class Target {
     /** Target-specific attribute string to pass along to the LLVM backend */
     std::string m_attributes;
 
+    /** Target-specific function attributes */
+    std::map<std::string, std::string> m_funcAttributes;
```
dbabokin: I'm not sure why m_funcAttributes needs to be map<string, string>, while it's effectively vector of tuples.

PR #1455, opened 2019-04-15
Updated fail_db.txt with failures on Windows

aneshlya:


dbabokin:
Please remove tests\transcendentals-10-0.ispc fails - the test supposed to run on Linux only. It doesn't work on Windows and Mac by design.

Also please remove timeout fails - they are due to too slow execution under SDE. So they should not count as real fails.

PR #1472, opened 2019-04-29
Enabling AARCH64 - Initial Commit.

DeepakRajendrakumaran:


run_tests.py
```diff
@@ -270,6 +273,8 @@ def run_test(testname):
                 else:
                     if options.arch == 'x86':
                         gcc_arch = '-m32'
+                    elif options.arch == 'aarch64':
```
aneshlya: Could you rework the condition starting on 271 line so it will be single condition with several elifs without nesting condition?

src/main.cpp
```diff
@@ -439,6 +439,13 @@ int main(int Argc, char *Argv[]) {
     LLVMInitializeARMAsmParser();
     LLVMInitializeARMDisassembler();
     LLVMInitializeARMTargetMC();
+
```
aneshlya: Since you're adding arch64 please modify comment on line 435

src/ispc.cpp
```diff
@@ -581,6 +588,9 @@ Target::Target(const char *arch, const char *cpu, const char *isa, bool pic, boo
     if (std::string(arch) == "x86_64")
         arch = "x86-64";
 
+    if (std::string(arch) == "aarch64")
```
aneshlya: Looks like these lines can be safely removed.

CMakeLists.txt
```diff
@@ -128,7 +128,7 @@ if (${LLVM_VERSION_NUMBER} VERSION_GREATER "3.3.0")
 endif()
 
 if (ARM_ENABLED)
```
aneshlya: Could you set ARM_ENABLED to ON by default so we would be able to check the builds in CI.

PR #1482, opened 2019-05-14
Fixed ospray build

aneshlya:
It fixes the link error: "relocation R_X86_64_PC32 against symbol stdout@@GLIBC_2.2.5..."


dbabokin:
OSPRay issue: https://github.com/ospray/ospray/issues/322

PR #1485, opened 2019-05-20
_fast implementations for rcp and rsqrt.

DeepakRajendrakumaran:


dbabokin:
1. We need tests and documentation update.
2. Would be good to use the same pattern for initializing a vector with single element - either insertelement or shuffle. insertelement seems more appropriate in this case.

PR #1489, opened 2019-05-21
Adding 'noinline' function qualifier.

DeepakRajendrakumaran:


dbabokin:
Looks good, but I would add:
1. Test for an error for "noinline" in the wrong context.
2. A warning for combination of "inline" and "noinline" on the same function. And a test for that.

PR #1503, opened 2019-06-12
Adding static intialization for varying

DeepakRajendrakumaran:


dbabokin:
The following is missing for now:
- TARGET_ELEMENT_WIDTH macro for element width in bytes (i.e. 4 for i32). It's not requited for this change-set strictly speaking, but it makes sense to bundle the changes with TARGET_WIDTH macro.
- initialization for structs.
- test coverage for new macros (used as macros, as otherwise programCount can be used). I think we need both (1) lit test compiled for multiple targets and (2) executable tests using TARGET_WIDTH for specializing for varying initialization for different width. Actually we can't use varying initialization without TARGET_WIDTH if code assumed to be portable.
- test that programCount equals to TARGET_WIDTH.
- document every varying initialization in the test with expected semantics.
- documentation update.

docs/ispc.rst
```diff
@@ -2432,6 +2432,13 @@ Variables can also be declared in ``for`` statement initializers:
 
     for (int i = 0; ...)
 
+Varying variables can be initialized with individual element values in braces.
+The number of values has to be equal to the target width:
```
dbabokin: Please mention that using static varying initialization is not portable across targets with different width. The recommended way to deal with it is to guard it with specific TARGET_WIDTH ifdef. I think the example should use use as well.

dbabokin: Please add description of TARGET_WIDTH and TARGET_ELEMENT_WIDTH to the table in line 846.

PR #1508, opened 2019-06-13
Add dump-file and debug-llvm options

aaupov:
1. Added --dump-file option to dump module IR to file after each pass (after doInitialization and runOnModule)
2. Added --debug-llvm option to turn on llvm debug info dumps
3. Fixed --debug-phase=first:last option interpretation


dbabokin:
First comment, I will have more later.
You need to update help message with description of new options.

dbabokin:
After the fixes looks good.

PR #1512, opened 2019-06-15
Fix build on trunk.

DeepakRajendrakumaran:


dbabokin:
- target .ll files don't use camel case naming, unless numbers are involved. So probably the name should be "convert_scale_to_const".
- does it make sense to define macros in util.m4?
- for 2x or 4x cases we can use "convert_scale_to_const" for individual gathers and assemble in the end. The expectation is that switch statements are optimized away relatively early, so it should not make big difference. If we optimize it late or it create problems for any optimizations, we better notice it earlier.
- LLVM IR doesn't have 80 chars limitation nor strict formatting rules. I suggest don't span a call into two line if it's reasonably sized. Also if spans into two line, parameters should start on the second line with some consistent ident (aligned with the first param, for example), same applies to switch statements ident.
- you definitely need a comment explaining why the trick is needed.
- the test should also check AVX512
- the test should check that switch did not survive.
- copyright date update is needed.

DeepakRajendrakumaran:
Made changes. Need to verify performance.

dbabokin:
Looks good now.

Issue #1408, opened 2019-01-23
ICE on Windows Compiling AVX512 KNL/SKX with Optimizations and Debug Symbols


Twinklebear:
Compiling the following example on Windows:

```c
export void add_lists(const uniform float a[], const uniform float b[],
		uniform float c[], const uniform int count)
{
	foreach (i = 0 ... count){
		c[i] = a[i] + b[i];
	}
}
```

When compiling with both debug symbols and optimization for the AVX512 KNL or SKX target:

```
ispc ./simple.ispc -g -o outtest -O2  --target=avx512skx-i32x16
```

Will crash the compiler on an LLVM error: "LLVM ERROR: unknown codeview register". The crash doesn't happen on Linux, and doesn't happen when building without optimizations.

dbabokin:
I suggest adding the test and closing the issue.

dbabokin:
Regression test was added.

PR #1517, opened 2019-06-21
Linux build in appveyor and test for #1408

aneshlya:


dbabokin:
Yay! We have Linux artifacts now!

PR #1520, opened 2019-06-25
KNC/KNL generic targets removal

aneshlya:


dbabokin:
Also, we need to change docs, but it may go as a separate PR.

dbabokin:
LGTM.

PR #1532, opened 2019-07-19
Removing AVX1.1 (IvyBridge) targets.

dbabokin:
Removing AVX1.1 targets to simplify compiler. These targets do not add much value nowadays. For most of the needs AVX1 targets can do the job. If hardware supports AVX2, then using AVX2 targets is always beneficial.

If someone really needs these targets, please file an issue.


dbabokin:
Good catch, I've updated the scripts.

PR #1534, opened 2019-07-23
Transition to Python3, fixing RunWithTimeout issues on Windows

aaupov:


dbabokin:
What about alloy.py and perf.py? We need to switch to python3 all scripts together.
Also we use python2 during build (stdlib2cpp.py and bitcode2cpp.py), we should move build to python3 as well - just to drop the dependancy from python2.

aaupov:
I ran 2to3 tool and all other scripts were reported as OK. Lit testing should also be python3-ready.

aneshlya:
Please change dependency in CMakelists file:
find_package(PythonInterp 2.7 REQUIRED)

aneshlya:
Looks like appveyor started to fail. Update .appveyor.yml with correct python version as well
Also changes are needed in scripts/build.sh

aaupov:
Travis testing is done, please check status.
@aneshlya

run_tests.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python3
```
aneshlya: This is the only python script where I see /usr/bin/env python3. others have /usr/bin/python.
I think other python scripts should have python 3 as well.

aaupov: Done in 31ea9a9

aneshlya:
Looks good to me.

aaupov:
@dbabokin please review and merge if things look good

dbabokin:
I've tried it on Mac. Just for the record: I had to update to the latest (3.15 from 3.10) CMake to make it work, otherwise python 3.7 was not found.

dbabokin:
alloy.py doesn't work, email packages are used incorrectly for python3.

I run alloy.py to build LLVM:
```
alloy.py -b --version=8.0
```

Here's seem to be the right patch:
```
--- a/alloy.py
+++ b/alloy.py
@@ -1082,10 +1082,10 @@ import multiprocessing
 import subprocess
 import re
 from shutil import copyfile
-from email.MIMEMultipart import MIMEMultipart
-from email.MIMEBase import MIMEBase
+from email.mime.multipart import MIMEMultipart
+from email.mime.base import MIMEBase
 from email.mime.text import MIMEText
-from email.Encoders import encode_base64
+from email.encoders import encode_base64
 # our drivers
 import run_tests
 import perf
```

dbabokin:
Merged manually with alloy fix.

PR #1541, opened 2019-08-06
Fixing floating point negation

DeepakRajendrakumaran:
There are 2 issues here-
1. We were not creating negative 0
2. Code was not optimal for certain fma scenarios-

A pattern like -c+a* b was not creating optimal code. We ended up creating additional assembly code to create 0 and subtract
xorps	%xmm1, %xmm1, %xmm1
vsubps	(%rcx), %ymm1, %ymm1
vfmadd231ps	(%rdx), %ymm0, %ymm1 # ymm1 = (ymm0 * mem) + ymm1

With change, this becomes
vmovups	(%rdx), %ymm1
vfmsub213ps	(%rcx), %ymm0, %ymm1 # ymm1 = (ymm0 * ymm1) - mem

See attached files for test case
[negTestFma.zip](https://github.com/ispc/ispc/files/3474545/negTestFma.zip)

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -1010,11 +1010,13 @@ static llvm::Value *lEmitNegate(Expr *arg, SourcePos pos, FunctionEmitContext *c
         return NULL;
 
     // Negate by subtracting from zero...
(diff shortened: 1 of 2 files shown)


dbabokin:
LGTM.

PR #1543, opened 2019-08-07
Improve readability of run_tests.py

aaupov:
In run_tests function:
1) Introduce Host and TargetConfig classes to reduce usage of global variables
2) extract parts into subroutines to improve the readability
3) added a few comments around multiprocessing Lock and Value usage

### DIFF
--- a/.appveyor-clean-cache.txt
+++ b/.appveyor-clean-cache.txt
@@ -38,3 +38,4 @@ by updates of docker files.
 08/27/2019: Updated dockers with tools for cross compilation
 11/26/2019: Updated docker to include fix for #1595
 01/07/2020: Updated LLVM 9.0 docker to use VS2019
+01/16/2020: Updated LLVM 8.0 docker with r372480 patch and LLVM 9.0 docker with 9.0.1 version instead of 9.0


dbabokin:
Generally looks good, but there a few things that need to be addressed before merging:
- default behaviour should not be changed with respect to printing passed tests. Default run assumes no fails and printing every passed test is not a wise choice.
- alloy.py fails to run tests. Here's an example of command line I use: 'alloy.py -r --only='current stability -O0 -O2 x86-64' --only-targets="sse avx" ' It yields the following fail:
```
_________________________STABILITY REPORT_________________________

Using test compiler: Intel(r) SPMD Program Compiler (ispc), 1.12.1dev (build commit f7ddd5256db20654 @ 20190826, LLVM 8.0.1)
Using C/C++ compiler: clang version 8.0.1

Warning:
**********
We don't have history of fails for compiler clang++8.0
All fails will be new!!!
**********
ERROR: Exception in execute_stability - maybe some test subprocess terminated before it should have
```
- run_test.py changed the way it reports the path to found ISPC. Before: `Testing ispc: /users/dybaboki/ispc/dbabokin_ispc/wt/p3/ispc` After: `Testing ispc: ../../../../../../../users/dybaboki/ispc/dbabokin_ispc/wt/p3/ispc`

The rest looks good.

dbabokin:
@aaupov Any update on that?

aaupov:
No updates so far

dbabokin:
Github decided on my behalf that I want to close this PR, instead of updating it and doesn't allow to reopen. I'll submit a new one with the same changes.

Issue #1576, opened 2019-10-01
--target=avx2-i32x16,avx512skx-i32x16 crashes ispc 1.12.0


mikemccarty-vertex:
Example:  https://ispc.godbolt.org/z/OumXQ5

Specifying only `--target=avx512skx-i32x16` or `--target=avx2-i32x16` works as expected.

dbabokin:
The source is irrelevant, it crashes when processing options. @DeepakRajendrakumaran please have a look.

DeepakRajendrakumaran:
The reason for the crash is during the removal of avx11, AVX11 was removed from ISA enum. This creates non continuous enums since we numbered the enums. We trying to iterate through all elements of the enum and convert it to string, ISAToTargetString() gets called with '3' which is not not supported.

dbabokin:
Fixed.
Thanks for reporting this!

PR #1577, opened 2019-10-01
Fix for #1576.

DeepakRajendrakumaran:
There doesn't seem to be any reason for us to set enum field values explicitly. Having the values set explicitly was causing a crash when iterating through enum with last value as upper limit since we removed AVX11.

### DIFF
--- a/tests/lit-tests/1576.ispc
+++ b/tests/lit-tests/1576.ispc
@@ -1,0 +1,6 @@
+// RUN: %{ispc} --target=avx2-i32x16,avx512skx-i32x16 %s > %t 2>&1
+// RUN: FileCheck --input-file=%t %s
+// CHECK-NOT: Please file a bug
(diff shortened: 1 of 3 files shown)


tests/lit-tests/1576.ispc
```diff
@@ -0,0 +1,5 @@
+// RUN: %{ispc} --target=avx2-i32x16,avx512skx-i32x16 %s > %t 2>&1
```
dbabokin: Is this going to work without FileCheck? Do we need FileCheck is we just checking that compiler is not crashing (returning 0)?

dbabokin:
LGTM, except test

tests/lit-tests/1576.ispc
DeepakRajendrakumaran: The test will fail even without the FileCheck since it crashes but I've added it and squashed the 2 separate commits.

Issue #1210, opened 2016-06-01
"writing dependencies to file" output


svenwoop:
ISPC outputs "writing dependencies to file XXX" when using the -MMM option. This output should be enabled only in verbose mode.

DeepakRajendrakumaran:
Closing this as fixed.

PR #1579, opened 2019-10-03
Fix for #1210 : Suppressing comment unless in "--debug" mode.

DeepakRajendrakumaran:


tests/lit-tests/1210.ispc
```diff
@@ -0,0 +1,5 @@
+// RUN: %{ispc} %s -MMM %t.dep -h %t.h > %t 2>&1
+// RUN: FileCheck --input-file=%t %s --implicit-check-not "writing dependencies"
```
dbabokin: Shouldn't we check for this string starting from capital letter not?

DeepakRajendrakumaran: Good catch. I've modified it.

Issue #1493, opened 2019-05-31
Warnings formatting doesn't take into account tabulations in the source code.


dbabokin:
If the source code line, which triggers the warning, contains tabs, the warnings message is printed as if the tab is a single space. As a result, highlighting of the part of the source code is done incorrectly.

PR #1580, opened 2019-10-03
Fix for #1493

DeepakRajendrakumaran:
In order to point to the correct code snippet we have to treat tab
as space since during lexing when we calculate the original column number,
tab and spaces are both treated as same.


dbabokin:
LGTM

PR #1592, opened 2019-10-28
Adding macro and modifying error when redefining uint*.

DeepakRajendrakumaran:


docs/ispc.rst
```diff
@@ -822,6 +822,9 @@ preprocessor runs:
   * - TARGET_ELEMENT_WIDTH
     - Element width in bytes, e.g., 4 for i32.
     - Static varying initialization.
+  * - ISPC_UINT_IS_DEFINED
+    - 1 or 0.
```
dbabokin: Value is 1.
It's either defined to 1 not not defined (which is not equivalent to being defined to 0).

PR #1601, opened 2019-11-09
Fix alignment of struct

### LINKED ISSUE #1600: Window examples assertion failed.
multiple example failed with
```
Assertion failed: sizeof(TaskInfo) % 32 == 0, file C:\Users\cloud\Desktop\ispc-v1.12.0-windows\examples\tasksys.cpp, line 196
```
with serial code.

cloudhan:
Fixed #1600

### DIFF
--- a/examples/tasksys.cpp
+++ b/examples/tasksys.cpp
@@ -171,7 +171,7 @@ typedef void (*TaskFuncType)(void *data, int threadIndex, int threadCount, int t
 
 // Small structure used to hold the data for each task
 #ifdef _MSC_VER
-__declspec(align(16))
+__declspec(align(32))
 #endif
     struct TaskInfo {
     TaskFuncType func;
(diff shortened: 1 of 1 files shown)


dbabokin:
LGTM.

@aneshlya any idea why we don't see this problem in our environment?

dbabokin:
After thinking a bit about this assert and TaskInfo alignment, I don't think we need any of them. It was added to unify with NVPTX support, but it's not really needed. I suggest dropping them, instead of fixing alignment value.

cloudhan:
Maybe it is because you are using a 32bit msvc?

```
struct TaskInfo {                      32bit   64bit
    TaskFuncType func;                 4       8
    void *data;v                       4       8
    int taskIndex;                     4       4
    int taskCount3d[3];                4*3     4*3
#if defined(ISPC_USE_CONCRT)
    event taskEvent;                   40      72
#endif 
};
```

for 32bit compiler, TaskInfo occupies 64bit without padding
for 64bit compiler, TaskInfo occupies 104bit, after padding 112bit, which is not a multiple of 32.

dbabokin:
ISPC_USE_CONCRT is not commonly used. I think the most used parallel runtimes are TBB and OpenMP. So it's typically 24 bytes for 32 bit environment and 32 bytes for 64 bit.

TaskInfo is internal data structure, which padding should not affect parallel runtime, as if it's passed to the runtime, it is used as opaque pointer. So I don't see the reason to have it aligned.

dbabokin:
@cloudhan thanks for spotting this problem and submitting the patch!

PR #1618, opened 2019-12-03
Cleanups

dbabokin:
* Remove remaining bits of -generic- support.
* Doc typo in foreach_active() example.
* Report error only using util function, do not report errors using fprintf(stderr, ...)
* Use Assert() instead of assert()
* Validate CMAKE_BUILD_TYPE


aneshlya:
Please fix typo in commit message 
"Report errors only using unil functions"

dbabokin:
I didn't want to touch cbackend.cpp at all - too much of old stuff. But probably it's good idea to fix it as well. I've pushed the updated changes.

PR #1621, opened 2019-12-11
Removed dead code optimization from -O0

scottlafetra:
Until the "volatile" keyword is in ispc, I can't have ispc removing variables I need in code for spacing reasons.

### DIFF
--- a/src/opt.cpp
+++ b/src/opt.cpp
@@ -467,7 +467,6 @@ void Optimize(llvm::Module *module, int optLevel) {
         optPM.add(CreateIsCompileTimeConstantPass(true));
         optPM.add(llvm::createFunctionInliningPass());
         optPM.add(CreateMakeInternalFuncsStaticPass());
-        optPM.add(llvm::createCFGSimplificationPass());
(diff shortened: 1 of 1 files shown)


dbabokin:
Could you please elaborate on the reasons why you need that? The way stdlib is designed, we would have **a lot** of dead code if we don't run dead code elimination. I'm not sure how much of this elimination is done by `GlobalDCEPass` and how much by `CFGSimplificationPass` though.

As a workaround, you can turn this pass of from command line:

```
ispc --off-phase=106 -O0 test.ispc
```

scottlafetra:
I'm having ISPC read through an array of structures that is passed to it by a pointer. Using a known offset is not feasible because I need to access multiple pieces of data in different parts of each structure in the array. Copying over the data into an ispc friendly structure is also not ideal, as that would most likely negate any perf benefits I would get from using ispc in the first place. 

That command line option looks like it does the trick for me though, thanks!

dbabokin:
I'm closing the PR, if you have a reproducer - please file a tracker and we'll fix it.

PR #1622, opened 2019-12-16
Dynamic registration of targets files

dbabokin:
Major redesign of how target files (target*.ll, builtins-c*.ll, dispatch*.ll) are linked to ISPC. With current change set targets only need to be linked to ISPC, after that they are automatically registered and can be used. This simplifies the build system, which now needs to manage multi-target builds (targeting multiple OSes and architectures).

As a side effect, I had to clean up the way we represent ispc-target and arch information. Overall this intended to be simplification and clarification of how we manage this.

This PR contains quite a few TODO, which I'm going to address in the following commits.


dbabokin:
Remaining TODOs summary:
target_enums.cpp:
```
    // TODO: ensure skx-i32x8 is not enabled and linked for earli LLVM version.
```

bitcode_lib.cpp:
```
// TODO: this is debug version: either remove or make it use friendly.
void BitcodeLib::print() const
```

target_registry.h
```
    // TODO: do we need separate dispatch module for Windows and for Unix (or any specific platform)?
    const BitcodeLib *m_dispatch;

    ...

    // Print user-friendly message about platforms support
    // TODO: implement.
    void printSupportMatrix() const;

    // TODO: implement
    std::string getSupportedArchs();
    std::string getSupportedTargets();
    std::string getSupportedOSes();

    // TODO: implement
    bool isSupported(ISPCTarget target, TargetOS os, Arch arch) const;
```

PR #1624, opened 2019-12-23
Support matrix

dbabokin:
* Print support matrix under --support-matrix
* Use information about dynamically link targets to report supported targets/arches/oses.

### DIFF
--- a/src/target_enums.cpp
+++ b/src/target_enums.cpp
@@ -315,7 +315,7 @@ TargetOS ParseOS(std::string os) {
 std::string OSToString(TargetOS os) {
     switch (os) {
     case TargetOS::windows:
-        return "Window";
+        return "Windows";
     case TargetOS::linux:
         return "Linux";
     case TargetOS::macos:
@@ -335,7 +335,7 @@ std::string OSToString(TargetOS os) {
 std::string OSToLowerString(TargetOS os) {
     switch (os) {
     case TargetOS::windows:
-        return "window";
+        return "windows";
     case TargetOS::linux:
         return "linux";
     case TargetOS::macos:
--- a/src/target_enums.cpp
+++ b/src/target_enums.cpp
@@ -292,7 +292,7 @@ bool ISPCTargetIsNeon(ISPCTarget target) {
 }
 
 TargetOS ParseOS(std::string os) {
-    std::string supportedOses = Target::SupportedOSes();
+    std::string supportedOses = g->target_registry->getSupportedOSes().c_str();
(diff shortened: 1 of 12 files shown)


src/target_enums.cpp
```diff
@@ -332,6 +332,26 @@ std::string OSToString(TargetOS os) {
     UNREACHABLE();
 }
 
+std::string OSToLowerString(TargetOS os) {
+    switch (os) {
+    case TargetOS::windows:
+        return "window";
```
aneshlya: Please fix "window"->"windows"

src/target_registry.cpp
```diff
@@ -135,3 +140,143 @@ const BitcodeLib *TargetLibRegistry::getISPCTargetLib(ISPCTarget target, TargetO
...
+    // OS names row
+    std::vector<std::string> os_names;
+    os_names.push_back("");
+    for (int j = (int)TargetOS::windows; j < (int)TargetOS::error; j++) {
+        os_names.push_back(OSToString((TargetOS)j));
+    }
+    table.push_back(os_names);
+
+    // Fill in the name, one target per the row.
+    for (int i = (int)ISPCTarget::sse2_i32x4; i < (int)ISPCTarget::error; i++) {
```
aneshlya: Do we have some guarantee that we will not change the order of targets in enums and sse2_i32x4 will be always the 1st one?
Maybe it makes sense to have "none" as the 1st value for targets and OSes enums if it does not introduce many changes.

src/target_registry.cpp
```diff
@@ -135,3 +140,143 @@ const BitcodeLib *TargetLibRegistry::getISPCTargetLib(ISPCTarget target, TargetO
...
+                    arch_list_os += ArchToString(arch);
+                }
+            }
+            arch_list_target.push_back(arch_list_os);
+            row.push_back(arch_list_os);
+        }
+        table.push_back(row);
+    }
+
+    // Colect maximum sizes for all columns
```
aneshlya: Colect->collect

src/target_registry.cpp
dbabokin: Iterating through enum values is ugly. I suggest keeping it this way. I we have more places, where we have to iterate through these enums, we'll need to come up with a better way.

src/target_registry.cpp
dbabokin: Fixed.

src/target_enums.cpp
dbabokin: That's really good catch. Thanks! Fixed.

PR #1640, opened 2020-01-23
Fix synchronization issue in run_tests.py.

dbabokin:
The pattern with checking Queue.empty() and doing Queue.get() without critical
section leads to dead locks. So, options are: use critical section or different
mechanism to pull the new job. This solution uses "STOP" markers in the Queue.
So, no check for empty queue is needed. The same queue is also used to indicate
that all tests are handled, as it's JoinableQueue. Hence getting rid of extra
Queue for signaling completion.

Also queue for signalling errors is removed. All exception will be reported as
Runtime failures and reported in the log.

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -413,7 +413,7 @@ def run_test(testname, host, target):
 # pull tests to run from the given queue and run them.  Multiple copies of
 # this function will be running in parallel across all of the CPU cores of
 # the system.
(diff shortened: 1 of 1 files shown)


dbabokin:
At some point I experienced problem with this fix on Windows, but I can't reproduce it anymore. So, now merging.

Issue #1595, opened 2019-10-30
ispc 1.12.0 compiler crash


mikemccarty-vertex:
I tried to simplify it as much as possible.  Commenting out the last line of `csClosestPoint()` avoids the crash.

Program also available at https://ispc.godbolt.org/z/5wyijf

```
struct Vec3f {
};

inline uniform float signedDist(uniform const Vec3f& pt) {
    return 0;
}

struct TriangleVerts
{
    Vec3f v0;
    Vec3f v1;
    Vec3f v2;
};

uniform TriangleVerts get_triangle_verts()
{
    uniform TriangleVerts tri;
    return tri;
}

void csClosestPoint()
{
    uniform TriangleVerts tri = get_triangle_verts();

    uniform const float d0 = signedDist(tri.v0);
}
```

DeepakRajendrakumaran:
The crash happens while running llvm's IPConstantpropogation pass. Can be reproduced with this -https://ispc.godbolt.org/z/45LBCg

dbabokin:
Please check if the bug is reproducible with LLVM 9.0 and trunk and file a bug against LLVM if it's not fixed yet.

DeepakRajendrakumaran:
It is reproducible with trunk(using opt trunk in the godbolt link above is one way to check).
 Holding off on filing a  bug for now as Craig is possibly adding a fix soon.

dbabokin:
Filing a bug against LLVM should do no harm and help Craig describe the problem in his patch. But I would say it's up to Craig to decide if he needs one.

DeepakRajendrakumaran:
https://bugs.llvm.org/show_bug.cgi?id=43857

DeepakRajendrakumaran:
Here is the change review - https://reviews.llvm.org/D69656 

Looks like the pass we are using is deprecated and we need to switch to -ipsccp: Interprocedural Sparse Conditional Constant Propagation

dbabokin:
@DeepakRajendrakumaran any update?

DeepakRajendrakumaran:
There are 2 parts to this-
1. The llvm fix to not crash with the current pass. This patch has already merged. I've tested it with 8.0 and it doesn't crash. Back-porting it to 8.0 is one option. This would be the short-term fix to stop the crash
2. Ideally, we need to look into using Interprocedural Sparse Conditional Constant Propagation instead.

dbabokin:
[...]

PR #1646, opened 2020-01-28
Add FreeBSD support

myfreeweb:
Since usage of the TargetOS enum was added, ispc stopped working on FreeBSD.
Add FreeBSD to TargetOS and related CMake files.

Plus, fix an issue with pthreads in tasksys (semaphore names must start with a slash on FreeBSD).

Also use `stdlib.h` in tasksys instead of `malloc.h` on both Linux and FreeBSD.
(malloc.h has been deprecated for ages and FreeBSD -CURRENT forbids its usage)

---

Tested on amd64 and aarch64.

---

cc @yurivict — you've updated the port to the latest git revision, and it builds, but it can't compile anything since this wasn't done :) btw, you should add aarch64 (and I guess armv7 but I haven't tested that) to the architectures, and replace the dump patches with `CMAKE_ON=ISPC_NO_DUMPS`. also `run-examples` is broken, examples are built with CMake, and it's just trying to run ninja in the source dirs. also of course remove the tasksys patch since I've fixed that here. and remove the stale comment about tests not supporting python 3.


examples/cmake/AddISPCExample.cmake
```diff
@@ -44,7 +44,7 @@ function(add_ispc_example)
     set(ISPC_OBJ_NAME "${CMAKE_CURRENT_BINARY_DIR}/${ISPC_SRC_NAME}_ispc${CMAKE_CXX_OUTPUT_EXTENSION}")
     set(ISPC_FLAGS ${example_ISPC_FLAGS})
     if (UNIX)
-        execute_process( COMMAND bash "-c" "uname -m | sed -e s/x86_64/x86/ -e s/i686/x86/ -e s/arm.*/arm/ -e s/sa110/arm/" OUTPUT_VARIABLE ARCH)
+        execute_process( COMMAND bash "-c" "uname -m | sed -e s/x86_64/x86/ -e s/amd64/x86/ -e s/i686/x86/ -e s/arm64/aarch64/ -e s/arm.*/arm/ -e s/sa110/arm/" OUTPUT_VARIABLE ARCH)
```
myfreeweb: (yeah, due to historical accident, `uname -m` reports `arm64` even though literally every other place refers to it as `aarch64`)

yurivict:
@myfreeweb Sorry, I didn't check the runtime. I will update the port.

dbabokin:
On the first glance the changes looks ok, but I'll need more time to play with them.

I also like the idea to have FreeBSD officially supported, as we have multiple requests from you.

Ideally we need to build it as part of CI, otherwise we'll keep breaking it. Are you aware about any public CI services, that support FreeBSD?

myfreeweb:
> Are you aware about any public CI services, that support FreeBSD?

[Cirrus](https://cirrus-ci.org/guide/FreeBSD/), [sr.ht](https://builds.sr.ht/)

yurivict:
Patches from the ispc port should be upstreamed too.

myfreeweb:
@yurivict see original post here — they should be removed after replacing with this. I fixed the pthread thing in the example, and

> replace the dump patches with CMAKE_ON=ISPC_NO_DUMPS

dbabokin:
Looks like Appveyor is going add support for FreeBSD: https://github.com/appveyor/ci/issues/2844

dbabokin:
I'll reiterate my comment on using stock LLVM (built with no ASSERT or DUMP enabled and without applying patches) - this way ISPC will miss some functionality and will contain known bugs.

yurivict:
We can't use custom build because LLVM doesn't easily build on BSDs without patches.

dbabokin:
How large these patches are? One of the possibilities is to include them to ISPC scripts building LLVM.

dbabokin:
Also, I would assume that LLVM community is interested to accept these patches for upstream.

yurivict:
Last time I looked - there are a lot of patches. I don't think LLVM community would accept them. With the LLVM add-on the ispc port becomes unmaintainable (high-maintenance).

dbabokin:
Is there a link where I can see these patches? I'd like to see the scale of the problem.

yurivict:
https://svnweb.freebsd.org/ports/head/devel/llvm-devel/files/

---------------------

But the correct solution is for you to either convince llvm folks to enable the functionality that you need by default, or to add a shared library to the llvm build that would contain the functionality that you need.

Obviously, to rebuild the whole llvm in the port is a wrong solution.

dbabokin:
I've glanced through the patches and most of them look ok in terms of potential to integrate to LLVM upstream. It will take quite a lot of talking and explaining why it's needed and probably rewriting some of them during review. So I'm wondering if this was already attempted or not?

yurivict:
I don't maintain the llvm ports so I wouldn't know.

dbabokin:
Is there a way to reach this person? The ability to build clang on FreeBSD is kind of deal breaker for ease of maintenance for products building on top of LLVM, ISPC included.

yurivict:
The maintainer's e-mail is in the Makefile one level up.

examples/cmake/AddISPCExample.cmake
dbabokin: iOS also refers to is as `arm64`

src/ispc.cpp
```diff
@@ -1127,6 +1127,22 @@ std::string Target::GetTripleString() const {
...
         break;
+    case TargetOS::freebsd:
+        if (m_arch == Arch::x86) {
+            triple.setArchName("i386");
+        } else if (m_arch == Arch::x86_64) {
+            triple.setArchName("amd64");
+        } else if (m_arch == Arch::arm) {
+            triple.setArchName("armv7");
+        } else if (m_arch == Arch::aarch64) {
+            triple.setArchName("aarch64");
```
dbabokin: Is it should be `aarch64` or `arm64` here?

dbabokin:
Also, please update Copyright notice to 2020 in the beginning of the affected files.

src/ispc.cpp
myfreeweb: `aarch64`. As I said, `arm64` is *only* featured in the `uname` output.

myfreeweb:
> We can't use custom build because LLVM doesn't easily build on BSDs without patches.

I really doubt that this is true, these patches are mostly about system installs (version suffixes like `llvm90` etc.). LLVM as a library included in a project should build fine, I'll try that.

yurivict:
Why don't you integrate your requirements in LLVM instead? It's a one-to-many situation. Instead of paining many, you can just fix one and many would benefit.

dbabokin:
> Why don't you integrate your requirements in LLVM instead? It's a one-to-many situation. Instead of paining many, you can just fix one and many would benefit.

Simple. We require two things:
1. LLVM patches. All of the them are backporting. I.e. patches for 8.0 are patches integrated in trunk and possible 9.0/10.0. We just discovered the problem too late and the fix was late for 8.0. But we still might want to build with 8.0. So manual backporting.
2. Building with ASSERTs - we naturally prefer to know about problems earlier (at compile time), rather than spend time investigating incorrect behavior. Also ASSERTs enable DUMPs, which we use as a way to dump intermediate representation and the way to understand what's going wrong. It's not super critical, but it's preferable way to build ISPC.

myfreeweb:
> Building with ASSERTs

Actually we were building llvm packages with assertions enabled for a long time (and might still be doing that). Doesn't seem intentional (!), it's been that way due to a CMake accident. A few compilers didn't like that very much.

> ASSERTs enable DUMPs

Doesn't seem to be the case.. (or maybe we are without assertions now)

> All of the them are backporting. I.e. patches for 8.0 are patches integrated in trunk and possible 9.0/10.0

Oh, well then using `llvm90` or `llvm-devel` in ports packaging is perfectly fine.

For personal use / CI, I still think that building your own LLVM from upstream source might just work.

yurivict:
A special version of llvm can't be just under ispc. It should be a separate port. You should ask the maintainer of llvm-devel to kindly fork with the options that you require. Then, it should be installed into a non-conflicting path so that all of them can be installed at the same time.

It's not a simple matter as you can see.

dbabokin:
LLVM as a collection of products (clang, lldb, etc), which are usable themselves - yes this way it needs to be a separate port with non-conflicting path, etc.

But for as LLVM is part of our build. It's a collection of spare part that we are using and end up linking to our binary. So this is completely different story. I doesn't need to be compatible to system clang, for example.

And again, the version that we build is just stock version with a number of back-ported patches, which fix specific problems important for us, nothing else. And building with ASSERTs enabled - it's just the flavor of the build (configuration options), nothing else.

myfreeweb:
Is anyone reading my comments? Do I have to repeat myself?

> Actually we were building llvm packages with assertions enabled for a long time (and might still be doing that).

so

> just stock version with a number of back-ported patches

means we can just use our version. (No need for the backported patches when the version is the most recent one.)

Why is the port even discussed here at all? This should not matter to the upstream project, we can package however we want, if the packaged llvm works fine in practice, we can keep using it.

yurivict:
> Why is the port even discussed here at all?

Because this is a part of the package that we are discussing here. Nothing can be just downloaded and built during the build a most upstreams seem to be implying. If you would like to use a custom version of llvm, it should be downloaded before ispc by the port.

myfreeweb:
> If you would like to use a custom version of llvm, it should be downloaded before ispc by the port.

Of course. But we would not like that, so let's just keep using `llvm90` instead of listening to @dbabokin :P It works fine.

dbabokin:
Ok, if you prefer to build without all the known fixes - that's your choice. I don't object, unless already fixed bugs are reported again. For example, for LLVM 9.0 we have on patch for this problem https://github.com/ispc/ispc/issues/1595

We'll get back to this problem when we will be enabling FreeBSD CI. I hope Appveyor introduce FreeBSD instances soon.

dbabokin:
For the PR itself, it looks good to me, just fix Copyright string to 2020 in the files where it's relevant.

dbabokin:
Ok, I've fixed copyright myself.

myfreeweb:
Nice!

@yurivict please also add `aarch64` to the arch list in the port (and remove the `tests are broken with python-3.6` comment)

Issue #1596, opened 2019-11-05
Misplaced "Initializer for global variable must be a constan" error


tkoziara:
When trying to compile such code:

```
/* a simple case */
static uniform float a = 1.0;
static uniform float * uniform pa = &a; /* this fails */

/* here is what I'd really like to do */
static const uniform float b[] = {0.};
static const uniform float c[] = {0.};
static const uniform float * uniform pv[] = {NULL, b, c}; /* this also fails */
```
the following errors are reported:
bug7.ispc:3:37: Error: Initializer for global variable "pa" must be a constant. 
bug7.ispc:8:46: Error: Initializer for global variable "pv" must be a constant. 

File attached:) Thank you:)
[bug7.zip](https://github.com/ispc/ispc/files/3808585/bug7.zip)

dbabokin:
Fixed by #1656

PR #1656, opened 2020-02-15
Fix for #1596.

DeepakRajendrakumaran:
This mostly has logic to identify compile time constants
which are global addresses and their variants.

Enables following initializations -
uniform int sz = sizeof(uniform int);

static uniform float a = 1.0;
static uniform float * uniform pa = &a;
static const uniform float * uniform parr[] = {NULL, &a, &a};

static const uniform float b[] = {0.,4., 3., 2., 1.};
static const uniform float * uniform pb[] = {NULL, b+3, b};

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -2647,6 +2647,43 @@ void BinaryExpr::Print() const {
     pos.Print();
 }
 
+llvm::Constant *BinaryExpr::GetConstant(const Type *type) const {
+
+    // Are we doing something like (basePtr + offset)[...] = ... for a Global
+    // Variable
+    if (!GetLValueType())
+        return NULL;
+
+    if (!((op == BinaryExpr::Op::Add) || (op == BinaryExpr::Op::Sub)))
+        return NULL;
+    else if (op == BinaryExpr::Op::Sub) {
+        if (const PointerType *pt1 = CastType<PointerType>(arg1->GetType())) {
+            return NULL;
+        }
+    }
+
+    if (const PointerType *pt0 = CastType<PointerType>(arg0->GetType())) {
+        llvm::Constant *c1 = arg0->GetConstant(pt0);
+        ConstExpr *cExpr = llvm::dyn_cast<ConstExpr>(arg1);
+        if ((cExpr == NULL) || (c1 == NULL))
+            return NULL;
+        llvm::Constant *c2 = cExpr->GetConstant(cExpr->GetType());
+        if (op == BinaryExpr::Op::Sub)
+            c2 = llvm::ConstantExpr::getNeg(c2);
+        llvm::Constant *c = llvm::ConstantExpr::getGetElementPtr(PTYPE(c1), c1, c2);
+        return c;
+    } else if (const PointerType *pt1 = CastType<PointerType>(arg1->GetType())) {
+        llvm::Constant *c1 = arg1->GetConstant(pt1);
+        ConstExpr *cExpr = llvm::dyn_cast<ConstExpr>(arg0);
+        if ((cExpr == NULL) || (c1 == NULL))
+            return NULL;
+        llvm::Constant *c2 = cExpr->GetConstant(cExpr->GetType());
+        llvm::Constant *c = llvm::ConstantExpr::getGetElementPtr(PTYPE(c1), c1, c2);
+        return c;
+    }
+
+    return NULL;
+}
 ///////////////////////////////////////////////////////////////////////////
 // AssignExpr
 
@@ -6950,8 +6987,27 @@ llvm::Constant *TypeCastExpr::GetConstant(const Type *constType) const {
     // 2. Converting function types to pointer-to-function types
     // 3. And converting these from uniform to the varying/soa equivalents.
     //
-    if (CastType<PointerType>(constType) == NULL)
+
+    if ((CastType<PointerType>(constType) == NULL) && (llvm::dyn_cast<SizeOfExpr>(expr) == NULL))
         return NULL;
+
+    llvm::Value *ptr = NULL;
+    if (GetBaseSymbol())
+        ptr = GetBaseSymbol()->storagePtr;
+
+    if (ptr && llvm::dyn_cast<llvm::GlobalVariable>(ptr)) {
+        if (CastType<ArrayType>(expr->GetType())) {
+            if (llvm::Constant *c = llvm::dyn_cast<llvm::Constant>(ptr)) {
+                llvm::Value *offsets[2] = {LLVMInt32(0), LLVMInt32(0)};
+                llvm::ArrayRef<llvm::Value *> arrayRef(&offsets[0], &offsets[2]);
+                llvm::Value *resultPtr = llvm::ConstantExpr::getGetElementPtr(PTYPE(c), c, arrayRef);
+                if (resultPtr->getType() == constType->LLVMType(g->ctx)) {
+                    llvm::Constant *ret = llvm::dyn_cast<llvm::Constant>(resultPtr);
+                    return ret;
+                }
+            }
+        }
+    }
 
     llvm::Constant *c = expr->GetConstant(constType->GetAsUniformType());
     return lConvertPointerConstant(c, constType);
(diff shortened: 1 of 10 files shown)


src/expr.cpp
```diff
@@ -2647,6 +2647,43 @@ void BinaryExpr::Print() const {
...
+
+    // Are we doing something like (basePtr + offset)[...] = ... for a Global
+    // Variable
+    if (!GetLValueType())
+        return NULL;
+
+    if (!((op == BinaryExpr::Op::Add) || (op == BinaryExpr::Op::Sub)))
+        return NULL;
+    else if (op == BinaryExpr::Op::Sub) {
+        if (const PointerType *pt1 = CastType<PointerType>(arg1->GetType())) {
```
dbabokin: Why defining and assigning to pt1 here? It's confusing and makes the intentions not clear.
Also adding comment why this case is rejects will not harm.

src/expr.cpp
```diff
@@ -7299,8 +7355,30 @@ llvm::Constant *AddressOfExpr::GetConstant(const Type *type) const {
     if (ft != NULL) {
         llvm::Constant *c = expr->GetConstant(ft);
         return lConvertPointerConstant(c, type);
-    } else
-        return NULL;
+    } else {
```
dbabokin: Removing "else" branch here, it would save 1 level of nesting.

src/expr.cpp
```diff
@@ -7358,6 +7436,26 @@ Expr *SizeOfExpr::Optimize() { return this; }
 
 int SizeOfExpr::EstimateCost() const { return 0; }
 
+llvm::Constant *SizeOfExpr::GetConstant(const Type *type) const {
```
dbabokin: I'm not sure why, but this produces incorrect code:
`uniform int s = sizeof (varying int);`
It treats "varying int" as uniform and initializes s with 4.

dbabokin: Also, there should be a warning or error, in case of assigning size of varying types to global variables when doing multi-target compilation with different data layouts.

src/ispc.cpp
```diff
@@ -1347,6 +1347,13 @@ void Target::markFuncWithTargetAttr(llvm::Function *func) {
     }
 }
 
+bool Target::IsGenericTypeLayoutIndeterminate(llvm::Type *type) {
```
dbabokin: Let's change existing instances of "(m_isa == Target::GENERIC && lGenericTypeLayoutIndeterminate(type) == true)" to IsGenericTypeLayoutIndeterminate.

tests/lit-tests/1596.ispc
```diff
@@ -0,0 +1,23 @@
+// RUN: %{ispc} %s > %t 2>&1
```
dbabokin: Lit tests must use --target option, otherwise they depend of the host.

tests/lit-tests/1596.ispc
```diff
@@ -0,0 +1,23 @@
...
+static const uniform float * uniform parr[] = {NULL, &a, &a};
+
+static const uniform float b[] = {0.,4., 3., 2., 1.};
+static const uniform float * uniform pb[] = {NULL, b+3, b};
+static const uniform float * uniform pb1[] = {NULL, b, b};
+
+varying float i = 80;
+varying float p_foo[] = {1, 4, 3, 2, 4, 3, 3, 3};
+varying float * uniform au = &p_foo[0];
+
```
dbabokin: Adding these lines to test crashes the compiler:
```
uniform int ii[][] = {{1, 2, 3}, {1, 2, 3}};
uniform int * uniform aaa = &ii[0][0];
```

src/expr.cpp
```diff
@@ -2647,6 +2647,43 @@ void BinaryExpr::Print() const {
...
+llvm::Constant *BinaryExpr::GetConstant(const Type *type) const {
+
+    // Are we doing something like (basePtr + offset)[...] = ... for a Global
+    // Variable
+    if (!GetLValueType())
+        return NULL;
+
+    if (!((op == BinaryExpr::Op::Add) || (op == BinaryExpr::Op::Sub)))
+        return NULL;
+    else if (op == BinaryExpr::Op::Sub) {
```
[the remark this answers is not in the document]

src/expr.cpp
DeepakRajendrakumaran: done

src/expr.cpp
DeepakRajendrakumaran: multi-target compilation causes some issues. I'll have to rework that a little.

src/expr.cpp

DeepakRajendrakumaran:
All the comments should be addressed now.

tests/1596.ispc
```diff
@@ -0,0 +1,37 @@
...
+// *parr[2] = 7
+// *pb[1] = 5
+// *pb1[2] = 2
+// *aaa = 4
+// sz = 4
+// sVar = 8 * WIDTH
+// sVar_loc = 1 * WIDTH
+// sVar_loc_global = 1 * WIDTH
+export void result(uniform float RET[]) {
+    RET[programIndex] = 7 * 7 * 5 * 2 * 4 * 4 * (8 * programCount) * (1 * programCount) * (1 * programCount);
```
dbabokin: RET has float type. To represent the value precisely it needs to fit mantissa, which is 23 bits.
In case of programCount = 64 (as in avx512skx-i8x64, for example), the result of this expression will require 34 bits (0x3D4000000).
The test will likely pass, as it's going to be rounded in the same way both times, but it's not very reliable way to test for equality. So please use other formula to compute the  "hash".  Either adding values or adding "weighted" values (1*x1 + 2*x2 + 3*x3 +....) is good enough IMHO.

src/expr.h
```diff
@@ -80,7 +80,7 @@ class Expr : public ASTNode {
         constant of the given type, this method should return the
         corresponding llvm::Constant value.  Otherwise it should return
         NULL. */
-    virtual llvm::Constant *GetConstant(const Type *type) const;
+    virtual std::pair<llvm::Constant *, bool> GetConstant(const Type *type) const;
```
dbabokin: You need comments explaining the new function interface and how it's supposed to be used.

src/expr.cpp
```diff
@@ -2647,6 +2648,53 @@ void BinaryExpr::Print() const {
     pos.Print();
 }
 
+std::pair<llvm::Constant *, bool> BinaryExpr::GetConstant(const Type *type) const {
```
dbabokin: This function vigorously needs comments explaining isNotValidForGlobal logic and the concept as a whole.

src/ispc.h
```diff
@@ -621,6 +623,9 @@ struct Globals {
 
     /** Lines for which warnings are turned off. */
     std::map<std::pair<int, std::string>, bool> turnOffWarnings;
+
+    /* If true, we are compiling for more than one target. */
+    bool isMultiTargetCompilation;
```
dbabokin: Why is this a flag, but not a function? `targets.size() > 1` check doesn't seem to be too expensive.

src/ispc.h
```diff
@@ -211,6 +211,8 @@ class Target {
     /** Mark LLVM function with target specific attribute, if required. */
     void markFuncWithTargetAttr(llvm::Function *func);
 
+    bool IsGenericTypeLayoutIndeterminate(llvm::Type *type);
```
dbabokin: Where is the documentation for this function?

src/expr.h
dbabokin: For the second argument I suggest using `enum class` consisting of two values - this way semantics is much more explicit and it's harder to misinterpret the result.

src/module.cpp
```diff
@@ -335,17 +336,26 @@ void Module::AddGlobalVariable(const std::string &name, const Type *type, Expr *
...
                         if (type->IsConstType())
                             // Try to get a ConstExpr associated with
                             // the symbol.  This llvm::dyn_cast can
                             // validly fail, for example for types like
                             // StructTypes where a ConstExpr can't
                             // represent their values.
                             constValue = llvm::dyn_cast<ConstExpr>(initExpr);
-                    } else
+                    } else if (skipInitialization)
+                        Error(initExpr->pos,
```
dbabokin: In case of single target compilation we need a warning.

src/ispc.h
DeepakRajendrakumaran: We don't really save the targets anywhere. It's passed into 'Module::CompileAndOutput' and we make decisions for targets individually. So we either store all targets somewhere or just set a flag when we know it's multi-target.

dbabokin: Ok, makes sense.

src/expr.h
DeepakRajendrakumaran: The problem with making it an 'enum class' is ORing values in case of recursive calls. Everytime GetConstant is called, it might recursively call GetConstant on part of that expresion and we need to ORvalue in current GetConstant with the called GetConstant. If its an enum, this becomes a little clunky. Let me know if I understand this correctly

dbabokin: Ok, let's keep it a bool.

src/expr.h
```diff
@@ -78,9 +78,10 @@ class Expr : public ASTNode {
 
     /** If this is a constant expression that can be converted to a
         constant of the given type, this method should return the
-        corresponding llvm::Constant value.  Otherwise it should return
-        NULL. */
-    virtual llvm::Constant *GetConstant(const Type *type) const;
+        corresponding llvm::Constant value and a flag denoting if it's
+        valid for multi-target compilation for use as an initializer.
```
dbabokin: for use as an initializer **of a global variable**.

Locals are not a problem.

src/expr.cpp
```diff
@@ -2647,6 +2648,66 @@ void BinaryExpr::Print() const {
...
+    // pointer addresses
+    // Case 1 : first argument is a pointer address.
+    // In this case as long as the second argument is a constant value, we are fine
+    // Case 2 : second argument is a pointer address.
+    // In this case, it has to be an addition with first argument as
+    // a constant value.
+    if (!((op == BinaryExpr::Op::Add) || (op == BinaryExpr::Op::Sub)))
+        return std::pair<llvm::Constant *, bool>(NULL, false);
+    if (op == BinaryExpr::Op::Sub) {
+        // Ignore cases where subtrahend is a PointerType
```
dbabokin: typo: "subtrahend"

src/expr.cpp
```diff
@@ -2647,6 +2648,66 @@ void BinaryExpr::Print() const {
...
+        // Eg. b - 5 is valid but 5 - b is not.
+        if (CastType<PointerType>(arg1->GetType()))
+            return std::pair<llvm::Constant *, bool>(NULL, false);
+    }
+
+    // 'isNotValidForMultiTargetGlobal' is required to let the caller know
+    // that the llvm::constant value returned cannot be used in case of
+    // multi-target compilation.
```
dbabokin: add "for initialization of globals. This is due to different constant values for different targets, i.e. computation involving sizeof() of varying types"

src/expr.h
DeepakRajendrakumaran: Done

src/expr.cpp

Issue #1583, opened 2019-10-07
Implement WebAssembly backend


dbabokin:
Implementing WebAssembly back-end for ISPC should be relatively straight-froward. I've heard from several customers that they are interested in experimenting with it.

dbabokin:
#1662 implements initial support for WebAssembly

dbabokin:
I think the grand-plan for enabling should look like the following:

- [x] Add target and basic infrastructure (PR #1662)
- [ ] Complete target-wasm-i32x4.ll, so that it has complete set of required library functions (possibly just stubs without real implementation). Hence programs will be able to compile without --nostdlib and will fail only when specific functionality is missing.
- [ ] Add test driver, similar to test_static.cpp, and support in run_tests.py. At this step we might want a script to install V8 dependency.
- [ ] Add CI support - I expect that it going to be pretty easy.
- [ ] Look at code and this about code generation.

dbabokin:
Tagging @aschrein, who contributed initial support (PR #1662) and I hope will keep working on it.

aschrein:
I've been investigating some artifacts and found that [WAVM](https://github.com/WAVM/WAVM) is a more robust alternative to V8. V8 is a widely used real-life engine and some testing on it must be performed but WAVM is much more lightweight and probably suits better for CI.
Regarding the artifact:
Opened an [issue](https://bugs.chromium.org/p/v8/issues/detail?id=10357) for V8. Right now it's unclear whether it's a corner case produced at LLVM IR level or a V8 implementation error. I'm inclined to say at this point that it's a V8 issue as I performed WASM target IR compilation to x86 by overriding the target triple in llc and the results were normal, also WAVM produces normal results.
```console
## Compile WASM target IR to x86 by overriding target triple
ispc runner.ispc -O3 --target=wasm-i32x4 --emit-llvm -o ispc.bc && \
llc --mtriple=x86_64-unknown-linux-gnu -filetype=obj ispc.bc -o ispc.o
```
Right now it works as no WASM specific intrinsics are emitted by ISPC and luckily no 32-bit specific optimizations are performed that'd break the compatibility. The code at question is not super complex just a lot of nested loops and inlining.

dbabokin:
[...]

PR #1662, opened 2020-03-03
Enable experimental Web Assembly support

aschrein:
This patch introduces new
* ispc target: 'wasm-i32x4'
* target OS: 'web'
* target architecture: 'wasm32'
---
Questions:
* Minimal LLVM version
  * Tested on 11.0.0
* Execution environment for testing
  * v8 has CLI
* Runtime library/wrappers
  * Freestanding is a viable option but sometimes LLVM could generate calls to memset and etc.
  * How users load modules? How to pass arguments to wasm-ld(--no-entry etc.) 
* Performance analysis/tracking - v8 has linux perf integration 
---
A way to make wasm objects using this patch:
```console
ispc $NAME.ispc --target=wasm-i32x4 --nostdlib --emit-llvm-text -o $NAME.ll
llc -O3 -filetype=obj $NAME.ll -o $NAME.o
wasm-ld --no-entry --export-all --allow-undefined -o $NAME.wasm $NAME.o
```
Then a wasm blob could be loaded in Javascript:
```js
// test.js
const bytes = read('test.wasm', 'binary');
(new WebAssembly.Module(bytes));
// EOF
```
x86 jitted disassembly could be obtained from debug v8 build(which could be installed using https://github.com/GoogleChromeLabs/jsvu):
```console
v8-debug --experimental-wasm-simd --print-wasm-code --no-liftoff --single-threaded --no-debug-code test.js
```


builtins/target-wasm32-i32x4.ll
```diff
@@ -0,0 +1,139 @@
+;;  Copyright (c) 2010-2019, Intel Corporation
```
cloudhan: 2020

dbabokin:
I've pushed commit to this pull request, which enables WebAssembly support in LLVM built by alloy.py. My experiments show that ISPC with WebAssembly support builds fine with LLVM 7.0+, but not with 6.0.

dbabokin:
Question about the architecture name `wasm32`. Why does it have suffix `32`? Is there `64`, at least theoretically?

src/builtins.cpp
```diff
@@ -940,11 +940,11 @@ void DefineStdlib(SymbolTable *symbolTable, llvm::LLVMContext *ctx, llvm::Module
     // Unlike regular builtins and dispatch module, which don't care about mangling of external functions,
     // so they only differentiate Windows/Unix and 32/64 bit, builtins-c need to take care about mangling.
     // Hence, different version for all potentially supported OSes.
-
-    const BitcodeLib *builtins = g->target_registry->getBuiltinsCLib(g->target_os, g->target->getArch());
-    Assert(builtins);
-    AddBitcodeToModule(builtins, module, symbolTable, warn);
-
+    if (g->target_os != TargetOS::web) {
```
dbabokin: Why not executing AddBitcodeToModule(), if you need `--nostdlib` anyway (at least before the target file has complete implementation of all require intrinsics)?

cmake/GenerateBuiltins.cmake
```diff
@@ -253,6 +253,10 @@ function (generate_target_builtins resultList)
             endforeach()
         endforeach()
     endforeach()
+    if (WASM_ENABLED)
```
dbabokin: Other targets are added in `CMakeFile.txt `in the project root dir (search for ISPC_TARGETS) and `cmake/GenerateBuiltins.cmake` has logic to treat them differently if needed. Could you please follow the same pattern?

dbabokin:
Also, is there a way to do print in WebAssembly? (at least when run in terminal)

aschrein:
#### Is there 64, at least theoretically?
AFAIK there are plans for it in the future. Also LLVM [calls it wasm32](https://github.com/llvm/llvm-project/search?q=wasm32&unscoped_q=wasm32)
#### Is there a way to do print in WebAssembly?
```c++
extern "C"
{
void println(char *str);
void print_hello() { println("hello from c++!"); }
}
```
is compiled into this:
```lisp
(module
 (memory $0 256 256)
 (data (i32.const 1024) "hello from c++!\00")
 (import "env" "println" (func $println (param i32 i32)))
 (export "print_hello" (func $print_hello))
 (func $print_hello
  (call $println
   (i32.const 1024)
   (i32.const 16)
  )
 )
)
```
Note that the module declares "env" "println" as imports.
In javascript:
```js
var memory = null;
function println(offset, length) {
    var bytes = new Uint8Array(memory.buffer, offset, length);
    var string = new TextDecoder('utf8').decode(bytes);
    console.log(string);
}
var importObject = {
    env: {
        println: println
    }
};
let bytes = read('test.wasm', 'binary');
let module = new WebAssembly.Module(bytes);
// Get exported heap from wasm module
// Note that once it's resized the object may change
// That's why emscripten_notify_memory_growth may be needed
memory = module.exports.memory;
let instance = new WebAssembly.Instance(module, importObject);
```
importObject delivers symbols to wasm module. In that way pretty much anything from stdlib could be implemented. By default wasm module exports a [WebAssembly.Memory](https://emscripten.org/docs/api_reference/module.html#Module.wasmMemory) object which could be resized at runtime. This object is a javascript array and everything that wasm does to memory is reflected on that object. Emscripten basically implements malloc in c++ which is then compiled to wasm and merged with user's sources. Wasm has some [intrinsics for memory management](https://github.com/llvm/llvm-project/blob/d0a767608773cfa3d8644c27796b63ff40962806/llvm/test/CodeGen/WebAssembly/memory-addr32.ll) Emscripten generates 'memory.grow' and 'memory.size' instructions.

src/builtins.cpp
aschrein: builtins.c is not compiled for wasm in this patch so Assert(builtins); fires. Maybe use Emscripten to compile it?

cmake/GenerateBuiltins.cmake
aschrein: It doesn't play well when I add "web" to the TARGET_OS_LIST_FOR_LL. It starts generating all sse, avx .ll files for  web.

src/builtins.cpp
aschrein: Is it possible to redistribute ispc with some version of Emscripten implementation of libc? It could be compiled to wasm or LLVM IR modules and linked against at the last stage.

dbabokin: Main question if emscripten license is compatible with BSD license. If it is, we can do that. Though, I'd like to avoid the dependency if possible. Probably we can have a separate implementation of print() in builtin.c for WASM, it also might be in form of LLVM IR (result of emscripten compiler). So we need a little more data before we can make an informed decision.

cmake/GenerateBuiltins.cmake
dbabokin: I see, let me try play with it and see how we can solve it.
And just for the background of UNIX/WINDOWS 32/64 macros which go to target .ll preprocessing - they were introduced solely to have proper versions of new/delete operator (they map to aligned malloc/free, which are different on UNIX / WINDOWS and sensitive to 32/64 bits). For the rest of the code of target library we don't care about these differences. So adding WEB OS to this list makes perfect sense to me, as memory allocation will be different.

src/builtins.cpp
dbabokin: I've added empty builtins.c for WASM, so it's a placeholder for future implementation.

For specific implementation we can decide after this PR is merged.

cmake/GenerateBuiltins.cmake
dbabokin: I've pushed commit, which implementation that I had in mind, when I commented that. Please have a look and check if it looks good to you.

aschrein:
The last patch works for my tests.
Runtime testing requires v8(or other simd enabled JS engine). Is it difficult to add the dependency on that to add an example to the codebase?
Lint tests are probably not required yet as there's no use of wasm specific intrinsics and the generated code is mostly identical to that of sse2-i32x4 target(except for some mask handling).
I also think that some performance investigation work is needed as simd support is relatively new in JS engines and from my observations code gen is far from being perfect. I compared ispc->sse2 vs ispc->wasm->v8->sse2 and there's plenty of opportunities for performance with no clear source of inefficiency(ispc?llvm?v8?). Theoretically there could be no gap for simple nested-loop-like benchmarks(except for memory manipulations/indirect calls/unstructured control flows etc) between native and wasm implementations as both are compiled anyways.

dbabokin:
I think the grand-plan for enabling should look like the following:
1.  Add target and basic infrastructure (this PR)
2. Complete target-wasm-i32x4.ll, so that it has complete set of required library functions (possibly just stubs without real implementation). Hence programs will be able to compile without `--nostdlib` and will fail only when specific functionality is missing.
3. Add test driver, similar to test_static.cpp, and support in run_tests.py. At this step we might want a script to install V8 dependency.
4. Add CI support - I expect that it going to be pretty easy.
5. Look at code and this about code generation.

For the code generation quality, I would expect that major difference is mask type definition - you use i1, while sse2/4 use i32. i32 mask is way more efficient for masking vectors and doing blending on sse2/4 and avx1/2.

I think this PR is more or less ready to be merge, with one minor issue:
- I need to exclude wasm folder in example from compiler package, that we build by default (I'll push the fix for it later today)

Also, yesterday I've pushed to the master the fix for the `alloy.py` script, which enables WebAssembly support . So docker files with pre-built LLVM for CI should be updated soon and we can start adding to CI tasks.

After closing this PR, we can continue design discussions in #1583

dbabokin:
I missed that this PR doesn't contain WASM examples, so this PR is good to be merged.

Lectem:
Would it be possible to enable webasm for binaries based on master ?
It would make it easier to test the feature on Windows.

I wanted to update the appveyor build script for that but didn't find where binaries are published, is it a manual process ?

dbabokin:
Binaries are published as artifacts in Appveyor.
So Appveyor files need to be changed to enabled WASM build. I'll do that.

PR #1671, opened 2020-03-09
Small fixes in the user's guide

johshoff:
Here are a few errors I found reading through the user's guide. (Great job on the guide, btw!)


dbabokin:
Thanks for the contribution! These are all good catches!

PR #1672, opened 2020-03-09
Doc fixes exclusive scan

johshoff:
Another documentation fix. This one is a little less straight forward, but I had to run examples to be sure how it worked.


dbabokin:
Good catch. Thanks!

Issue #1673, opened 2020-03-10
In LLVM 11.0+ optimization breaks LLVMVectorIsLinear


aschrein:
[This optimization](https://github.com/llvm/llvm-project/blob/1badf7c33a5d01900c77646f750e2ea11ad8bf5a/llvm/lib/Transforms/InstCombine/InstCombineAddSub.cpp#L1313) is triggered on LLVM 11.
```c++
// A+B --> A|B iff A and B have no bits set in common.
if (haveNoCommonBitsSet(LHS, RHS, DL, &AC, &I, &DT))
  return BinaryOperator::CreateOr(LHS, RHS);
```
```asm
%iter_val24 = add <4 x i32> %smear_counter23, <i32 0, i32 1, i32 2, i32 3>
; is transformed into
%iter_val24 = or <4 x i32> %smear_counter23, <i32 0, i32 1, i32 2, i32 3>
```
And lVectorIsLinear does not handle bop->getOpcode() == llvm::Instruction::Or which leads to gathers being generated.
In LLVM proving that two values have no bits in common involves quite a lot of code. I propose just changing the order of optimizations as a temporal remedy:
```c++
// At src/opt.cpp:517
// From
       if (g->opt.disableGatherScatterOptimizations == false && g->target->getVectorWidth() > 1) {
            optPM.add(llvm::createInstructionCombiningPass(), 210);
            optPM.add(CreateImproveMemoryOpsPass());
        }
// To
        if (g->opt.disableGatherScatterOptimizations == false && g->target->getVectorWidth() > 1) {
            optPM.add(CreateImproveMemoryOpsPass());
            optPM.add(llvm::createInstructionCombiningPass(), 210);
        }
```

aschrein:
>> In LLVM proving that two values have no bits in common involves quite a lot of code
Actually this code is exported and could be reused. I'll look if we could leverage it.

DeepakRajendrakumaran:
https://github.com/DeepakRajendrakumaran/ispc/commits/llvm11 fixes a few other cases. This si not the final change and still needs some addition. Will update issue here once change is final.

dbabokin:
@DeepakRajendrakumaran is #1683 fixing all known issues? I think we can close this issue.

DeepakRajendrakumaran:
Yep..this particular issue should be fixed. I did not see any warnings as of last week.

PR #1674, opened 2020-03-10
Fix for A+B-->A|B breaking gather elimination in LLVM 11+

aschrein:
Fix for #1673

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -334,7 +334,7 @@ static void lPrint(const char *type, bool isError, SourcePos p, const char *fmt,
     // exact same error message.  If so, return, so we don't redundantly
     // print it and annoy the user.
     static std::set<std::string> printed;
-    if (isError && printed.find(formattedBuf) != printed.end()) {
+    if (printed.find(formattedBuf) != printed.end()) {
         free(errorBuf);
         free(formattedBuf);
         return;
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -334,7 +334,7 @@ static void lPrint(const char *type, bool isError, SourcePos p, const char *fmt,
     // exact same error message.  If so, return, so we don't redundantly
     // print it and annoy the user.
     static std::set<std::string> printed;
-    if (printed.find(formattedBuf) != printed.end()) {
+    if (isError && printed.find(formattedBuf) != printed.end()) {
         free(errorBuf);
         free(formattedBuf);
         return;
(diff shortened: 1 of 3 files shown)


aschrein:
Probably it makes sense to enable this patch for all LLVM versions as there's no test tasks for 11

src/util.cpp
```diff
@@ -334,7 +334,7 @@ static void lPrint(const char *type, bool isError, SourcePos p, const char *fmt,
     // exact same error message.  If so, return, so we don't redundantly
     // print it and annoy the user.
     static std::set<std::string> printed;
-    if (printed.find(formattedBuf) != printed.end()) {
+    if (isError && printed.find(formattedBuf) != printed.end()) {
```
dbabokin: Why do need that?

We don't want to duplicate both errors and warnings.

dbabokin:
The fix looks reasonable, but running it with trunk I don't see improvements. I.e. all build of examples is still producing warnings and "make check-all" yields 4 fails.

Can you also add a lit test?

DeepakRajendrakumaran:
Some additional notes-

This and it’s following changes in LLVM is what seem to cause it - https://reviews.llvm.org/D73435

The patch fixes some cases. But there are others which still remains

The case below is an example where an add got converted to OR. Something like this is fixed

varying float
IntersectLightsWithTileMinMax(uniform float light_positionView_z_array[],uniform int32 numLights)
{
        float light_positionView_z;
       foreach (lightIndex = 0 ... numLights)
              light_positionView_z = light_positionView_z_array[lightIndex];

      return light_positionView_z;
}

src/util.cpp
aschrein: This routine is used for debug output. I think we do need debug messages whether they're duplicate or not.

aschrein:
> "make check-all" yields 4 fails

Those tests fail under LLVM 11 independent of this patch. No fails with LLVM 9.

> all build of examples is still producing warnings

I've been looking at this simple example:
```c++
```

src/util.cpp
dbabokin: It's used for error, warnings (including performance warnings), and debug prints.

We definitely don't want duplication for errors and warnings. Debug prints should allow duplication.

The reason for duplicated errors and warnings is that we duplicate extern functions (masked and unmasked version) and they trigger the same messages, which is very annoying and confusing.

aschrein: Ok, I removed this change. It requires some extra efforts.

PR #1675, opened 2020-03-11
Doc fixes for FAQ and perfguide

johshoff:
See comments on individual changes.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -451,7 +451,7 @@ For example, if you run the command:
    ispc foo.ispc -o foo.o --target=sse2,sse4-x2,avx-x2
 
 Then four object files will be generated: ``foo_sse2.o``, ``foo_sse4.o``,
-``foo_avx.o``, and ``foo.o``.[#]_  Link all of these into your executable, and
+``foo_avx.o``, and ``foo.o`` [#]_. Link all of these into your executable, and
 when you call a function in ``foo.ispc`` from your application code,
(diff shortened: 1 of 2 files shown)


docs/faq.rst
```diff
@@ -451,7 +451,7 @@ For example, if you run the command:
    ispc foo.ispc -o foo.o --target=sse2,sse4-x2,avx-x2
 
 Then four object files will be generated: ``foo_sse2.o``, ``foo_sse4.o``,
-``foo_avx.o``, and ``foo.o``.[#]_  Link all of these into your executable, and
```
johshoff: This change is really about inserting a space before `[#]_` so it will be parsed correctly. Moving the `.` is just a minor style change that could be left out.

docs/perfguide.rst
```diff
@@ -341,7 +341,7 @@ the "simple" array indexing calculation below:
 
 Since the index ``i`` is a varying value, the program instances in the gang
 will in general be reading different locations in the array ``x``.  Because
-current CPUs have a "gather" instruction, the ``ispc`` compiler has to
+not all CPUs have a "gather" instruction, the ``ispc`` compiler has to
```
johshoff: I'm thinking specifically of `vgatherdps` and friends here.

dbabokin:
Thanks for fixing it. Also note, that perf guide and faq were not updated too long and probably need more extensive update.

johshoff:
I see. Even if they were somewhat outdated, they were really helpful. Thanks!

PR #1680, opened 2020-03-13
WIP: Add wasm example

aschrein:
An example that uses emscripten to link against their libc and to link the executable. It's rather ugly at this stage. I don't know how to map the Makefile to the usual ispc CMakelists.txt structure for examples.
This PR's priority is to start a discussion rather than being merged.
The reason for the acrobatics around builtins.c is because emscripten linkage model uses WASM-Objects rather than LLVM IR that ispc uses.
Note that you need emcc(Emscripten SDK) installed.


dbabokin:
I've added a change, which installs ``wasm`` folder in the examples, only if WASM is enabled.

Integrating CMake to examples' CMake doesn't seem to be a big deal. ``examples/wasm/CMakefile.txt`` should take care about finding the tools, like ``emcc``. It also include sub directories with specific examples, like hello_world. It shouldn't be a big deal, I think. I can help with that, but I need to get handy with Emscripten SDK first.

I also suggest adding README.md to ``examples/wasm`` with description of prerequisites and compile instructions.

dbabokin:
@aschrein I've rebased this branch.
``examples/wasm/Makefile`` needs to be updated.
We may abandon this pull request, or use it as a base to add the examples, that can now be compiled.

dbabokin:
I'm closing this PR, as there it seems to be outdated. Please submit new PR if there's an interest in adding WASM example to the project.

PR #1694, opened 2020-03-22
Fixes for static analysis tool remarks

dbabokin:


src/module.cpp
```diff
@@ -815,18 +815,18 @@ bool Module::writeOutput(OutputType outputType, OutputFlags flags, const char *o
     if (diBuilder && (outputType != Header) && (outputType != Deps))
         lStripUnusedDebugInfo(module);
 
+    // SIC! (verifyModule() == TRUE) means "failed", see llvm-link code.
+    if ((outputType != Header) && (outputType != Deps) && (outputType != HostStub) && (outputType != DevStub) &&
+        module && llvm::verifyModule(*module))
```
aneshlya: We should do verifyModule after adding debug flag to module.

PR #1695, opened 2020-03-22
[WASM] Experimental stdlib build

aschrein:
This patch allows building a wider range of sources for WASM target and adds emscripten based  build command for builtins.c. This is enough to build volume_rendering/rt/volume_rendering examples with help of emscripten and slight modifications to tasksys.cpp.
```console
ispc ..deferred/kernels.ispc --math-lib=system  --opt=fast-masked-vload --opt=fast-math --opt=force-aligned-memory \
           --emit-c++ -h kernels_ispc.h -O3 --target=wasm-i32x4 --emit-llvm-text -o kernels.ispc.ll && \
emcc -DWASM ..deferred/main.cpp  -I./ -O3 -s EXPORTED_FUNCTIONS='["_main"]' -c -o main.o && \
emcc -DWASM ..deferred/common.cpp  -I./ -O3 -s EXPORTED_FUNCTIONS='["_main"]' -c -o common.o && \
emcc -DWASM ..deferred/dynamic_c.cpp  -I./ -O3 -s EXPORTED_FUNCTIONS='["_main"]' -c -o dynamic_c.o && \
emcc -DWASM -DISPC_USE_PTHREADS -s USE_PTHREADS=1 ../tasksys.cpp   -I./  -O2 -s EXPORTED_FUNCTIONS='["_main"]' -c -o tasksys.o && \
llc -filetype=obj kernels.ispc.ll -o kernels.ispc.o && \
emcc -O3 main.o kernels.ispc.o common.o dynamic_c.o tasksys.o -s USE_PTHREADS=1 \
                    -s PTHREAD_POOL_SIZE=12 \
                    --preload-file ...deferred/data/pp1280x720.bin@/home/dummy/pp1280x720.bin \
                    -s TOTAL_MEMORY=268435456 -o index.html
```


dbabokin:
What is the correct way to install emscripten on Linux?
I've tried going ``apt-get install emscripten`` and then running ``emcc`` to initialize. After that it issue a warning when executed:
```
WARNING  root: LLVM version appears incorrect (seeing "(tags/RELEASE_700/final)", expected "3.4")
INFO     root: (Emscripten: Running sanity checks)
```

Is it expected?

cmake/GenerateBuiltins.cmake
```diff
@@ -230,16 +230,26 @@ function(builtin_to_cpp bit os_name arch supported_archs supported_oses resultFi
...
-        add_custom_command(
-            OUTPUT ${output}
-            COMMAND ${CLANG_EXECUTABLE} ${target_flags} -w -m${bit} -emit-llvm -c ${inputFilePath} ${wasm_define} -o - | (\"${LLVM_DIS_EXECUTABLE}\" - || echo "builtins.c compile error")
-                | \"${Python3_EXECUTABLE}\" bitcode2cpp.py c --type=builtins-c --runtime=${bit} --os=${os_name} --arch=${target_arch} --llvm_as ${LLVM_AS_EXECUTABLE}
-                > ${output}
-            DEPENDS ${inputFilePath} bitcode2cpp.py
-            WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+            add_custom_command(
+                OUTPUT ${output}
+                COMMAND ${EMCC_EXECUTABLE} -DWASM -s WASM_OBJECT_FILES=0 -c ${inputFilePath} -emit-llvm -c -o - -s EXPORTED_FUNCTIONS='["___wasm_do_print", "___wasm_clock"]'
```
dbabokin: You need to escape quotes, otherwise they are lost:
```
'[\"___wasm_do_print\", \"___wasm_clock\"]'
```

dbabokin: I've push the fix to your branch.

dbabokin: Not it builds for me locally, but Travis still has an error, though it formally passes. I guess error needs to be properly propagated.

aschrein:
> What is the correct way to install emscripten on Linux?

I followed the [instructions](https://emscripten.org/docs/getting_started/downloads.html):
```console
# Get the emsdk repo
git clone https://github.com/emscripten-core/emsdk.git
# Enter that directory
cd emsdk
# Fetch the latest version of the emsdk (not needed the first time you clone)
git pull
# Download and install the latest SDK tools.
./emsdk install latest
# Make the "latest" SDK "active" for the current user. (writes ~/.emscripten file)
./emsdk activate latest
# Activate PATH and other environment variables in the current terminal
source ./emsdk_env.sh
```

cmake/GenerateBuiltins.cmake
aschrein: Probably -s EXPORTED_FUNCTIONS='["___wasm_do_print", "___wasm_clock"]' is not even needed for LLVM IR output. It makes sense for WASM objects because it needs to have something for export. I was just thinking emscripten might put some attributes on the functions and if they're not exported it could be dead coded so I followed a safe path.

aschrein:
hmm. I wonder what is the right way to pick emscripten version. Installing latest on CI is definitely not the best idea. Emscripten SDK ships with it's own version of LLVM toolchain(11 for the latest) and the current script may tamper with paths on the global environment. We need the version to be stable and compatible on IR level with the LLVM that ISPC uses for compilation.

dbabokin:
That's very good question. We need to have ``emcc`` compatible with ``llvm-dis`` that we use and then compatible with LLVM version we use to consume the IR. It probably doesn't need to be exactly the same version. If it compiles and runs, it's very good sign that it's good enough match. Though, if we can let Emscripten use our LLVM, this is probably the best case for compatibility.

Do you know if LLVM_ROOT defines the LLVM that is used by Emscripten? If it is sufficient to switch it over to our version of LLVM, we just need to invoke ``emcc --version`` after adding our LLVM build to the ``PATH``.

Other than that, let's move ``install_emscripten.sh`` to ``scripts`` folder. We don't have all scripts moved there yet, but that's the idea.

aschrein:
I don't think the llvm version of emcc is controllable. The other path is to use wasm objects for linking. In the same way spirv is used to solve incompatible llvm versions in some tools. I already tried that approach and it works. Just build a dummy builtins.c with proxy functions and keep another builtins.c with implementation as a .wasm object. Then link it at the final stage. The downside is that it requires the user to explicitly feed builtins.o to the linker

builtins/builtins.c
```diff
@@ -73,7 +73,7 @@
 // #define _SC_NPROCESSORS_ONLN 58
 // long sysconf(int);
 #endif // !_MSC_VER
-
+#endif // WASM
```
dbabokin:
```suggestion
#endif // !WASM
```

dbabokin:
Changes look good to me.
@aschrein thanks for contribution!

Issue #1631, opened 2020-01-16
Missing const pointer overloads for aos_to_soaX


est77:
Sample code fragment:

```c
const uniform float* uniform src = src_pixels;
for (uniform unsigned int i = 0; i < width; i += programCount)
{
    float r, g, b, a;
    aos_to_soa4(&src[i], &r, &g, &b, &a);
}
```

fails to compile with the following error:

Error: Unable to find any matching overload for call to function "aos_to_soa4". 
Passed types: (const uniform float * uniform, varying float * uniform, varying float * uniform, varying float * uniform, varying float * uniform) 

casting the first argument to (uniform float* uniform) fixes the error.

DeepakRajendrakumaran:
There are a couple of issues here
1. Our algorithm for finding matching function doesn't correctly identify matching function with const pointer argument
2. Even after matching the function, our type conversion has trouble realizing this is a valid conversion.

Have an initial fix but needs to check different scenarios to ensure it doesn't break anything.

DeepakRajendrakumaran:
Closing this as fixed - https://github.com/ispc/ispc/commit/9422b4649ed87cfaf3c4b3e50c7f4bddbedaa756

Reopen if not.

PR #1699, opened 2020-03-24
Fix for #1631 - conversion between const and non-const types.

DeepakRajendrakumaran:


tests/lit-tests/1631.ispc
```diff
@@ -0,0 +1,44 @@
+//; RUN: %{ispc} %s -o %t.o --nowrap --target=avx2-i32x8 2>&1 | FileCheck %s
+//; CHECK: Warning: Converting from const pointer type "const uniform float * uniform" to pointer type "uniform float *
+//uniform" for function call argument discards const qualifier.
```
dbabokin, line 3: You can't split this to extra line, it's basically ignored.

DeepakRajendrakumaran: That was a mistake. The first line should be enough to check. Took out the 2nd part.

Issue #1659, opened 2020-02-20
extract/insert functions are missing for bool type


dbabokin:
We need the following functions in stdlib:
```
uniform bool extract(bool x, uniform int i);
bool insert(bool x, uniform int i, uniform bool v);
```

dbabokin:
Fixed by #1700

dbabokin:
Reopening - need to fix documentation. @DeepakRajendrakumaran

DeepakRajendrakumaran:
Just this - https://github.com/ispc/ispc/pull/1716 or something else needed?

dbabokin:
We can close now.

PR #1700, opened 2020-03-25
Fix for #1659 - adding bool implementation for insert and extract.

DeepakRajendrakumaran:


dbabokin:
The changes look good, but I prefer to put it on hold before we clear up definition of ``bool`` type.

Issue #1654, opened 2020-01-31
Need to limit maximum number of errors emitting by ISPC


dbabokin:

DeepakRajendrakumaran:
What exactly does this mean?

dbabokin:
An options similar to clang ``-ferror-limit=n``

dbabokin:
Fixed.

PR #1701, opened 2020-03-25
Issue #1654 : Adding option to limit number of errors emitted

DeepakRajendrakumaran:


src/main.cpp
```diff
@@ -118,6 +118,7 @@ static void lPrintVersion() {
     printf("    [--emit-llvm-text]\t\t\tEmit LLVM bitcode file as output in textual form\n");
     printf("    [--emit-obj]\t\t\tGenerate object file file as output (default)\n");
     printf("    [--force-alignment=<value>]\t\tForce alignment in memory allocations routine to be <value>\n");
+    printf("    [--ferror-limit=<value>]\t\tLimit maximum number of errors emitting by ISPC to <value>\n");
```
dbabokin: We don't use "f" prefix for options, I think it needs to be ``--error-limit``

PR #1703, opened 2020-03-25
Change definition of static linking

dbabokin:


CMakeLists.txt
```diff
@@ -370,14 +370,10 @@ else()
     set_target_properties(${PROJECT_NAME} PROPERTIES LINK_FLAGS "-z noexecstack -z relro -z now -Wl,--gc-sections")
 endif()
 
-if (ISPC_STATIC_STDCXX_LINK)
+if (ISPC_STATIC_STDCXX_LINK OR ISPC_STATIC_LINK)
     set_target_properties(${PROJECT_NAME} PROPERTIES LINK_FLAGS "-static-libgcc -static-libstdc++")
```
aneshlya: Here set_target_properties setting will rewrite LINK_FLAGS, not append to them. Consider using target_link_options instead.

.travis.yml
```diff
@@ -64,6 +64,9 @@ env:
 before_install:
   - sudo apt-get update
   - sudo apt-get install -y libc6-dev-i386 g++-multilib lib32stdc++6
+  - wget https://cmake.org/files/v3.17/cmake-3.17.0-Linux-x86_64.sh && mkdir /opt/cmake && sh cmake-3.17.0-Linux-x86_64.sh --prefix=/opt/cmake --skip-license && sudo ln -s /opt/cmake/bin/cmake /usr/local/bin/cmake && rm cmake-3.17.0-Linux-x86_64.sh
```
aneshlya: Please update in the same way Dockerfile for Ubuntu

aneshlya: BTW why do we use cmake-3.17 here when minimum required version is 3013?

dbabokin: It's already in Ubuntu Dockerfile, I've borrowed it from there. 3.17 is just because it's the latest.

PR #1712, opened 2020-04-13
Changing bool type in memory to i8 and <i8 x WIDTH>.

DeepakRajendrakumaran:


src/type.h
```diff
@@ -202,8 +202,9 @@ class Type {
         syntax. */
     virtual std::string GetCDeclaration(const std::string &name) const = 0;
 
-    /** Returns the LLVM type corresponding to this ispc type */
-    virtual llvm::Type *LLVMType(llvm::LLVMContext *ctx) const = 0;
+    /** Returns the LLVM type corresponding to this ispc type.
+     *  'isStorageType; should be set to 'true' if asking for storage LLVM type.*/
```
dbabokin, line 206: typo ; => '

src/type.cpp
```diff
@@ -1008,7 +1010,7 @@ llvm::Type *PointerType::LLVMType(llvm::LLVMContext *ctx) const {
             if (baseType->IsVoidType())
                 ptype = LLVMTypes::VoidPointerType;
             else
-                ptype = llvm::PointerType::get(baseType->LLVMType(ctx), 0);
+                ptype = llvm::PointerType::get(baseType->LLVMType(ctx, isStorageType), 0);
```
dbabokin, line 1013: I think it needs to be always storage type here, as pointer points to the storage (memory) by definition.

src/ctx.cpp
```diff
@@ -2072,7 +2091,19 @@ llvm::Value *FunctionEmitContext::LoadInst(llvm::Value *ptr, const char *name) {
...
 
     AddDebugPos(inst);
-    return inst;
+
+    llvm::Value *loadVal = inst;
+    // bool type is stored as i8. So, it requires some processing.
+    if ((type != NULL) && (type->IsBoolType())) {
+        if (CastType<AtomicType>(type) != NULL)
+            loadVal = SwitchBoolSize(loadVal, inst->getType(), type->LLVMType(g->ctx));
+        if ((CastType<VectorType>(type) != NULL) && (type->IsBoolType())) {
```
dbabokin, line 2100: I think "else if" here is more appropriate.
Also, please use {} even for single line if, it's less error prone.

src/ctx.cpp
```diff
@@ -2302,6 +2344,26 @@ llvm::Value *FunctionEmitContext::gather(llvm::Value *ptr, const PointerType *pt
...
+            // This is needed when array of bool is passed in from cpp side
+            // TRUE in clang is '1'. This is zero extended to i8.
+            // In ispc, this is uniform * varying which after gather becomes
+            // varying bool. Varying bool is ispc is '-1'. The most
+            // significant bit being set to 1 is important for blendv
+            // operations to work as expected.
+            if (ptrType->GetBaseType()->IsUniformType()) {
+                gatherCall = TruncInst(gatherCall, LLVMTypes::Int1VectorType);
+                gatherCall = SExtInst(gatherCall, llvmReturnType);
+            } else
```
dbabokin, line 2360: {} for single line else statement - for style uniformness.

src/ctx.cpp
```diff
@@ -2589,6 +2679,17 @@ void FunctionEmitContext::StoreInst(llvm::Value *value, llvm::Value *ptr) {
     llvm::PointerType *pt = llvm::dyn_cast<llvm::PointerType>(ptr->getType());
     AssertPos(currentPos, pt != NULL);
 
+    if ((ptrType != NULL) && (ptrType->IsBoolType())) {
+        if ((CastType<AtomicType>(ptrType) != NULL))
+            value = SwitchBoolSize(value, value->getType(), ptrType->LLVMStorageType(g->ctx));
+        if (CastType<VectorType>(ptrType) != NULL) {
```
dbabokin, line 2685: "else if" and {} around single line statement under if condition.

src/expr.cpp
```diff
@@ -3814,6 +3854,7 @@ std::pair<llvm::Constant *, bool> ExprList::GetConstant(const Type *type) const
                 return std::pair<llvm::Constant *, bool>(NULL, false);
             }
             llvm::Type *llvmType = elementType->LLVMType(g->ctx);
+            ;
```
dbabokin, line 3857: need to remove: ";"

src/expr.cpp
```diff
@@ -6442,16 +6507,28 @@ static llvm::Value *lUniformValueToVarying(FunctionEmitContext *ctx, llvm::Value
...
+                // If member is a uniform bool, it needs to be truncated to i1 since
+                // uniform  bool in IR is i1 and i8 in struct
+                // Consider switching to just a broadcast for bool
+                if ((elemType->IsBoolType()) && (CastType<AtomicType>(elemType) != NULL)) {
+                    v = ctx->TruncInst(v, LLVMTypes::BoolType);
+                }
+                v = lUniformValueToVarying(ctx, v, elemType, pos);
+                // If the extracted element if bool and varying needs to be
+                // converted back to i8 vector to insert into varying struct.
+                if ((elemType->IsBoolType()) && (CastType<AtomicType>(elemType) != NULL))
```
dbabokin, line 6529: {} for style uniformness.

PR #1713, opened 2020-04-14
Fixing assert condition : Reference type can also return valid bool value.

DeepakRajendrakumaran:
…alue


DeepakRajendrakumaran:
This was the crash - https://godbolt.org/z/cNAw29

PR #1714, opened 2020-04-16
SGEMM: fix MT version issues

jzielins:
This commit adds additional validation of input arguments (matrix sizes) to prevent errors in algorithms that require matrix size to be a multiple of SIMD width. It also modifies the mechanism of

### DIFF
--- a/examples/sgemm/SGEMM_kernels.ispc
+++ b/examples/sgemm/SGEMM_kernels.ispc
@@ -36,6 +36,14 @@
 
 #define TILE_SIZE 32
 
+export uniform int SGEMM_get_program_count() {
+    return programCount;
+}
+
+export uniform int SGEMM_get_tile_size() {
+    return TILE_SIZE;
+}
+
 // Naive implementation. The outer foreach achieves some parallelism.
 export void SGEMM_naive(uniform float matrixA[], uniform float matrixB[], uniform float matrixC[],
                         uniform int M, uniform int N, uniform int K) {
@@ -58,14 +66,15 @@ export void SGEMM_naive(uniform float matrixA[], uniform float matrixB[], unifor
     }
 }
 
+
 // Naive implementation with task parallelism added. The matrix rows are divided into chunks
 // which are calculated by separate tasks.
 static task void SGEMM_naive_task(uniform float matrixA[], uniform float matrixB[], uniform float matrixC[],
                                     uniform int M, uniform int N, uniform int K) {
     // Determine workset for this task instance:
     uniform unsigned int uNumRowsPerTask = M / taskCount;
-    uniform unsigned int uRowStart = uNumRowsPerTask * taskIndex;
-    uniform unsigned int uRowEnd = uRowStart + uNumRowsPerTask;
+    uniform unsigned int uRowStart       = uNumRowsPerTask * taskIndex;
+    uniform unsigned int uRowEnd         = uRowStart + uNumRowsPerTask;
 
     foreach (m = uRowStart ... uRowEnd) {
         uniform unsigned int n, k;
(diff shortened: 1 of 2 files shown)


examples/sgemm/SGEMM_kernels.ispc
```diff
@@ -85,8 +94,9 @@ static task void SGEMM_naive_task(uniform float matrixA[], uniform float matrixB
 
 export void SGEMM_naive_withTasks(uniform float matA[], uniform float matB[], uniform float matC[],
                                     uniform int M, uniform int N, uniform int K) {
-    // Generate more tasks than threads, to enable some load balancing:
-    uniform int numTasks = num_cores() * 4;
+    // Task must have at least programCount columns to process, but we require the matrix
+    // K size to be a miltiple of programCount, so it's safe to divide here
+    uniform int numTasks = M / programCount;
```
dbabokin, line 99: Either comment or code is wrong. Comment talks about K, code uses M.

examples/sgemm/SGEMM_kernels.ispc
```diff
@@ -403,7 +412,7 @@ static task void SGEMM_tileNoSIMDIntrin_task(uniform float matrixA[], uniform fl
 export void SGEMM_tileNoSIMDIntrin_withTasks(uniform float matA[], uniform float matB[], uniform float matC[],
     uniform int M, uniform int N, uniform int K) {
     // Generate more tasks than threads, to enable some load balancing:
-    uniform int numTasks = num_cores()*4;
+    uniform int numTasks = M / programCount;
```
dbabokin, line 415: Some use ``K / programCount``, some use ``M / programCount``. I think it requires a note in comments to explain this.

examples/sgemm/SGEMM_main.cpp
```diff
@@ -231,7 +232,18 @@ int main(int argc, char **argv) {
...
         printf("%s\n", argv[0]);
         printf("\nInvalid number of inputs\n");
-        getchar();
+        exit(-1);
+    }
+
+    int programCount = SGEMM_get_program_count();
+    int tileSize = SGEMM_get_tile_size();
+    if (K % programCount != 0 || K % tileSize != 0) {
+        printf("\nNumber of columns in Matrix B must be a multiple of %d and %d!\n", programCount, tileSize);
```
dbabokin, line 241: Let's have additional clarification what these numbers are. i.e.
``"...of %d (target width) and %d (tile size)!\n"``

examples/sgemm/SGEMM_main.cpp
```diff
@@ -231,7 +232,18 @@ int main(int argc, char **argv) {
...
+        exit(-1);
+    }
+
+    if (M % programCount != 0) {
+        printf("\nNumber of rows in Matrix A must be a multiple of %d!\n", programCount);
```
dbabokin, line 246: Same here:
``"... of %d (target width)!\n"``

examples/sgemm/SGEMM_main.cpp
```diff
@@ -243,7 +255,7 @@ int main(int argc, char **argv) {
     init_matrix_rand(matrixB, N, K, 10.0f);
     float *matrixC;
     matrixC = (float *)malloc(M * K * sizeof(float));
-    init_matrix_rand(matrixC, M, K, 0.0f);
```
dbabokin, line 246: why is this change?

examples/sgemm/SGEMM_kernels.ispc
jzielins: Ctrl-C, Ctrl-V error, fixed.

jzielins: Ctrl-C, Ctrl-V error, fixed.

examples/sgemm/SGEMM_main.cpp
jzielins: Done.

examples/sgemm/SGEMM_main.cpp
jzielins: Code cleanup. I thought that using rand with limit 0.0f is a little bit of overkill.  init_matrix(..., value) seems more logical here, although functionally it's the same thing. Additionally, the init_matrix() function was not used in the code, so we should either remove or use it.

examples/sgemm/SGEMM_main.cpp
jzielins: Good point, fixed

examples/sgemm/SGEMM_kernels.ispc
jzielins: True. Fixed.

examples/sgemm/SGEMM_main.cpp
dbabokin: Ah, you are right! We definitely need to use init_matrix() here.

PR #1718, opened 2020-04-20
[WASM] Add v8 install script to CI

aschrein:
* Add v8 install script to CI
* Fix LLVM 11.0(trunk) build

Unfortunately LLVM trunk had some changes that prevent building ISPC so I had to fix some of the issues. The fixes are ad-hoc and it'd be better to wrap changed API in a separate function that handles all the differences.
If the initial version doesn't break the CI I'll push WASM testing to this or a new PR

### DIFF
--- a/scripts/install_v8.sh
+++ b/scripts/install_v8.sh
@@ -2,6 +2,5 @@ WD=`pwd`
 sudo apt install nodejs && \
 npm install jsvu -g && \
 export PATH="${HOME}/.jsvu:${PATH}" && \
-jsvu --os=linux64 --engines=v8-debug && \
-alias d8=v8-debug
+jsvu --os=linux64 --engines=v8-debug
 cd $WD
--- a/scripts/install_v8.sh
+++ b/scripts/install_v8.sh
@@ -1,5 +1,5 @@
 WD=`pwd`
-sudo apt install nodejs && \
+sudo apt-get install nodejs && \
 npm install jsvu -g && \
 export PATH="${HOME}/.jsvu:${PATH}" && \
 jsvu --os=linux64 --engines=v8-debug
--- a/scripts/install_v8.sh
+++ b/scripts/install_v8.sh
@@ -5,8 +5,8 @@ fetch v8 &&  \
 cd v8 &&  \
 gclient sync &&  \
 ./build/install-build-deps.sh &&  \
-./tools/dev/v8gen.py x64.optdebug &&  \
-ninja -C out.gn/x64.optdebug &&  \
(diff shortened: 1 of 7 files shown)


aschrein:
CI build of V8 ran out of time. Will try another compile command

aschrein:
No luck with building v8 from sources. I'll try jsvu

scripts/install_v8.sh
```diff
@@ -0,0 +1,6 @@
+WD=`pwd`
+sudo apt install nodejs && \
```
dbabokin, line 2: Using ``apt-get`` is more conventional way to install something. In latest distributions they come together, but in older ones that might be an issue.

src/cbackend.cpp
```diff
@@ -4857,8 +4857,15 @@ llvm::Value *SmearCleanupPass::getShuffleSmearValue(llvm::Instruction *inst) con
 
     // Check that the shuffle is a broadcast of the element of the first vector,
     // i.e. mask vector is vector with equal elements of expected size.
-    if (!(mask && (mask->isNullValue() || (shuffleInst->getMask()->getSplatValue() != 0)) &&
-          llvm::isa<llvm::VectorType>(mask->getType()) &&
+
+    if (!(mask &&
+          (mask->isNullValue() ||
+#if ISPC_LLVM_VERSION == ISPC_LLVM_11_0
```
dbabokin, line 4863: ``ISPC_LLVM_VERSION == ISPC_LLVM_11_0`` needs to be ``ISPC_LLVM_VERSION >= ISPC_LLVM_11_0``, as it will suddenly break when the trunk switch to 12.0.
Same for other instances this check.

src/opt.cpp
```diff
@@ -1269,8 +1275,14 @@ static llvm::Value *lGetBasePtrAndOffsets(llvm::Value *ptrs, llvm::Value **offse
     // Looking for %gep_offset = shufflevector <8 x i64> %0, <8 x i64> undef, <8 x i32> zeroinitializer
     llvm::ShuffleVectorInst *shuffle = llvm::dyn_cast<llvm::ShuffleVectorInst>(ptrs);
     if (shuffle != NULL) {
+#if ISPC_LLVM_VERSION == ISPC_LLVM_11_0
+        llvm::Value *indices = shuffle->getShuffleMaskForBitcode();
+        llvm::Value *vec = shuffle->getOperand(1);
```
dbabokin, line 1280: This line remans the same, no reason to keep it under ifdef.

dbabokin:
Sorry for the delay with merge. Changes look good to me, but I'd like to postpone the merge until we release v1.13. This should happen within a day or two.

PR #1721, opened 2020-04-20
Link with shared benchmark's lib

aneshlya:


Lectem:
Hi, I'm looking into running benchmarks on a machine other than the one doing the build.
However the changes of this PR assume that the user necessarily has the google benchmark  so installed.

This means that I can not fetch the build artifacts and run the benchmarks in it.
Is there a specific reason for using the shared version of the library instead of the static one ? (size ? performance impact?)

dbabokin:
License. If linking statically, we would need to include it in our license / bill of materials. The license for Google Benchmark is not restrictive and compatible with ours, but it would require some paperwork. I think we should add it.

Issue #1719, opened 2020-04-20
Division by zero leads to ICE


Vsevolod-Livinskij:
ISPC crashes on valid code:
```c
export void test(uniform int64 var_7) {
    var_7 = -25 % 0;
}
```

Error:
> ispc func.ispc --target=avx2-i32x8
/ispc/src/main.cpp(360): FATAL ERROR: Unhandled signal sent to process; terminating.

ISPC commit: 9422b4649ed87cfaf3c4b3e50c7f4bddbedaa756
LLVM version 10.0.0

dbabokin:
backtrace:
```
  * frame #0: 0x00000001000875a4 ispc`BinaryExpr::Optimize() + 28836
    frame #1: 0x000000010003a356 ispc`WalkAST(ASTNode*, bool (*)(ASTNode*, void*), ASTNode* (*)(ASTNode*, void*), void*) + 1414
    frame #2: 0x0000000100039f3f ispc`WalkAST(ASTNode*, bool (*)(ASTNode*, void*), ASTNode* (*)(ASTNode*, void*), void*) + 367
    frame #3: 0x0000000100039e9f ispc`WalkAST(ASTNode*, bool (*)(ASTNode*, void*), ASTNode* (*)(ASTNode*, void*), void*) + 207
    frame #4: 0x000000010003a255 ispc`WalkAST(ASTNode*, bool (*)(ASTNode*, void*), ASTNode* (*)(ASTNode*, void*), void*) + 1157
    frame #5: 0x00000001000b4727 ispc`Function::Function(Symbol*, Stmt*) + 247
    frame #6: 0x0000000100039c2e ispc`AST::AddFunction(Symbol*, Stmt*) + 62
    frame #7: 0x000000010000c59c ispc`yyparse() + 40508
    frame #8: 0x00000001000cc1ab ispc`Module::CompileFile() + 299
    frame #9: 0x00000001000d5479 ispc`Module::CompileAndOutput(char const*, Arch, char const*, std::__1::vector<ISPCTarget, std::__1::allocator<ISPCTarget> >, Module::OutputFlags, Module::OutputType, char const*, char const*, char const*, char const*, char const*, char const*, char const*) + 233
    frame #10: 0x00000001000c9596 ispc`main + 9254
    frame #11: 0x00007fff6c33ccc9 libdyld.dylib`start + 1
    frame #12: 0x00007fff6c33ccc9 libdyld.dylib`start + 1
```

PR #1745, opened 2020-05-03
Fix for #1719.

DeepakRajendrakumaran:


src/expr.cpp
```diff
@@ -2034,12 +2034,21 @@ static ConstExpr *lConstFoldBinaryIntOp(BinaryExpr::Op op, const T *v0, const T
...
-        FOLD_OP_REF(BinaryExpr::Mod, %, TRef);
         FOLD_OP_REF(BinaryExpr::Shl, <<, TRef);
         FOLD_OP_REF(BinaryExpr::Shr, >>, TRef);
         FOLD_OP_REF(BinaryExpr::BitAnd, &, TRef);
         FOLD_OP_REF(BinaryExpr::BitXor, ^, TRef);
         FOLD_OP_REF(BinaryExpr::BitOr, |, TRef);
+    case BinaryExpr::Mod:
+        for (int i = 0; i < count; ++i) {
+            if (v1[i] == 0) {
+                Warning(pos, "Division by zero encountered in expression.");
```
dbabokin, line 2045: It's not division, clang gives the following message: "warning: remainder by zero is undefined".
For division it has: "division by zero is undefined".

src/expr.cpp
```diff
@@ -2086,10 +2095,11 @@ static ConstExpr *lConstFoldBinaryArithOp(BinaryExpr::Op op, const T *v0, const
     case BinaryExpr::Div:
```
dbabokin, line 2095: Div is handled here, while mod in lConstFoldBinaryIntOp(). This is inconsistent. This needs to be handle in the same place (either of them).

src/expr.cpp
DeepakRajendrakumaran: Whoops..will change it - I checked with gcc : https://godbolt.org/z/WYMXnn .
Clang was different - https://godbolt.org/z/D9M9tw

src/expr.cpp
DeepakRajendrakumaran: That'd by design
lConstFoldBinaryIntOp : Constant fold the binary integer operations that aren't also applicable  to floating-point types. So, Mod goes here

lConstFoldBinaryArithOp - For those applicable for float. Hence div

src/expr.cpp

src/expr.cpp
dbabokin: Ok, looks good.

PR #1750, opened 2020-05-06
[WASM] Enable tests

aschrein:
V8 is switching to new simd binary opcodes (https://github.com/v8/v8/commit/41fbbd12a35d4f7bb89859c91da4b7b88e9d2d74) due to this PR (https://github.com/WebAssembly/simd/pull/209) so versions of emscripten and v8 have to be aligned to the recent compatible for now. Later the versions would have to be bumped manually.

expected failures:
* tests/clock.ispc - the kernel is too fast for low precision timer available for web 

To run:
```console
python3 run_tests.py --arch=wasm32 --target=wasm-i32x4
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,7 +64,7 @@ before_install:
   - sudo apt-get update
   - sudo apt-get install -y libc6-dev-i386 g++-multilib lib32stdc++6
   - if [ -n "$WASM_FLAGS" ]; then source scripts/install_emscripten.sh && emcc --version; fi
-  - if [ -n "$WASM_FLAGS" ]; then source scripts/install_v8.sh && v8-debug -e "console.log(\"V8 TEST\")"; fi
+  - if [ -n "$WASM_FLAGS" ]; then source scripts/install_v8.sh && v8 -e "console.log(\"V8 TEST\")"; fi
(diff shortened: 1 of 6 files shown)


aschrein:
AppVeyor fail:
```console
Downloading winflexbison3 
  from 'https://sourceforge.net/projects/winflexbison/files/old_versions/win_flex_bison-2.5.18.zip'
ERROR: The remote file either doesn't exist, is unauthorized, or is forbidden for url 
```
???
doesn't seem reasonable. I'll try to trigger another build.

dbabokin:
> doesn't seem reasonable. I'll try to trigger another build.

I saw the same yesterday, looks like some connection problem.

Do I understand correctly, that with LLVM 10.0 we will stay on old version of ``emcc`` and with LLVM trunk we'll need to switch back to ``emcc`` trunk again?

Also, looks like I'll need to change #1749 as well.

aschrein:
> Do I understand correctly, that with LLVM 10.0 we will stay on old version of emcc and with LLVM trunk we'll need to switch back to emcc trunk again?

Once emcc/LLVM catches up with new wasm encodings v8 would have to be bumped and probably wasm binaries produced by LLVM 10 will not work anywhere as is. There's an option with  binary translation which is not that complicated though. As of now llvm trunk still spits old [encodings](https://github.com/llvm/llvm-project/blob/ac00376a13e9b637a8bc6b9e2ae6fd668e01b0f9/llvm/test/MC/WebAssembly/simd-encodings.s#L88) . `f32x4.splat` [0xfd,0x12] is old [0xfd, 0x13] is new

PR #1753, opened 2020-05-11
Dockerfiles uses required CMake version

robertmaynard:
As ISPC now requires CMake 3.13 make sure the docker image has the correct version.

Additionally update to download from the github release page as it has better download speeds.


robertmaynard:
I just noticed that the v1.13 release used CMake 3.14, so do you mind using a newer version or do you want your docker images to use the minimum required?

dbabokin:
I don't have strong preference, but minimum required is probably better, because it helps verifying that minimum required used in CMakeLists.txt is the correct one.

dbabokin:
Thanks for contribution!

Issue #1638, opened 2020-01-22
@<filename> doesn't support quotes arguments


Unit2Ed:
To workaround issues with command-lines being too long, I'm using @<filename>-style compiler invocation. Unfortunately this doesn't work well with arguments that contain spaces - eg `-I`.

This is my reduced sample that demonstrates the problem:

```
"C:\MyProject\Source\Test.ispc"
-h "C:\MyProject\Source\Test.ispc.generated.h"
--arch=x86_64
--target=avx2,avx,sse4,sse2
-I "C:\MyProject\Third Party\Source"
```
Produces the following output:
```
C:\MyProject\Third Party\ispc>ispc.exe "@C:\MyProject\Build\Test.ispc.response"
Multiple input files specified on command line: ""C:\MyProject\Source\Test.ispc"" and "Party\Source"".
```

ispc is splitting the last argument in two - `-I "C:\MyProject\Third` and `Party\Source"`.

When written onto the command-line normally this isn't an issue because the OS/C-runtime takes care of splitting the arguments into individual strings, keeping quoted parts together.

dbabokin:
We'll definitely need to fix that. But I assume if we fix an issue with number of arguments, this will not block you, right?

Unit2Ed:
Unfortunately the amount of command-line I need to pass is 53KB, which exceeds the Windows limit of 32KB. At the moment an @ file is the only option I have available.

I'm working to reduce the amount, but it requires some major structural changes to the project to get a decent win.

dbabokin:
@Unit2Ed #1641 should fix the problem. Could you please try the latest build and confirm that it works for you?

Unit2Ed:
@dbabokin Yep, that's fixed it - thanks for the quick turnaround!

PR #1754, opened 2020-05-12
Add a collection of tests for ISPC response file handling

robertmaynard:
I couldn't find any test that verified the fix to https://github.com/ispc/ispc/issues/1638 so I went ahead and added one. This behavior is important to the project I am working on.

At the same time I noticed that ISPC didn't have any tests verifying missing or nested response files so I added those as well.


dbabokin:
Looks good!

By the way, if your project happens to be open source, feel free to post a link to it, so we can build it as part of release testing.

PR #1758, opened 2020-05-19
Fixes for LLVM trunk.

jiapei100:
…for the compatibility of LLVM 10.0.0


dbabokin:
Please apply ``clang-format -i src/ctx.cpp``

DeepakRajendrakumaran:
Hi. Quick question: Looking at the LLVM API, I see that it makes sense to use valueOrNone() when there is a chance if undefined. Did you run into any cases where this was an issue/did you have any failures due to this?

dbabokin:
I'm not sure if adding more commits was intentional for this PR. I suggest not to expand the scope of this PR and keep it for llvm::MaybeAlign issue only. You'll need to updated your branch for this.

Original changes look good to me, but would be interesting to know what triggered these changes. @jiapei100 have you checked alignment info and was wrong?

dbabokin:
@DeepakRajendrakumaran thanks for looking at this and amending with missing fixes to make trunk work again.

src/llvmutil.h
```diff
@@ -51,6 +51,12 @@ class PHINode;
 class InsertElementInst;
 } // namespace llvm
 
+#if ISPC_LLVM_VERSION >= ISPC_LLVM_11_0
+#define LLVMVECTOR llvm::FixedVectorType
```
dbabokin, line 55: It's probably better to have type alias instead of a macro here.

PR #1780, opened 2020-06-02
Fix #1769 - ISPC asserts when doing logical and/or with vectors.

### LINKED ISSUE #1769: ISPC asserts when doing logical and/or with vectors.
https://godbolt.org/z/WRzS4z 

Seems to have been an existing problem in older versions as well.

DeepakRajendrakumaran:
Closes #1769

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -1640,11 +1640,10 @@ llvm::Value *lEmitLogicalOp(BinaryExpr::Op op, Expr *arg0, Expr *arg1, FunctionE
     // right side of the expression is a) relatively simple, and b) can be
     // safely executed with an all-off execution mask, then we just
     // evaluate both sides and then the logical operator in that case.
-    // FIXME: not sure what we should do about vector types here...
     bool shortCircuit =
-        (EstimateCost(arg1) > PREDICATE_SAFE_IF_STATEMENT_COST || SafeToRunWithMaskAllOff(arg1) == false ||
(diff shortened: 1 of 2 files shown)


src/expr.cpp
```diff
@@ -1642,9 +1642,9 @@ llvm::Value *lEmitLogicalOp(BinaryExpr::Op op, Expr *arg0, Expr *arg1, FunctionE
     // evaluate both sides and then the logical operator in that case.
     // FIXME: not sure what we should do about vector types here...
```
dbabokin, line 1643: FIXME is not relevant anymore.

src/expr.cpp
```diff
@@ -1642,9 +1642,9 @@ llvm::Value *lEmitLogicalOp(BinaryExpr::Op op, Expr *arg0, Expr *arg1, FunctionE
...
     bool shortCircuit =
-        (EstimateCost(arg1) > PREDICATE_SAFE_IF_STATEMENT_COST || SafeToRunWithMaskAllOff(arg1) == false ||
-         CastType<VectorType>(type0) != NULL || CastType<VectorType>(type1) != NULL);
-    if (shortCircuit == false) {
```
dbabokin, line 1647: Generating always if conditions for vectors in not a good idea - I think safety check should work the same way it works for scalars. I.e. if no unsafe operations involved (pointer access, fp trapping operations), we should do the same as for scalars.

dbabokin: Ok, I see that we check mask and etc for varying and need to do the same for vectors, so current fix is good enough.

PR #1781, opened 2020-06-06
Fix #1768 - different rounding rules for uniform and varying integer division.

### LINKED ISSUE #1768: Uniform and varying types have different rounding rules.
ISPC uses different rounding approaches for uniform and varying types. Uniform types are rounded to 0 and varying types are rounded to minus inf. It happens in case of negative values that have a remainder. 

> \>$ispc foo.ispc -o foo.o --target=avx2-i32x4 && clang++ main.cpp foo.o && ./a.out
-1
[-2,-2,-2,-2]

Reproducer:
foo.ispc
```c++
uniform int i = -13;

export void foou() {
    uniform int inp = i;
    uniform int res = inp / 12;
    print("%\n", res);

}

export void foov() {
    varying int inp = i;
    varying int res = inp / 12;
    print("%\n", res);
}
```

main.cpp
```c++
extern "C" { void foou(); void foov(); };

int main() {
    foou();
    foov();
}
```

ISPC version:
1.14.0dev (build commit 763adc5a026a3d07 @ 20200526, LLVM 10.0.0)

DeepakRajendrakumaran:
Disabling fast division in ispc to let the llvm codegen handle it.
This needs to be verified using benchmark to ensure division is not slower.


benchmarks/01_trivial/04_fastdiv.cpp
```diff
@@ -0,0 +1,113 @@
+#include <benchmark/benchmark.h>
+#include <bitset>
+#include <cstdint>
+#include <stdio.h>
+
+#include "../common.h"
+#include "04_fastdiv_ispc.h"
+
+static Docs docs("Check fast_idiv implmentation of stdlib functions:\n"
+                 "[int8, int16, int32] x [2, 13, 16, 127] x [-ve, +ve] versions.\n"
```
dbabokin, line 10: - you don't need different power of 2, they trigger the same code.
- is int64 triggering fastdiv? I think it makes sense to have it to see it's perfomance, even if it doesn't trigger fastdiv.
- positive and negative means that algorithm is covering positive and negative vlaues. I.e. ``signed int`` should trigger algorithm covering both negative and positive, while ``unsigned int`` should trigger algorithm, which covers only positives. I'm not sure that range analysis takes care about it and is used in fastdiv.

benchmarks/01_trivial/04_fastdiv.ispc
```diff
@@ -0,0 +1,121 @@
+
+
+export void fastdiv_int32_positive_13(uniform int32 *uniform dst, uniform int count) {
```
dbabokin, line 3: I suggest having single macro for generating thsse functions.

benchmarks/01_trivial/04_fastdiv.cpp
DeepakRajendrakumaran: 1. I meant 16 as power of 2, 13 as non-power of 2 and 2 and 127 as outer bounds of values. I have no logical reasoning for thinking the outer bounds will behave any differently but decided to have those anyway. I have taken out 2 in the update
2. int64 does not trigger fast_idiv. But have added it for comparison.
3 Not sure about this. My intention was to just test the algorithm with negative values . I have changed to use uint value. Our earlier implementation for this seem to be a bit of a mess - https://godbolt.org/z/fDfuJf  : see warning

benchmarks/01_trivial/04_fastdiv.ispc
DeepakRajendrakumaran: done

benchmarks/01_trivial/04_fastdiv.cpp
dbabokin: - comments were not changed
- 13 and 127 triggered code. which looked different, but if you don't have reasoning leave only 13.
- you should not have two different functions for testing positive and negative values. You should either pass these values or produce them internally without a negation operator. I think if you pass them in another array (say src[]), that would be better. Otherwise it creates unnecessary complications and confuses.
- README still needs an update.

DeepakRajendrakumaran: 1. Comments changed
2. Done
3. This more complicated. Adding a src[] makes it messier in my opinion. The numerator needs to be a varying  variable and cannot be an array so src[i]/divVal doesn't work. Using the src[] is some other way means adding a gather too. I have tweaked it to make it a little more cleaner.
4. Updated..its missing 03_popcnt too.

dbabokin: 1-2. ok.
3. ``src[i]/divVal`` works perfectly fine, as ``src[i]`` is varying value in case of varying ``i``. Avoiding extra perations in the loops makes it cleaner test for division. Even though range arithmetic will likely be negligible, but it's better to avoid it. memory load from L1 should be very fast, especially as HW can overlap different iterations. 
4. Yes, I'll add popcnt description.

benchmarks/01_trivial/README.md
```diff
@@ -2,3 +2,4 @@
 
 - ``01_aossoa`` - all variations of ``aos_to_soa``, stdlib and ISPC source implementations.
 - ``02_soaaos`` - all variations of stdlib implmentation ``soa_to_aos``. TODO: ISPC source implementations.
+- ``04_fastdiv`` - division operations which used to trigger ``__fast_idiv`` in ispc.
```
dbabokin, line 5:
```suggestion
- ``04_fastdiv`` - integer division by a constant is handled by an algorithm, which produces a code sequence without actual division operation. Current implementation relies on code generator to do the right thing on every specific platform. This benchmark tests perfomance integer division by a constant.
```

benchmarks/01_trivial/04_fastdiv.cpp
dbabokin: If you add ``src[]``, you'll be able to handle sign and unsigned version in uniform way.

DeepakRajendrakumaran: 3. Fair enough. I have changed the ispc side.

benchmarks/01_trivial/04_fastdiv.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+#include "../common.h"
+#include "04_fastdiv_ispc.h"
+
+static Docs docs("Check fast_idiv implmentation of stdlib functions:\n"
+                 "[int8, uint8, int16, uint16, int32, uint32, int64, uint64] x [13, 16] versions.\n"
+                 "Conditions to trigger fast_idiv:\n"
+                 " - The value being divided must be an int8/16/32.\n"
+                 " - The divisor must be the same compile-time constant value for all of the vector lanes.\n"
+                 " - The divisor must be >= 2 and <128 (for 8-bit divides), and <256 otherwise.\n"
```
dbabokin, line 14:
```suggestion

```

dbabokin:
After the latest commit, that I pushed to the branch the changes are ready to be merged.

PR #1784, opened 2020-06-08
Fix  #1747 - Change base type for varying bool vector to int8.

### LINKED ISSUE #1747: ISPC 1.13 fails to compile sources which 1.12 compiles fine
Attached are the files which compiles fine with ISPC 1.12, but with 1.13 following error is printed:
Assertion failed: castIsValid(getOpcode(), S, Ty) && "Illegal SExt", file D:\users\aneshlya\llvm\llvm-10.0\llvm\lib\IR\Instructions.cpp, line 3326

Thanks for looking into it :)
[repro.zip](https://github.com/ispc/ispc/files/4577056/repro.zip)

DeepakRajendrakumaran:
varying bool vector is stored in IR as array type. So, need to be
treated similar to array type in IR.

### DIFF
--- a/tests/lit-tests/1747.ispc
+++ b/tests/lit-tests/1747.ispc
@@ -1,0 +1,3 @@
+// RUN: %{ispc} --target=host %s > %t 2>&1
+typedef bool<2> bool2;
+void MakeBool2(bool2 arg1) { return; }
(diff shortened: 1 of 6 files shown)


tests/lit-tests/1747.ispc
```diff
@@ -0,0 +1,4 @@
+// RUN: %{ispc} --target=sse4-i32x8 %s > %t 2>&1
+// REQUIRES: X86_ENABLED
```
dbabokin, line 2: This is not specific for x86, so use ``--target=host``, this is the current way to run on both ARM and x86. We may change it, but for now just use host.

DeepakRajendrakumaran:
> `tests/short-vec-15.ispc` is failing.

Yep. Fixed it.

tests/lit-tests/1747.ispc
DeepakRajendrakumaran: Done.

PR #1787, opened 2020-06-09
Fix #1739 - Print error when accessing vector elements incorrectly.

### LINKED ISSUE #1739: ISPC crashes on code that does not type check (array of short vectors)
I happened to type in the following code (simplified) in my project:
```
float foo(float<4> x[4])
{
    return x.a;
}
```
This code does not type check, so I expect to get a compilation error. However, ISPC crashed with the following message:
```
ispc -O3 --target=avx2-i32x8 --arch=x86-64 bug.ispc 
bug.ispc:3:12: Error: Assertion failed (/build/ispc/src/ispc-1.12.0/src/expr.cpp:4586): "vt != NULL". 
    return x.a;
           ^^^

***
***
/build/ispc/src/ispc-1.12.0/src/main.cpp(353): FATAL ERROR: Unhandled signal sent to process; terminating.
Aborted (core dumped)
```
I am using `Intel(r) SPMD Program Compiler (ispc), 1.12.0 (build  @ 20200414, LLVM 10.0.0)` on Linux.

DeepakRajendrakumaran:
Fix #1739

### DIFF
--- a/tests/lit-tests/1739.ispc
+++ b/tests/lit-tests/1739.ispc
@@ -1,0 +1,4 @@
+//; RUN: not %{ispc} --target=host %s --nowrap 2>&1 | FileCheck %s
+
+//; CHECK: Error: Member operator "." can't be applied to pointer type "varying float<4>". Did you mean to use "->"?
+float foo(float<4> foo_arg[4]) { return foo_arg.a; }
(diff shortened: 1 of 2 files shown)


tests/lit-tests/1739.ispc
```diff
@@ -0,0 +1,4 @@
+//; RUN: not %{ispc} --target=avx2-i32x8 --target=host %s --nowrap 2>&1 | FileCheck %s
```
dbabokin, line 1: Only one ``--target=...`` is needed here.

DeepakRajendrakumaran: Done

PR #1805, opened 2020-06-23
Fixing bool->non bool conversions for avx512.

DeepakRajendrakumaran:
Using 'SwitchBoolSize()' for switching between bools and non-bool types
and removing legacy check.


dbabokin:
LGTM.

PR #1811, opened 2020-07-04
Fix #1806 - treat bool as unsigned, not as signed.

### LINKED ISSUE #1806: ISPC produces wrong code with bool type iterator.
The loop should be execuded once while ISPC skips it. This happens on a lot of different targets.
> \>$ ispc func.ispc -o func.o && clang++ func.o driver.cpp && ./a.out
BAD

Reproducer:
func.ispc
```c++
export void test() {
    for (uniform bool i = 0; i < (uniform bool)1; i += 1)
    {
        print("OK\n");
        return;
    }
    print("BAD\n");
}
```

driver.cpp
```c++
extern "C" { void test(); }

int main() {
    test();
}
```

ISPC version:
1.14.0dev (build commit 1bc127f35c0d0780 @ 20200623, LLVM 10.0.0) + da679386632e25bfb9d82de7f4cd9b9d88e8c937

dbabokin:
This affects how true (1) is treated. In case of unsigned it's 1. In case of signed, it's treated as -1.

There's another problem with boolean </> comparison - we do compare without promoting to i32, which is inconsistent with C/C++. If it causes a problem again, we should change it.

### DIFF
--- a/src/type.cpp
+++ b/src/type.cpp
@@ -174,7 +174,7 @@ bool AtomicType::IsIntType() const {
 }
 
 bool AtomicType::IsUnsignedType() const {
-    return (basicType == TYPE_UINT8 || basicType == TYPE_UINT16 || basicType == TYPE_UINT32 ||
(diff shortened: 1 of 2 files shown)


dbabokin:
Fixes #1806

PR #1813, opened 2020-07-07
Fixing access to bool pointer elements for struct.

DeepakRajendrakumaran:


tests/lit-tests/bool_struct_member.ispc
```diff
@@ -0,0 +1,8 @@
+// RUN: %{ispc} %s --target=host > %t 2>&1
+// RUN: FileCheck --input-file=%t %s
+// CHECK-NOT: Please file a bug report at https://github.com/ispc/ispc/issues
```
dbabokin, line 3:
```suggestion
// RUN: %{ispc} %s --target=host --nostdlib
```

PR #1820, opened 2020-07-08
Switching to __vectorcall calling convention for windows for x64.

DeepakRajendrakumaran:


src/ctx.cpp
```diff
@@ -2995,11 +2995,26 @@ llvm::Value *FunctionEmitContext::CallInst(llvm::Value *func, const FunctionType
...
-        llvm::Instruction *ci = llvm::CallInst::Create(func_type, func, argVals, name ? name : "", bblock);
+        llvm::CallInst *callinst = llvm::CallInst::Create(func_type, func, argVals, name ? name : "", bblock);
 #else
-        llvm::Instruction *ci = llvm::CallInst::Create(func, argVals, name ? name : "", bblock);
+        llvm::CallInst *callinst = llvm::CallInst::Create(func, argVals, name ? name : "", bblock);
 #endif
 
+        // We could be dealing with a function pointer in which case this will not be a 'llvm::Function'.
+        // If 'llvm::Function', use same calling convention as the actual function definition. It's
+        // important we do this since  prebuilt stdlib functions does not use vectorcall on any OS.
```
dbabokin, line 3005: This sounds like we need to change calling convention for stdlib as well. It should not have bug effect on the resulting code, as most of the functions will be inlined anyway, but I see no reason not to use ``vectorcall`` convetion for stdlib and loose effiency.

src/func.cpp
```diff
@@ -464,6 +464,8 @@ void Function::GenerateIR() {
 
                 llvm::Function *appFunction = llvm::Function::Create(ftype, linkage, functionName.c_str(), m->module);
                 appFunction->setDoesNotThrow();
+                if (g->calling_conv == CallingConv::x86_vectorcall)
```
dbabokin, line 467: In this changeset the pattern "if if (g->calling_conv == CallingConv::x86_vectorcall) XXX->setCallingConv(llvm::CallingConv::X86_VectorCall);" appears at least 4 times. It worth adding a function to do that. This will also may check if the state of calling convention is not properly initialized"

examples/perfbench/perfbench.cpp
```diff
@@ -37,14 +37,24 @@
 #pragma warning(disable : 4244)
 #pragma warning(disable : 4305)
 #endif
+#ifdef _WIN32
+#define WINDOWS
+#endif
+#ifdef _WIN64
+#define WINDOWS
+#endif
```
dbabokin, line 45: #if defined(_WIN32) || defined(_WIN64)

examples/perfbench/perfbench_serial.cpp
```diff
@@ -33,7 +33,18 @@
 
 #include <math.h>
 
+#ifdef _WIN32
+#define WINDOWS
+#endif
+#ifdef _WIN64
+#define WINDOWS
+#endif
```
dbabokin, line 41: #if defined(_WIN32) || defined(_WIN64)

examples/perfbench/perfbench.cpp
```diff
@@ -37,14 +37,24 @@
...
 
 #include "../timing.h"
 #include <algorithm>
 #include <stdio.h>
 
 #include "perfbench_ispc.h"
 
+#ifdef WINDOWS
+typedef void(__vectorcall FuncType)(float *, int, float *, float *);
```
dbabokin, line 54: I think defining macro ``CALLING_CONV`` (or something like that) and setting it's value to either empty or ``__vectorcall`` is better options from readablility prospective.

test_static.cpp
```diff
@@ -72,6 +72,17 @@
 #define ARRAY_SIZE 256
 
 extern "C" {
+#ifdef ISPC_IS_WINDOWS
+extern int __vectorcall width();
```
dbabokin, line 76: Same here, ``CALLING_CONV`` macro looks a better option.

src/main.cpp
```diff
@@ -104,6 +104,7 @@ static void lPrintVersion() {
     PrintWithWordBreaks(cpuHelp, 16, TerminalWidth(), stdout);
     printf("    [-D<foo>]\t\t\t\t#define given value when running preprocessor\n");
     printf("    [--dev-stub <filename>]\t\tEmit device-side offload stub functions to file\n");
+    printf("    [--disable-vectorcall]\t\t\t\tDisable vectorcall calling convention.\n");
```
dbabokin, line 107: This make sense on Windows only, so this needs to be enabled only for Windows.

src/ctx.cpp
DeepakRajendrakumaran: stdlib.ispc will be compiled with vectorcall. Problem is with the .ll files in builtin.

DeepakRajendrakumaran: p.s - yes, the code in .ll will be inlined. So, it's not a problem eventually.

src/target_enums.cpp
```diff
@@ -382,3 +382,11 @@ TargetOS GetHostOS() {
     return TargetOS::error;
 #endif
 }
+
+CallingConv GetDefaultCallingConv() {
```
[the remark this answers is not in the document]
DeepakRajendrakumaran: Done

src/main.cpp
```diff
@@ -423,6 +424,15 @@ static int ParsingPhaseName(char *stage, ArgErrors &errorHandler) {
     }
 }
 
+static void resetCallingConv(bool disableVectorCall) {
```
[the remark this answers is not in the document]
DeepakRajendrakumaran: Done

src/ctx.cpp
DeepakRajendrakumaran: Left it as it is. Let me know if you disagree.

src/func.cpp
DeepakRajendrakumaran: Done

examples/perfbench/perfbench.cpp
DeepakRajendrakumaran: Done

examples/perfbench/perfbench.cpp
DeepakRajendrakumaran: Done

src/main.cpp
DeepakRajendrakumaran: Done

DeepakRajendrakumaran:
I see a crash which I need to fix. Will update when I do.

examples/perfbench/perfbench_serial.cpp
DeepakRajendrakumaran: Done

src/ctx.cpp
dbabokin: I think it's ok, but if see any non-inlined instances, we need to reconsider this.

tests/lit-tests/1319.ispc
```diff
@@ -1,4 +1,4 @@
-//; RUN: %{ispc} %s -h %t.h -o /dev/null
+//; RUN: %{ispc} %s -h %t.h --disable-vectorcall -o /dev/null
```
dbabokin, line 1: ``-o /dev/null`` is not needed here and can be removed.
I suggest instead of adding ``--disable-vectorcall`` modifying a check to allow ``__vectorcall`` in the string.

tests/lit-tests/vectorcall-1.ispc
```diff
@@ -0,0 +1,27 @@
+// RUN: %{ispc}  %s --emit-asm --target=avx2-i32x8 -o - | FileCheck %s -check-prefix=CHECK_AVX2
```
dbabokin, line 1: Comment describing what you are checking here is needed. I.e. "Check that calling convention is not adding any overhead and function is optimized to only add and ret instructions"

tests/lit-tests/vectorcall-2.ispc
```diff
@@ -0,0 +1,15 @@
+
+// RUN: %{ispc} %s -h %t1.h --target=host -o /dev/null
```
dbabokin, line 2: ``-o /dev/null`` is not needed.
Add ``--target-os=windows`` to make sure that it's compiled for Windows, otherwise it might fail in cross-compiler testing. Though right now we don't target Windows from non-Windows.

tests/lit-tests/vectorcall-3.ispc
```diff
@@ -0,0 +1,34 @@
+// RUN: %{ispc}  %s --emit-asm --target=avx2-i32x8 --disable-vectorcall -o - | FileCheck %s -check-prefix=CHECK_AVX2
```
dbabokin, line 1: Comment on the intent is needed. Especially, given that you are checking very specific code sequence.

src/main.cpp
```diff
@@ -423,6 +424,14 @@ static int ParsingPhaseName(char *stage, ArgErrors &errorHandler) {
     }
 }
 
+static void setCallingConv(bool disableVectorCall) {
```
dbabokin, line 427: Need a check for ``uninitialized`` state as well.

src/main.cpp
```diff
@@ -104,6 +104,7 @@ static void lPrintVersion() {
     PrintWithWordBreaks(cpuHelp, 16, TerminalWidth(), stdout);
     printf("    [-D<foo>]\t\t\t\t#define given value when running preprocessor\n");
     printf("    [--dev-stub <filename>]\t\tEmit device-side offload stub functions to file\n");
+    printf("    [--disable-vectorcall]\t\t\t\tDisable vectorcall calling convention on windows.\n");
```
dbabokin, line 107: "windows" -> "Windows"

src/main.cpp
```diff
@@ -657,6 +667,8 @@ int main(int Argc, char *Argv[]) {
                 errorHandler.AddError("Unsupported value for --target-os, supported values are: %s",
                                       g->target_registry->getSupportedOSes().c_str());
             }
+        } else if (!strcmp(argv[i], "--disable-vectorcall")) {
```
dbabokin, line 670: Add a warning that it's ignored if the target OS is not Windows, see similar warnings for ``--dllexport`` around line 930.

tests/lit-tests/vectorcall-3.ispc
DeepakRajendrakumaran: I would have settled for a more simple check if I could use something like CHECK-NEXT-NOT to make sure the first instruction is not a mov. If you think the check is too specific, I can just check for a mov

dbabokin: I like your idea. Checking that no "mov" are generated is actually what we need here.
This way you can simplify the check and make it universal for any target.

dbabokin:
The changes look good to me and I think the PR is ready to be merged after fixign the tests (as proposed).
Aslo we may want to document this somewhere, but I'm not sure about the best place for it.

tests/lit-tests/vectorcall-3.ispc
DeepakRajendrakumaran: In this case, it would be a check to ensure a 'mov' is created for windows when vectorcall is used. I've made the change.

src/parse.yy
```diff
@@ -1134,6 +1134,11 @@ specifier_qualifier_list
                       "function declarations.");
                 $$ = $2;
             }
+            else if ($1 == TYPEQUAL_VECTORCALL) {
+                Error(@1, "\"vectorcall\" qualifier is illegal outside of "
```
dbabokin, line 1138: I suggest adding test for that.

dbabokin:
Why diverging from C/C++ spelling ``__vectorcall`` to ``vectorcall``. I suggest keeping the same spelling to avoid confusion.

src/parse.yy
DeepakRajendrakumaran: The qualifier error mechanism doesn't work right now for any qualifiers. Will file a bug and add test sepaately.

DeepakRajendrakumaran:
> Why diverging from C/C++ spelling `__vectorcall` to `vectorcall`. I suggest keeping the same spelling to avoid confusion.

This is done

dbabokin:
When you think it's ready to be merged, please remove "WIP" from the title.

PR #1827, opened 2020-07-17
Use WIX CPack generator for Windows

aneshlya:


dbabokin:
- seems that `CPACK_WIX_UPGRADE_GUID` needs to be set to allow upgrades.
- seems that setting `CPACK_WIX_CMAKE_PACKAGE_REGISTRY` should be useful.

The point of switching to msi package is to make it possible for ISPC to be installed by `winget`, which mean subitting a manifest to https://github.com/microsoft/winget-pkgs When this PR is merged, we need to open an issue to do that.

CMakeLists.txt
```diff
@@ -589,8 +590,12 @@ if (ISPC_PREPARE_PACKAGE)
 
     # CPack configuration
     if (WIN32)
-        set(CPACK_GENERATOR "ZIP")
+        set(CPACK_GENERATOR "WIX")
         set(ISPC_SYSTEM_NAME "windows")
+        # WIX generator expects version in format x.y.z where x, y, z are numbers
+        set(CPACK_PACKAGE_VERSION "${ISPC_VERSION_MAJOR}.${ISPC_VERSION_MINOR}.${ISPC_VERSION_PATCH}")
+		set(CPACK_WIX_UPGRADE_GUID "EDD858F2-19B0-4E5B-B9B9-2A09B85D451C")
```
dbabokin, line 597: Identation.

CMakeLists.txt
```diff
@@ -589,8 +590,12 @@ if (ISPC_PREPARE_PACKAGE)
...
+		set(CPACK_WIX_CMAKE_PACKAGE_REGISTRY "Intel(R) Implicit SPMD Program Compiler")
```
dbabokin, line 598: I think the package name should be `ISPC`, so that `find_package(ISPC)` would return this package.

Though, I might be reading docs incorrectly. Could you please check if this works?

aneshlya: Since we do not provide cmake configuration file as part of our installation, find_package(ISPC) will not work.

PR #1829, opened 2020-07-21
Fix compiling with GCC (#1828)

### LINKED ISSUE #1828: [GCC/Linux] Building fails because of clang only compile flag
I'm trying to build ispc with GCC 9.2.0, however it fails with:
`x86_64-pc-linux-gnu-g++: error: unrecognized command line option ‘-fno-strict-float-cast-overflow’; did you mean ‘-fno-sanitize=float-cast-overflow’?`

After googling the flag, this seems to be a clang only flag at the moment.
I'm guessing this could be solved with a simple "if clang" check.

DarkDefender:
This fixes #1828.

Using an `if/else` for Clang and GCC is fine I hope.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -355,11 +355,11 @@ if (UNIX)
     target_compile_options(${PROJECT_NAME} PRIVATE -Wall -Wno-sign-compare -Wno-unused-function -Werror ${LLVM_CPP_FLAGS})
     # Security options
     if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-      # Using the clang compiler
-      set(COMPILER_SPECIFIC_FLAGS -fno-strict-float-cast-overflow)
+        # Using the clang compiler
+        set(COMPILER_SPECIFIC_FLAGS -fno-strict-float-cast-overflow)
     else()
-      # Assume that we are using GCC
-      set(COMPILER_SPECIFIC_FLAGS -fno-sanitize=float-cast-overflow)
+        # Assume that we are using GCC
+        set(COMPILER_SPECIFIC_FLAGS -fno-sanitize=float-cast-overflow)
     endif()
     target_compile_options(${PROJECT_NAME} PRIVATE ${COMPILER_SPECIFIC_FLAGS} -fstack-protector-strong
                            -fdata-sections -ffunction-sections -fno-delete-null-pointer-checks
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -354,7 +354,14 @@ target_include_directories(${PROJECT_NAME} PRIVATE
 if (UNIX)
     target_compile_options(${PROJECT_NAME} PRIVATE -Wall -Wno-sign-compare -Wno-unused-function -Werror ${LLVM_CPP_FLAGS})
     # Security options
-    target_compile_options(${PROJECT_NAME} PRIVATE -fstack-protector-strong -fno-strict-float-cast-overflow
+    if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+      # Using the clang compiler
```
dbabokin, line 358: The rest of the file has indentation by 4 spaces, please fix.

DarkDefender: Ah, my bad.
Fixed it now.
Do you want me to squash the commits?

dbabokin: Don't worry about it, I'll squash it on the merge.

CMakeLists.txt
```diff
@@ -354,7 +354,14 @@ target_include_directories(${PROJECT_NAME} PRIVATE
 if (UNIX)
     target_compile_options(${PROJECT_NAME} PRIVATE -Wall -Wno-sign-compare -Wno-unused-function -Werror ${LLVM_CPP_FLAGS})
     # Security options
-    target_compile_options(${PROJECT_NAME} PRIVATE -fstack-protector-strong -fno-strict-float-cast-overflow
+    if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+        # Using the clang compiler
+        set(COMPILER_SPECIFIC_FLAGS -fno-strict-float-cast-overflow)
+    else()
+        # Assume that we are using GCC
+        set(COMPILER_SPECIFIC_FLAGS -fno-sanitize=float-cast-overflow)
```
aneshlya, line 362: I think the right option for gcc is "-fno-strict-overflow"

DarkDefender: Hmm, I've tried googling both flags and it seems like both Clang and GCC supports `-fno-strict-overflow` and `-fno-sanitize=float-cast-overflow`.

I didn't manage to find a elaborate explanation for `no-strict-float-cast-overflow` and the difference with `no-sanitize=float-cast-overflow`.

Perhaps this is a case were we don't really need this `if/else` at all and can just use flags that are supported by both (with the same end result)?

aneshlya: Right. Let's keep -fno-strict-overflow for both Clang and GCC.
Thanks for your findings!

DarkDefender: Done! :)

aneshlya: Since our CI failed after adding this option I've dived into it and found out that we do not need to use -fno-strict-overflow if -fwrapv is used. So I removed -fno-strict-overflow.

PR #1837, opened 2020-07-29
Switch default on x64 Windows to non __vectorcall.

DeepakRajendrakumaran:


src/main.cpp
```diff
@@ -670,8 +675,11 @@ int main(int Argc, char *Argv[]) {
                 errorHandler.AddError("Unsupported value for --target-os, supported values are: %s",
                                       g->target_registry->getSupportedOSes().c_str());
             }
-        } else if (!strcmp(argv[i], "--disable-vectorcall")) {
-            disableVectorCall = true;
+        } else if (!strcmp(argv[i], "--no-vectorcall")) {
+            vectorCall = VectorCallStatus::disabled;
+         }
+        else if (!strcmp(argv[i], "--vectorcall")) {
+            vectorCall = VectorCallStatus::enabled;
```
DeepakRajendrakumaran, line 681: Add check to see if --vectorcall and --no-vectorcall are used together?

tests/lit-tests/vectorcall-5.ispc
```diff
@@ -1,5 +1,6 @@
 // This test checks function signature of externs with '__vectorcall' to ensure it has 'x86_vectorcallcc'.
-// RUN: %{ispc}  %s --emit-llvm-text --target=host -o - | FileCheck %s
+// RUN: %{ispc}  %s --emit-llvm-text --vectorcall --target=host -o - | FileCheck %s
+// RUN: %{ispc}  %s --emit-llvm-text --no-vectorcall --target=host -o - | FileCheck %s
```
DeepakRajendrakumaran, line 3: This fails for the --no-vectorcall option because check looks for vectorcall

src/main.cpp
dbabokin: I think it's ok to allow using them both and have the last one win, it's pretty standard behavior. We have it for --target as well.

tests/lit-tests/vectorcall-5.ispc
dbabokin: I've changed the test, we'll need to change behavior later.

PR #1862, opened 2020-08-31
Switching from constprop to instsimplify.

DeepakRajendrakumaran:


dbabokin:
I suggest just unconditionally switch to ``llvm::createInstSimplifyLegacyPass``, instead of doing that LLVM 12+. I see no reasons to have two code paths.

PR #1863, opened 2020-09-03
Fix VectorType to match llvm change.

DeepakRajendrakumaran:


dbabokin:
I see several fails on soa-16/soa-17 test on AVX512. They seem to be not flaky. They should not be related to this fix, so I'm merging. But we will need to have a look at them.

PR #1866, opened 2020-09-11
Fix  #1826 : Modify check for left shift when constant folding.

### LINKED ISSUE #1826: "1<<31" should not trigger a warning
``1<<31`` should not trigger a warning. Even through, 1 is treated as ``int32``, left shift should not trigger a warning, as long as result is not exeding 32 bits. ISPC should follow C/C++ in this behavior.

Reproducer:
```c
uniform unsigned int32 foo() {
    uniform unsigned int32 u = 1<<31;
    return u;
}
```

```bash
>ispc test.ispc --target=avx2-i32x8
test.ispc:2:32: Warning: Binary expression with type "const uniform int32" can't represent value.
    uniform unsigned int32 u = 1<<31;
                               ^^^^^
```

DeepakRajendrakumaran:


src/expr.cpp
```diff
@@ -2039,6 +2039,22 @@ const Type *BinaryExpr::GetType() const {
         }                                                                                                              \
         break
 
+template <typename T> static T countLeadingZeros(T val) {
+
+    T leadingZeros = 0;
```
dbabokin, line 2044: I think ``leadingZeros`` and return type should be ``int``, not ``T``.

src/expr.cpp
```diff
@@ -2039,6 +2039,22 @@ const Type *BinaryExpr::GetType() const {
...
+    T size = sizeof(T) * CHAR_BIT;
```
dbabokin, line 2045: sizeof(T) is ``size_t``, so you will be downconverting in some cases when assigning to ``T`` type. Some compilers will have warnings. I suggest ``size`` to make ``size_t`` and explicitly convert ``1 << (size - 1)`` to ``T``.

src/expr.cpp
```diff
@@ -2048,11 +2064,20 @@ static ConstExpr *lConstFoldBinaryIntOp(BinaryExpr::Op op, const T *v0, const T
...
-        FOLD_OP_REF(BinaryExpr::Shl, <<, TRef);
         FOLD_OP_REF(BinaryExpr::Shr, >>, TRef);
         FOLD_OP_REF(BinaryExpr::BitAnd, &, TRef);
         FOLD_OP_REF(BinaryExpr::BitXor, ^, TRef);
         FOLD_OP_REF(BinaryExpr::BitOr, |, TRef);
+
+    case BinaryExpr::Shl:
+        for (int i = 0; i < count; ++i) {
+            result[i] = (v0[i] << v1[i]);
+            if (v1[i] > (T)countLeadingZeros(v0[i])) {
```
dbabokin, line 2075: This is a little weird, LHS generally speaking doesn't have type ``T``. I.e. to shift ``char`` you don't need to shift by ``char``, you will shift it by ``int``.

But for the sake of this transformation this is irrelevant, as we just dealing with constant, which need to be less or equal to 64.

Nevertheless, casting to ``T`` is not needed here, as comparison operator in C requires standard integer promotion, which mean at least 32 bit integers will be compared.

PR #1871, opened 2020-09-14
WIP: fixes #1755 - Fixing bool varying initilaization.

### LINKED ISSUE #1755: Crash during varying initialization
https://godbolt.org/z/7G9bA8

This existed from 1.12 when varying initialization was added.

Here's the reproducer:
```c
void a(uniform bool aFOO) { 
	bool t = {false, aFOO, aFOO, aFOO, aFOO, aFOO, aFOO, aFOO};
} 
```

DeepakRajendrakumaran:


tests/lit-tests/1755.ispc
```diff
@@ -0,0 +1,6 @@
+//; RUN: %{ispc} %s -o %t.o --nowrap --target=avx2-i32x8 2>&1
```
dbabokin, line 1: Either, a restriction for running only on X86 is missing or you need to have ``--target=host`` and ifdef chan to cover compilation width for 4/8/16.

src/ctx.cpp
```diff
@@ -2847,12 +2847,12 @@ void FunctionEmitContext::scatter(llvm::Value *value, llvm::Value *ptr, const Ty
     AssertPos(currentPos,
               pt != NULL || CastType<AtomicType>(valueType) != NULL || CastType<EnumType>(valueType) != NULL);
 
-    llvm::Type *type = value->getType();
-    llvm::Type *llvmStorageType = type;
+    llvm::Type *llvmStorageType = value->getType();
+    ;
```
dbabokin, line 2851: dangling ``;``

src/type.cpp
```diff
@@ -279,6 +281,18 @@ const AtomicType *AtomicType::GetAsUniformType() const {
     return asUniformType;
 }
 
+const AtomicType *AtomicType::GetAsUniformStorageType() const {
+
```
dbabokin, line 285: We typically don't use vertical spacing in the begging of the function.

src/type.cpp
```diff
@@ -279,6 +281,17 @@ const AtomicType *AtomicType::GetAsUniformType() const {
     return asUniformType;
 }
 
+const AtomicType *AtomicType::GetAsUniformStorageType() const {
+    Assert(basicType != TYPE_VOID);
+    if (variability == Variability::Uniform)
+        return this;
+
+    asUniformType = new AtomicType(basicType, Variability::Uniform, isConst);
```
dbabokin, line 289: It doesn't seem right to
(1) create new ``AtomicType`` object every single time we call ``AtomicType::GetAsUniformStorageType()`` regardless it was already initialized or not;
(2) right different types to ``asUniformType`` in ``GetAsUniformStorageType()`` and ``GetAsUniformType()``. I.e. if we call ``GetAsUniformStorageType()`` before ``GetAsUniformType()``, it will change (incorrectly) the type returned by ``GetAsUniformType()``.

dbabokin:
It's WIP for now, as transition between types are not context independent (`asUniformStorageType` may be initialized differently on different code paths for the same type). Type system should clearly define how bool, mask, and storage concepts relates to each other and to uniform and varying concepts.

DeepakRajendrakumaran:
Created separate PR which is a better fix.

dbabokin:
#1923 supersedes this PR.

PR #1880, opened 2020-10-06
fix #548 : Adding support for loop unroll pragmas.

### LINKED ISSUE #548: feature: #pragma unroll support
Is there support for #pragma unroll [n] for loop unrolling, like in CUDA, e.g.

```
#pragma unroll 2
for (uniform int i = 0; i < N; i++)
 { .. do stuff .. }
```

where `N` can be either a runtime or a compile-time parameter? If not, are any plans to add this feature and/or how much work is required to accomplish this?

DeepakRajendrakumaran:


src/lex.ll
```diff
@@ -706,64 +714,186 @@ lCComment(SourcePos *pos) {
...
+
+    const char *currChar = fromUserReq.data();
+    yylval->pragmaAttributes = new PragmaAttributes();
+    yylval->pragmaAttributes->aType = PragmaAttributes::AttributeType::pragmaloop;
+    int count = -1;
+
+    lNextValidChar(pos, currChar);
+
+    if (isNounroll) {
+        if (*currChar == '\n') {
```
dbabokin, line 736: The line may end by ``\r\n``. Would that be a problem?

src/parse.yy
```diff
@@ -1667,7 +1681,26 @@ initializer_list
       }
     ;
 
+attribute_list
```
dbabokin, line 1684: Should we chain pragmas? We accept only one pragma and have not check for pragma compatibility.
I don't care about pragma compatibility much, if it doesn't crash the program. But the following example does:
```c
void foo();
void bar() {
#pragma nounroll
#pragma unroll (8)
  for(uniform int i = 0; i<100; i++) {
    foo();
  }
}
```

src/lex.ll
```diff
@@ -706,64 +714,186 @@ lCComment(SourcePos *pos) {
...
+            yylval->pragmaAttributes->unrollType = Globals::pragmaUnrollType::nounroll;
+            pos->last_column = 1;
+            pos->last_line++;
+            return;
+        }
+        pos->last_column = 1;
+        pos->last_line++;
+        Error(*pos, "extra tokens at end of '#pragma nounroll'.");
```
dbabokin, line 744: I would expect it to be a warning.

src/lex.ll
```diff
@@ -706,64 +714,186 @@ lCComment(SourcePos *pos) {
...
+
+    bool popPar = false;
+    if (*currChar == '(') {
+        popPar = true;
+        currChar++;
+        ++pos->last_column;
+    }
+
+    char *endPtr = NULL;
+#if defined(ISPC_HOST_IS_WINDOWS) && !defined(__MINGW32__)
```
dbabokin, line 764: ``std::stoi()`` seem to be better alternative here, at least it's cross-platform.
It also doesn't return 0 on failure, but instead indicates it through not changed pointer (at least documentation says that).

src/stmt.cpp
```diff
@@ -80,6 +80,9 @@ void ExprStmt::EmitCode(FunctionEmitContext *ctx) const {
 }
 
 Stmt *ExprStmt::TypeCheck() { return this; }
+void ExprStmt::SetLoopAttribute(std::pair<Globals::pragmaUnrollType, int> lAttr) {
+    Error(pos, "Illegal '#Pragma' - expected a loop to follow '#pragma unroll/nounroll'.");
```
dbabokin, line 84: I'm not sure why ``#Pragma`` starts from capital latter.

src/stmt.h
```diff
@@ -66,7 +66,10 @@ class Stmt : public ASTNode {
     // also provide a default implementation of Optimize(), since most
     // Stmts don't have anything to do here.
     virtual Stmt *Optimize();
+
     virtual Stmt *TypeCheck() = 0;
+
+    virtual void SetLoopAttribute(std::pair<Globals::pragmaUnrollType, int>) = 0;
```
dbabokin, line 72: Why not defining a default behavior (not accepting the pragma) and redefining it only for loop statments?

tests/lit-tests/loop_nounroll.ispc
```diff
@@ -0,0 +1,99 @@
+// Test to check '#pragma nounroll' functionality for different loop statemnts.
+
+// RUN: %{ispc} %s --target=avx2-i32x8 --nowrap -O0 --emit-llvm-text -o - | FileCheck %s -check-prefix=CHECKO0
```
dbabokin, line 3: Any reason to have it for specific x86 target? This should generally be the same for any target?

tests/lit-tests/loop_nounroll.ispc
```diff
@@ -0,0 +1,99 @@
...
+// RUN: %{ispc} %s --target=avx2-i32x8 --nowrap -O2 --emit-llvm-text -o - | FileCheck %s -check-prefix=CHECKO2
+// REQUIRES: X86_ENABLED
+
+// CHECKO0: br label %for_test, !llvm.loop
```
dbabokin, line 7: Label names are not reliable. I suggest wildcarding the label names.

tests/lit-tests/loop_nounroll.ispc
```diff
@@ -0,0 +1,99 @@
...
+
+// CHECKO2: call void @goo_for___uni
+// CHECKO2-NOT: call void @goo_for___uni
+
+extern void goo_for(uniform int);
```
dbabokin, line 12: When there are several functions in the same test, it's good to verify that the check matches the text of specific function. I.e. you need to match beginning of the function first: ``define void @goo_for___``

tests/lit-tests/loop_nounroll.ispc
dbabokin, line 3: If no stdlib is used in the test, it's better to have ``--nostdlib`` in the arguments - it speeds up the compilation.

src/parse.yy
DeepakRajendrakumaran: There is a mistake there. But I think the right way to chain would be in grammar rule for statement. Sfould be:

statement 
       : attribute_list statement
         { ....}
        | statement_base
       ;
This will allow us to handle new pragmas if added here and chain them. In this particular case, throw error since we are trying to set opp unroll commands for same loop.

dbabokin: Agree. Also probably better names here are:
statement_base -> statement
statement -> attributed_statement

src/lex.ll
```diff
@@ -706,64 +714,186 @@ lCComment(SourcePos *pos) {
     Error(*pos, "unterminated comment");
 }
 
+static void lNextValidChar(SourcePos *pos, char const*&currChar) {
+    while (*currChar == ' ') {
```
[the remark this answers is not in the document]
DeepakRajendrakumaran: In lexer we seem to treat space, '\t' and '\r' as whitespace. The lexer line is

WHITESPACE [ \t\r]+

tests/lit-tests/loop_nounroll.ispc
DeepakRajendrakumaran: The names are different for all function. They all start with goo..but it's goo_for vs goo_foreach_unique etc. But yeah..can match the define too

src/lex.ll
DeepakRajendrakumaran: std::stoi throws an exception for characters...something like 'aa'. We can use a try, catch block I suppose. Plus the code I used is the same as we use for identifying integers elsewhere in the lexer. If we plan to change, maybe we should take care of both together?

src/stmt.cpp
```diff
@@ -874,14 +887,17 @@ void DoStmt::EmitCode(FunctionEmitContext *ctx) const {
...
+        ctx->setLoopUnrollMetadata(branchInst, loopAttribute, pos);
     } else {
         // For the varying case, update the mask based on the value of the
         // test.  If any program instances still want to be running, jump
         // to the top of the loop.  Otherwise, jump out.
         llvm::Value *mask = ctx->GetInternalMask();
         ctx->SetInternalMaskAnd(mask, testValue);
-        ctx->BranchIfMaskAny(bloop, bexit);
+        branchInst = ctx->BranchIfMaskAny(bloop, bexit);
+        printf("\n VARYING \n");
```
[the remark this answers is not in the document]
DeepakRajendrakumaran: Fixed

src/lex.ll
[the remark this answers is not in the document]
DeepakRajendrakumaran: I treat all 3 as whitespace now.

src/lex.ll
DeepakRajendrakumaran: Since we treat \r as whitespace, \r\n won.\'t be  a problem. In theory we should ignore the \r as whitespace in 'lNextValidChar' and get to \n

src/lex.ll
```diff
@@ -706,64 +714,186 @@ lCComment(SourcePos *pos) {
...
+/** Consume line starting with '#pragma' and decide on next action based on
+ * directive.
+ */
+static bool lConsumePragma(YYSTYPE *yylval, SourcePos *pos) {
+    char c;
+    std::string userReq;
+    do {
+        c = yyinput();
+        ++pos->last_column;
+    } while (c == ' ');
```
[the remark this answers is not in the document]

src/lex.ll
```diff
@@ -706,64 +714,186 @@ lCComment(SourcePos *pos) {
...
+        return true;
+    }
+    else if (ignoreWarning == userReq.substr(0, ignoreWarning.size())) {
+        pos->last_column += ignoreWarning.size();
+        lPragmaIgnoreWarning(pos, userReq.erase(0, ignoreWarning.size()));
+        return false;
+    }
+
+    // Ignore pragma : invalid directive provided.
+    Warning(*pos, "Undefined #pragma.");
```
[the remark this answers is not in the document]
DeepakRajendrakumaran: Done.

src/parse.yy
DeepakRajendrakumaran: Done + change made to identify conflicting chaining of loop pragmas. The conflict will be identified in setloopattribute()

src/lex.ll
DeepakRajendrakumaran: changed

src/stmt.h
DeepakRajendrakumaran: Done

tests/lit-tests/loop_nounroll.ispc
DeepakRajendrakumaran: Done

tests/lit-tests/loop_nounroll.ispc
DeepakRajendrakumaran: Done

tests/lit-tests/loop_nounroll.ispc
DeepakRajendrakumaran: Done

DeepakRajendrakumaran: Done. Changed to host target

src/parse.yy
```diff
@@ -1667,6 +1681,25 @@ initializer_list
...
```
dbabokin, line 1684: I think ``attribute_list`` is now misleading. In this definition it's not a list. It's a single node.
Just "pragma" or "single_pragma" might be a better name.

src/lex.ll
dbabokin: Need test coverage for that.

src/lex.ll
dbabokin: If this is already the code used in this file, then ok.

tests/lit-tests/loop_nounroll.ispc
dbabokin: You need a check for function start for O0 as well.

src/stmt.h
```diff
@@ -81,6 +84,7 @@ class ExprStmt : public Stmt {
     void Print(int indent) const;
 
     Stmt *TypeCheck();
+
```
dbabokin, line 87: There's a bunch of unintended new empty line in this file. Would be good to remove.

src/stmt.cpp
```diff
@@ -1924,6 +1947,12 @@ Stmt *ForeachStmt::TypeCheck() {
     return anyErrors ? NULL : this;
 }
 
+void ForeachStmt::SetLoopAttribute(std::pair<Globals::pragmaUnrollType, int> lAttr) {
+    if (loopAttribute.first != Globals::pragmaUnrollType::none)
+        Error(pos, "Multiple '#pragma unroll/nounroll' directives used.");
```
dbabokin, line 1952: Need test coverage for this error.

src/stmt.cpp
```diff
@@ -65,6 +65,10 @@
 
 Stmt *Stmt::Optimize() { return this; }
 
+void Stmt::SetLoopAttribute(std::pair<Globals::pragmaUnrollType, int> lAttr) {
+    Error(pos, "Illegal '#pragma' - expected a loop to follow '#pragma unroll/nounroll'.");
```
dbabokin, line 69:
```suggestion
    Error(pos, "Illegal pragma - expected a loop to follow '#pragma unroll/nounroll'.");
```

DeepakRajendrakumaran: Done

src/stmt.cpp
DeepakRajendrakumaran: Done

src/lex.ll
DeepakRajendrakumaran: Added

tests/lit-tests/loop_nounroll.ispc
DeepakRajendrakumaran: Added

src/stmt.h
DeepakRajendrakumaran: Changed

src/stmt.cpp
DeepakRajendrakumaran: Added

src/stmt.cpp
DeepakRajendrakumaran: Changed

dbabokin:
``#pragma unroll`` effect on ``for`` and ``do-while`` loops with ``uniform`` condition is as expected.
``#pragma unroll`` effect on ``for`` loops with varying condition and ``foreach`` is not as expected. Unrolling analysis doesn't understand ``varying`` conditions and just makes multiple copies of loop body, without changing iterator or condition. As a result, after each body there's a condition in the end of the basic block. This eliminates most of unrolling benefits. I suggest disabling it for ``foreach`` as part of this commit and working on enabling correct transformation in a separate PR.
For ``foreach_active`` and ``foreach_unique`` I don't think that there's any point of doing ``#pragma unroll``. I suggest disallowing it.

PR #1891, opened 2020-10-19
don't assume that ncurses & zlib are system libraries

aumuell:
While trying to create a spack package for ispc (https://github.com/spack/spack/pull/19361), I had the problem that zlib and ncurses themselves had been installed via spack. But then, the assumption that they can be found in a system directory, is wrong.

This patch solves the issue when linking dynamically to these libraries.

Spack is a distro-agnostic package manager which compiles software from source, see https://spack.io.


aneshlya (approved):

aneshlya:
Thanks for the fix!

Issue #791, opened 2014-05-20
libispc - ISPC in library form


breakin:
A feature request! (Sorry if you've heard it before... I know I requested it from Matt Pharr when ISPC was new...)

In addition to the ISPC-compiler a C-library could be created. It would use the JIT capabilities of LLVM to let the library user synthesize and execute programs. This would be useful when using ISPC as a shader system in an application. This setup would not work for everyone since there are security implications, but it would be super useful for some.

The library would need to:
- Receive ISPC-sources (preferably no disk access required)
- Enumerate exported ISPC-function (and their arguments)
- Let the user get function pointer to an exported (and JIT:ed) ISPC-function so he can call it
- Be a standalone C-API, pre-compiled on all major platforms with no dependencies (on say dynamic CRT).

The hard part as I see it is in enumerating the argument to the functions. It would need a way to "talk about" all the possible arguments to exported ISPC-functions.

I realize that using ispc.exe with emit-llvm and then hooking it up with LLVM yourself it possible, but that is hard work for most end-users (especially if they are multi-platform).

inducer:
:+1:

hlide:
+1. This is the big reason I do not use ISPC.

capr:
ISPC in lib form would be cool. Building it to dynamically link against an existing LLVM shared lib would be even cooler. This would allow me to use both [terra](http://terralang.org/) and ispc in the same program without duplicating (not to mention recompiling) LLVM. Thanks for thinking about this.

dbabokin:
Dynamically linking against existing LLVM libs is not possible on practice, as ISPC is tightly coupled with particular version of LLVM libs (LLVM API is not stable). We have ifdefs to be able to build with different LLVM versions. Also we sometimes apply patches to LLVM that we use to make it work well for us.

capr:
Yes, and yet it's the only way to have multiple pieces of software that depend on LLVM in the same app (as long as you build them all against the exact same version of LLVM, I understand that). As for patches, are those API-breaking? Is there anything that can be done about it? Adding LLVM twice to an app is a 25 MB hit to the binary.

dbabokin:
Patches are only for stability. We are trying to avoid them, but we have to back-port the fixes to already release versions of LLVM.

Anyway, deciding on static vs dynamic linking is lower order problem than building ispc as library.

damiendr:
I would also very much like this. I use ISPC as a JIT target and the CLI overhead is an issue.

[...]

PR #1893, opened 2020-10-28
WIP: libispc + refactor

tay10r:
Dating all the way back to #791, this PR is addressing the feature request for ISPC has a library.

It was attempted in #1196 but lost steam.

There is a large amount of refactoring that needs to get done in order to make this a reentrant library. For example, `SourcePos`, which seemed innocent at first, contains a default constructor that relies on global variables and is used almost everywhere. The goal of this PR to (in order):

 - remove usage of the `g` and `m` global pointers
 - update with c++17 features
 - replace functions like `Error()` with a extensible `DiagnosticConsumer` class

I chose to advance the standard to c++17 because the attributes it introduces are really good refactoring tools. The newer classes, `string_view` and `optional` are also extremely useful.

The `DiagnosticConsumer` class is similar to the one in [clang](https://clang.llvm.org/doxygen/classclang_1_1DiagnosticConsumer.html) but will look something like this:

```cxx
class DiagnosticConsumer {
public:
  virtual ~DiagnosticConsumer() {}
  virtual void consume(const Diagnostic &) const = 0;
};
```

And `Diagnostic` can be almost as simple as:

```cxx
enum class Severity { ... };

struct Diagnostic final {
  Severity severity = Severity::Error;
  const char *path = "";
  std::size_t line = 1;
  std::size_t column = 1;
  const char *description = "";
};
```

You can either leave this open and comment/review on the progress or close it if you already know it is not going to happen.
Either way I'm happy to hear feedback.


dbabokin:
What compiler are you using?
Building locally on macOS, I have following compilation errors (same as Travis macOS build error):
```
.../build/./lex.cpp:1302:2: error: ISO C++17 does not allow 'register' storage class specifier [-Wregister]
        register yy_state_type yy_current_state;
        ^~~~~~~~~
```

Also I'm a bit puzzled why Linux builds do not find <optional>. We seem to have C++17 lib.

tay10r:
@dbabokin I'm using Clang-10 and GCC 9.3.0 (from Ubuntu 20.04). I've been meaning to look more at the CI errors, but I'm caught up on doing some work on the Flex and Bison parser. I didn't see what was causing the missing C++17 features, I'm curious as well.

tay10r:
@dbabokin I believe the reason that error is popping up is because Xcode probably has an outdated version of flex (see [this relevant issue](https://github.com/westes/flex/issues/132). Newer versions of flex remove the `register` storage specifier, which was removed in C++17.

tay10r:
I've moved the work I'm doing to a seperate repository because:

 - It's all new code anyway, it's a rewrite of the current lexer and parser
 - So that it be licensed under Apache 2.0, similar to most other Intel libraries and LLVM

If I get it to a reasonable working state, I'll write up another PR that is hopefully much less intrusive.

dbabokin:
I'm still interested in porting refactoring changes back to ISPC repo, I think they are useful and right thing to do. I'll look at CI fails and will try to fix them.

As for changing lexer and parser - this is an interesting one. I'm wondering why you need that? I.e. there are plenty of reason to do that. I'm just wondering what is the motivation in your case.

Also, what are you going to use as lexer and parser? I'm specifically interested in this question, as adding more C++ features will require that. And I tired unifying ISPC front-end with clang in the past, but failed. Partly due to attempt to use the whole clang (i.e. lexer, parser, semantic analysis and AST) and looks like the decision to use clang AST was not a good one - it's not really fitted to represent ISPC type system (uniform and varying concept). Though, probably I did it in the wrong way and there's a better way to do that.

tay10r:
> I'm still interested in porting refactoring changes back to ISPC repo, I think they are useful and right thing to do. I'll look at CI fails and will try to fix them.

I'd be happy to reopen so you can cherry pick the refactoring changes (some of the later ones are new code and not part of the refactor). Or I can submit a new PR, whatever is helpful.

> As for changing lexer and parser - this is an interesting one. I'm wondering why you need that? I.e. there are plenty of reason to do that. I'm just wondering what is the motivation in your case.

 - With the direction that C++ is going (for example, constexpr new and constexpr virtual), I thought it would be neat to eventually have a parsing library that was fully constexpr. As I've currently written it, the scanner is fully constexpr but the parser would require `constexpr virtual` for compile time parsing.
 - There's a lot more control over error conditions when using C++ and a recursive descent parser.
 - Not requiring Flex and Bison can make it easier to build on Windows

> Also, what are you going to use as lexer and parser? I'm specifically interested in this question, as adding more C++ features will require that.

Using a hand written scanner and parser. I was using Flex and Bison for a short bit, then I tried out Lemon, and finally went with hand writing them.

> And I tired unifying ISPC front-end with clang in the past, but failed. Partly due to attempt to use the whole clang (i.e. lexer, parser, semantic analysis and AST) and looks like the decision to use clang AST was not a good one - it's not really fitted to represent ISPC type system (uniform and varying concept). Though, probably I did it in the wrong way and there's a better way to do that.

If there is then it's beyond me. I've been avoiding using the clang library because it takes a long time to build and the API changes a lot.

dbabokin:
> I'd be happy to reopen so you can cherry pick the refactoring changes (some of the later ones are new code and not part of the refactor). Or I can submit a new PR, whatever is helpful.

I can try the changes from closed PR, so it's should be ok.

> With the direction that C++ is going (for example, constexpr new and constexpr virtual), I thought it would be neat to eventually have a parsing library that was fully constexpr. As I've currently written it, the scanner is fully constexpr but the parser would require constexpr virtual for compile time parsing.

`Constexpr`s are fun. But is there a practical use case for fully constexpr parser?

> There's a lot more control over error conditions when using C++ and a recursive descent parser.

Totally agree!

> Not requiring Flex and Bison can make it easier to build on Windows.

Yes, totally. Though alternatives like ANTLR are also implying dependencies. That's why I asked about your solution.

> Using a hand written scanner and parser. I was using Flex and Bison for a short bit, then I tried out Lemon, and finally went with hand writing them.

You are brave! Flex and Bison have numerous limitations, I don't care much about lexer, but hand-written parser can do much better than Bison.

> If there is then it's beyond me. I've been avoiding using the clang library because it takes a long time to build and the API changes a lot.

I guess it comes mostly from the complexity of C++ itself and there's nothing we can do about it. C++ is too big and complex.

You also mentioned that you'd like to have your solution under Apache 2.0. Would be good to be able to use it in BSD projects, like ISPC. Personally, I like Apache 2.0 better, but we have no choice for historical reasons.

tay10r:
> Constexprs are fun. But is there a practical use case for fully constexpr parser?

Hard to say, I've yet to write one. I think one practical use might be in the case that an ISPC program is in a C++ string literal, like what's often the case with GLSL. A fully constexpr parser may be able to detect syntax errors at compile time instead of runtime.

I may open another issue or PR when I've got enough of it done to show a minimal working example of the new parser. It'll probably take a little while though.

PR #1899, opened 2020-11-06
Add testing with LLVM 11.0 and other fixes.

dbabokin:
- Adding testing with LLVM 11.0 to Appveyor and Travis.
- Removing testing with LLVM 8.0 from Appveyor and Travis.
- Changing emcc testing for Linux to use fixed version instead of trunk (as we do on Windows).
- Fix alloy.py to fail uniformly with native testing and SDE.
- Fix site template to say that ISPC compiler for CPU *and GPU*.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -32,13 +32,15 @@
 configuration: Release
 
 environment:
-  LLVM_LATEST: 10.0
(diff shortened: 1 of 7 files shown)


aneshlya (approved):
LGTM

PR #1907, opened 2020-11-15
calcite - benchmarks in CI

Lectem:
As discussed with @dbabokin, we at [siliceum](https://siliceum.com) are currently working on [calcite](https://calcite.siliceum.com), a CI performance reporting tool that helps tracking performance regressions in CI.
We offered to provide one of our own stabilized machines for ISPC to test and experiment, and it has already been added to the ispc project Github Actions runners list.

This pull request does the following:

- Allow to install the googlebench library and hence make it available from the build artifacts
  - This is NOT installed in the official packages, so no changes here
  - It is however done in the new workflow that executes when something gets pushed/merged to master, or when a pull request is updated
  - When installation is done, we also patch the `third-party-programs.txt` with google bench licence (Apache2, same as the LLVM Project, so there should be no issue here)
- Add a new github actions workflow
  - Executes when code is pushed to master or in a pull request
  - Builds ispc using LLVM10 (should this now be using LLVM11?) with benchmarks, packages it with benchmarks and googlebench in an artifact (`ispc_llvm10_linux`)
  - Artifact is downloaded on our runner, selected using the `bench` label. This can be changed in this file and from the project settings if needed
  - Benchmarks are ran 10times each
    - This is so that we can obtain statistically significant results. It could probably be lower, but untested (it depends on the performance profiles of the benchmarks).
    - This currently takes ~15minutes, for a total workflow time of ~21 minute including the build.
      - We could: 
        - reduce the number of repetitions per benchmark
        - reduce the minimum time per benchmark globally
        - Select less size combinations for the benchmarks, or disable a few for CI only (based on a define or `CI` env variable?)
        - patch googlebench to give us intermediate loop results (might be done by siliceum at some point)
        - Use a library that actually computes required time correctly based on the clock resolution (nanobench?) or send a patch google bench to do this. (might be done by siliceum at some point)
        - Decide it's ok for now
  - Benchmark results are packaged in an artifact (`ispc_llvm10_linux_bench_results`)
  - Benchmark results are downloaded an another job, then sent to *calcite*
  - If this was triggered by a pull request, a *calcite* workflow is triggered
    - Right now, this compares the current PR merge commit against the destination branch (master) commit used when creating the merge commit. This is why we also run this gh actions workflow on pushes to master
    - If configured to do so, *calcite* will comment the pull request with the detected changes (regressions or improvements). It can also send emails, slack messages, etc...
  - Any benchmark/calcite job failure will not cause the workflow to fail for now.

Right now, only @dbabokin can navigate on the `calcite` platform (detailed diff, benchmark history...) as an account is required. If more accounts are needed for the other maintainers, please get in touch with me.

Of course if you have any question or feedback, feel free to contact me.


Lectem:
@dbabokin It seems the secret `CALCITE_TOKEN` was not set correctly ?

dbabokin:
The token is set, but there's a disclaimer there:
```
Secrets are environment variables that are encrypted. Anyone with collaborator access to this repository can use these secrets for Actions.
Secrets are not passed to workflows that are triggered by a pull request from a fork.
```

Please use the fixes from #1903 as a base for the pipeline. It has a few stability changes, which are required. Also it's preferable to use LLVM 11 now.

dbabokin:
I'm closing this PR, as #1918 included most of the changes.

PR #1908, opened 2020-11-16
fixes #1906 : Reimplementing packed_load/packed_store + adding i64 support.

### LINKED ISSUE #1906: packed_[store|load]_active() for 64 bit types
`packed_store_active`, `packed_load_active`, `packed_store_active2`, `packed_load_active2` need to be extended to handle for i64 types.
Ideally we need to handle all basic type, but this request is specifically for 64 bits.

Related LLVM bug: https://bugs.llvm.org/show_bug.cgi?id=48147

DeepakRajendrakumaran:


dbabokin:
It's WIP, as code quality and performance need to be reviewed. Basically we need to identify the assembly instruction that we want to generate and make sure they are generated.

dbabokin:
I've updated LLVM builds in CI with the latest patches, so it should be ok to add tests that verify that intrinsics work as expected.

DeepakRajendrakumaran:
> I've updated LLVM builds in CI with the latest patches, so it should be ok to add tests that verify that intrinsics work as expected.

Tests are already in there. It was only failing for avx512skx-i32x8 without the patch - tests/packed-load-64-1.ispc for example was one of the failures.

stdlib.ispc
```diff
@@ -1302,43 +1302,80 @@ static unsigned int64 exclusive_scan_or(unsigned int64 v) {
 ///////////////////////////////////////////////////////////////////////////
 // packed load, store
 
+static inline uniform int
+packed_load_active(uniform unsigned int64 a[],
+                   varying unsigned int64 * uniform vals) {
+    return __packed_load_activei64(a, vals, (UIntMaskType)__mask);
+}
+
+static inline uniform int
```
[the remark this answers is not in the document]
DeepakRajendrakumaran: What do you think should be legal combinations? Specifically, what are the legal cross combinations outside of the obvious ones

int32, int32
uint32, uint32
int64, int64
uint64, uint64

What are the cross combination to add to above?

builtins/util.m4
```diff
@@ -6302,6 +6348,24 @@ done:
...
 ')
 
+')
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; packed load and store functions
+;;
+;; These define functions to emulate those nice packed load and packed store
+;; instructions.  For packed store, given a pointer to destination array and 
+;; an offset into the array, for each lane where the mask is on, the
```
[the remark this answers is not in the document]
DeepakRajendrakumaran: Yeah. Those were pre-existing comments. I do not think they are correct.

stdlib.ispc
```diff
@@ -1302,43 +1302,80 @@ static unsigned int64 exclusive_scan_or(unsigned int64 v) {
 ///////////////////////////////////////////////////////////////////////////
 // packed load, store
 
+static inline uniform int
+packed_load_active(uniform unsigned int64 a[],
```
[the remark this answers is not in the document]
DeepakRajendrakumaran: I re-grouped them.

stdlib.ispc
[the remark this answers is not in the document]
dbabokin: (float, float) and (double, double) are definitely valid combinations.

One might try to use shorted types (int8/int16) and these are incorrect uses. I think it's ok not to test it.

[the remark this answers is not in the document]
dbabokin: (float, float) and (double, double) appears to be not a valid combinations, as it's not value type, but pointer. They can be handled through explicit conversions through.

So please ignore this comment, we don't need to test this.

benchmarks/01_trivial/05_packed_load_store.cpp
```diff
@@ -0,0 +1,191 @@
+#include <benchmark/benchmark.h>
+#include <bitset>
+#include <cstdint>
+#include <stdio.h>
+
+#include "../common.h"
+#include "05_packed_load_store_ispc.h"
+
+static Docs docs("Check packed_load_active/packed_store_active implmentation of stdlib functions:\n"
+                 "[int32, uint32, int64, uint64] x [all, even, four, greater] versions.\n");
```
dbabokin, line 10: Looking at `[all, even, four, greater]` I think they are good for experiments, but doesn't worth having all of them running regularly - too much noise, which doesn't have enough information.
I suggest to keep `all` and the version, which stores/loads a single value - the reasoning is that we'd need to measure two extremes. To have a single value version I suggest comparing `programIndex == idx`, where `idx` is function argument - to prevent dead code elimination from removing parts of intrinsic.
Running `all` will have emphasis on doing more memory operations, while single value version will have more emphasis on overhead, like using `popcnt()` to count the active lanes.

builtins/util.m4
```diff
@@ -6302,6 +6262,24 @@ done:
...
+;; packed load and store functions
+;;
+;; These define functions to emulate those nice packed load and packed store
+;; instructions.  For packed store, given a pointer to destination array and 
+;; a varying value, for each lane where the mask is on, the
+;; corresponding value for that lane is stored into packed locations in the
+;; destination array.  For packed load, each lane that has an active mask
+;; loads a sequential value from the array.
+;;
+;; $1: 'TRUE' if intrinsics can be used for '__packed_store_active2'.
```
dbabokin, line 6277: This is correct statement, but would be good to elaborate a little bit. And `intrinsics` -> `LLVM intrinsics`, otherwise it might be confusing.

tests/packed-load-64-1.ispc
```diff
@@ -0,0 +1,13 @@
+export void f_f(uniform float RET[], uniform float aFOO[]) {
+    uniform unsigned int64 a[programCount];
+    #pragma ignore warning(perf)
```
dbabokin, line 3: Why do we have this pragma here, but not in other tests?

tests/packed-load-64-2.ispc
```diff
@@ -0,0 +1,15 @@
+export void f_f(uniform float RET[], uniform float aFOO[]) {
+    uniform int64 a[programCount];
+    a[programIndex] = aFOO[programIndex];
+    int64 aa = 15;
+    uniform int count = 0;
+    if (programIndex < 2)
+        count += packed_load_active(a, &aa);
```
dbabokin, line 7: `count` is dead code and is never checked.

dbabokin (commented):
Also, on Gen new 64 bit intrinsics will fail. Would be good to add them to `fail_db.txt` (by running them on Gen) or add implementation to `util-genx.ll`

tests/packed-load-64-1.ispc
DeepakRajendrakumaran: We get warnings because of 'unsigned -> other type' conversions here

tests/packed-load-64-2.ispc
DeepakRajendrakumaran: I basically used the same tests as i32. So, I have removed count for both. Existing test and the new one.

benchmarks/01_trivial/05_packed_load_store.cpp
DeepakRajendrakumaran: Done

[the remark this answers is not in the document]
DeepakRajendrakumaran: Done

builtins/util.m4
DeepakRajendrakumaran: Done

tests/packed-load-64-1.ispc
DeepakRajendrakumaran: For this, there is an 'unsigned <--> other type' conversion. So, it caused a warning.

tests/packed-load-64-2.ispc
DeepakRajendrakumaran: Yeah. I removed it for this and the original int32 test from which this test was taken from.

benchmarks/01_trivial/05_packed_load_store.ispc
```diff
@@ -0,0 +1,63 @@
...
+    export uniform int32 packed_load_##T_ISPC##_all(uniform T_ISPC *uniform src, uniform T_ISPC *uniform dst,          \
+                                                    uniform int count) {                                               \
+        uniform int32 num = 0;                                                                                         \
+        foreach (i = 0...count) {                                                                                      \
+            varying T_ISPC var = 0;                                                                                    \
+            num += packed_load_active(&src[num], &var);                                                                \
+            dst[i] = var;                                                                                              \
+        }                                                                                                              \
+        return programCount;                                                                                           \
```
dbabokin, line 13: Did you mean to return `num`?

benchmarks/01_trivial/05_packed_load_store.ispc
```diff
@@ -0,0 +1,63 @@
...
+            varying T_ISPC var = src[i];                                                                               \
+            if (programIndex == index)                                                                                 \
+                num += packed_store_active(&dst[num], var);                                                            \
+        }                                                                                                              \
+        return programCount;                                                                                           \
+    }
+
+PACKEDLOAD_ALL(int32)
+PACKEDLOAD_ONE(int32)
+PACKEDSTORE_ALL(int32)
```
dbabokin, line 57: We need `packed_store_active2()` next to `packed_store_active()`.

benchmarks/01_trivial/05_packed_load_store.ispc
DeepakRajendrakumaran: Ahh...I tried a slightly diff signature. It's not needed since I added the 'Width()' function. Should be 'void' return now

tests/packed-load-64-2.ispc
dbabokin: No, checking that return value is correct is important, otherwise we may easily miss the problem in popcnt() implementation. I think both tests need it.

benchmarks/01_trivial/05_packed_load_store.ispc
DeepakRajendrakumaran: I meant to keep it as a void return. Anyway I modified it to return num and check it in the cpp side.

benchmarks/01_trivial/05_packed_load_store.ispc
DeepakRajendrakumaran: Added

tests/packed-load-64-2.ispc
DeepakRajendrakumaran: I think I idea behind these tests originally was each test tests something. For eg. 'packed-load-64-3.ispc' checks the popcnt calculation. Same with 'packed-load-64-5.ispc ' - it also kind of tests the popcnt in it's own way. 
So even though not all tests check the popcnt value, so do

dbabokin: Ok. It's always good to spell out the intention of the test in the comments. Keep that in mind for new tests.

benchmarks/01_trivial/05_packed_load_store.ispc
dbabokin: It needs to be consumed. Either returned or stored somewhere. Otherwise popcnt logic will be removed by dead code. Returning value, on the other side, will require adopting test for GPU. But we'll need to have a dedicated effort for that anyway.

dbabokin:
It's ready to be merge now.

Performance gains for `packed_load_active()` are substantial (1-2.5x) when emulating, i.e. non-AVX512.
Performance for `packed_store_active()` mostly improved (up to 40%), but there are regressions (up to 20%) for some access patterns (AVX2).
`packed_store_active2()` and all of the intrinsics on AVX512 are not affected.

PR #1918, opened 2020-11-25
calcite - benchmarks in CI - retry

dbabokin:
This is slightly modified PR #1907, which doesn't have visibility of organization secrets for some reason.
Resubmitting should fix that.


dbabokin:
Using `pull_request_target` event should enable this pipeline, but requires the workflow to be already committed to master, so I'm pushing the changes and will verify the result by making another PR to test.

PR #1920, opened 2020-11-25
Update calcite cli to 0.7.2 (no more out of date dependencies)

Lectem:
As asked, this updates calcite-cli to be using the latest version of `class-transformer` since 0.2.3 had a CVE.


Lectem:
I also changed the calcite trigger step to run with the new `pull_request_target` instead of `pull_request`.

dbabokin (approved):
LGTM!

PR #1923, opened 2020-11-30
fixes #1755.

### LINKED ISSUE #1755: Crash during varying initialization

DeepakRajendrakumaran:
The hack to return uniform type for bool varying was needed earlier since returning i1 causes a crash when trying to store it. But it's not a problem anymore since now we are storing both uniform and varying as i8 irrespective of mask type.

### DIFF
--- a/src/type.cpp
+++ b/src/type.cpp
@@ -248,31 +248,7 @@ const AtomicType *AtomicType::GetAsUniformType() const {
         return this;
 
     if (asUniformType == NULL) {
-        BasicType bType = basicType;
-        // Varying bool needs to be handled differently since LLVMTypes::BoolVectorType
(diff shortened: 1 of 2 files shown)


DeepakRajendrakumaran:
Testing out the fix.

DeepakRajendrakumaran:
https://github.com/ispc/ispc/issues/1620 : this is the bug which caused the original hack to be put in place.

PR #1927, opened 2020-12-02
[CI] Use safe defaults for calcite benchmarks diff/regressions policies

Lectem:
Due to an oversight during refactoring, diff/regression policies for the calcite datapoints were not set.
The calcite CLI default values were not meant to be used (will be fixed in an upcoming update) as it leads to very noisy results.
The slightest difference that passed our statistical tests would be considered as regressions/improvements, even if the change was as small as 1%. This is of course not what one would want for such benchmarks.

Following a general guideline, we now use the the median to compute the importance of the changes, with a threshold set at 20%.
This is a bit high, but takes into account the fact by default benchmarks (perhaps not the case for ISPC, but this is a general guideline) do not take into account for CPU issues such as the JCC erratum.

The threshold could probably be lowered at a later point in time, but for now I'd rather use a high bar to avoid noise.

### DIFF
--- a/.calcite/calcite.config.js
+++ b/.calcite/calcite.config.js
@@ -1,6 +1,18 @@
 const fs = require('fs');
 const path = require('path');
 
+const defaultDatapointPolicies = {
+    // Median is a safe default to use, see https://doc.calcite.siliceum.com/performance/stability/statistics.html#descriptive-statistics for why we do not pick the mean.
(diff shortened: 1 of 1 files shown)


dbabokin:
20% seem really quite high. I would expect something around 5% to be a threshold. But looks good for now to watch how this plays out.

dbabokin (approved):

PR #1931, opened 2020-12-04
Automatically use CMakes 3.19 ISPC support for examples

robertmaynard:


dbabokin:
For broader review context I'm adding the list of commits to CMake, which add ISPC support enabled in this PR:
https://github.com/Kitware/CMake/commit/34cc6acc81758e29f8c88607c21ab11d8807f87c Add ISPC compiler support to CMake
https://github.com/Kitware/CMake/commit/e783bf8aa6f90ddb1458c7b1a78c5d0225c1496a ISPC: Support ISPC header generation byproducts and parallel builds
https://github.com/Kitware/CMake/commit/2368f46ba42efab00f7a0c7d1b24301f19196176 ISPC: Support building with the MSVC toolchain
https://github.com/Kitware/CMake/commit/8976817d6d4cf8ba5755ab24e2760c4c9cb485de ISPC: Update help documentation to include ISPC 
https://github.com/Kitware/CMake/commit/5a1750017e65727660167ae1953746a3fd3d2c7b ISPC: Add compiler launcher support
https://github.com/Kitware/CMake/commit/a020787a9b7ede3aa490345bd446e469fb1713d2 ISPC: Support generation for multiple instruction sets
https://github.com/Kitware/CMake/commit/c83711d9bd6f6fdbe534f475727bf80f7e8a8603 ISPC tests will now run on any hardware that only supports SSE2 
https://github.com/Kitware/CMake/commit/a83521e082f11fd08027e0f440408914bd83a5eb ISPC: Use the `obj` file extension for objects on windows
https://github.com/Kitware/CMake/commit/8de145cae18a8c368ffce60c511d6f6424c55aa2 ISPC: DynamicLibrary test now passes on windows.
https://github.com/Kitware/CMake/commit/bf88a94d88556ecd93e5612afe3fad99334579fb ISPC: CompilerLauncher tests work properly with x86 builds
https://github.com/Kitware/CMake/commit/e17e2c6a517cf40b41014cdf3e239c0b21da868e ISPC: CheckCompilerFlags and CheckSourceCompiles support ISPC

aneshlya:
Hi @robertmaynard,

Thanks a lot for your pull request, it looks great!
I'm worried a little bit about headers extensions change: "_ispc.h" to ".ispc.h". Historically in all our examples, guides and docs we recommended to use "_ispc.h" extension. I don't want to force people to update their source code with new header names if possible. Can the extension be changed to "_ispc.h"? Or can the header name be configurable from user's side?

aneshlya (changes requested):

examples/cmake/AddISPCExampleModern.cmake
```diff
@@ -0,0 +1,88 @@
...
+#
+# ispc ADDISPCTest.cmake
+#
+function(add_ispc_example)
+    set(options USE_COMMON_SETTINGS)
+    set(oneValueArgs NAME ISPC_SRC_NAME DATA_DIR)
+    set(multiValueArgs ISPC_IA_TARGETS ISPC_ARM_TARGETS ISPC_FLAGS TARGET_SOURCES LIBRARIES DATA_FILES)
+    cmake_parse_arguments("example" "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN} )
+
+    set(ISPC_KNOWN_TARGETS "sse2" "sse4" "avx1-" "avx2" "avx512knl" "avx512skx" "neon")
```
aneshlya, line 42: Is this line needed now?

aneshlya:
We also have GPU support in ISPC for which .spv and .zebin files are produced. You can see specific options/commands here: https://github.com/ispc/ispc/blob/master/ispcrt/cmake/ispc.cmake#L222. Would it be possible to support ISPC for GPU in cmake as well?

pbrubaker:
> Hi @robertmaynard,
> 
> Thanks a lot for your pull request, it looks great!
> I'm worried a little bit about headers extensions change: "_ispc.h" to ".ispc.h". Historically in all our examples, guides and docs we recommended to use "_ispc.h" extension. I don't want to force people to update their source code with new header names if possible. Can the extension be changed to "_ispc.h"? Or can the header name be configurable from user's side?

I have been using `.isph` as the extension for the vscode and VS extensions, and Embree uses it too.  https://www.embree.org/downloads.html

examples/cmake/AddISPCExampleModern.cmake
robertmaynard: No it isn't needed, I will drop it.

examples/cmake/AddISPCExampleModern.cmake
```diff
@@ -0,0 +1,88 @@
...
+        ${example_TARGET_SOURCES}
+      )
+
+    set_property(TARGET ${example_NAME} PROPERTY ISPC_INSTRUCTION_SETS "${ISPC_TARGETS}")
+    target_compile_options(${example_NAME} PRIVATE $<$<COMPILE_LANGUAGE:ISPC>:${example_ISPC_FLAGS}>)
+    target_compile_options(${example_NAME} PRIVATE $<$<COMPILE_LANGUAGE:ISPC>:--arch=${ISPC_ARCH}>)
+
+    if (UNIX)
+      set(arch_flag "-m${ISPC_ARCH_BIT}")
+      target_compile_options(${example_NAME} PRIVATE $<$<COMPILE_LANGUAGE:C,CXX>:${arch_flag}>)
```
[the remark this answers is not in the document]
robertmaynard: I intended to replace the manual usage of `--fpie` and `--pic` with CMake's `POSITION_INDEPENDENT_CODE` abstraction but didn't finish this as the tests built cleanly locally without any PIC flags.

I have corrected this, and now will build with PIC enabled.

robertmaynard:
> Hi @robertmaynard,
> 
> Thanks a lot for your pull request, it looks great!
> I'm worried a little bit about headers extensions change: "_ispc.h" to ".ispc.h". Historically in all our examples, guides and docs we recommended to use "_ispc.h" extension. I don't want to force people to update their source code with new header names if possible. Can the extension be changed to "_ispc.h"? Or can the header name be configurable from user's side?

Given that `_ispc.h` is the current recommended practise I can update CMake to use this pattern. I am not sure that we would want this to be user-configured at this point.

robertmaynard:
> We also have GPU support in ISPC for which .spv and .zebin files are produced. You can see specific options/commands here: https://github.com/ispc/ispc/blob/master/ispcrt/cmake/ispc.cmake#L222. Would it be possible to support ISPC for GPU in cmake as well?

This is something that could be covered by future collaboration between the ISPC and CMake's teams. I would be hestitant to add support right now, as I thought this feature was still experimental and the compiler byproducts/output are still subject to change.

jeffamstutz:
I think it is too early to jump to the conclusion that the current GPU output artifacts is the right approach for composing with CMake. I'd like to see GPU outputs (`.spv` and `.zebin` alike) end up as embedded binary data that is accessed via functions exported in a `_ispc.h` header first before making build systems more complicated.

jeffamstutz:
> I am not sure that we would want this to be user-configured at this point.

@robertmaynard what is the downside of providing something like `CMAKE_ISPC_HEADER_SUFFIX` to make generated header names configurable? I definitely understand wanting to minimize the number of points of configuration in any feature (that's reason enough to be hesitant here), but I'm just curious if there's something else too.

aneshlya:
I completely agree that we don't need to add GPU support to build system right now. Just wanted to ensure that you know that we have GPU support and we'll need CMake changes for that in the future.

aneshlya:
> Given that _ispc.h is the current recommended practise I can update CMake to use this pattern. I am not sure that we would want this to be user-configured at this point.

Looks like even in this pull request review we have different opinions about what extension should be. So it seems that to have it configurable will be quite useful. @dbabokin, what do you think?

examples/cmake/AddISPCExampleModern.cmake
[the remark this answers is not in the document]
aneshlya: Thanks! Works without any issues on my system.

aneshlya (approved):

robertmaynard:
>  what is the downside of providing something like CMAKE_ISPC_HEADER_SUFFIX to make generated header names configurable? I definitely understand wanting to minimize the number of points of configuration in any feature (that's reason enough to be hesitant here), but I'm just curious if there's something else too.

It is partially about the nubmer of configuration points, and also making sure we have captured the necessary requirements.
My initial thought was we could go with a hard-coded "_ispc.h" but I agree it something like `CMAKE_ISPC_HEADER_SUFFIX` would be the better approach.

Since we want to capture the "_ispc.h" the algorithm for `CMAKE_ISPC_HEADER_SUFFIX` would be to take the input source file ( `example.ispc` ) and drop the right-most extension plus `.`, and replace it with `CMAKE_ISPC_HEADER_SUFFIX`. 

This would produce the following transformations:

| CMAKE_ISPC_HEADER_SUFFIX  | input | output |
| ------------- | ------------- | ------------- |
| "_ispc.h"  | simple.ispc  | simple_ispc.h |
| "_ispc.h"  | simple.ispc  | simple_ispc.h |
| ".ispch"  |  simple.ispc  | simple.ispch |

Note: Should we drop only the right most extension or drop all extensions? I think we want only to drop the right most as the transformation of `complicated.ver1.ispc` into `complicated_ispc.h` could clobber headers generated by `complicated.ver2.ispc`

jeffamstutz:
I agree that dropping the right-most extension is what makes the most intuitive sense and is the most robust/future-proof.

jeffamstutz:
I would think for compatibility with CMake 3.19.1, `CMAKE_ISPC_HEADER_SUFFIX` would default to `".ispc.h"`?

robertmaynard:
> I would think for compatibility with CMake 3.19.1, `CMAKE_ISPC_HEADER_SUFFIX` would default to `".ispc.h"`...or is a breaking changing between 3.19.1 and 3.19.2 for this not a big deal?
> 
> FWIW, defaulting to `".ispc.h"` makes sense to me.

Luckily since ISPC was introduced in CMake 3.19 we can break 3.19.0 and 3.19.1 without issue. Personally getting the right defaults now, are worth the very short term pain caused by the breakage.

dbabokin:
@pbrubaker I think there's some confusion here. `*.isph` is an ISPC header file, which compliments `*.ispc` files and contains ISPC code, i.e. they are included by `*.ispc` files. While `*.ispc.h` / `*_ispc.h` in question are generated C/C++ interoperability headers (contain C/C++ code).

pbrubaker:
Yeah, I figured out my mistake, sorry for detailing.

dbabokin:
For the choice between `*_ispc.h` and `*.ispc.h` - it's difficult one, as it's really a matter of taste. Some people tend to name files with multiple dots, while others avoid it by using underscore.

I would vote for underscore as a default and making it configurable. Mostly because this is the way we advertise this pattern for a long time and hence this is the way how people are already using it.

robertmaynard:
> For the choice between `*_ispc.h` and `*.ispc.h` - it's difficult one, as it's really a matter of taste. Some people tend to name files with multiple dots, while others avoid it by using underscore.
> 
> I would vote for underscore as a default and making it configurable. Mostly because this is the way we advertise this pattern for a long time and hence this is the way how people are already using it.

I have opened a merge request that changes the default header suffix to be `_ispc.h` and also provides a configuration point.
https://gitlab.kitware.com/cmake/cmake/-/merge_requests/5597

dbabokin:
@robertmaynard Do you know when the next minor update for CMake 3.19 is coming?

robertmaynard:
> @robertmaynard Do you know when the next minor update for CMake 3.19 is coming?

There is a collection of regressions queued up for 3.19.2, and once they are resolved we should release. I can't give a firm date, but it will be 'soon'.

dbabokin:
This sounds good. 

Can we use `3.19.2` in `CMAKE_VERSION VERSION_GREATER_EQUAL 3.19` or it can be only `3.19`?

I suggest changing this PR to match your changes in CMake and merge it.

robertmaynard:
> This sounds good.
> 
> Can we use `3.19.2` in `CMAKE_VERSION VERSION_GREATER_EQUAL 3.19` or it can be only `3.19`?
> 
> I suggest changing this PR to match your changes in CMake and merge it.

I have updated to activate only on `3.19.2`. I would like to be safe and hold off on merging this till 3.19.2 is released just in case something goes wrong and we need to yank the ISPC changes.

dbabokin:
Sounds good for waiting till 13.19.2 is released.

Should we also roll back the changes in convert `_ispc.h` to `.ispc.h` in this PR?

Also I think we need a CI pipeline, which would test CMake 3.19.2 or later. Either @kraszkow or I will take care about it, once it's released.

robertmaynard:
> Should we also roll back the changes in convert `_ispc.h` to `.ispc.h` in this PR?

I have rolled back to using `_ispc.h`

dbabokin:
`3.19.2` is out. I gave it a try and it fail to build examples with ISPC ([log](https://github.com/dbabokin/ispc/runs/1573066422?check_suite_focus=true)):
```
-- The ISPC compiler identification is unknown
CMake Error at examples/cmake/AddISPCExample.cmake:77 (enable_language):
  No CMAKE_ISPC_COMPILER could be found.

  Tell CMake where to find the compiler by setting either the environment
  variable "ISPC" or the CMake cache entry CMAKE_ISPC_COMPILER to the full
  path to the compiler, or to the compiler name if it is in the PATH.
Call Stack (most recent call first):
  examples/CMakeLists.txt:41 (include)


-- Configuring incomplete, errors occurred!
```

Which is kind of expected... As if we are trying to find ISPC, when we are configuring ISPC build, it's obviously cannot be found... Any ideas how to handle it gracefully?

robertmaynard:
> `3.19.2` is out. I gave it a try and it fail to build examples with ISPC ([log](https://github.com/dbabokin/ispc/runs/1573066422?check_suite_focus=true)):
> 
> ```
> -- The ISPC compiler identification is unknown
> CMake Error at examples/cmake/AddISPCExample.cmake:77 (enable_language):
>   No CMAKE_ISPC_COMPILER could be found.
> 
>   Tell CMake where to find the compiler by setting either the environment
>   variable "ISPC" or the CMake cache entry CMAKE_ISPC_COMPILER to the full
>   path to the compiler, or to the compiler name if it is in the PATH.
> Call Stack (most recent call first):
>   examples/CMakeLists.txt:41 (include)
> 
> 
> -- Configuring incomplete, errors occurred!
> ```
> 
> Which is kind of expected... As if we are trying to find ISPC, when we are configuring ISPC build, it's obviously cannot be found... Any ideas how to handle it gracefully?

Yes we are caught in a chicken and egg problem, as CMake expects that compiler to exist before it is executed.

A couple of thoughts come to mind:
1. We could use the Legacy fallback mode when building ISPC itself. This doesn't allow ISPC to leverage the new feature, but when packaged users will be able too.
2. We could do a fancier version of above, but where if you re-run CMake after you have built the ISPC executable the examples will detect that and switch over to the `Modern` path. This will require some additional changes as we will need to make sure that all the examples depend on the `ISPC` target so they aren't built before the executable on subsequent build invocations.

jeffamstutz:
One option is to use `ExternalProject` to do a unique CMake invocation on examples that executes only after the build of ISPC itself. This would lend itself to the examples being written as a stand-alone CMake build, and invoking them as a part of building ISPC itself is then only one way of building them.

@robertmaynard what do you think?

robertmaynard:
> One option is to use `ExternalProject` to do a unique CMake invocation on examples that executes only after the build of ISPC itself. This would lend itself to the examples being written as a stand alone CMake build, and invoking them as a part of building ISPC itself is then only one way of building them.
> 
> @robertmaynard what do you think?

I prefer this approach now that you bring it up, as it creates an excellent separation of the two components ( ISPC the compiler, and the ISPC examples ).

dbabokin:
> I prefer this approach now that you bring it up, as it creates excellent separation of the two components ( ISPC the compiler, and the ISPC examples ).

I agree, this is the way!

robertmaynard:
I have refactored `examples` and `examples_genx` to use ExternalProject so we can properly use the ISPC compiler we just built.

CMakeLists.txt
```diff
@@ -543,12 +543,32 @@ if (ISPC_INCLUDE_UTILS)
     endif()
 endif()
 
-if (ISPC_INCLUDE_EXAMPLES AND NOT ISPC_PREPARE_PACKAGE)
-    add_subdirectory(examples)
+if ( ISPC_INCLUDE_EXAMPLES AND NOT ISPC_PREPARE_PACKAGE)
```
dbabokin, line 546:
```suggestion
if (ISPC_INCLUDE_EXAMPLES AND NOT ISPC_PREPARE_PACKAGE)
```

CMakeLists.txt
```diff
@@ -543,12 +543,32 @@ if (ISPC_INCLUDE_UTILS)
...
+    include(ExternalProject)
+    ExternalProject_Add(ispc_examples
+        PREFIX examples
+        DEPENDS ispc
+        SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/examples"
+        CMAKE_CACHE_ARGS
+            -DISPC_BUILD:BOOL=TRUE
+            -DCMAKE_ISPC_COMPILER:PATH=$<TARGET_FILE:ispc>
+            -DISPC_EXECUTABLE:PATH=$<TARGET_FILE:ispc>
```
dbabokin, line 555:
```suggestion
            -DISPC_EXECUTABLE:PATH=$<TARGET_FILE:ispc>
            -DCMAKE_C_COMPILER:PATH=${CMAKE_C_COMPILER}
            -DCMAKE_CXX_COMPILER:PATH=${CMAKE_CXX_COMPILER}
```

By default we should use clang used for ISPC build.

CMakeLists.txt
```diff
@@ -543,12 +543,32 @@ if (ISPC_INCLUDE_UTILS)
...
-    add_subdirectory(examples/portable/genx)
+    include(ExternalProject)
+    ExternalProject_Add(ispc_genx_examples
+        PREFIX genx_examples
+        DEPENDS ispc
+        SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/examples/portable/genx"
+        CMAKE_CACHE_ARGS
+            -DISPC_BUILD:BOOL=TRUE
+            -DCMAKE_ISPC_COMPILER:PATH=$<TARGET_FILE:ispc>
+            -DISPC_EXECUTABLE:PATH=$<TARGET_FILE:ispc>
```
dbabokin, line 569:
```suggestion
            -DISPC_EXECUTABLE:PATH=$<TARGET_FILE:ispc>
            -DCMAKE_C_COMPILER:PATH=${CMAKE_C_COMPILER}
            -DCMAKE_CXX_COMPILER:PATH=${CMAKE_CXX_COMPILER}
```

examples/CMakeLists.txt
```diff
@@ -36,7 +36,7 @@
 cmake_minimum_required(VERSION 3.12)
 
 set (PROJECT_NAME ispc_examples)
-project(${PROJECT_NAME})
+project(${PROJECT_NAME} C CXX)
```
dbabokin, line 40:
```suggestion

if(CMAKE_BUILD_TYPE)
    # Validate build type
    set(CONFIGURATION_TYPES "Debug;Release;RelWithDebInfo")

    string(FIND "${CONFIGURATION_TYPES}" "${CMAKE_BUILD_TYPE}" MATCHED_CONFIG)
    if (${MATCHED_CONFIG} EQUAL -1)
         message(FATAL_ERROR "CMAKE_BUILD_TYPE (${CMAKE_BUILD_TYPE}) allows only the following values: ${CONFIGURATION_TYPES}")
    endif()
else(NOT CMAKE_BUILD_TYPE)
    set(CMAKE_BUILD_TYPE "Release")
    message(STATUS "Build type not specified: Use Release by default.")
endif(CMAKE_BUILD_TYPE)

```

Another side effect of moving examples to external project is loosing default build type, while it's critical to build as `Release` by default, as its used for performance evaluation. This is copy-paste of ISPC CMakeLists.txt.

dbabokin (changes requested):
`examples/cmake/*` files use mix of 2 and 4 indentation, style need to be more consistent. In ISPC we typically use 4 spaces indentation.

robertmaynard:
> `examples/cmake/*` files use mix of 2 and 4 indentation, style need to be more consistent. In ISPC we typically use 4 spaces indentation.

I believe I corrected everything to 4 space identation, and have included all your suggested fixes

dbabokin (approved):

dbabokin:
@robertmaynard Thank you for the contribution!

Issue #1930, opened 2020-12-04
Generate fat binaries on OSX


clusty:
Feature Request:
Generating x86_64 and arm64 binaries on OS X is very easy with clang, since it supports passing multiple architecture flags on the same command.

Would it be possible to do the same thing with ISPC.
This would simplify a great deal the build scripts for OSX.

dbabokin:
Do you mean supporting fat binary for x86 and M1? 

We do not support ARM64 on macOS at the moment. Though it might be working out of the box using iOS target. I didn't have a change to test it.

Could you give example how you would compile it using clang, so we would be on the same page?

clusty:
To compile for both x86 and M1 with clang and generate a fat binary, all is needed is to pass **-arch x86_64 -arch arm64** on the compile line.
That was my plan: to test building for iOS, for x86 and then merging the binaries together with **lipo obj1*.o -create -output objFat.o**
As you can see, this makes the OS X building unnecessarily complicated :)

PR #1943, opened 2020-12-10
Enable macOS + Arm64/Neon target for M1 Macs

Twinklebear:
This PR enables support for a macOS + Arm64 target to support M1 based macs.

I think this PR is needed to enable #1930 but I'm not sure it resolves that, I didn't test creating a fat binary for x86 + arm. That might need some additional work, but this PR will at least allow the compilation of a macOS + arm binary with ISPC.

This turned out to be a really minor tweak, but let me know if anything else should be changed here

### DIFF
--- a/.calcite/calcite.config.js
+++ b/.calcite/calcite.config.js
@@ -17,8 +17,9 @@ const fromGoogleBenchFile = (filePath, builder, calciteContext) => {
     const data = calciteContext.readJson(filePath);
     
     const testSuiteName = path.basename(data.context.executable);
-    
-    data.benchmarks.forEach((bench) => {
+
+    const benchsNoAggregate = data.benchmarks.filter(bench => bench.run_type === 'iteration');
+    benchsNoAggregate.forEach((bench) => {
         builder.addOrUpdateDataPoint(
             testSuiteName ,
             bench.run_name,


src/ispc.cpp
```diff
@@ -1334,10 +1334,13 @@ std::string Target::GetTripleString() const {
     case TargetOS::macos:
         // asserts
         if (m_arch != Arch::x86_64) {
-            Error(SourcePos(), "macOS target supports only x86_64.");
+            triple.setArchName("amd64");
+        } else if (m_arch == Arch::aarch64) {
+            triple.setArchName("arm64");
```
dbabokin, line 1339: `arm64` is also an alias for `aarch64`, though LLVM differentiates between arm64 and aarch64 targets. I'm not fully following this logic, but I agree that in this context `arm64` is the right naming here.

dbabokin (changes requested):
Thanks for the contribution. The changes generally look good with except the minor change mention below.

We definitely would like to support M1 chip, this will probably require also adding new CPU definition, though I'm afraid it will be supported only by LLVM trunk.

CI fails are due to MacOSX SDK used for cross build, we should update it to the latest to make it work.

I also would like to postpone merging this commit before I can test it myself (I already ordered hardware), because we have a minor release coming next week and I don't want to add the target that I cannot test myself.

src/ispc.cpp
Twinklebear: I wasn't quite sure which to pick here, I saw that iOS used arm64 as the string but it looks like aarch64 also works fine so I can change this to the preferred string

Twinklebear:
Thanks @dbabokin ! I fixed those issues, no problem on postponing the merge until you can test it more. In this PR I really just tried making it not an error to have macOS + arm and things seemed to build and run ok. I am able to run the ISPC examples and Embree + ISPC stuff with this but I didn't do any in depth validation testing.

Twinklebear:
Just a small update, I tried merging this with the latest master to bring in the changes for 1.15 and noticed that ISPC was outputting x86_64 binaries instead of aarch64, so there may be something conflicting there or with llvm 11 that needs to change to output Arm binaries correctly again

src/ispc.cpp
```diff
@@ -1334,10 +1334,13 @@ std::string Target::GetTripleString() const {
     case TargetOS::macos:
         // asserts
         if (m_arch != Arch::x86_64) {
-            Error(SourcePos(), "macOS target supports only x86_64.");
+            triple.setArchName("x86_64");
+        } else if (m_arch == Arch::aarch64) {
+            triple.setArchName("aarch64");
```
dbabokin, line 1339:
```suggestion
            triple.setArchName("arm64");
```

dbabokin:
I guess it's changing the name of arch, which screwed the things up. So even though `arm64` is a synonym for `aarch64`, looks like it's not that simple. I've committed the change that reverts it. Try it now. I don't remember any other changes that might affect it.

Also, rebasing to ispc:master should solve Appveyor problem, as I've turned off macOS cross target on Windows and Linux.

dbabokin (approved):

Twinklebear:
It looks like it might be related to a change from LLVM 10 to 11, when I build this PR rebased on top of master, but change the build script to still fetch LLVM 10 I'm able to build the examples. When using LLVM 11, I get the same link error that the object files were compiled for x86_64.

dbabokin:
How do you compile examples?
This command line seems to produce proper ARM assembly (on my intel-based Mac, with LLVM 11).
```
bin/ispc foo.ispc --target=neon --emit-asm -o foo.s
```

Twinklebear:
Pretty much the same way, just with `--target=neon-i32x4 --arch=aarch64` and outputting an object file instead of asm. Let me check what asm comes out with LLVM 11, will need a second to rebuild it haha

dbabokin:
I was wrong, I tried it on Linux machine, instead of Mac (I need more coffee this morning). My Mac is 10.15, while for M1 support I'll need to update to BigSur.

dbabokin:
I've built with new (BigSur) SDK and can reproduce it now:
```
> cmake .. -DISPC_MACOS_SDK_PATH=/Library/Developer/CommandLineTools/SDKs/MacOSX11.0.sdk
> make
<...>
Error: Unknown arch.
make[2]: *** [examples/aobench_instrumented/ao_instrumented_ispc.h] Error 1
```

Twinklebear:
I think that's the same I'm getting then, the full error when trying to build the examples is:

```
[ 42%] Built target ispc
[ 42%] Linking CXX executable ../../bin/aobench
ld: warning: ignoring file ao_ispc.o, building for macOS-arm64 but attempting to link with file built for unknown-x86_64
Undefined symbols for architecture arm64:
  "_ao_ispc", referenced from:
      _main in ao.cpp.o
  "_ao_ispc_tasks", referenced from:
      _main in ao.cpp.o
ld: symbol(s) not found for architecture arm64
clang-11: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [bin/aobench] Error 1
make[1]: *** [examples/aobench/CMakeFiles/aobench.dir/all] Error 2
make: *** [all] Error 2
```

If I just use the ispc compiler built with LLVM 11 to compile an object file:

```
bin/ispc ../examples/aobench/ao.ispc --target=neon-i32x4 --arch=aarch64 -o ao.o
```

Then the output object file is actually for x86_64:

```
% file ao.o
ao.o: Mach-o 64-bit object x86_64
```

src/ispc.cpp
```diff
@@ -1334,10 +1334,13 @@ std::string Target::GetTripleString() const {
     case TargetOS::macos:
         // asserts
         if (m_arch != Arch::x86_64) {
```
dbabokin, line 1336: Fix for correct selection of macOS arch.
```suggestion
        if (m_arch == Arch::x86_64) {
```

dbabokin:
I'm not sure why it was working before :) Should work now, at least on my Mac it works.

dbabokin:
@Twinklebear The changes look good now, but let me add tests and do some clean up. I'll push-force rebased / cleaned up changes to your branch.

dbabokin:
Hmmmm, doing push-force to the source branch of this PR I occasionally push "master", instead of "my branch" to target "master". So PR closed and I can't updated the branch anymore. I've resubmitted the changes in PR #1954.

Twinklebear:
Shoot I'm not sure how it was working earlier before either based on that line haha. I'm just rebuilding to check now but https://github.com/ispc/ispc/pull/1954/files looks good

Twinklebear:
No worries about the force push/closing of the PR

PR #1954, opened 2020-12-22
Enable macOS + Arm64/Neon target for M1 Macs (resubmitting PR)

dbabokin:
This is resubmission of https://github.com/ispc/ispc/pull/1943


dbabokin:
@Twinklebear Could you please verify that this branch works well for you on M1 Mac?
Please do `make` and then `make check-all`.

One thing that is missing in this PR is check that used MacOS SKD is 11.0 or later.

Twinklebear:
Cool will do, just starting up the build now

Twinklebear:
The two checks in `macos_targets.ispc`:

```
// RUN: %{ispc} %s --arch=x86-64  --target-os=macos --target=avx2-i32x8 --emit-asm -o - | FileCheck %s --check-prefix=CHECK_X86

// RUN: not %{ispc} %s --arch=x86 --target-os=macos --target=avx2-i32x8 --nowrap -o %t.o 2>&1 | FileCheck %s --check-prefix=CHECK_ERROR
```

Initially failed on the mini since I didn't have X86 support enabled by by default in cmake (https://github.com/ispc/ispc/blob/master/CMakeLists.txt#L52). After enabling that make check-all passes, but those checks should probably not run if x86 support wasn't built, since the x86/x86-64 arch flags won't be supported.

dbabokin:
Heh, I'm taking take x86 availability for granted :) 
I think it's final version now. Should work with old and new macOS SDK (for old ARM target is disabled), with x86 target enabled and disabled on macOS.

Twinklebear:
Yeah I think it looks good as well!

dbabokin:
@Twinklebear Thank you for the contribution!

PR #1956, opened 2020-12-23
Github Actions: refactor flows, add testing CI

kraszkow:


.github/workflows/ispc-ci.yml
```diff
@@ -0,0 +1,212 @@
+name: ISPC CI workflow
+
+on:
+  schedule:
+    # Run daily - test sse2-avx512 targets @ -O0/-O1/-O2
+    - cron:  '0 7 * * *'
+  push:
```
dbabokin, line 7: I think we don't need it for `push`

.github/workflows/ispc-ci.yml
```diff
@@ -0,0 +1,212 @@
...
+      if: failure()
+      with:
+        name: fail_db.${{matrix.target}}.txt
+        path: fail_db.txt
+
+    - name: Check
+      run: |
+        # Print fails to the log.
+        git diff
+        # Exit with error code is there are fails.
```
dbabokin, line 168: I know it was copied from existing file, but would be good to fix the typo: `is` -> `if`

.github/workflows/ispc-ci.yml
```diff
@@ -0,0 +1,212 @@
...
```
dbabokin, line 210: `is` -> `if`

.github/workflows/linux-nightly-trunk.yml
```diff
@@ -78,26 +77,21 @@ jobs:
...
-        bin/ispc --support-matrix
-        make check-all
-        make ispc_benchmarks && make test
+        .github/workflows/scripts/check-ispc.sh
 
     - name: Upload package
       uses: actions/upload-artifact@v2
       with:
-        name: ispc_llvm_trunk_linux
+        name: ispc_trunk_linux
```
dbabokin, line 89: It was intentionally `ispc_llvm_trunk_linux`, as `ispc_trunk_linux` refers to ISPC trunk package (regardless LLVM version used).

Both names are terrible and misleading, but `ispc_llvm_trunk_linux` is a better one.

.github/workflows/ispc-ci.yml
```diff
@@ -0,0 +1,212 @@
+name: ISPC CI workflow
```
dbabokin, line 1: The name is not informative.
It belongs to ISPC repo, so `ISPC` in the name is not useful.
It's a GitHub Action, so it's `CI` anyway.
And it's workflow, so no need to say `workflow` :)

Right now it's Linux only, but I assume we will add Windows and macOS to the same pipeline, so probably we should avoid saying Linux.

Then probably just `Tests`?

.github/workflows/scripts/build-ispc.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash -e
+echo PATH=$PATH
+mkdir build && cd build
```
Lectem, line 3: I think it's a good time to use the cmake commands to avoid calling mkdir and make.

See https://github.com/ispc/ispc/blob/0593138b0091464be9681df539ca692a142cfe75/.github/workflows/linux-benchmarks.yml#L47-L48

Note that it will make it easier to switch to ninja for faster build at a later point.

.github/workflows/scripts/check-ispc.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash -e
+cd build
+bin/check_isa
+bin/ispc --support-matrix
+make check-all
```
Lectem, line 5: Should we also replace those calls to make by `cmake --build . --target check-all ispc_benchmarks test` ?

Lectem (commented):
Should we also move [linux-benchmarks.yml](.github/workflows/linux-benchmarks.yml) to be using this ? Or wait for this PR to be merged and then update it ?

.github/workflows/scripts/build-ispc.sh
dbabokin: I don't have much experience with `cmake --build`, so I was just doing it the old way.
But probably it worthy try it and see if we get any problems.

.github/workflows/scripts/check-ispc.sh
dbabokin: Sounds like a good idea.

.github/workflows/ispc-ci.yml
```diff
@@ -0,0 +1,212 @@
...
+                    (github.event_name == 'workflow_dispatch' && github.event.inputs.flow_type == 'full') }}
+
+      # define rule when to run smoke flow
+      RUN_SMOKE: ${{ github.event_name == 'push' ||
+                    (github.event_name == 'workflow_dispatch' && github.event.inputs.flow_type == 'smoke') }}
+    steps:
+    - id: set-tests-matrix
+      run: |
+         # set tests matrix depends on flow
+         $RUN_SMOKE && echo "::set-output name=matrix::${TARGETS_SMOKE}" || true
```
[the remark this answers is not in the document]
kraszkow: Done.

.github/workflows/scripts/build-ispc.sh
```diff
@@ -0,0 +1,5 @@
...
+cmake .. -DISPC_PREPARE_PACKAGE=ON -DISPC_INCLUDE_BENCHMARKS=ON
+make -j4 package
```
dbabokin, line 5:
```suggestion
cmake -B build -DISPC_PREPARE_PACKAGE=ON -DISPC_INCLUDE_BENCHMARKS=ON
cmake --build build --target package -j4
```

.github/workflows/scripts/check-ispc.sh
```diff
@@ -0,0 +1,6 @@
...
+make ispc_benchmarks && make test
```
dbabokin, line 6:
```suggestion
cmake --build . --target check-all ispc_benchmarks test
```

dbabokin (approved):
LGTM

dbabokin:
> Should we also move [linux-benchmarks.yml](.github/workflows/linux-benchmarks.yml) to be using this ? Or wait for this PR to be merged and then update it ?

@Lectem yes, would be good to switch to use the same scripts where possible. Note that `scripts/build-ispc.sh` doesn't include benchmarks in the install, so probably not this one.

PR #1961, opened 2021-01-02
CI adjustments and editorial changes

dbabokin:
* reduce amount (and time) of `smoke` testing
* Introduce testing with `-g` in `full` mode
* enabled LLVM10 testing for `smoke`
* Fix ISPCRT version to match ISPC version
* Editorial changes


kraszkow (approved):
I've couldn't find "linux-test-debug-llvm11" run example but if it works then it looks good for me 👍

PR #1966, opened 2021-01-05
Remove LLVM 8.0 and LLVM 9.0.

DeepakRajendrakumaran:


dbabokin:
Missing changes in Appveyor.
Our Travis is not running anymore, but would be good to edit it as well.

alloy.py
```diff
@@ -121,24 +121,14 @@ def checkout_LLVM(component, version_LLVM, target_dir, from_validation, verbose)
...
-    # git: "origin/release/9.x"
+    # An example of using branch (instead of final tag) is the following (for 10.0):
+    # git: "origin/release/10.x"
     if  version_LLVM == "trunk":
         GIT_TAG="main"
     elif  version_LLVM == "11_0":
         GIT_TAG="llvmorg-11.0.0"
     elif  version_LLVM == "10_0":
         GIT_TAG="llvmorg-10.0.1"
-    elif  version_LLVM == "9_0":
```
dbabokin, line 132: Let's keep the ability to build LLVM - that might be useful sometimes. For example to bootstrap LLVM on old OS. I had to do that on macOS 10.10, where to build LLVM 10.0 you need to do that in several steps 6.0->8.0->10.0.

DeepakRajendrakumaran: Removed the alloy changes to build LLVM. Added change to disable appveyor run for 9.0 since it now fails. Waiting to see if it skips LLVM 9.0 now.

dbabokin: Appveyor changes look good, please also Fix `.travis.yml`, it's straightforward enough to fix that without checking.

dbabokin:
Missing parts: CMake (grep for `LLVM_VERSION_NUMBER`) and lit tests (tests themselves and config).

alloy.py
DeepakRajendrakumaran: Travis fixed too and tests passing.

PR #1968, opened 2021-01-06
fixes #1963: Removes incorrect perf warnings.

### LINKED ISSUE #1963: unsigned integer to floating point conversion performance warning is unnecessary on some targets.
ISPC sometimes warns about conversion from unsigned integer to floating point when it is unnecessary.

```
<source>:3:25: Performance Warning: Conversion from unsigned int to 
        float is slow. Use "int" if possible 
    varying float foo = src[programIndex];
                        ^^^^^^^^^^^^^^^^^
```

We probably don't need it at all for AVX512.  In AVX2 it looks like `vcvtdq2ps` is generated for 8-bit and 16-bit integers, signed and unsigned - so the warning doesn't need to be present on those types.

Also, in converting from `int64` and `uint64` to `float` the compiler doesn't generate a warning about possible data loss.  Just the perf warning.

Example Code:

https://ispc.godbolt.org/z/cGWdzn

DeepakRajendrakumaran:


tests/lit-tests/1963.ispc
```diff
@@ -0,0 +1,33 @@
+// This test does uint->float varying typecast and verifies that performance
+// warning is printed only for relevant cases.
+
+//; RUN: %{ispc} %s --emit-asm -o %t.sse2.s --nowrap --target=sse2 2>&1 | FileCheck %s -check-prefix=CHECK_SSE2_WARNING
```
dbabokin, line 4: Do not use shot form of target, i.e. `sse2`, `avx2` - they are depricated.

src/ispc.h
```diff
@@ -407,6 +409,9 @@ class Target {
 
     /** Indicates whether the target has FP64 support. */
     bool m_hasFp64Support;
+
+    /** Indicates whether the target has uint32 -> float cvt support **/
+    bool m_hascvtu32Support;
```
dbabokin, line 414: I don't like the name, as it's in line with other properties, which define **functionality**. This one is explicitly for warnings. I would name it `m_warnFtoU32IsExpensive`.

dbabokin (commented):
The rest looks good.

tests/lit-tests/1963.ispc
DeepakRajendrakumaran: Switched to sse2-i32x4 and avx2-i32x8.

src/ispc.h
DeepakRajendrakumaran: Changed name to 'm_warnFtoU32IsExpensive'.

dbabokin:
Looks good now.

One more thing, though. It's 2021 now, which means it's time to update copyright in the affected files :)

PR #1969, opened 2021-01-06
[Calcite] Remove the need for package.json 

Lectem:
Dependabot currently warns about security issues if calcite-cli is outdated because we use a `package.json` file to install it, which is not needed.

This pull request uses `npx` instead to directly install/cache the version we want when needed.
This has the advantage of not requiring `package(-lock).json` files and make it easier to change the version of the calcite CLI.
The only downside is that the calcite cli version is now stored in the github action instead of somewhere near the `calcite.config.js` file.


dbabokin:
@Lectem Thank you for the quick fix!

PR #1972, opened 2021-01-07
fixes #1836 - adds aos_to_soa2/soa_to_aos2 library implementation.

### LINKED ISSUE #1836: aos_to_soa2/soa_to_aos2 for floats and doubles.
This need came up on Twitter today as well as in some of the work I'm doing with another game developer using Complex Numbers and FFTs.  It'd be handy to have for both floats and doubles, but a float version would be the highest priority.

https://twitter.com/noel_austin/status/1288131255360528393

DeepakRajendrakumaran:


benchmarks/01_trivial/01_aossoa.cpp
```diff
@@ -63,6 +64,24 @@ template <typename T> static void check3(T *dst, int count) {
     }
 }
 
+template <typename T> static void check2(T *dst, int count) {
```
dbabokin, line 67: Looks like it was a bad idea in the first place to have `check3()` and `check4()` as separate functions (blame on me), as aos-"factor" may be passed as a parameter. Could you please refactor it to make it a single function for `check2()`/`check3()`/`check4()` by passing an extra argument to the function?

benchmarks/01_trivial/02_soaaos.cpp
```diff
@@ -20,8 +20,23 @@ static Docs docs("soa_to_aos*_stdlib_<type> - test for stdlib implimentation for
 // 256 * sizeof (int) << 12 = 4 Mb - expected to reside in L3.
 #define ARGS4 Arg(256)->Arg(256 << 4)->Arg(256 << 7)->Arg(256 << 12)
 #define ARGS3 Arg(192)->Arg(192 << 4)->Arg(192 << 7)->Arg(192 << 12)
+#define ARGS2 Arg(128)->Arg(128 << 4)->Arg(128 << 7)->Arg(128 << 12)
 
 // Helper functions
+template <typename T> static void init2(T *src, T *dst, int count) {
```
dbabokin, line 26: Same here, could you please refactor it be a single `init()` function?

dbabokin (commented):
Looks good, but please do refactoring in the benchmarks as noted below.

benchmarks/01_trivial/01_aossoa.cpp
DeepakRajendrakumaran: Done

benchmarks/01_trivial/02_soaaos.cpp
DeepakRajendrakumaran: Done

PR #1978, opened 2021-01-12
WIP fixes #1886 : fixing typecast with variability not specified.

### LINKED ISSUE #1886: Wrong or not intuitive type cast
In the following code example I would expect that the varying function will be called. ```float``` by default is varying, so a cast to ```(float)``` should be ```(varying float)```. Instead ISPC casts it into ```(uniform float)```. The example code should print:
```
uniform func
varying func
[15.000000,15.000000,15.000000,15.000000,15.000000,15.000000,15.000000,15.000000,15.000000,15.000000,15.000000,15.000000,15.000000,15.000000,15.000000,15.000000]
```
but instead creates an infinite recursion.

Tested with ISPC 1.14.1 on Linux.

#### test.ispc
```c++
float func(float A, float B) {
        print("varying func\n");
        return A * B;
}

float func(float A, uniform float B) {
        print("uniform func\n");
        return func(A, (float)B); // adding 'varying' in the type cast fixes the problem.
}

export void test() {
        float A = 5;
        uniform float B = 3;
        print("%\n", func(A, B));
}
```

#### test.cpp
```c++
extern "C" void test(void);

int main(int argc, const char** argv) {
        test();
        return 0;
}
```

#### build steps
```bash
ispc -o test.o test.ispc
g++ test.o test.cpp
./a.out
```

DeepakRajendrakumaran:


dbabokin (commented):
Looking at this PR I realized that the fix is too aggressive. For most of the cases it's convenience feature, which has conservative default (i.e. don't change variability). If the variability needs to be changed, it will be changed by implicit convert. There's no possibility for mistake. So it's a good feature, which  maks the language more compact and easier to write on.

The case with function call is different. In the context of function argument this type of conversion matters and may lead to different overload resolution. So it's potentially dangerous.

Ideal fix would be to warn users if the convert leads to different variability of the function argument and suggest to use explicit variability in this case. Documentation also needs to be clarified to explain this behavior explicitly.

I'm wondering if there are other contexts which need explicit variability resolution. It *seems* that calling a function is only such case.

dbabokin:
We can probably close this PR for future reference and implement new solution (with a warning) as a new PR.

DeepakRajendrakumaran:
Closing this PR based on above comments.

PR #1979, opened 2021-01-12
Add windows workflow for Github Actions CI

kraszkow:


dbabokin (approved):
Looks great, thanks!

A few changes that still needed (keep it here as a todo for the next commit):
- fail_db artifacts need to be named differently for 10 and 11 (same for Linux)
- SDE needs to be installed on Windows.
- would be good to have "check environment" step as on Linux. Information about host CPU is critical for understanding why something passes or fails.

PR #1981, opened 2021-01-14
fixes #1821 - Removing LLVMGetName() and switching to LLVM::Twine.

### LINKED ISSUE #1821: ISPC code needs refactoring to start using llvm::Twine()
For specifying the name of LLVM instructions ISPC uses ``const char*``, while LLVM moved to use ``llvm::Twine`` class, which is very nice way to avoid creation of temporary objects. ISPC need use ``llvm::Twine``.

The problem with ``const char*`` is that it's very difficult to manage temporay objects lifetime: ``LLVMGetName()`` (see [here](https://github.com/ispc/ispc/blob/151a11348dc8f9a909d43f66f1826bbf232c7641/src/llvmutil.cpp#L1563)) uses ``strdup()`` to create the temporary object, which needs to be freed, if not used. So moving to ``llvm::Twine()`` should solve the problem of memory leaks.

DeepakRajendrakumaran:
Planning to switch to llvm::Twine for the entire repo for (const char *) in stages. This is stage 1.

### DIFF
--- a/src/ctx.cpp
+++ b/src/ctx.cpp
@@ -1020,8 +1020,8 @@ void FunctionEmitContext::EmitCaseLabel(int value, bool checkMask, SourcePos pos
     llvm::BasicBlock *bbCaseImpl = NULL;
     if (emitGenXHardwareMask()) {
         // Create basic block with actual case implementation
-        std::string bbName = bbCase->getName().str() + "_impl";
-        bbCaseImpl = CreateBasicBlock(bbName.c_str(), bbCase);
+        llvm::Twine bbName = llvm::Twine(bbCase->getName()) + "_impl";
+        bbCaseImpl = CreateBasicBlock(bbName, bbCase);
     }
 #endif
 
@@ -1185,7 +1185,7 @@ bool FunctionEmitContext::initLabelBBlocks(ASTNode *node, void *data) {
     if (ctx->labelMap.find(ls->name) != ctx->labelMap.end())
         Error(ls->pos, "Multiple labels named \"%s\" in function.", ls->name.c_str());
     else {
-        llvm::BasicBlock *bb = ctx->CreateBasicBlock(ls->name.c_str());
+        llvm::BasicBlock *bb = ctx->CreateBasicBlock(ls->name);
         ctx->labelMap[ls->name] = bb;
     }
     return true;
@@ -1299,7 +1299,7 @@ llvm::Value *FunctionEmitContext::Any(llvm::Value *mask) {
     // We can actually call either one, since both are i32s as far as
     // LLVM's type system is concerned...
     llvm::Function *fmm = mm[0]->function;
-    return CallInst(fmm, NULL, mask, LLVMGetName(mask, "_any"));
+    return CallInst(fmm, NULL, mask, llvm::Twine(mask->getName()) + "_any");
 }
 
 llvm::Value *FunctionEmitContext::All(llvm::Value *mask) {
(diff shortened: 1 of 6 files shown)


src/ctx.cpp
```diff
@@ -779,8 +779,8 @@ void FunctionEmitContext::jumpIfAllLoopLanesAreDone(llvm::BasicBlock *target) {
         allDone = MasksAllEqual(finishedLanes, blockEntryMask);
     }
 
-    llvm::BasicBlock *bAll = CreateBasicBlock("all_continued_or_breaked");
-    llvm::BasicBlock *bNotAll = CreateBasicBlock("not_all_continued_or_breaked");
+    llvm::BasicBlock *bAll = CreateBasicBlock(llvm::Twine(llvm::Twine("all_continued_or_breaked")));
```
dbabokin, line 782: Why `twine(twine(...))`?

dbabokin: Do we need to construct `Twine` explicitly? I think if the function accepts `Twine` as an argument, then it will be constructed if a matching `Twine` constructor is found. I.e. no need to change invocation.

DeepakRajendrakumaran: Ah. I cleaned it up.

src/ctx.cpp
```diff
@@ -1364,11 +1364,12 @@ llvm::Value *FunctionEmitContext::MasksAllEqual(llvm::Value *v1, llvm::Value *v2
 #else
     if (g->target->getArch() == Arch::wasm32) {
         llvm::Function *fmm = m->module->getFunction("__wasm_cmp_msk_eq");
-        return CallInst(fmm, NULL, {v1, v2}, LLVMGetName("wasm_cmp_msk_eq", v1, v2));
+        return CallInst(fmm, NULL, {v1, v2}, (llvm::Twine("wasm_cmp_msk_eq", v1->getName()) + v2->getName()));
```
dbabokin, line 1367: Should we have underscores inserted between "wasm_cmp_msk_eq", v1->getName() and v2->getName()?

src/ctx.cpp
```diff
@@ -2248,20 +2273,22 @@ llvm::Value *FunctionEmitContext::SwitchBoolSize(llvm::Value *value, llvm::Type
     llvm::Value *newBool = value;
     if (g->target->getDataLayout()->getTypeSizeInBits(fromType) >
         g->target->getDataLayout()->getTypeSizeInBits(toType)) {
-        if (name == NULL)
-            name = LLVMGetName(value, "_switchBool");
+        if (name.isTriviallyEmpty()) {
```
dbabokin, line 2276: Should we do that once before the if statement?

src/ctx.cpp
DeepakRajendrakumaran: I changed the way I added strings and added the underscore

src/ctx.cpp
DeepakRajendrakumaran: I changed it. But not to before if statement. 
1. There are conditions where from and to types have same size where we won't go into either the if or else condition
2. If it's outside, a new llvm:Twine will have to be created.

src/ctx.cpp
```diff
@@ -1364,11 +1364,12 @@ llvm::Value *FunctionEmitContext::MasksAllEqual(llvm::Value *v1, llvm::Value *v2
 #else
     if (g->target->getArch() == Arch::wasm32) {
         llvm::Function *fmm = m->module->getFunction("__wasm_cmp_msk_eq");
-        return CallInst(fmm, NULL, {v1, v2}, LLVMGetName("wasm_cmp_msk_eq", v1, v2));
+        return CallInst(fmm, NULL, {v1, v2}, llvm::Twine("wasm_cmp_msk_eq_") + v1->getName() + "_" + v2->getName());
```
dbabokin, line 1367: This seems to be dependent on the order of expression evaluation, which is not guaranteed.
I.e. if evaluated from right to left, it will yield an error.
Adding extra braces for ordering expressions seems to be ok-ish solution here.

DeepakRajendrakumaran: Done

src/ctx.cpp
dbabokin: Creating Twine is very cheap - that's by design. So if Twine is created and then discarded it should not be expensive.
Anyway, I'm fine with current code.

PR #1984, opened 2021-01-18
Fix fails reporting in ISPC CI

dbabokin:
`alloy.py` return non-zero error code only if there were problems during testing (i.e. the process failed to complete, etc). Failing tests are not triggering non-zero error code. So `if: failure()` would not trigger in case of fails.

`git diff --exit-code` produces non-zero exit code if there are differences, so the job will be indicated as failing and fail log will be uploaded.


kraszkow (approved):
LGTM.

PR #1985, opened 2021-01-19
fixes #1921 :  fix + lit test.

### LINKED ISSUE #1921: Anonymous structs cause undefined error in ISPC
Hey

When I try to compile the following example, ISPC crashes without any meaningful error message (see below).

## Code
```cpp
typedef struct {
    varying float x;
    varying int8 y;
} float_int8_t;

uniform float_int8_t init(float x, int8 y) {
    uniform float_int8_t out = {x, y};
    return out;
}
```

## ISPC Output
```
/usr/local/src/ispc/src/module.cpp:862: Assertion failed: "m->errorCount > 0".
***
***
/usr/local/src/ispc/src/main.cpp(372): FATAL ERROR: Unhandled signal sent to process; terminating.
Aborted (core dumped)
```

The code compiles, when the first line is changed to ```typedef struct float_int8 {```.

DeepakRajendrakumaran:


src/type.cpp
```diff
@@ -1555,9 +1556,8 @@ StructType::StructType(const std::string &n, const llvm::SmallVector<const Type
         // Create a unique anonymous struct name if we have an anonymous
         // struct (name == ""), or if we are creating a derived type from
```
dbabokin, line 1557: 1. comment needs to be update.
2. do we still create "derived type from an anonymous struct"?

src/type.cpp
```diff
@@ -1599,6 +1599,15 @@ StructType::StructType(const std::string &n, const llvm::SmallVector<const Type
             lStructTypeMap[mname]->setBody(elementTypes);
         }
     }
+    // Create a unique anonymous struct name if we have an anonymous struct (name == "")
+    // Ensuring struct is created with a name, prevents each use of original
+    // struct from having different names causing type match errors.
+    if (name == "") {
+        char buf[16];
+        snprintf(buf, sizeof(buf), "$anon%d", count);
```
dbabokin, line 1607: We will create two names per anonymous struct - here and in the code above. Is intended?

src/type.cpp
DeepakRajendrakumaran: Yeah. The comment/derived part is still valid.  When variability is specified, this path is taken.

src/type.cpp
DeepakRajendrakumaran: It'll not create two names. If it goes through the earlier path(i.e variability is mentioned, name won't be " "(empty) anymore. If unbounded, name will be set here..otherwise at the earlier place.

src/type.cpp
dbabokin: How does this works, as it's refer to "e.g. the varying variant--name == '$'" part, which was removed?

src/type.cpp
dbabokin: Ok, makes sense.

src/type.cpp
DeepakRajendrakumaran: '$' was a kind of hack to identify later that we are dealing with anonymous struct. I never found  a case where that code was hit. If this was intended for creating separate names for anonymous when variability is mentioned, then that doesn't make sense either since we mangle the name with variability later too(lmangleStructName). The only place where we were using '$' was for printing name. I changed implementation is flag it as anonymous which I think is better.

tests/lit-tests/1921.ispc
```diff
@@ -0,0 +1,21 @@
+// RUN: %{ispc} --target=host --nowrap %s > %t 2>&1
+// RUN: FileCheck --input-file=%t %s
+// CHECK-NOT: Please file a bug
```
dbabokin, line 3:
```suggestion
// RUN: %{ispc} --target=host %s -o %t.o
```

Note the `-o`, which triggers code generation, which also might be buggy.

src/type.cpp
dbabokin: Ok, looks like there's no way for ISPC to generate names starting with $.

But the comment about `varying variant--name == '$'` is misleading now.

PR #1986, opened 2021-01-20
fixes #1886 : Add warning for possible ambiguous function calls.

### LINKED ISSUE #1886: Wrong or not intuitive type cast

DeepakRajendrakumaran:


dbabokin:
First of all, the logic in the fix is not correct.
```c
uniform float u;
varying float v;
foo(((float)u) + u); // should trigger a warning, but will not.
```

Second, documentation should have a clear description of this problem. People will google for this warning and should get there.

Third, the warning should point out to the specific type convert.

DeepakRajendrakumaran:
> First of all, the logic in the fix is not correct.
> 
> ```c
> uniform float u;
> foo(((float)u) + u); // should trigger a warning, but will not.
> ```
> 
> Second, documentation should have a clear description of this problem. People will google for this warning and should get there.
> 
> Third, the warning should point out to the specific type convert.

Second and third points can be addressed once we decide on exact nature of change

First point - That was by design but your point makes sense. If it's a binary operator with one ambiguous and one uniform, we still have to show warning. I'll make the change

dbabokin:
@DeepakRajendrakumaran Please have a look at the changes that I pushed. If they look good to you, it's ready to be merged.

DeepakRajendrakumaran:
> @DeepakRajendrakumaran Please have a look at the changes that I pushed. If they look good to you, it's ready to be merged.

LGTM

PR #1992, opened 2021-01-22
Return back zip cpack generator on windows

aneshlya:
msi and zip files are generated simultaneously.


dbabokin:
I think we need both. Not sure about default, probably it should be zip. Both we need msi as well.

PR #2000, opened 2021-01-29
Turning on HoistCommonInsts flag for CFGSimplificationPass.

DeepakRajendrakumaran:


src/opt.cpp
```diff
@@ -482,6 +482,10 @@ void Optimize(llvm::Module *module, int optLevel) {
     }
     optPM.add(llvm::createIndVarSimplifyPass());
 
+#if ISPC_LLVM_VERSION > ISPC_LLVM_11_0
```
dbabokin, line 485: This needs to be >= LLVM 12 to handle correctly LLVM 11.1

```suggestion
#if ISPC_LLVM_VERSION >= ISPC_LLVM_12_0
```

src/opt.cpp
```diff
@@ -517,7 +521,11 @@ void Optimize(llvm::Module *module, int optLevel) {
 #endif
         optPM.add(llvm::createFunctionInliningPass());
         optPM.add(CreateMakeInternalFuncsStaticPass());
+#if ISPC_LLVM_VERSION <= ISPC_LLVM_11_0
```
dbabokin, line 524: The same, LLVM 11.1 needs to be handled correctly.
So it needs to be ">= 12.0" and "else".

dbabokin (changes requested):

src/opt.cpp
DeepakRajendrakumaran: Done

src/opt.cpp
DeepakRajendrakumaran: Done

dbabokin (approved):

dbabokin:
LGTM

PR #2012, opened 2021-02-05
Turn on SinkCommonInst and move hoist past ISPC memory Improvement pass.

DeepakRajendrakumaran:


dbabokin:
I think we need a test to specific code transformations, which happened / didn't happen.

PR #2018, opened 2021-02-16
Adding 'llvm.assume' intrinsic and stdlib option.

DeepakRajendrakumaran:


docs/ispc.rst
```diff
@@ -3999,6 +4000,16 @@ for an optimized release build), use the ``--opt=disable-assertions``
...
 
 
+Compiler Optimization Hints
+---------------------------
+
+The ``ispc`` standard library includes a mechanism for adding ``assume()``
+statements to ``ispc`` program code. The ``assume()`` function takes a
+single uniform boolean expression as an argument. This expression is
+assumed to be ``true`` and this information will be used for optimization
+when possible. This feature is currently only available on CPU targets.
```
dbabokin, line 4010: Other points to make here:
- code generation should not be affected by `assume()` and not runtime checks are done, the information is used for optimizations only.
- list of examples that should work.

src/stmt.cpp
```diff
@@ -3657,6 +3650,44 @@ void AssertStmt::EmitCode(FunctionEmitContext *ctx) const {
     free(errorString);
 }
 
+void AssertStmt::EmitAssumeCode(FunctionEmitContext *ctx, const Type *type) const {
+
```
dbabokin, line 3654: Typically we don't have extra vertical space in the beginning of the function.
```suggestion
```

builtins/util.m4
```diff
@@ -5686,6 +5686,15 @@ define void @__memset64(i8 * %dst, i8 %val, i64 %len) alwaysinline {
     ret void
 }
 
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; assume
+declare void @llvm.assume(i1)
+
+define void @__do_assume_uniform(i1 %test) alwaysinline {
```
dbabokin, line 5693: Do we need a mask argument?

tests/lit-tests/assume_uniform.ispc
```diff
@@ -0,0 +1,94 @@
...
+//; CHECK_SSE2-NOT: test
+
+//; CHECK_AVX2: foo2
+//; CHECK_AVX2-NOT: test
+
+//; CHECK_AVX512: foo2
+//; CHECK_AVX512-NOT: test
+
+// Case 2 : pointer null check removed.
+void bar2(uniform int * uniform a) {
```
dbabokin, line 35:
```suggestion
inline void bar2(uniform int * uniform a) {
```

builtins/util.m4
DeepakRajendrakumaran: For uniform? I based it off assert implemetation. We just use the condition to make decision

```
declare i32 @puts(i8*)
declare void @abort() noreturn

define void @__do_assert_uniform(i8 *%str, i1 %test, <WIDTH x MASK> %mask) {
  br i1 %test, label %ok, label %fail

fail:
  %call = call i32 @puts(i8* %str)
  call void @abort() noreturn
  unreachable

ok:
  ret void
}
```

dbabokin: yeah, this is correct.

tests/lit-tests/assume_uniform.ispc
DeepakRajendrakumaran: Done.

src/stmt.cpp
DeepakRajendrakumaran: Done.

docs/ispc.rst
DeepakRajendrakumaran: 1. 'code generation should not be affected by assume()' - this might be a confusing statement. Code generation is affected if optimization happens, right? Do you think "code generation should not be aborted by assume and no runtim..." works better?
2. Will add snippets from lit test.

dbabokin: Right, this might be confusing. "The condition used in `assume()` statement will not be code generated and does not imply runtime checks. It will solely be used as optimization hint, if compiler is able to use this information."

dbabokin:
Please manually check that documentation is successfully "compiled" after your changes and look good. To compile documentation use `docs/build.sh` script on Linux.

docs/ispc.rst
```diff
@@ -3999,6 +4000,84 @@ for an optimized release build), use the ``--opt=disable-assertions``
...
+        assume(((uniform uint64)((void*)Source1) & (32 * TARGET_WIDTH)-1) == 0);
+        assume(((uniform uint64)((void*)Result) & (32 * TARGET_WIDTH)-1) == 0);
+        uniform AlignedFloat S1;
+        S1[programIndex] = Source1[programIndex];
+        const uniform AlignedFloat R = S1;
+        Result[programIndex] = R[programIndex];
+   }
+
+The ``assume()`` hint informs the compiler that memories used for loading and
+storing are aligned. This results in aligned instructions instead of
```
dbabokin, line 4078:
```suggestion
The ``assume()`` hint informs the compiler that memory location used by loads and
stores are aligned. This results in aligned instructions instead of
```

DeepakRajendrakumaran: Done!

DeepakRajendrakumaran:
> Please manually check that documentation is successfully "compiled" after your changes and look good. To compile documentation use `docs/build.sh` script on Linux.

I did not know that. Tested.

docs/ispc.rst
DeepakRajendrakumaran: Done

PR #2023, opened 2021-02-19
Adding trunc() to stdlib.

DeepakRajendrakumaran:


builtins/util.m4
```diff
@@ -1533,6 +1533,41 @@ ret <16 x double> %ret
...
+}
+
+declare <WIDTH x double> @llvm.trunc.v`'WIDTH`'f64(<WIDTH x double> %val)
+define <WIDTH x double> @__trunc_varying_double(<WIDTH x double> %val) {
+  %r = call <WIDTH x double> @llvm.trunc.v`'WIDTH`'f64(<WIDTH x double> %val)
+  ret <WIDTH x double> %r
+}
+')
+
+truncate()
```
dbabokin, line 1569: Even though, instantiating `truncate()` macro unconditionally in the body of util.m4 works, it's the bad way to have it. `ceil/floor/round` are implemented differently (this would be good to refactor and unify, but this is the way it is now) in every builtins file specific for every targets. `truncate()` macro instantiation ideologically need to be next to `ceil/floor/round` implementation.

Would be good to unify `ceil/floor/round` as a separate commit, also. But eventually I see it as calling some `roundings()` macro in all the targets.

dbabokin (commented):
To properly test `floor/ceil/round/trunc` the following values need to be tested:
- positive with fraction in range (0, 0.5)
- positive with fraction in range (0.5, 1.0)
- negative with fraction in range (0, 0.5)
- negative with fraction in range (0.5, 1.0)

Also special cases need to be checked (they need to be preserved):
- +/- 0.0
- +/- inf
- +/-NaN

I suggest creating similar tests for `floor/ceil/round/trunc`.

builtins/util.m4
DeepakRajendrakumaran: I'll add the trucate(0 instantiation near floor/ceil

tests/trunc-double-uniform.ispc
```diff
@@ -0,0 +1,127 @@
+
```
dbabokin, line 3: double tests should be skipped on DG1/TGLLP

tests/trunc-double-uniform.ispc
```diff
@@ -0,0 +1,127 @@
...
+export void f_f(uniform float RET[], uniform float aFOO[]) {
+
```
dbabokin, line 5: We don't typically use vertical white space in he beginning of the function.

tests/trunc-double-uniform.ispc
```diff
@@ -0,0 +1,127 @@
...
+    uniform float error = 0;
+    RET[programIndex] = 0;
+    uniform double deltaLow = 0.000001d;
+    uniform double deltaHigh = 0.999999d;
+    uniform double testVal;
+    uniform double resVal;
+    uniform double baseVal;
+
+    //Case 1 : Generic value 1 < val < +inf with deltaLow.
```
dbabokin, line 14: I would expect it to be formatted like `// Case 1: bla-bla` - note whitespace after // and no white space before `:`.

tests/trunc-double-uniform.ispc
```diff
@@ -0,0 +1,127 @@
...
+
+    //Case 4 : Generic value -inf < val < -1 with deltaHigh.
+    baseVal = -aFOO[programCount - 1];
+    testVal = baseVal - deltaHigh;
+    resVal = trunc(testVal);
+    if (baseVal != resVal) {
+        error++;
+    }
+
+    //Case 5 : Value 0 < val < +0.5.
```
dbabokin, line 46: Is there a difference between 0 < x < 1 and 1 < x +inf? This seems to be pretty much the same, I don't think it needs to be tested separately.

tests/trunc-double-uniform.ispc
```diff
@@ -0,0 +1,127 @@
+
+export void f_f(uniform float RET[], uniform float aFOO[]) {
+
+    uniform float error = 0;
+    RET[programIndex] = 0;
+    uniform double deltaLow = 0.000001d;
+    uniform double deltaHigh = 0.999999d;
```
dbabokin, line 9: No need to make these values small / close to 1. `0.2` and `0.8` would be better.
Small / close to 1 values are confusing and might make people think that this is somehow important and means something.

tests/trunc-double-uniform.ispc
```diff
@@ -0,0 +1,127 @@
...
+    resVal = trunc(testVal);
+    if (testVal != resVal) {
+        error++;
+    }
+
+    //Case 13 : Value -nan.
+    baseVal = -aFOO[programCount - 1];
+    testVal = sqrt(baseVal);
+    resVal = trunc(testVal);
+    if (!isnan(resVal)) {
```
dbabokin, line 112: It would be good to check that it's the same -nan as well (i.e. the sign didn't change):
`!isnan(resVal) || resVal != testVal `

tests/trunc-double-varying.ispc
```diff
@@ -0,0 +1,78 @@
...
+    checkVal = baseVal;
+    if (baseVal < 0) {
+        checkVal = -abs(baseVal + 1); // -abs() Needed for -0.0
+    }
+    if (checkVal != resVal) {
+        error++;
+    }
+
+    //Case 2 : Covers range  -(programCount/2 -1) < val < +(programCount/2) including -1 < val < +1 with deltaHigh.
+     // So, this covers +0.0, -0.0, -inf < val < 0 with decimal less than 0.5, 0 < val < +inf with decimal greater than 0.5
```
dbabokin, line 30: Identation.

dbabokin (changes requested):

tests/trunc-double-uniform.ispc
DeepakRajendrakumaran: It's skipped. I added in the next commit.

tests/trunc-double-uniform.ispc
DeepakRajendrakumaran: Added check for sign bit.

tests/trunc-double-varying.ispc
DeepakRajendrakumaran: Done.

tests/trunc-double-uniform.ispc
DeepakRajendrakumaran: Done

tests/trunc-double-uniform.ispc
DeepakRajendrakumaran: Added comment explaining

tests/trunc-double-uniform.ispc
DeepakRajendrakumaran: Changed

tests/trunc-double-uniform.ispc
DeepakRajendrakumaran: Removed.

tests/trunc-double-varying.ispc
dbabokin: This fixed in other tests, but not in this one.

tests/trunc-double-uniform.ispc
dbabokin: This is fixed in most of the cases, but not in all. Please go through all the tests and fix it.

tests/trunc-double-uniform.ispc
dbabokin: Two tests still not updated.

tests/ceil-double-varying.ispc
```diff
@@ -0,0 +1,72 @@
...
+    RET[programIndex] = 0;
+    double deltaLow = 0.2d;
+    double deltaHigh = 0.8d;
+    double testVal;
+    double resVal;
+    double baseVal;
+    double checkVal = 0;
+
+    // Case 1: Covers range  -(programCount/2 -1) < val < +(programCount/2) including -1 < val < +1 with deltaLow.
+    // So, this covers +0.0, -0.0, -inf < val < 0 with decimal greater than 0.5, 0 < val < +inf with decimal less than 0.5
```
dbabokin, line 14: This doesn't cover `+0.0` and `-0.0` as special cases. I.e. it doesn't test that `ceil(+0.0) == +0.0`. So they need to be checked separately.

dbabokin (commented):
The rest looks good. After adding +/-0.0 varying cases it can be merged.

tests/ceil-double-varying.ispc
DeepakRajendrakumaran: It's good to go now.

dbabokin (approved):
LGTM now!

Issue #2015, opened 2021-02-10
CMake for examples need to be able to do cross builds.


dbabokin:
CMake for examples need to be able to do cross builds for another OS and another architecture. Specifically, cases like Windows->Linux and x86->ARM need to be covered.

jmbuzano:
Working on this one

PR #2024, opened 2021-02-19
CMake: Check MSVC before adding compiler and linker options.

Biswa96:
This sets correct options while compiling with GCC or Clang in Win32.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -406,18 +406,18 @@ target_include_directories(${PROJECT_NAME} PRIVATE
                            ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_CFG_INTDIR})
 
 # Compile options
-if (UNIX)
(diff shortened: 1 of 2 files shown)


dbabokin:
This makes sense to me and looks like it's part of the fix for #2015.
@aneshlya could please you have a look at this?

aneshlya (approved):
LGTM

Biswa96:
Speaking of cross-compilation, I compiled this in msys2/mingw-w64 environment. Though ispc is compiled it can't be used to compile other projects showing random errors with stdlib.ispc.

dbabokin:
>Though ispc is compiled it can't be used to compile other projects showing random errors with stdlib.ispc.

This means that target files from builtins/*.ll were built incorrectly.

I'm not sure that building ispc itself in msys2/mingw-w64 environment has any practical significance. More practical goal is to build ispc using clang toolchain in MSVC. And the most useful thing is to build examples for other platforms (Linux/PS4/Android).

dbabokin:
This PR doesn't affect the build with clang toolchain on Windows, it works both before and after the fix. I'm merging this PR.

PR #2041, opened 2021-03-18
Removing 'dead' pass.

DeepakRajendrakumaran:
This is essentially dead code. 'FixBooleanSelectPass::fixSelect' is not called and the 'runOnFuction()' is basically a dummy call.


dbabokin:
LGTM.

PR #2042, opened 2021-03-19
Use python 3.6 in Ubuntu 16.04 docker images to allow LLVM trunk build

dbabokin:


dbabokin:
Manually triggered GH Action for LLVM trunk build has passed CMake config stage, so it's good to merge.

PR #2045, opened 2021-03-22
Fixes #294: Adding alloca() implementation.

### LINKED ISSUE #294: provide alloca() in stdlib

DeepakRajendrakumaran:


src/lex.ll
```diff
@@ -434,6 +436,7 @@ return { RT; return TOKEN_RETURN; }
 soa { RT; return TOKEN_SOA; }
 signed { RT; return TOKEN_SIGNED; }
 sizeof { RT; return TOKEN_SIZEOF; }
+alloca  { RT; return TOKEN_ALLOCA; }
```
dbabokin, line 439:
```suggestion
alloca { RT; return TOKEN_ALLOCA; }
```

src/expr.cpp
```diff
@@ -7708,6 +7708,63 @@ std::pair<llvm::Constant *, bool> SizeOfExpr::GetConstant(const Type *rtype) con
...
+    const Type *t = expr ? expr->GetType() : NULL;
+    if (t != NULL)
+        printf(" [type %s]", t->GetString().c_str());
+    printf(")");
+    pos.Print();
+}
+
+Expr *AllocaExpr::TypeCheck() {
+    if (expr == NULL) {
+        Error(pos, "\"allocate()\" cannot have a \"NULL\" argument.");
```
dbabokin, line 7743: It should be `alloca()` in the error message, not `allocate()`

tests/allocate.ispc
```diff
@@ -0,0 +1,19 @@
+// rule: skip on arch=genx32
+// rule: skip on arch=genx64
+
+void check_val(uniform int *uniform alloca_ptr, uniform float RET[], uniform float aFOO[]) {
```
dbabokin, line 4: I assume you meant to make this function `noinline`

src/expr.cpp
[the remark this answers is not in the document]
DeepakRajendrakumaran: The following code snippet will trigger this-
void *ptr = alloca(nullptr);

It's not a likely scenario but it is still a user error
That's why I thought it makes sense as a user error message.

But looking at clang implementation, it just gets converted to a alloca for size 0 - https://godbolt.org/z/nrxhK7qq9

src/ctx.cpp
```diff
@@ -2591,6 +2591,43 @@ void FunctionEmitContext::addGSMetadata(llvm::Value *v, SourcePos pos) {
...
+    }
+
+    llvm::AllocaInst *inst = NULL;
+    if (atEntryBlock) {
+        // We usually insert it right before the jump instruction at the
+        // end of allocaBlock
+        llvm::Instruction *retInst = allocaBlock->getTerminator();
+        AssertPos(currentPos, retInst);
+        unsigned AS = llvmFunction->getParent()->getDataLayout().getAllocaAddrSpace();
+        inst = new llvm::AllocaInst(llvmType, AS, size, name, retInst);
```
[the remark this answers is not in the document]
DeepakRajendrakumaran: We essentially need 'alloca i8, i64 <size>, align 16' for this and this is the corresponding LLVM alloca instruction even for alloca(1) -> https://godbolt.org/z/4KqxYhTdM.  Let me know if I misunderstood the question

tests/allocate.ispc
DeepakRajendrakumaran: Done

src/expr.cpp

src/lex.ll

src/ctx.cpp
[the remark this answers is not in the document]
dbabokin: Right, we need an array. Non-array is for allocating space for local variables, which is different use case for alloca.

src/expr.cpp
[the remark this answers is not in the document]
dbabokin: I don't understand this. `nullptr` is an expression converted to `0`, which means AllocaExpr will have an `expr`, which will be a constant zero. Missing an expression is another story and it seems that it could possibly be only in case of broken compiler logic / something caused an error during AST build and for some reason we didn't crashed (though we should).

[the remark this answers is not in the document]
dbabokin: > It's made more complicated by the fact that we might end up with a null for expr if we cannot evaluate an expr during compile time. So, it's a possibility that it's a compiler error as well

I sounds like we should crash when we can't evaluate the expression, not here.

tests/allocate.ispc
```diff
@@ -0,0 +1,19 @@
+// rule: skip on arch=genx32
+// rule: skip on arch=genx64
+
+void noinline check_val(uniform int *uniform alloca_ptr, uniform float RET[], uniform float aFOO[]) {
+    if (alloca_ptr[programIndex] == aFOO[programIndex]) {
+        RET[programIndex] = 0;
+    }
+}
+
+export void f_f(uniform float RET[], uniform float aFOO[]) {
```
dbabokin, line 10: I suggest testing alloca() in different CFG contexts in the function, not just in entry basic block. I.e. under the if, in the loop (don't do many iterations!).

src/expr.cpp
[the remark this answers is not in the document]
DeepakRajendrakumaran: Okay. I changed the check. Now, we check for nullexpr in parser and handle it there.

tests/allocate.ispc
DeepakRajendrakumaran: Added new test conditions.

dbabokin (commented):
Looks good, but missing docs update.

DeepakRajendrakumaran:
> Looks good, but missing docs update.

Do you think it should be under 'The ISPC Standard Library'? If yes, do create a new section called 'Stack memory allocation'? I ask because 'Dynamic Memory Allocation' is a separate section under 'Expression'

dbabokin:
Yep, I think it should be in standard library section. It's standard library from user prospective, even though the implementation is not really a library solution, but rather a builtin.

PR #2047, opened 2021-03-24
Adding check for PS4 + macro.

DeepakRajendrakumaran:


src/module.cpp
```diff
@@ -2296,6 +2296,9 @@ void Module::execPreprocessor(const char *infilename, llvm::raw_string_ostream *
     opts.addMacroDef("ISPC");
     opts.addMacroDef("PI=3.1415926535");
 
+    if (g->enableLLVMIntrinsics) {
+        opts.addMacroDef("ISPC_LLVM_INTRINSICS_DEFINED");
```
dbabokin, line 2300: I think `ISPC_LLVM_INTRINSICS_ENABLED` is a better name.

dbabokin (commented):
Also need to add docs for the define.

src/module.cpp
DeepakRajendrakumaran: Done!. I added some info in doc as well but the build script for doc fails with something above.

PR #2050, opened 2021-03-26
Adding option to discard value names.

DeepakRajendrakumaran:


src/main.cpp
```diff
@@ -207,6 +207,7 @@ static void lPrintVersion() {
     printf("    [--print-target]\t\t\tPrint target's information\n");
     printf("    [--fuzz-test]\t\t\tRandomly perturb program input to test error conditions\n");
     printf("    [--fuzz-seed=<value>]\t\tSeed value for RNG for fuzz testing\n");
+    printf("    [--[no-]discard-value-names]\t\tDo not discard/Discard value names when generating LLVM IR.\n");
```
dbabokin, line 210: An option with [no-] should be sorted in alphabetical order as if [no-] is not present.
Could you sort the whole output for `devUsage()`? The sorting broken currently.

dbabokin (changes requested):
Could you please rebase the PR and add this to newly added `tests/lit-tests/full_mask.ispc`, which also uses variable names and needs them to pass.

src/main.cpp
DeepakRajendrakumaran: Done

dbabokin (approved):

dbabokin:
LGTM

PR #2056, opened 2021-04-06
builtins/cpu: Give __num_cores C-linkage when building WASM

MarijnS95:
The `__num_cores` intrinsic is not found when compiling with stdlib for WASM:

    stdlib.ispc:1765:12: Error: Undeclared symbol "__num_cores". Did you mean "num_cores"?

This function was moved from a c to cpp file in 2f6d077e ("New print implementation with support of genx.") but `extern "C"` wasn't added to every function declaration, leading to a mangled symbol name that cannot be found by its demangled C name.


dbabokin:
Good catch. Thanks for the PR.

PR #2058, opened 2021-04-09
Fixes #2038 : Perform integer promotion for shift.

### LINKED ISSUE #2038: Incorrect code behavior - compiler bug
I have provided a short example of some ISPC source code which causes incorrect compilation / bad code behaviour.

It seems as though the logical not operator isn't being evaluated when inlined.

You can comment out the EXECUTE_BAD_CODE macro to see the expected behaviour of the Test() function.

https://godbolt.org/z/Peq5Tb

DeepakRajendrakumaran:


aneshlya:
Closing this PR since it's not finished. We may revisit it in a future.

Issue #2052, opened 2021-03-31
Suboptimal codegen for ARM NEON


akaStiX:
While investigating codegen for ARM I found a few issues.
Let's start with a simple Vector transformation by Matrix:
```
uniform FVector4 VectorTransformVector2(const uniform FVector4 &VecP, const uniform FMatrix &Matrix)
{
	uniform FVector4 Result;
    const uniform FVector4 *uniform M = (const uniform FVector4 *uniform) &Matrix;

	// Mul by the matrix
	Result = VectorMultiply(M[0], VecP.V[0]);
	Result = VectorMultiplyAdd(M[1], VecP.V[1], Result);
	Result = VectorMultiplyAdd(M[2], VecP.V[2], Result);
	Result = VectorMultiplyAdd(M[3], VecP.V[3], Result);

	return Result;
}
```
this generates ideal code for AVX2:
```
        vbroadcastss    xmm0, dword ptr [rdi]
        vbroadcastss    xmm1, dword ptr [rdi + 4]
        vbroadcastss    xmm2, dword ptr [rdi + 12]
        vmulps  xmm1, xmm1, xmmword ptr [rsi + 16]
        vmulps  xmm2, xmm2, xmmword ptr [rsi + 48]
        vfmadd231ps     xmm1, xmm0, xmmword ptr [rsi] # xmm1 = (xmm0 * mem) + xmm1
        vbroadcastss    xmm0, dword ptr [rdi + 8]
        vfmadd231ps     xmm2, xmm0, xmmword ptr [rsi + 32] # xmm2 = (xmm0 * mem) + xmm2
        vaddps  xmm0, xmm1, xmm2
        ret
```
Now let's look at ARM:
```
        ext     v1.16b, v0.16b, v0.16b, #8
        and     v3.8b, v0.8b, v1.8b
        mov     x8, x0
        ld1r    { v0.4s }, [x8], #4
        ldp     q2, q1, [x1]
        mov     w9, v3.s[1]
        fmul    v0.4s, v2.4s, v0.4s
        ld1r    { v2.4s }, [x8]
        fmov    w8, s3
        tst     w8, w9
        fmla    v0.4s, v1.4s, v2.4s
        ldp     q1, q4, [x1, #32]
        ldp     s2, s3, [x0, #8]
        fmla    v0.4s, v1.4s, v2.s[0]
        fmla    v0.4s, v4.4s, v3.s[0]
        ret
```
For comparison this is code written by me using NEON intrinsics:
```
[...]

PR #2059, opened 2021-04-14
[Arm64] use across lanes reduction instructions

sebpop:


builtins/target-neon-i32x4.ll
```diff
@@ -258,6 +258,175 @@ define <WIDTH x double> @__sqrt_varying_double(<WIDTH x double>) nounwind readno
...
+                      <2 x i32> <i32 0, i32 1>
+  %v1 = shufflevector <4 x i64> %0, <4 x i64> undef,
+                      <2 x i32> <i32 2, i32 3>
+  %sum = add <2 x i64> %v0, %v1
+  %m = call i64 @llvm.aarch64.neon.saddv.i64.v2i64(<2 x i64> %sum)
+  ret i64 %m
+}
+
+define i64 @__reduce_min_int64(<4 x i64>) nounwind readnone alwaysinline {
+  reduce4(i64, @__min_varying_int64, @__min_uniform_int64)
```
dbabokin, line 414: Should these functions that still use `reduce4` not be duplicated?

builtins/target-neon-i32x4.ll
```diff
@@ -258,6 +258,175 @@ define <WIDTH x double> @__sqrt_varying_double(<WIDTH x double>) nounwind readno
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; reductions
 
+dnl $1: scalar type
```
dbabokin, line 261: I had to look up M4 docs to figure out what `dnl` is, we typically use just comments.
Though it's a matter of taste, I'm ok with it if you prefer it this way.

dbabokin (commented):
Are you going to do the similar change for other neon targets? Would be good to have it uniform across targets.

Generally it looks good to me, but I need to play with it on ARM machines as our CI on ARM is not working at the moment (you can see TravisCI fails, which you can ignore).

builtins/target-neon-i32x4.ll
sebpop: You're right.  I will push a patch to fix these.

sebpop:
> Are you going to do the similar change for other neon targets? Would be good to have it uniform across targets.
> 

Unfortunately Arm32 cannot be modified in a similar way: the NEON cross lane reduction instructions have been added to Arm64, and they do not exist in Arm32.

> Generally it looks good to me, but I need to play with it on ARM machines as our CI on ARM is not working at the moment (you can see TravisCI fails, which you can ignore).

By updating my tree to push the PR I have seen that there are a few new tests failing on Arm64 (with and without my patch):

Failing Tests (6):
    ispc :: 1886.ispc
    ispc :: 1921.ispc
    ispc :: discard_value_names.ispc
    ispc :: full_mask.ispc
    ispc :: loop_nounroll.ispc
    ispc :: loop_unroll.ispc

I will investigate which commit started failing these tests on Arm64.

I also need to run some more correctness and performance tests and make sure all the patterns in the bug report are caught and fixed by the patch.

sebpop:
git-bisect points to the following patches causing the fails on Arm64:
> ispc :: 1886.ispc
```
commit 85d9f0b0d8126dfba721b8ddf7f2cd2a7d97ada9
Author: Deepak Rajendrakumaran <deepak.rajendrakumaran@intel.com>
Date:   Tue Jan 19 13:21:44 2021 -0800

    fixes #1886 : patch to add warning + lit test.
```

> ispc :: 1921.ispc
```
commit a657b70c4873740c959d5aa4470e6aebc1997550
Author: Deepak Rajendrakumaran <deepak.rajendrakumaran@intel.com>
Date:   Mon Jan 18 18:32:29 2021 -0800

    fixes #1921 : fix + lit test.

    This change ensures that when we create an anonymous struct, we create it with a name so that the type added to symbol table has a name.
```
> ispc :: full_mask.ispc
```
commit 62f5a6c37bf26fc25a7cf81868052d3472874610
Author: Arina Neshlyaeva <arina.neshlyaeva@intel.com>
Date:   Tue Mar 23 23:07:30 2021 -0700

    Do not generate function mask when it is not required
```

The following 3 fails are due to 4cd0ecf7913199b58d1bd27b82526d7b9164fba2
> ispc :: loop_nounroll.ispc
> ispc :: loop_unroll.ispc
> ispc :: discard_value_names.ispc
```
commit 4cd0ecf7913199b58d1bd27b82526d7b9164fba2
Author: Deepak Rajendrakumaran <deepak.rajendrakumaran@intel.com>
Date:   Fri Mar 26 04:07:02 2021 -0700

    Adding option to discard value names.
```

sebpop:
> > Are you going to do the similar change for other neon targets? Would be good to have it uniform across targets.
> 

Yes, I will impact the changes to the other Arm64 neon targets:
target-neon-i16x8.ll
target-neon-i32x4.ll
target-neon-i32x8.ll
target-neon-i8x16.ll

dbabokin:
Yes, I meant these targets (i16x8, i8x32, i32x8) in ARM64 mode.

Failing tests look like a test problems. We used to test ARM in Travis, but it was turned off about 5 months ago due to changes in their policy for open source projects and came back online for us just yesterday. I'll take care about these tests.

sebpop:
> Failing tests look like a test problems. We used to test ARM in Travis, but it was turned off about 5 months ago due to changes in their policy for open source projects and came back online for us just yesterday. I'll take care about these tests.

Sorry for the noise: all tests are passing on Arm64.

After I removed my old build dirs and re-run cmake, the failed tests are passing.
I am not sure what was stale in those build dirs when I updated my git tree.

dbabokin:
I've updated `.travis.yml` so now it runs on ARM64, please rebase and do `git push --force` to your branch.

Also, I just realized that NEON you basically saying that instructions are different on ARM and aarch64. I thought that NEON ISA is independent from 32/64. Is that correct that NEON ISA is different for 32 and 64 bit arch?

sebpop:
> I've updated `.travis.yml` so now it runs on ARM64, please rebase and do `git push --force` to your branch.
> 
> Also, I just realized that NEON you basically saying that instructions are different on ARM and aarch64. I thought that NEON ISA is independent from 32/64. Is that correct that NEON ISA is different for 32 and 64 bit arch?

Yes, the "vertical" reduction instructions `addv, maxv, minv` have been added to a64:
https://developer.arm.com/documentation/ddi0487/ga
These instructions are not in the spec for a32:
https://developer.arm.com/documentation/ddi0406/latest

I pushed an updated patch that enables target={neon-i32x4, neon-i32x8}.  The other two are not supported on arm64:
```
static const bool lIsTargetValidforArch(ISPCTarget target, Arch arch) {
[...]
    } else if (target == ISPCTarget::neon_i8x16 || target == ISPCTarget::neon_i16x8) {
        if (arch != Arch::arm)
            ret = false;
    } else if (target == ISPCTarget::neon_i32x4 || target == ISPCTarget::neon_i32x8) {
        if (arch != Arch::arm && arch != Arch::aarch64)
            ret = false;
```
I looked over target-neon-i8x16.ll and target-neon-i16x8.ll and I haven't seen anything that could not be supported on a64.
I will send a separate patch that enable these two targets on arm64.

sebpop:
With the two patches I see all the tests in bug #2052 generating good code.
Also the patches passed `make check-all` and
```
./run_tests.py -v --arch=aarch64 --target=neon-i32x4 --opt=O2
./run_tests.py -v --arch=aarch64 --target=neon-i32x8 --opt=O2
```
with no new fails.

sebpop:
Rebased on main and re-tested on a64.

dbabokin:
> Yes, the "vertical" reduction instructions `addv, maxv, minv` have been added to a64:
> https://developer.arm.com/documentation/ddi0487/ga
> These instructions are not in the spec for a32:
> https://developer.arm.com/documentation/ddi0406/latest

Got it. So it's the matter of the difference of NEON as defined in ARMv7 (32bit only) vs ARMv8 (32 and 64 bit).
Looking at the spec and Wikipedia, it's not clear for me if ARMv8 new NEON instructions are available in 32 bit mode.
This also raises the questions if we need to support ARMv7 at all. This seems to be too old and mostly irrelevant for SIMD acceleration. But, of course, this is out of scope of this PR.

> With the two patches I see all the tests in bug #2052 generating good code.

Does it make sense to add a lit test for good instruction sequence? I'm not sure if this is required / reasonable.

src/ispc.cpp
```diff
@@ -1016,7 +1016,7 @@ Target::Target(Arch arch, const char *cpu, ISPCTarget ispc_target, bool pic, boo
         this->m_dataTypeWidth = 8;
         this->m_vectorWidth = 16;
         this->m_hasHalf = true; // ??
-        this->m_maskingIsFree = false;
```
dbabokin, line 1019: Why is this change? NEON doesn't have mask registers, so masking is not "free".

sebpop: You are right.

`m_maskingIsFree` is used in https://github.com/ispc/ispc/blob/main/src/func.cpp#L340-L350
which is an optimization under which ispc creates a branch for `all_on` and `some_on`, and simplification of the branch condition sometimes leads to code without mask.

For AArch64 simplifycfg is not able to remove the branch as the back-end thinks that it is more beneficial to form FMAs in each branch: https://reviews.llvm.org/D7299 
The branch is removed once the back-end was able to form the FMA, however that happens very late and by that time DCE does not run anymore to remove the `tst` block of code that looks useless in bug #2052.

dbabokin: Thanks for the explanation. This sounds like reasonable motivation to have this. I've been looking at this all_on/all_off optimization recently and thinking if it worth keeping. This seems to add more problems than have benefits. This is one more reasons to disable it by default.

If you'd like to keep it - add it as a separate commit and add comment that it's workaround for this problem.

sebpop: I opened a separate PR #2064 to disable this optimization on Arm64.

dbabokin:
@sebpop Thanks a lot for your contribution!

PR #2064, opened 2021-04-22
[arm64] disable Coherent Control Flow optimization - bug #2052

sebpop:


dbabokin:
The failing test is due to setting `disableCoherentControlFlow` based on host architecture, but not target architecture. Setting the *default* based on target arch is problematic, through. As default is set long before target is known... I think we should just rewrite this options when target is know. Let me think about it a bit to see if we can handle it more gracefully.

sebpop:
> The failing test is due to setting `disableCoherentControlFlow` based on host architecture, but not target architecture. Setting the _default_ based on target arch is problematic, through. As default is set long before target is known... I think we should just rewrite this options when target is know. Let me think about it a bit to see if we can handle it more gracefully.

I updated the PR to look like the previous version of the patch.
To disable the CCF optimization the patch relies on the target by checking whether `arch == Arch::aarch64`.

src/ispc.cpp
```diff
@@ -1016,7 +1016,7 @@ Target::Target(Arch arch, const char *cpu, ISPCTarget ispc_target, bool pic, boo
         this->m_dataTypeWidth = 8;
         this->m_vectorWidth = 16;
         this->m_hasHalf = true; // ??
-        this->m_maskingIsFree = false;
+        this->m_maskingIsFree = (arch == Arch::aarch64);
```
dbabokin, line 1019:
```suggestion
        // https://github.com/ispc/ispc/issues/2052
        // AArch64 disables Coherent Control Flow optimization because of a bug in
        // LLVM aarch64 back-end that reduces the efficiency of simplifyCFG.
        // Branches added by CCF can only be removed after the back-end formed
        // fused-multiply-adds.  This reduces the quality of code as most of scalar
        // optimizations will not apply.  FIXME: Consider turning this optimization
        // back on after https://reviews.llvm.org/D100963 gets committed to LLVM-13.
        this->m_maskingIsFree = (arch == Arch::aarch64);
```

dbabokin:
Thank you for your contribution and sorry for long review time.

PR #2074, opened 2021-04-30
Fix sporadic GH Actions test hangs on Windows

dbabokin:
For some reason `alloy.py` doesn't return properly on Windows. It reaches `sys.exit(n)` and hangs. This means that trying to clean up (release file handlers, flush streams, etc) it gets locked. In GH Actions we have 1-2 hangs (out of 84 jobs) per nightly run almost daily.

The fix is to use `os._exit(n)` instead, which exits the script, but doesn't release all resources.

The PR also contains several improvements in exception handling in `alloy.py` and a fix for a problem with `pickle`, which was previously hidden my exception handlers.

### DIFF
--- a/alloy.py
+++ b/alloy.py
@@ -942,6 +942,7 @@ def Main():
         os.rename(f_date, date_name)
         print_debug("Logs are in " + date_name + "\n", False, "")
     except Exception as e:
+        # Do not return non-zero exit code here, as it's not a critical error and testing might be considered successful.
         print_debug("Exception: " + str(e), False, stability_log)
 
     if current_OS == "Windows":
--- a/alloy.py
+++ b/alloy.py
@@ -944,6 +944,10 @@ def Main():
     except Exception as e:
         print_debug("Exception: " + str(e), False, stability_log)
 
+    if current_OS == "Windows":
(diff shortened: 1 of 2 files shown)


dbabokin:
CC @kraszkow - you might be interested to see that we need to use the trick with not releasing handlers on Windows to not hang.

alloy.py
```diff
@@ -927,14 +929,26 @@ def Main():
...
+    try:
         os.chdir(current_path)
         date_name = "alloy_results_" + datetime.datetime.now().strftime('%Y_%m_%d_%H_%M_%S')
         if os.path.exists(date_name):
             alloy_error("It's forbidden to run alloy two times in a second, logs are in ./logs", 1)
         os.rename(f_date, date_name)
         print_debug("Logs are in " + date_name + "\n", False, "")
-        exit(return_status)
+    except Exception as e:
+        print_debug("Exception: " + str(e), False, stability_log)
```
dbabokin, line 945:
```suggestion
        # Do not return non-zero exit code here, as it's not a critical error and testing might be considered successful.
        print_debug("Exception: " + str(e), False, stability_log)
```

aneshlya (approved):

Issue #2082, opened 2021-05-17
[Bisected] ISPC fails to build because of alignment issues.


DarkDefender:
1ff78753b480eeed08f234460c8f6b9121a5e885 breaks building ISPC on my system.
I tried using both llvm 11 and 12 with the same result.

I get the following warnings and error message:
```
/var/tmp/portage/dev-lang/ispc-9999/work/ispc-9999/src/opt.cpp: In member function ‘bool ImproveMemoryOpsPass::runOnBasicBlock(llvm::BasicBlock&)’:
/var/tmp/portage/dev-lang/ispc-9999/work/ispc-9999/src/opt.cpp:3149:108: warning: ‘mlInfo[i].lImproveMaskedLoad(llvm::CallInst*, llvm::BasicBlock::iterator)::MLInfo::align’ may be used uninitialized in this function [-Wmaybe-uninitialized]
 3149 |                 llvm::MaybeAlign(g->opt.forceAlignedMemory ? g->target->getNativeVectorAlignment() : info->align)
      |                                                                                                      ~~~~~~^~~~~
At global scope:
cc1plus: note: unrecognized command-line option ‘-Wno-deprecated-register’ may have been intended to silence earlier diagnostics
cc1plus: note: unrecognized command-line option ‘-Wno-c99-extensions’ may have been intended to silence earlier diagnostics
```
```
ispc: /usr/lib/llvm/12/include/llvm/Support/Alignment.h:137: llvm::MaybeAlign::MaybeAlign(uint64_t): Assertion `(Value == 0 || llvm::isPowerOf2_64(Value)) && "Alignment is neither 0 nor a power of 2"' failed.
/var/tmp/portage/dev-lang/ispc-9999/work/ispc-9999/src/main.cpp(392): FATAL ERROR: Unhandled signal sent to process; terminating.
***
*** Please file a bug report at https://github.com/ispc/ispc/issues
*** (Including as much information as you can about how to reproduce this error).
*** You have apparently encountered a bug in the compiler that we'd like to fix!
***
```

I'm unsure what information I should provide. (I don't want to spam this report with useless information)
@aneshlya do you have any ideas of what it could be based on the bisected commit?
If I need to provide more info or test patches, I'll gladly do so. :)

dbabokin:
Fixed by #2084

DarkDefender:
Thanks! :D

PR #2084, opened 2021-05-17
Fixed alignment issues when using gcc

aneshlya:
This is the fix for https://github.com/ispc/ispc/issues/2082
To avoid uninitialized error from GCC I'm moving structs definition level up from the condition.
For successful compilation, it is required to declare genx specific intrisnics for CPU targets.


dbabokin:
LGTM

PR #2086, opened 2021-05-27
WIP : fp16 draft implementation.

DeepakRajendrakumaran:


dbabokin:
Would be good to have documentation update as part of this commit. This would make review easier as well.

docs/ispc.rst
```diff
@@ -1524,7 +1524,7 @@ The following reserved words from C89 are also reserved in ``ispc``:
 ``ispc`` additionally reserves the following words:
 
 ``bool``, ``delete``, ``export``, ``cdo``, ``cfor``, ``cif``, ``cwhile``,
-``false``, ``foreach``, ``foreach_active``, ``foreach_tiled``,
+``false``, ``float16``, ``foreach``, ``foreach_active``, ``foreach_tiled``,
 ``foreach_unique``, ``in``, ``inline``, ``noinline``, ``__vectorcall``, ``int8``, ``int16``,
```
dbabokin, line 1528: Looks like we missed `uint*` identifiers here, would be good to add them as well.

docs/ispc.rst
```diff
@@ -1639,7 +1639,7 @@ doesn't support character constants.
 The following identifiers are reserved as language keywords: ``bool``,
 ``break``, ``case``, ``cdo``, ``cfor``, ``char``, ``cif``, ``cwhile``,
 ``const``, ``continue``, ``default``, ``do``, ``double``, ``else``,
-``enum``, ``export``, ``extern``, ``false``, ``float``, ``for``,
+``enum``, ``export``, ``extern``, ``false``, ``float``, ``float16``, ``for``,
 ``foreach``, ``foreach_active``, ``foreach_tiled``, ``foreach_unique``,
 ``goto``, ``if``, ``in``, ``inline``, ``noinline``, ``__vectorcall``, ``int``, ``int8``,
```
dbabokin, line 1644: Also missing `uint*` here...

src/ispc.h
```diff
@@ -146,6 +146,14 @@ struct SourcePos {
     bool operator==(const SourcePos &p2) const;
 };
 
+/** Structure to represent float16 constant value. */
+struct float16Ty {
+    float16Ty() : isStr(true) {}
+    float fval;
```
dbabokin, line 152: I'm wondering if we can use `APFloat` class for storing `float16` value right aways, without go through `float` type.

dbabokin (commented):
This also would be good to rebase on top of the trunk.

src/ispc.h
DeepakRajendrakumaran: I'm not storing it as float when the parser identifies a fp16 constant. In these cases I directly use this str to create a  APFloat. The reason for the option to store as float is there are othe ways bu which an fp16 is created

1. Reading fp16 directly from parser/lexer(*fp16 suffix) - this case we simple store it as  str and create APFloat directly from this
2. Parser read a float/double/int etc and it is stored as suct in AST. This si then typecast to fp16. In theses cases, tehe existing float/double/int constant is stored as such. I cannot create a APFloat from string in these cases. The option I went with was store these as float and use APFloat to convert it to falf

src/type.cpp
```diff
@@ -361,6 +368,9 @@ std::string AtomicType::Mangle() const {
     case TYPE_UINT32:
         ret += "u";
         break;
+    case TYPE_FLOAT16:
+        ret += "h";
```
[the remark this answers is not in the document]
DeepakRajendrakumaran: This si covered by lit test where we check for function signature with float16 arg

src/ispc.h
DeepakRajendrakumaran: Done

tests/lit-tests/float16_uniform.ispc
```diff
@@ -0,0 +1,51 @@
+//; RUN: %{ispc} %s --target=host --emit-llvm-text -O0 -o - | FileCheck %s
+//; REQUIRES: X86_ENABLED
+
+//; CHECK: define half @foo0___unh(half
```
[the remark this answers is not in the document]
DeepakRajendrakumaran: Added +0, -0, -inf, +inf

tests/lit-tests/float16_uniform.ispc
```diff
@@ -0,0 +1,51 @@
...
+//; CHECK: define void @foo1___REFunhun_3C_unh_3E_(half* noalias %arg0, half* noalias %arg1
+//; CHECK: store half 0xH4252, half*
+//; CHECK: fmul half %arg
+void foo1(uniform float16 &arg0, uniform float16 *uniform arg1) {
+    arg0 = 3.16f16;
+    *arg1 = *arg1 * arg0;
+}
+
+//; CHECK: define half @foo2___unbunhunh(i1 %cond, half %arg0, half %arg1
+//; CHECK: if_then
```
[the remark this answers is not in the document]
DeepakRajendrakumaran: changed.

docs/ispc.rst
DeepakRajendrakumaran: Added.

docs/ispc.rst
DeepakRajendrakumaran: Added

src/stmt.cpp
```diff
@@ -3021,6 +3021,12 @@ static ExprWithType lProcessPrintArgType(Expr *expr) {
         type = expr->GetType();
     }
 
+    if (Type::Equal(baseType, AtomicType::UniformFloat16)) {
```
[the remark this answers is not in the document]
DeepakRajendrakumaran: Done - it get extended to float and printed

tests/lit-tests/float16_uniform.ispc
```diff
@@ -0,0 +1,78 @@
+//; RUN: %{ispc} %s --target=host --emit-llvm-text -O0 -o - | FileCheck %s
+//; REQUIRES: X86_ENABLED
```
dbabokin, line 2: Why does this require x86?

tests/lit-tests/float16_uniform.ispc
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// Tests float16 -> wider float type conversion.
+//; CHECK: define double @foo5___unh(half %arg0,
+//; CHECK: fpext half %{{[a-zA-Z_][a-zA-Z0-9_]*}} to double
+uniform double foo5(uniform float16 arg0) {
+    return arg0;
+}
+
+// Creates +0, =0, +inf and -inf.
```
dbabokin, line 58: typo: -0

docs/ispc.rst
```diff
@@ -1721,6 +1722,7 @@ basic types:
 * ``int``: 32-bit signed integer; may also be specified as ``int32``.
 * ``unsigned int``: 32-bit unsigned integer; may also be specified as
   ``unsigned int32``, ``uint32`` or ``uint``.
+* ``float16``: 16-bit floating point value
```
dbabokin, line 1725: `fp16` suffix and how `float16` literals are supposed to be written needs to be documented.

src/stmt.cpp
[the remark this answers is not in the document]
dbabokin: I mean we need runtime test, which actually print float16 values.

tests/lit-tests/float16_uniform.ispc
DeepakRajendrakumaran: Fixed

tests/lit-tests/float16_uniform.ispc
DeepakRajendrakumaran: This should have been removed when I added gen support - removed now

docs/ispc.rst
```diff
@@ -1729,7 +1730,7 @@ general" of the two types, with the following precedence:
 ::
 
   double > uint64 > int64 > float > uint32 > int32 >
-      uint16 > int16 > uint8 > int8 > bool
+      float16 > uint16 > int16 > uint8 > int8 > bool
```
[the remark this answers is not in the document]
DeepakRajendrakumaran: Added tests.

src/stmt.cpp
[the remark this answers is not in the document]
DeepakRajendrakumaran: Added runtime test

src/type.cpp
```diff
@@ -413,6 +423,9 @@ std::string AtomicType::GetCDeclaration(const std::string &name) const {
     case TYPE_UINT32:
         ret += "uint32_t";
         break;
+    case TYPE_FLOAT16:
+        ret += "__fp16";
```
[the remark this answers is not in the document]
DeepakRajendrakumaran: Added test.

docs/ispc.rst
DeepakRajendrakumaran: Added to doc

tests/lit-tests/float16_interface.ispc
```diff
@@ -0,0 +1,6 @@
+// RUN: %{ispc} %s --target=host -h %t.h
+// RUN: FileCheck --input-file=%t.h %s
+
+// CHECK: extern void export_float16(__fp16 arg0, __fp16 * arg1);
```
dbabokin, line 4: Now as we have test for header files, I see that it needs to be customized for different compilers.

`clang` and `gcc` support `__fp16`, while MSVC seems to not. I guess we need to create and typedef, which would map to different underlying types depending on the host compiler. For MSVC `uint16_t` seem to be the safe choice. They map to `uint16_t` for DirectX as well: https://docs.microsoft.com/en-us/windows/win32/dxmath/half-data-type

This change may go as a separate fix.

tests/log-uniform-float16.ispc
```diff
@@ -0,0 +1,17 @@
+// rule: skip on arch=x86
+// rule: skip on arch=x86_64
```
dbabokin, line 2: It's bad idea to have both `x86-64` and `x86_64` here. It needs to be fixed in `run_tests.py`. Please add arch string normalization around line 113 (https://github.com/ispc/ispc/blob/main/run_tests.py#L113), if arch string is `x86_64`, then `x86-64` needs to be used.

tests/print_uniform-f16.ispc
```diff
@@ -0,0 +1,11 @@
+// rule: skip on arch=x86
+// rule: skip on arch=x86_64
+// rule: skip on arch=x86-64
+export void print_uf(uniform float a) {
+    uniform float16  uniFloatVal   =  (float16)a + 4.75f16;
+    print("Test uniform float: %\n", uniFloatVal);
```
dbabokin, line 6: Could you please run `clang-format`  on the new tests?

tests/lit-tests/float16_uniform.ispc
```diff
@@ -0,0 +1,87 @@
+//; RUN: %{ispc} %s --target=host --emit-llvm-text -O0 -o - | FileCheck %s
+
+// Tests float16 constant parsing in *e*f16 form.
```
dbabokin, line 3: typo: `*e*f16` -> `*.*f16`

src/lex.ll
```diff
@@ -504,9 +509,15 @@ L?\"(\\.|[^\\"])*\" { lStringConst(&yylval, &yylloc); return TOKEN_STRING_LITERA
     return TOKEN_DOUBLE_CONSTANT;
 }
 
-
 {FLOAT_NUMBER} {
     RT;
+    std::string val(yytext);
+    std::string fp16("f16");
+    if (val.size() >= fp16.size() && ((val.compare(val.size() - fp16.size(), fp16.size(), fp16) == 0)
+           || (val.compare(val.size() - fp16.size(), fp16.size(), "F16") == 0))) {
```
dbabokin, line 517: Let's do `f16` and `F16` checks uniformly - create a `F16` string as well.
It's equivalent, but uniformity is a good thing in the source code.

docs/ispc.rst
```diff
@@ -1614,6 +1614,14 @@ Floating-point constants can optionally have a "f" or "F" suffix (``ispc``
 currently treats all floating-point constants as having 32-bit precision,
 making this suffix not currently have an effect.)
 
+Half precision(16 bit) floating-point constants are represented by a "f16"
+or "F16" suffix. ``float16`` can be specified only in one of two ways
+
+::
+
+  float16 one = 1e0f16;  // 1.0
+  float16 one  = 1.f16;  // 1.o
```
dbabokin, line 1623: typo: `1.o` -> `1.0`

tests/lit-tests/float16_uniform.ispc
DeepakRajendrakumaran: That's nota  type. I amusing exponential format for f16 constant here.

tests/lit-tests/float16_interface.ispc
DeepakRajendrakumaran: Yep. Will add this as a separate commit.

tests/log-uniform-float16.ispc
DeepakRajendrakumaran: Changed

tests/print_uniform-f16.ispc
DeepakRajendrakumaran: Done

src/lex.ll

docs/ispc.rst
DeepakRajendrakumaran: Done

PR #2090, opened 2021-06-04
CMake and CI updates

### LINKED ISSUE #2080: test failed about testing Ispc Functions
hello,I meet a problem about test failed  about testing Ispc Functions
see:
```console
[root@centos8 tests]# ispc c-test-108.ispc
Warning: No --target specified on command-line. Using default system target "neon-i32x4".
Error: Invalid architecture "x86-64"
       Options: .
Error: arch = x86-64 and target = neon-i32x4 is not a valid combination.
[root@centos8 tests]# ispc --arch=aarch64 c-test-108.ispc
Warning: No --target specified on command-line. Using default system target "neon-i32x4".
Error: Invalid architecture "aarch64"
       Options: .
/tmp/root/spack-stage/spack-stage-ispc-1.14.1-qvpvty5hcf2lqk2oic4umtfsbmo23l5a/spack-src/src/ispc.cpp(1041): FATAL ERROR: Problem with target (neon-i32x4)
***
***
/tmp/root/spack-stage/spack-stage-ispc-1.14.1-qvpvty5hcf2lqk2oic4umtfsbmo23l5a/spack-src/src/main.cpp(372): FATAL ERROR: Unhandled signal sent to process; terminating.
Aborted (core dumped)
```
Therefore, I want to know why the current test cases don't support the aarch64 platform?
### Information on your system
[root@centos8 tests]# spack debug report
* **Spack:** 0.16.1-1624-a0b5dcca3c
* **Python:** 3.6.8
* **Platform:** linux-centos8-aarch64
* **Concretizer:** original

dbabokin:
* Update `sde` version used in CI `8.59`->`8.63`.
* Fix #2080: fail in CMake if no targets are specified for ISPC.
* Remove `-Werror` from default build flags, add it manually in CI instead.
* Suppress warnings during Google Benchmark build, as this appears to be a problem with clang trunk.

### DIFF
--- a/.github/workflows/scripts/build-ispc.sh
+++ b/.github/workflows/scripts/build-ispc.sh
@@ -1,4 +1,4 @@
 #!/bin/bash -e
 echo PATH=$PATH
-cmake -B build -DISPC_PREPARE_PACKAGE=ON -DISPC_INCLUDE_BENCHMARKS=ON
+cmake -B build -DISPC_PREPARE_PACKAGE=ON -DISPC_INCLUDE_BENCHMARKS=ON -DCMAKE_CXX_FLAGS=-Werror
 cmake --build build --target package -j4
--- a/docker/ubuntu/cpu_ispc_build/Dockerfile
+++ b/docker/ubuntu/cpu_ispc_build/Dockerfile
@@ -61,7 +61,7 @@ RUN apt-get -y update && apt-get install -y m4 bison flex zlib1g-dev libc6-dev-i
 # Configure ISPC build
 RUN mkdir build_$LLVM_VERSION
 WORKDIR build_$LLVM_VERSION
-RUN cmake ../ -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_INSTALL_PREFIX=/usr/local/src/ispc/bin-$LLVM_VERSION
+RUN cmake ../ -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_INSTALL_PREFIX=/usr/local/src/ispc/bin-$LLVM_VERSION -DCMAKE_CXX_FLAGS=-Werror
 
 # Build ISPC
 RUN make ispc -j8 && make check-all && make install
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -133,7 +133,7 @@ for:
   before_build:
     - sh: |-
         cd $ISPC_HOME && mkdir build && cd build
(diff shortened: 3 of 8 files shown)


docker/ubuntu/cpu_ispc_build/Dockerfile
```diff
@@ -61,7 +61,7 @@ RUN apt-get -y update && apt-get install -y m4 bison flex zlib1g-dev libc6-dev-i
 # Configure ISPC build
 RUN mkdir build_$LLVM_VERSION
 WORKDIR build_$LLVM_VERSION
-RUN cmake ../ -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_INSTALL_PREFIX=/usr/local/src/ispc/bin-$LLVM_VERSION
+RUN cmake ../ -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_INSTALL_PREFIX=/usr/local/src/ispc/bin-$LLVM_VERSION -DCMAKE_CXX_FLAGS=-Werror
```
aneshlya, line 64: Don't we need to modify other Dockerfiles as well?

.github/workflows/scripts/build-ispc.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash -e
 echo PATH=$PATH
-cmake -B build -DISPC_PREPARE_PACKAGE=ON -DISPC_INCLUDE_BENCHMARKS=ON
+cmake -B build -DISPC_PREPARE_PACKAGE=ON -DISPC_INCLUDE_BENCHMARKS=ON -DCMAKE_CXX_FLAGS=-Werror
```
aneshlya, line 3: Why do we add -DCMAKE_CXX_FLAGS=-Werror to github actions but do not add to appveyor and travis builds?

docker/ubuntu/cpu_ispc_build/Dockerfile
dbabokin: I don't think so. This Dockerfile is used in CI and that's why we need `-Werror` there. Though I don't have strong opinion here.

.github/workflows/scripts/build-ispc.sh
dbabokin: Good point. I'll add them. I kinda forgot that we have anything besides Github Actions :)

aneshlya (approved):

PR #2097, opened 2021-06-09
Apple cpu definitions

dbabokin:
* Adding definition for Apple ARM CPUs: A7, A10-A14 (missing A8 and A9, as they are just aliases for A7).
* Change default CPU target for ARM macOS and iOS.

Note, there's not M1 definition in open source LLVM right now, so using the latest A-series definition, which is good enough approximation for our purposes.


aneshlya (approved):
Looks good to me. Probably we should mention new CPU targets in docs.

PR #2109, opened 2021-06-17
Add retry for apt-get failures

dbabokin:


kraszkow (approved):
LGTM. I don't see better solution for sporadic failures in apt we have.

PR #2110, opened 2021-06-23
Fixing float16 tests.

DeepakRajendrakumaran:
This commit includes
1. Fixing precision check for float16 pow/exp tests
2. Fixing range of arg for exp() tests so that values remain in valid range for float16
3. Removing unnecessary skip
4. Cleaning up reference value calculation for better understanding.


dbabokin:
LGTM.

PR #2117, opened 2021-07-07
Fixes #2116: Fixing crash while using unary expression with reference type as rvalue.

### LINKED ISSUE #2116: Fatal error during generation in version 1.16.0 release
Hi!

There seems to be a bug in version 1.16.0 release.

Reproducer snippet:

`void func(int& param_inout)
{
    while(param_inout-- > 0)
    {
    }
}
`
Will lead to _FATAL ERROR: Unhandled signal sent to process; terminating._

Best regards,
Lars

DeepakRajendrakumaran:
This change contains the following
- Unary operations on reference type should update the referenced address and return value(updated or not depending on post/pre). This means, the unary expressions 'type' is the reference target type and not the reference type. This commit fixes this.
-  Test to check the behavior and crash. 
With no fix : https://ispc.godbolt.org/z/rov7axnqW
This test also checks for correctness to ensure that prefix/postfix properties hold.

### DIFF
--- a/tests/lit-tests/2116.ispc
+++ b/tests/lit-tests/2116.ispc
@@ -10,7 +10,7 @@ void foo0(uniform int &param_inout) {
     }
 }
 
-// Case 1: 'if' in 'foo1' uses result of pre inc operation '++param_inout'
+// Case 2: 'if' in 'foo1' uses result of pre inc operation '++param_inout'
 void foo1(uniform int &param_inout) {
     if (++param_inout > 0) {
         param_inout++;
--- a/tests/lit-tests/2116.ispc
+++ b/tests/lit-tests/2116.ispc
@@ -1,0 +1,18 @@
+// Issue #2116 : This test ensures that using the result of
+// unary expression on a reference type for other operations
+// doesn't cause a crash.
+// RUN: %{ispc} %s -O2 --target=host
+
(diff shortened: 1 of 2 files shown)


dbabokin:
The fix looks good. But the test, besides checking that it compiles without a crash, checks that the result is folded to 1 or 2, which seem to be absolutely unrelated and complicates the test. Why is that? Is it somehow affected by the return type? If not, I'd like to have more straightforward test. Otherwise it's just misleading and confusing (for the person who will be potentially investigating it when it fails).

tests/lit-tests/2116.ispc
```diff
@@ -0,0 +1,18 @@
...
+// RUN: %{ispc} %s -O2 --target=host
+
+// Case 1: 'if' in 'foo0' uses result of post inc operation 'param_inout++'
+void foo0(uniform int &param_inout) {
+    if (param_inout++ > 0) {
+        param_inout++;
+    }
+}
+
+// Case 1: 'if' in 'foo1' uses result of pre inc operation '++param_inout'
```
dbabokin, line 13:
```suggestion
// Case 2: 'if' in 'foo1' uses result of pre inc operation '++param_inout'
```

dbabokin:
LGTM

PR #2120, opened 2021-07-14
Improving macOS SDK version detection

dbabokin:
`xcrun` is known to fail to report SDK version properly for standalone SDK installations and in some default conditions (without executing `xcode-select -s <path>` first). So the detection code tries `xcrun` first and looks at folder name as a fallback option.

Thanks to @chrstphrchvz for bringing up this problem and helping with investigation in #2115.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -146,24 +146,37 @@ endif()
 
 set(ISPC_MACOS_ARM_TARGET OFF)
 if (ISPC_MACOS_TARGET AND ISPC_MACOS_SDK_PATH)
-    if (IS_SYMLINK ${ISPC_MACOS_SDK_PATH})
-        # Follow symlinks, relative and absolute.
-        file(READ_SYMLINK "${ISPC_MACOS_SDK_PATH}" ISPC_MACOS_SDK_PATH_NEW)
-        if (NOT IS_ABSOLUTE "${ISPC_MACOS_SDK_PATH_NEW}")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -146,24 +146,37 @@ endif()
...
 set(ISPC_MACOS_ARM_TARGET OFF)
 if (ISPC_MACOS_TARGET AND ISPC_MACOS_SDK_PATH)
-    if (IS_SYMLINK ${ISPC_MACOS_SDK_PATH})
-        # Follow symlinks, relative and absolute.
-        file(READ_SYMLINK "${ISPC_MACOS_SDK_PATH}" ISPC_MACOS_SDK_PATH_NEW)
-        if (NOT IS_ABSOLUTE "${ISPC_MACOS_SDK_PATH_NEW}")
-            get_filename_component(ISPC_MACOS_SDK_PATH_DIR "${ISPC_MACOS_SDK_PATH}" DIRECTORY)
-            set(ISPC_MACOS_SDK_PATH_NEW "${ISPC_MACOS_SDK_PATH_DIR}/${ISPC_MACOS_SDK_PATH_NEW}")
+    # Get macOS SDK version.
+    # xcrun is known not to work for standalone SDK or in the default enviroment
```
aneshlya, line 150: typo enviroment -> environment

aneshlya (approved):

CMakeLists.txt
dbabokin:
```suggestion
    # xcrun is known not to work for standalone SDK or in the default environment
```

PR #2123, opened 2021-07-16
Switching to APFloat for representing 'ConstExpr' for all floating point types.

DeepakRajendrakumaran:
Switching ConstExpr representation of `float` and `double` types to APFloat. This is done for two reasons:
1. it's the right thing to do, as internal representation using `float` and `double` types is error prone and depends on build options and may have portability problems
2. consistency with handling of `float16` type
Note: these changes do not completely eliminate the use of host logic involving `float` and `double` types.

Along with switching `float` and `double` to APFloat, all optimizations / foldings available for these types were also enabled for `float16` (specifically for BinaryExpr and SelectExpr)

Other changes included in this PR:
- regression test for #2148 
- fix for the bug with constant folding of select expression with `double` and `bool` types
- enabling fast-math optimizations (a/C => a * (1/C); a/b => a * rcp(b)) for `float16` and `double`. This good for consistency with `float`. We should consider moving fast-math optimizations to backend.
- refactoring to reduce copy-paste code and improve consistency

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -1187,7 +1187,7 @@ Expr *UnaryExpr::Optimize() {
             int count = constExpr->GetValues(v);
             for (int i = 0; i < count; ++i)
                 v[i].changeSign();
-            return new ConstExpr(constExpr, v);
+            return new ConstExpr(type, v, pos);
         }
     }
     case BitNot: {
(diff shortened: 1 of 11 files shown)


src/expr.cpp
```diff
@@ -5312,53 +5403,55 @@ ConstExpr::ConstExpr(const Type *t, bool *b, SourcePos p) : Expr(p, ConstExprID)
...
+// The llvm::APFloat value is guaranteed to be a double.
+ConstExpr::ConstExpr(ConstExpr *old, std::vector<llvm::APFloat> v) : Expr(old->pos, ConstExprID) {
     type = old->type;
 
     AtomicType::BasicType basicType = getBasicType();
 
     switch (basicType) {
     case AtomicType::TYPE_BOOL:
         for (int i = 0; i < Count(); ++i)
-            boolVal[i] = (v[i] != 0.);
```
dbabokin, line 5323: This is an interesting case. Old behavior and new behavior are different - comparing for equality will succeed only for `+0.0`, but not for `-0.0`. While new version will handle both values. I'm not sure what is correct behavior. I guess old behavior is more correct, as it will match runtime behavior. Would be good how C/C++ compiler behave.

dbabokin (commented):
Generally looks good. It worth to note that it covers switching to APFloat for ConstExpr logic only, and does not intend to cover unification of parsing, which should be a separate changeset.

src/expr.cpp
DeepakRajendrakumaran: That part can actually be further simplified. Will upload a change. But to confirm behavior - https://godbolt.org/z/MPMo1shx6

DeepakRajendrakumaran: I've modified it since that line of code would not have been hit anyway.

dbabokin: Mu comment here was incorrect, According to IEE754, comparisons shall ignore the sign of zero (so +0 == −0).

aneshlya (approved):

Issue #2122, opened 2021-07-14
Drop LLVM 10 support


dbabokin:
LLVM 10 support appears to be quite problematic - there's something fishy with LLVM 10 builds on Windows:
- it doesn't build on Windows with the latest VS2019. `16.4` is known to work. The latest (`16.10`?) is known NOT to work.
- when enabling C++17 as a build language for ISPC it yields broken compiler when linked against LLVM 10 (#1904).

Github CI Windows workers have the latest VS2019, so they are unable to build LLVM10 (the latest failing pipeline is [here](https://github.com/ispc/ispc/actions/runs/1029401595)).

So I propose dropping support for LLVM 10. This will also unblock enabling C++17 as a default C++ dialect for ISPC builds (#1904).

dbabokin:
I've been pointed out that to build with VS2019 16.4 (or any other specific version), we can use this Action: [microsoft/setup-msbuild](https://github.com/marketplace/actions/setup-msbuild) like this:
```yaml
- name: Add msbuild to PATH
  uses: microsoft/setup-msbuild@v1.0.2
  with:
    vs-version: '[16.4,16.5)'
```

Sounds like a good workaround if we need it in the future.

dbabokin:
One more consideration. It's not possible to build any LLVM more recent than 10 on `macOS 10.10`. We support `macOS 10.12` and dropped `10.10` support in `v1.15.0` release, though.

This should not be an issue, but it's good to spell it out explicitly.

dbabokin:
Fixed by #2135

PR #2135, opened 2021-08-05
CI updates / removing LLVM 10 support

dbabokin:
* Support for LLVM 10 was removed. We were not testing with LLVM 10 for a while, also LLVM 10 is known to have problems building on Windows with the latest VS2019.
* SDE version was upgraded `8.63.0` => `8.69.1`
* Added Windows testing to nightly pipeline with LLVM 13.
* Added pipeline to build LLVM 13 reuse.


dbabokin:
Fixes #2122

PR #2157, opened 2021-09-10
Fix #2156: Insert shuffle instruction after its arguments are defined, but not before

### LINKED ISSUE #2156: write_same_loc.ispc test is failing for several targets
`write_same_loc.ispc` test is failing in CI for `avx2-i8x32`, `avx512knl-i32x16`, `avx512skx-i32x8`, `avx512skx-i32x16`, `avx512skx-i8x64`, `avx512skx-i16x32` when compiler for `x64` arch (i.e. 32 bit) with optimizations (O1 or O2).

It's failing in compile time with the message:
```
Instruction does not dominate all uses!
  %12 = shl <8 x i32> %b_load_to_int32_broadcast, <i32 2, i32 0, i32 0, i32 0, i32 0, i32 0, i32 0, i32 0>
  %shuffle = shufflevector <8 x i32> %12, <8 x i32> undef, <8 x i32> zeroinitializer
in function f_fu___un_3C_unf_3E_un_3C_unf_3E_unf
LLVM ERROR: Broken function found, compilation aborted!
/users/dybaboki/ispc/src/main.cpp(393): FATAL ERROR: Unhandled signal sent to process; terminating.
***
***
Aborted (core dumped)
```

dbabokin:
I'm not sure why it worked earlier, but the fixed code is clearly wrong, as it inserts the instruction before its first argument is defined.

### DIFF
--- a/src/opt.cpp
+++ b/src/opt.cpp
@@ -1458,7 +1458,7 @@ static llvm::Value *lGetBasePtrAndOffsets(llvm::Value *ptrs, llvm::Value **offse
                 // We expect here ConstantVector as
                 // <i64 4, i64 undef, i64 undef, i64 undef, i64 undef, i64 undef, i64 undef, i64 undef>
                 llvm::ConstantVector *cv = llvm::dyn_cast<llvm::ConstantVector>(bop_var->getOperand(1));
-                llvm::Value *shuffle_offset = NULL;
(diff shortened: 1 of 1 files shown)


aneshlya (approved):

dbabokin:
Fixes #2156

PR #2162, opened 2021-09-14
Tweak aos_to_soa documentation

CiderMan:
Following on from discussion #2159, this is an update to the documentation making clear the memory access behaviour of the aos_to_soa/soa_to_aos functions.


dbabokin:
Looks great. Thanks for the contribution!

PR #2196, opened 2021-11-20
Added draft of invoke_sycl design doc

aneshlya:


dbabokin:
Let's use code-block syntax like here: https://github.com/kiith-sa/RestructuredText-tutorial/blob/master/README.rst#source-code-highlighting

Issue #2198, opened 2021-11-22
Launching a Task Function through a Function Pointer Fails Assert in ISPC


Twinklebear:
It might not be valid to use tasks and function pointers in this way, but the following code triggers an assert in ISPC:

```c
typedef task void (*TaskFn)();

struct Test {
    TaskFn fn;
};

task void example() {
    print("%\n", taskIndex0);
}

export void callit() {
    Test test;
    test.fn = example;
    launch[1] test.fn();
}
```

This fails with the assertion:
```
repro.ispc:14:15: Error: Assertion failed (/usr/local/src/ispc/src/ctx.cpp:3562): "llvm::isa<llvm::Function>(callee)".
    launch[1] test.fn();
              ^^^^^^^

***
*** Please file a bug report at https://github.com/ispc/ispc/issues
*** (Including as much information as you can about how to reproduce this error).
*** You have apparently encountered a bug in the compiler that we'd like to fix!
***
/usr/local/src/ispc/src/main.cpp(392): FATAL ERROR: Unhandled signal sent to process; terminating.
Aborted (core dumped)
```

PR #2199, opened 2021-11-24
Allow uniform function pointers for launch expression

aneshlya:
Fix for https://github.com/ispc/ispc/issues/2198


dbabokin:
LGTM, but one remark - let's add a lit test to test error message.

dbabokin (approved):

PR #2206, opened 2021-12-08
Update vim syntax file

dbabokin:
Update vim syntax highlighting file to accommodate ISPC language changes, and fix inaccuracies:
- add `assert`, `assume`, `task`, and `unmasked` to statements
- moved `new` and `delete` from types to statements
- remove `reference` from statement list (not sure why it was there)
- add `uint8`, `uint16`, `uint32`, `uint64`, and `float16` types
- add `in` and `...` operators
- add `noinline` and `__vectorcall` as "storage class". Note `inline` is part of C syntax
- add all macros defined in ISPC.
- add llvm instrinsic calls support (@llvm.x86.blah-blah)
- properly define ISPC integer literals. Key differences are binary format support and [kMG] suffixes. Note that the suffix syntax is more restrictive that real patterns accepted by ISPC. The intent is to fix the parser to match this syntax file
- properly define ISPC fp literals, including `float16` literals. These literals are not exactly what is supported by C. The differences are: (1) `f16`/`F16` suffix, (2) `d`/`D` suffix, which is not supported by C in general, but exists as gcc extension only (not supported by vim C syntax), (3) hexadecimal are more restrictive that C++17 hexadecimal (though they are not in C anyway), (4) Fortran double format (it was there already, I've compacted 3 rules to 1).

Things that need improvement:
- operator `...` is not recognized when used without spaces around it, i.e. `foreach (i = 1...100) {}`. The problem is that `1.` will be matched by FP literals pattern (those from C syntax file). To fix this we need remove `cFloat` definition from C and keep our own only. And finish it with the rule "not match an extra trailing dot", i.e. `\.\@!`. This is not possible when including C syntax.

After committing these changes and letting them sit for a while, I plan to start discussion on vim mailing list about adding this to vim official distribution.

CC @Twinklebear as I noticed you have a dedicated repo for ISPC vim syntax.

### DIFF
--- a/contrib/ispc.vim
+++ b/contrib/ispc.vim
@@ -1,7 +1,8 @@
 " Vim syntax file
-" Language:	ISPC
-" Maintainer:	Andreas Wendleder <andreas.wendleder@gmail.com>
-" Last Change:	2016 May 04
+" Language:		ISPC
+" Maintainer:		Dmitry Babokin <dmitry.y.babokin@intel.com>
+" Previous Maintainer:	Andreas Wendleder <andreas.wendleder@gmail.com>
+" Last Change:		December 7, 2021
 
 " Quit when a syntax file was already loaded
 if exists("b:current_syntax")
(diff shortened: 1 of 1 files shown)


Twinklebear:
Awesome! Yeah I use https://github.com/Twinklebear/ispc.vim to manage it as a plugin with Vundle. I can copy these changes over there (or if you prefer to open a PR that also works). Having ISPC as part of the official vim distribution would be great.

dbabokin:
I haven't realized that this repo is used for Vundle! I think we should have an official syntax repo for Vundle and other editors.

Twinklebear:
An official repo would be great to have!

PR #2224, opened 2022-01-12
Prefetchw support

aneshlya:
- a family of `prefetchw_*()` standard library functions, which map to "prefetch in anticipation of write" instructions, when it's available in the hardware, or regular prefetch intrinsics otherwise.
- definition of `skylake` CPU (client Skylake, i.e. AVX2 CPU). It's not much different from the default for AVX2 - `haswell`. But given the amount of existing CPUs on the market built as derivatives of Skylake microarchitecture, it's good to have the possibility to do micro tuning for it.


tests/lit-tests/prefetchw.ispc
```diff
@@ -0,0 +1,15 @@
+// Check that prefetchw instruction is generated.
+// RUN: %{ispc} %s --emit-asm --target=avx2-i32x8 -o - | FileCheck %s -check-prefix=CHECK_PREFETCHW
```
dbabokin, line 2:
```suggestion
// RUN: %{ispc} %s --emit-asm --target=avx2-i32x8 -o - | FileCheck %s -check-prefix=CHECK_PREFETCHW

// REQUIRES: X86_ENABLED
```

dbabokin (approved):

PR #2250, opened 2022-01-31
Centos 8 Dockerfile - lack of ncurses-static workaround

DVLP:
Building static ncurses and symlinking  so -lcurses works as -lncurses(libcurses.a linked to libncurses.a). Not required for LLVM-only version but required for package build

### DIFF
--- a/docker/centos/8/cpu_ispc_build/Dockerfile
+++ b/docker/centos/8/cpu_ispc_build/Dockerfile
@@ -17,9 +17,12 @@ RUN dnf -y update && dnf install -y wget yum-utils gcc gcc-c++ git python3 make 
     dnf clean -y all
 
 # These packages are required if you need to link IPSC with -static.
-# TODO: build ncurses static version manually.
 RUN dnf -y --enablerepo=powertools install libstdc++-static zlib-static && \
     dnf clean -y all
+
+# Ncurses for g++ --static -lcurses -ltinfo. Despite setting --enable-overwrite it doesn't link(only static not linked?) libncurses as libcurses so symlinked manually
+RUN git clone https://github.com/ThomasDickey/ncurses-snapshots.git --depth=1 && cd ncurses-snapshots && ./configure --with-termlib --with-libtool --with-libtool-opts=-static --enable-static --enable-overwrite && make -j8 && make install && \
+  ln -s /usr/lib/libncurses.a /usr/lib/libcurses.a && ln -s /usr/lib/libncurses++.a /usr/lib/libcurses++.a
 
 # Download and install required version of cmake (3.14 for x86, 3.20 for aarch64, as earlier versions are not available as binary distribution) for ISPC build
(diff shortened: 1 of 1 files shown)


dbabokin:
Were you able to build the docker image? I'm trying it now and I get the following:
```
CentOS Linux 8 - AppStream
Error: Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist
```

I'm not sure if this is because I have a problem with network connectivity (`centos:7` updates well, though) or it's because CentOS is EOL. Probably it's the later, as I see other people experienced the same: https://forums.centos.org/viewtopic.php?f=54&t=78708

dbabokin:
I tried building using `rockylinux:8` as a base, instead of `centos:8` and it seems to be a good drop-in replacement, we should probably switch to it going forward.

A couple things that popped up:
1. `libtool` is needed to build ncurses, so it needs to be added to the list of packages installed before installing ncurses.
2. Using `main` branch of ncurses sounds like a bad idea, we need to settle on a specific stable tag.

DVLP:
Centos 8 packages repo was brutally switched off. I understand EOL but not such abrupt shutdown. Anyway yes rockylinux is pretty good but was quite slow today. Today it must be very overloaded due to people switching in panic so I'm staying with Rocky for now. I also tried alma and oracle.

1. Yes, libtool must be added. I pulled this snippet from a modified image dependent on a custom Emscripten etc. Btw is there a more straightforward way to build ISPC in docker with WASM enabled? My method to replace the base image seems like an overkill.
2. Makes sense, but it's going to be slightly more complicated when using --depth=1 flag. Using `git clone`(without --depth=1) and then `git checkout` means downloading the whole repo with all the branches and history. Not very efficient.

dbabokin:
To build ISPC (in docker or not) it's not necessary to build LLVM from scratch. The docker is to demonstrate how to bootstrap full environment in a proper way. But a faster way is to use pre-built LLVM that we use in our CI. It's stored in releases of this repo: https://github.com/ispc/llvm-project, it's properly patched and is build with x86, ARM, and WASM support. Though for development purposes it's better to use Ubuntu and preferably some recent version. The Linux version LLVM binaries that we have are built in Ubuntu 18.04.

Combination of `--depth=1` and `--branch=v6_3` seem to work well. I would prefer to have `v6_3`, rather than `v6_3_20220129`, as pulling some intermediate development state is not a good idea, unless we are focusing on ncureses development or concerned with some specific bug.

dbabokin:
Would be good to commit this PR. @DVLP could you please update the distro to `rockylinux:8` and have a fixed tag (`v6_3` probably?) used for ncurses?

DVLP:
Ok I pushed these two changed but it needs testing before merging. I'm leaving it building now and will confirm tomorrow if it built

dbabokin:
I've added missing `libtool` and now it works well.

Issue #2243, opened 2022-01-27
Instruction selection problem with shuffle(double, int) with LLVM 12


dbabokin:
The problem is reproducible with `v1.17.0` and `v1.16.1` release binaries (they both use LLVM 12 as a backend). It's not reproducible with LLVM 13, i.e. with "trunk" in Compiler Explorer.

Here's code snippet:
```c
double b(double c, int d) {
  return shuffle(c, d);
}
```

```
> ispc -g  --target=avx1-i32x4 --cpu=ps4 --arch=x86-64 -g -O0  t.ispc -o output.s --emit-asm
LLVM ERROR: Cannot select: 0x6a8e778: v2i64 = truncate 0x6a45ae0, stdlib.ispc:357:32 @[ t.ispc:2:10 ]
  0x6a45ae0: v4i32,ch = load<(dereferenceable load 16 from %ir.i2.i)> 0x6a45a10, FrameIndex:i64<6>, undef:i64, stdlib.ispc:357:32 @[ t.ispc:2:10 ]
    0x6a6ab78: i64 = FrameIndex<6>
    0x6a69e10: i64 = undef
In function: b___vydvyi
/usr/local/src/ispc/src/main.cpp(396): FATAL ERROR: Unhandled signal sent to process; terminating.
***
*** Please file a bug report at https://github.com/ispc/ispc/issues
*** (Including as much information as you can about how to reproduce this error).
*** You have apparently encountered a bug in the compiler that we'd like to fix!
***
Aborted (core dumped)
```

Compiler explorer [link](https://ispc.godbolt.org/z/5Ps7so8Ya)

dbabokin:
Here's simplified LLVM IR: 
[reduced.tar.gz](https://github.com/ispc/ispc/files/7954793/reduced.tar.gz)

To run:
```
llc bugpoint-reduced-simplified.ll -mcpu=haswell
```

dbabokin:
Here's the fix responsible for the fix of this problem in LLVM 13.0, it applies clearly to LLVM 12.0:
https://github.com/llvm/llvm-project/commit/13a51e017c09

dbabokin:
Fixed by #2251

PR #2251, opened 2022-01-31
Fix for #2243. Patch for LLVM 11 & 12 and a regression test

dbabokin:
LLVM 12 binaries were updated, so CI is expected to pass with new regression test.


aneshlya (approved):
LGTM!

dbabokin:
Hmmm, LLVM 11 fails with new regression test. It was passing locally for me. I will investigate.

dbabokin:
LLVM 11 is indeed failing. The same patch needs to be applied to LLVM 11 as well. I've pushed the changes and will rebuild LLVM 11 and rerun the testing.

dbabokin:
LLVM 11.1 binaries were updated.

PR #2258, opened 2022-02-02
Fixes to support FreeBSD build

### LINKED ISSUE #2244: Tests fail on FreeBSD with return code -11
ISPC tests fail on FreeBSD:
```
Running 8 jobs in parallel. Running 1543 tests.
Test ./tests/test-24.ispc failed (return code -11)                    
Test ./tests/typedef-4.ispc failed (return code -11)                  
Test ./tests/gather-int32-4.ispc failed (return code -11)             
Test ./tests/int64-min-1.ispc failed (return code -11)                
Test ./tests/cfor-test-77.ispc failed (return code -11)               
Test ./tests/funcptr-uniform-4.ispc failed (return code -11)            
Test ./tests/local-atomics-6.ispc failed (return code -11)            
Test ./tests/padds-i16.ispc failed (return code -11)                  
```

Version: 1.17.0
FreeBSD 13

dbabokin:
- support `amd64` as output of `uname -m` during CMake build to identify if this is x86 build
- use `sh` instead of `bash` where it doesn't matter
- identify FreeBSD as Linux in examples and `test_static.cpp`, as it needs to use the same memory allocation routines
- fail in examples and `test_static.cpp` if the host OS in not detected (we should fail, as otherwise it will not work anyway)

Fixes #2244


test_static.cpp
```diff
@@ -35,10 +35,12 @@
 #define ISPC_IS_WASM
 #elif defined(_WIN32) || defined(_WIN64)
 #define ISPC_IS_WINDOWS
-#elif defined(__linux__)
+#elif defined(__linux__) || defined(__FreeBSD__)
```
aneshlya, line 38: Please add a check of FreeBSD platform in test_static_l0.cpp and report that platform is not supported.

aneshlya (approved):

test_static.cpp
dbabokin: fixed, please have a look if it's what you expected

aneshlya (approved):

PR #2268, opened 2022-02-18
CI updates: VS2022, SDE 9.0.0

dbabokin:
- use VS2022 as a default toolchain on Windows. Would need to use `generator="Visual Studio 16 2019"` for `alloy.py` to use VS2019.
- Actions / Appveyor with LLVM13 to use VS2022. LLVM 11 & 12 to stay on VS2019.
- SDE version upgraded to 9.0.0. They use another download location now (sigh...)
- GnuWin32 download not to use quite mode, as it's failing frequently lately

The decision to move to VS2022 is debatable. And we might choose to postpone it. Though I don't see the reasons not to move to VS2022 now. Earlier we move, more experience we'll get with it before releasing.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -37,7 +37,7 @@ environment:
   matrix:
     - APPVEYOR_BUILD_WORKER_IMAGE: Ubuntu1804
       LLVM_VERSION: latest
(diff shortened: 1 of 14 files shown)


dbabokin:
> Please fix the path to Windows packages in Appveyor in Readme.md

That's a good catch!

aneshlya (approved):

PR #2274, opened 2022-03-06
don't assume that ncurses is a system library on macOS

aumuell:
When compiling ispc via Spack (https://spack.io) on macOS, it fails because it assumes that Curses is a system library. This patch fixes that by depending on CMake's find_package for detecting the library location also on macOS.

This is similar as the situation on Linux, which was already fixed in #1891.


dbabokin:
Before the patch:
```
bin/ispc:
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1281.100.1)
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
	/usr/lib/libncurses.5.4.dylib (compatibility version 5.4.0, current version 5.4.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 902.1.0)
```

After the patch:
```
> otool -L bin/ispc
bin/ispc:
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1281.100.1)
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
	/usr/lib/libncurses.5.4.dylib (compatibility version 5.4.0, current version 5.4.0)
	/usr/lib/libform.5.4.dylib (compatibility version 5.4.0, current version 5.4.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 902.1.0)
```

Any idea why we got `libform` dependency?

dbabokin:
The difference in the link command on macOS is as follows. Before:
```
-lz -lcurses
```

After:
```
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.1.sdk/usr/lib/libz.tbd /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.1.sdk/usr/lib/libcurses.tbd /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.1.sdk/usr/lib/libform.tbd
```

I don't understand this `*.tbd` thing, but it seems to be correct. But where this `libform` comes from and why is not clear to me.

aumuell:
Thank you for looking into this!

> I don't understand this `*.tbd` thing, but it seems to be correct. But where this `libform` comes from and why is not clear to me.

The libform comes from CURSES_LIBRARIES, which is set by FindCurses.cmake to include both libcurses/libncurses and libform. When inspecting FindCurses.cmake, I also found that it also discovers libtinfo and stores its location in CURSES_EXTRA_LIBRARY. So I force pushed an updated patch that uses those two CMake variables directly instead of CURSES_LIBRARIES.

dbabokin:
Just for the record, here's what I see on Linux. `libform.so.6` has dropped from the dependencies.

Before:
```
> ldd bin/ispc
	linux-vdso.so.1 (0x00007fff4b760000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007faf99aa2000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007faf99a7f000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007faf99a63000)
	libtinfo.so.6 => /lib/x86_64-linux-gnu/libtinfo.so.6 (0x00007faf99a33000)
	libncurses.so.6 => /lib/x86_64-linux-gnu/libncurses.so.6 (0x00007faf99a0a000)
	libform.so.6 => /lib/x86_64-linux-gnu/libform.so.6 (0x00007faf999f5000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007faf99813000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007faf996c4000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007faf996a9000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007faf994b7000)
	/lib64/ld-linux-x86-64.so.2 (0x00007faf99ab4000)
```
After
```
> ldd bin/ispc
	linux-vdso.so.1 (0x00007fff82ffd000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fd658375000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fd658352000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007fd658336000)
	libncurses.so.6 => /lib/x86_64-linux-gnu/libncurses.so.6 (0x00007fd65830d000)
	libtinfo.so.6 => /lib/x86_64-linux-gnu/libtinfo.so.6 (0x00007fd6582dd000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fd6580f9000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fd657faa000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fd657f8f000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fd657d9d000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fd658387000)
```

dbabokin:
I'm merging the PR.

Though I have a side question - is it possible to build LLVM as part of ISPC package, not to reuse the pre-built LLVM during Spack build? Also, for macOS you need to enable ARM.

Issue #1212, opened 2016-06-09
ispc return code incorrect with #error preprocessor directive


kjeffery:
With this simple ispc file:

```
#error Error 1
#error Error 2

void f()
{
}
```

The return code of ispc reports success.

```
[kjeffery@xlimepony tmp]$ /rel/third_party/intelispc/1.9.0/ispc error.ispc -o error.o
Warning: No --target specified on command-line. Using default system target "avx1.1-i32x8". 
error.ispc:2:2: error: Error 1
#error Error 1
 ^
error.ispc:3:2: error: Error 2
#error Error 2
 ^
[kjeffery@xlimepony tmp]$ echo $?
0
```

dbabokin:
`#error` seems to be handled by preprocessor, but it's not clear how this error state is communicated back to the preprocessor caller. Should be easy to fix, but requires some investigation.

dbabokin:
It's basically same as #800

dbabokin:
Fixed by #2280

Issue #800, opened 2014-06-06
Preprocessor failure does not terminate


mattwalsh:
Consider this program; it is compile-able save for the extraneous #endif.

```
export void J(
    uniform float a[],
    uniform float b[],
    uniform int N) {

}

#endif
```

If I compile this with ispc (ispc testcase.ispc -o testcase.o) I get this output...

```
testcase.ispc:9:2: error: #endif without #if
#endif
```

this is a valid error.  However, this error does not stop compilation.  In fact a .o file is generated and the return code ($?) is zero.

I believe this should error out, not even try to produce the .o file and return an error to the shell.

dbabokin:
Fixed by #2280

PR #2280, opened 2022-03-11
Preprocessor Fixes and Updates

### LINKED ISSUE #2257: Add -E switch, which dumps preprocessed output
ISPC is used with large files with multiple includes and multiple ifdefs. To debug them, it's very useful to have `-E` switch.

e3m3:
Closes #2257, #1212, and #800.

Changes:
* Adds `-E` and `--ignore-preprocessor-errors` flags to CLI.
* Updates `execPreprocessor` to return a status code.

Updated: 03/18/2022.

### DIFF
--- a/docs/ispc.rst
+++ b/docs/ispc.rst
@@ -663,6 +663,16 @@ To generate a text assembly file, pass ``--emit-asm``:
 To generate LLVM bitcode, use the ``--emit-llvm`` flag.
 To generate LLVM bitcode in textual form, use the ``--emit-llvm-text`` flag.
 
+To generate a stub after running the preprocessor, use the ``-E`` flag.
+
+::
+
+    ispc foo.ispc -E -o foo.i  OR  ispc foo.ispc -E -o foo.ispi
+
+In this mode, the preprocessor will assume ``stdout`` if no output file is
+specified.
+Both ``.i`` or ``.ispi`` are valid suffixes.
+
 Optimizations are on by default; they can be turned off with ``-O0``:
 
 ::
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -691,6 +691,7 @@ int main(int Argc, char *Argv[]) {
             g->generateDebuggingSymbols = true;
         } else if (!strcmp(argv[i], "-E")) {
             g->onlyCPP = true;
+            ot = Module::CPPStub;
         } else if (!strcmp(argv[i], "--emit-asm"))
             ot = Module::Asm;
         else if (!strcmp(argv[i], "--emit-llvm"))
@@ -1025,6 +1026,10 @@ int main(int Argc, char *Argv[]) {
         Warning(SourcePos(), "Both -M and -MMM specified on the command line. "
                              "-MMM takes precedence.");
         flags &= Module::GenerateMakeRuleForDeps;
+    }
+
+    if (g->onlyCPP && outFileName == nullptr) {
+        outFileName = "-"; // Assume stdout by default (-E mode)
     }
 
     if (outFileName == NULL && headerFileName == NULL &&
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -135,7 +135,7 @@ static void lPrintVersion() {
     printf("    [--host-stub <filename>]\t\tEmit host-side offload stub functions to file\n");
     printf("    [-h <name>/--header-outfile=<name>]\tOutput filename for header\n");
     printf("    [-I <path>]\t\t\t\tAdd <path> to #include file search path\n");
-    printf("    [--ignore-preprocessor-errors]\t\t\tSuppress errors from the preprocessor\n");
+    printf("    [--ignore-preprocessor-errors]\tSuppress errors from the preprocessor\n");
     printf("    [--instrument]\t\t\tEmit instrumentation to gather performance data\n");
     printf("    [--math-lib=<option>]\t\tSelect math library\n");
     printf("        default\t\t\t\tUse ispc's built-in math functions\n");
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -94,13 +94,11 @@ static void lPrintVersion() {
 #endif
 }
 
-
 // Command line argument constants
 static constexpr std::string_view ARG_NAME_ONLY_PREPROCESSOR{"--onlycpp"};
 static constexpr std::string_view ARG_DESC_ONLY_PREPROCESSOR{"Run only the preprocessor (if enabled)"};
 static constexpr std::string_view ARG_NAME_ONLY_PREPROCESSOR_SHORT{"-E"};
 static constexpr std::string_view ARG_DESC_ONLY_PREPROCESSOR_SHORT{"An alias for [--onlycpp]"};
-
 
 [[noreturn]] static void usage(int ret) {
     lPrintVersion();
@@ -715,7 +713,8 @@ int main(int Argc, char *Argv[]) {
 #endif
         else if (!strcmp(argv[i], "--enable-llvm-intrinsics")) {
             g->enableLLVMIntrinsics = true;
-        } else if (!strncmp(argv[i], ARG_NAME_ONLY_PREPROCESSOR_SHORT.data(), ARG_NAME_ONLY_PREPROCESSOR_SHORT.size())) { // -E
+        } else if (!strncmp(argv[i], ARG_NAME_ONLY_PREPROCESSOR_SHORT.data(),
+                            ARG_NAME_ONLY_PREPROCESSOR_SHORT.size())) { // -E
             g->onlyCPP = true;
         } else if (!strcmp(argv[i], "-I")) {
             if (++i != argc) {
@@ -771,7 +770,8 @@ int main(int Argc, char *Argv[]) {
             else {
                 errorHandler.AddError("Unknown --math-lib= option \"%s\".", lib);
             }
-        } else if (!strncmp(argv[i], ARG_NAME_ONLY_PREPROCESSOR.data(), ARG_NAME_ONLY_PREPROCESSOR.size())) { // --only-preprocessor
+        } else if (!strncmp(argv[i], ARG_NAME_ONLY_PREPROCESSOR.data(),
+                            ARG_NAME_ONLY_PREPROCESSOR.size())) { // --only-preprocessor
             g->onlyCPP = true;
         } else if (!strncmp(argv[i], "--opt=", 6)) {
             const char *opt = argv[i] + 6;
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -94,12 +94,6 @@ static void lPrintVersion() {
 #endif
 }
 
-// Command line argument constants
-static constexpr std::string_view ARG_NAME_ONLY_PREPROCESSOR{"--onlycpp"};
-static constexpr std::string_view ARG_DESC_ONLY_PREPROCESSOR{"Run only the preprocessor (if enabled)"};
-static constexpr std::string_view ARG_NAME_ONLY_PREPROCESSOR_SHORT{"-E"};
-static constexpr std::string_view ARG_DESC_ONLY_PREPROCESSOR_SHORT{"An alias for [--onlycpp]"};
-
 [[noreturn]] static void usage(int ret) {
     lPrintVersion();
     printf("\nusage: ispc\n");
@@ -121,6 +115,7 @@ static constexpr std::string_view ARG_DESC_ONLY_PREPROCESSOR_SHORT{"An alias for
     printf("    [--dllexport]\t\t\tMake non-static functions DLL exported.  Windows target only\n");
     printf("    [--dwarf-version={2,3,4}]\t\tGenerate source-level debug information with given DWARF version "
            "(triggers -g).  Ignored for Windows target\n");
+    printf("    [-E]\t\t\t\tRun only the preprocessor (if enabled)\n");
     printf("    [--emit-asm]\t\t\tGenerate assembly language file as output\n");
     printf("    [--emit-llvm]\t\t\tEmit LLVM bitcode file as output\n");
     printf("    [--emit-llvm-text]\t\t\tEmit LLVM bitcode file as output in textual form\n");
@@ -135,7 +130,6 @@ static constexpr std::string_view ARG_DESC_ONLY_PREPROCESSOR_SHORT{"An alias for
     printf("    [--enable-llvm-intrinsics]\t\tEnable experimental feature to call LLVM intrinsics from ISPC "
            "source code\n");
     printf("    [--error-limit=<value>]\t\tLimit maximum number of errors emitting by ISPC to <value>\n");
-    printf("    [%s]\t\t\t\t%s\n", ARG_NAME_ONLY_PREPROCESSOR_SHORT.data(), ARG_DESC_ONLY_PREPROCESSOR_SHORT.data());
     printf("    [--force-alignment=<value>]\t\tForce alignment in memory allocations routine to be <value>\n");
     printf("    [-g]\t\t\t\tGenerate source-level debug information\n");
     printf("    [--help]\t\t\t\tPrint help\n");
@@ -143,6 +137,7 @@ static constexpr std::string_view ARG_DESC_ONLY_PREPROCESSOR_SHORT{"An alias for
     printf("    [--host-stub <filename>]\t\tEmit host-side offload stub functions to file\n");
     printf("    [-h <name>/--header-outfile=<name>]\tOutput filename for header\n");
     printf("    [-I <path>]\t\t\t\tAdd <path> to #include file search path\n");
+    printf("    [--ignorecpp]\t\t\tSuppress errors from the preprocessor (if enabled)\n");
     printf("    [--instrument]\t\t\tEmit instrumentation to gather performance data\n");
     printf("    [--math-lib=<option>]\t\tSelect math library\n");
     printf("        default\t\t\t\tUse ispc's built-in math functions\n");
@@ -164,7 +159,6 @@ static constexpr std::string_view ARG_DESC_ONLY_PREPROCESSOR_SHORT{"An alias for
     printf("        -O0\t\t\t\tOptimizations disabled.\n");
     printf("        -O1\t\t\t\tOptimization for size.\n");
     printf("        -O2/O3\t\t\t\tOptimization for speed.\n");
-    printf("    [%s]\t\t%s\n", ARG_NAME_ONLY_PREPROCESSOR.data(), ARG_DESC_ONLY_PREPROCESSOR.data());
     printf("    [--opt=<option>]\t\t\tSet optimization option\n");
     printf("        disable-assertions\t\tRemove assertion statements from final code.\n");
     printf("        disable-fma\t\t\tDisable 'fused multiply-add' instructions (on targets that support them)\n");
@@ -697,6 +691,8 @@ int main(int Argc, char *Argv[]) {
             g->noPragmaOnce = true;
         else if (!strcmp(argv[i], "-g")) {
             g->generateDebuggingSymbols = true;
+        } else if (!strcmp(argv[i], "-E")) {
+            g->onlyCPP = true;
         } else if (!strcmp(argv[i], "--emit-asm"))
             ot = Module::Asm;
         else if (!strcmp(argv[i], "--emit-llvm"))
@@ -713,18 +709,17 @@ int main(int Argc, char *Argv[]) {
 #endif
         else if (!strcmp(argv[i], "--enable-llvm-intrinsics")) {
             g->enableLLVMIntrinsics = true;
-        } else if (!strncmp(argv[i], ARG_NAME_ONLY_PREPROCESSOR_SHORT.data(),
-                            ARG_NAME_ONLY_PREPROCESSOR_SHORT.size())) { // -E
-            g->onlyCPP = true;
         } else if (!strcmp(argv[i], "-I")) {
             if (++i != argc) {
                 lParseInclude(argv[i]);
             } else {
                 errorHandler.AddError("No path specified after -I option.");
             }
-        } else if (!strncmp(argv[i], "-I", 2))
+        } else if (!strncmp(argv[i], "-I", 2)) {
             lParseInclude(argv[i] + 2);
-        else if (!strcmp(argv[i], "--fuzz-test"))
+        } else if (!strcmp(argv[i], "--ignorecpp")) {
+            g->ignoreCPP = true;
+        } else if (!strcmp(argv[i], "--fuzz-test"))
             g->enableFuzzTest = true;
         else if (!strncmp(argv[i], "--fuzz-seed=", 12))
             g->fuzzTestSeed = atoi(argv[i] + 12);
@@ -770,9 +765,6 @@ int main(int Argc, char *Argv[]) {
             else {
                 errorHandler.AddError("Unknown --math-lib= option \"%s\".", lib);
             }
-        } else if (!strncmp(argv[i], ARG_NAME_ONLY_PREPROCESSOR.data(),
-                            ARG_NAME_ONLY_PREPROCESSOR.size())) { // --only-preprocessor
-            g->onlyCPP = true;
         } else if (!strncmp(argv[i], "--opt=", 6)) {
             const char *opt = argv[i] + 6;
             if (!strcmp(opt, "fast-math"))
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -62,8 +62,6 @@
 #include <llvm/Support/TargetSelect.h>
 #include <llvm/Support/ToolOutputFile.h>
 
-#include <string_view>
-
 using namespace ispc;
 
 #ifdef ISPC_HOST_IS_WINDOWS
@@ -115,7 +113,7 @@ static void lPrintVersion() {
     printf("    [--dllexport]\t\t\tMake non-static functions DLL exported.  Windows target only\n");
     printf("    [--dwarf-version={2,3,4}]\t\tGenerate source-level debug information with given DWARF version "
            "(triggers -g).  Ignored for Windows target\n");
-    printf("    [-E]\t\t\t\tRun only the preprocessor (if enabled)\n");
+    printf("    [-E]\t\t\t\tRun only the preprocessor\n");
     printf("    [--emit-asm]\t\t\tGenerate assembly language file as output\n");
     printf("    [--emit-llvm]\t\t\tEmit LLVM bitcode file as output\n");
     printf("    [--emit-llvm-text]\t\t\tEmit LLVM bitcode file as output in textual form\n");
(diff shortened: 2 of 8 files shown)


dbabokin:
To fix formatting issue run clang-format like this:
```
clang-format -i src/main.cpp src/module.h
```

aneshlya:
Please add a lit test to tests/lit-tests checking -E switch and termination does not happen anymore.

src/main.cpp
```diff
@@ -92,6 +94,14 @@ static void lPrintVersion() {
 #endif
 }
 
+
+// Command line argument constants
+static constexpr std::string_view ARG_NAME_ONLY_PREPROCESSOR{"--onlycpp"};
```
dbabokin, line 99: What's the point of having this as a `string_view` vs just using the string literal in the `printf()`?

src/main.cpp
```diff
@@ -92,6 +94,14 @@ static void lPrintVersion() {
...
+static constexpr std::string_view ARG_DESC_ONLY_PREPROCESSOR{"Run only the preprocessor (if enabled)"};
+static constexpr std::string_view ARG_NAME_ONLY_PREPROCESSOR_SHORT{"-E"};
+static constexpr std::string_view ARG_DESC_ONLY_PREPROCESSOR_SHORT{"An alias for [--onlycpp]"};
```
dbabokin, line 102: I think `-E` is sufficient and not need for `--onlycpp`.
In general we prefer to have verbose spelling for the switches, except for very standard ones. And `-E` is this kind of standard switches, that every programmer understands without explanation.

dbabokin (commented):
As Arina mentioned, we need lit tests.

And we need to fix build problems with LLVM 11/12.

src/main.cpp
e3m3: I also generally prefer having the verbose spelling. I wasn't familiar with the `-E` switch, so I added the alias, but I can remove it if preferred.

src/main.cpp
e3m3: I like this style (explicitly named constants) for a few reasons:
* It is portable (uses `std`),
* It is resolved at compile-time,
* It leverages type info/behavior, for example the calls `strncmp` which are safer than calls to `strcmp`,
* It reduces the line size and clutter of the conditional logic in the body (max number of line characters is determined by naming convention and not the string contents),
* It allows for better IDE interactions (e.g. typing `ARG_` is enough to autocomplete contextually when filling out the conditional logic).

Happy to change this back to inline literals.

src/main.cpp
dbabokin: yeah, I prefer to remove it. I would put it this way - I would expect everyone who might need it to be familiar with `-E` for C/C++ compilers. And I don't want to have too many switches, just to keep things simple. "Less is more" :)

At the same time, I've talked to engineers who compile a lot of ISPC code through Compiler Explorer and they really like the option to ignore preprocessor failures. I.e. they copy-paste huge file and don't bother if all of the `#include` directives are removed. I suggest adding `--ignore-cpp-errors` or `--ignore-preprocessor-errors`. @JeffRous any opinion on the switch naming?

e3m3: > yeah, I prefer to remove it. I would put it this way - I would expect everyone who might need it to be familiar with `-E` for C/C++ compilers. And I don't want to have too many switches, just to keep things simple. "Less is more" :)
> 
> At the same time, I've talked to engineers who compile a lot of ISPC code through Compiler Explorer and they really like the option to ignore preprocessor failures. I.e. they copy-paste huge file and don't bother if all of the `#include` directives are removed. I suggest adding `--ignore-cpp-errors` or `--ignore-preprocessor-errors`. @JeffRous any opinion on the switch naming?

To clarify, currently `--onlycpp` runs the preprocessor and exits the compiler after it completes. It does not suppress errors (that occur within macro code). Is the behavior you suggest: to add a flag that runs the entire ISPC compile jobs (assuming it can complete) while suppressing errors from `execPreprocessors`?

e3m3: And also remove the alias?

src/main.cpp
dbabokin: My main concern is readability of resulting code, i.e. right now it's almost 1:1 what we have in the source file vs the output of the `--help`. With ARG_* thing, it's difficult to see the complete picture and catch obvious things looking at source code.

`strcmp`/`strncmp` argument is fair, though for relatively short things, like `-E`, it's less important.

As description string is used only once in the code and parameter name only twice (in the help message and in the matching logic), adding the wrappers will significantly increase code size (in terms of source lines / number of characters in the source, not executable size) and will compromise readability.

So, I suggest doing it old way. Or consider redoing the whole thing, if/when we get too many switches, so readability of the help message / source code will not matter that much.

src/main.cpp
dbabokin: I suggest:
- removing the alias, i.e. `--onlycpp`
- add new switch to preserve old behavior of ignoring preprocessor errors. I.e. the compilation should not fail if preprocessor returned an error. Typical use case if not failing if the `#include "my_header.isph"` was not able fide the file to include.

To be clear, these changes are unrelated, I should have mentioned them in two separate comments to avoid confusion.

tests/lit-tests/2280.ispc
```diff
@@ -0,0 +1,30 @@
...
+// RUN: not %{ispc} --target=host --nowrap --nostdlib -g -O0         %s                     2>&1 | FileCheck %s -check-prefix=CHECK_ERROR_0
+// RUN: not %{ispc} --target=host --nowrap --nostdlib -g -O0         %s                     2>&1 | FileCheck %s -check-prefix=CHECK_ERROR_0
+
+// RUN: not %{ispc} --target=host --nowrap --nostdlib -g -O0         %s           -DDO_ELSE 2>&1 | FileCheck %s -check-prefix=CHECK_ERROR_1
+// RUN: not %{ispc} --target=host --nowrap --nostdlib -g -O0         %s           -DDO_ELSE 2>&1 | FileCheck %s -check-prefix=CHECK_ERROR_1
+
+// RUN: not %{ispc} --target=host --nowrap --nostdlib -g -O0         %s -E        -DDO_ELSE 2>&1 | FileCheck %s -check-prefix=CHECK_ERROR_1
+// RUN: not %{ispc} --target=host --nowrap --nostdlib -g -O0         %s --onlycpp -DDO_ELSE 2>&1 | FileCheck %s -check-prefix=CHECK_ERROR_1
+
+// REQUIRES: X86_ENABLED
```
dbabokin, line 13: X86_ENABLED is not needed, this functionality should work on any target.

tests/lit-tests/2280.ispc
```diff
@@ -0,0 +1,30 @@
+// RUN:     %{ispc} --target=host --nowrap --nostdlib -g -O0 -o %t.o %s -E
```
dbabokin, line 1: A few comments here
`-g -O0` is not essential here, so no reason to use it.
`--nowrap` is needed only when you are matching the output for a specific error message, so it's not wrapped depending on the terminal settings.
`--nostdlib` is good to use, as it basically reduces compile time.

dbabokin, line 1: I suggest checking specific things that where done right and can be detected in the preprocessor out put, specifically:
- inlining done right
- macros are substituted. I.e. pass `-DMYMACRO=MACRO_VALUE` and detect that `MACRO_VALUE` presents in the output.

tests/lit-tests/2280.ispc
```diff
@@ -0,0 +1,30 @@
...
+// RUN:     %{ispc} --target=host --nowrap --nostdlib -g -O0 -o %t.o %s --onlycpp
+
+// RUN: not %{ispc} --target=host --nowrap --nostdlib -g -O0         %s                     2>&1 | FileCheck %s -check-prefix=CHECK_ERROR_0
+// RUN: not %{ispc} --target=host --nowrap --nostdlib -g -O0         %s                     2>&1 | FileCheck %s -check-prefix=CHECK_ERROR_0
```
dbabokin, line 5: What's the different between these two runs?

e3m3: Forgot to remove

tests/lit-tests/2280.ispc
e3m3: Based on this, wouldn't is be safest to default to `--nowrap` so that terminal settings are never assumed (and therefore can be safely ignored)?

dbabokin: Wrapping is good for human beings. They are bad for parsers.

e3m3:
Checks passed. Ready for review.

src/module.cpp
```diff
@@ -2285,6 +2290,7 @@ void Module::execPreprocessor(const char *infilename, llvm::raw_string_ostream *
 
     llvm::IntrusiveRefCntPtr<clang::DiagnosticIDs> diagIDs(new clang::DiagnosticIDs);
     clang::DiagnosticsEngine *diagEngine = new clang::DiagnosticsEngine(diagIDs, diagOptions, diagPrinter);
```
e3m3, line 2292: For my understanding: Do we care about the leaks in the preprocessor? I think this leaks for every module processed.

src/main.cpp
```diff
@@ -134,6 +137,7 @@ static void lPrintVersion() {
     printf("    [--host-stub <filename>]\t\tEmit host-side offload stub functions to file\n");
     printf("    [-h <name>/--header-outfile=<name>]\tOutput filename for header\n");
     printf("    [-I <path>]\t\t\t\tAdd <path> to #include file search path\n");
+    printf("    [--ignorecpp]\t\t\tSuppress errors from the preprocessor (if enabled)\n");
```
e3m3, line 140: Other names welcome.

src/main.cpp
```diff
@@ -62,6 +62,8 @@
 #include <llvm/Support/TargetSelect.h>
 #include <llvm/Support/ToolOutputFile.h>
 
+#include <string_view>
+
```
e3m3, line 66: Forgot to remove. Remove?

tests/lit-tests/2280.ispc
```diff
@@ -0,0 +1,33 @@
+// RUN:     %{ispc} %s --target=host --nostdlib --emit-llvm-text -E                       -o -
```
e3m3, line 1: Wasn't sure how to check for null output (empty stdin && no generated files). Checks for compiler status code.

tests/lit-tests/2280.ispc
```diff
@@ -0,0 +1,33 @@
...
+// RUN:     %{ispc} %s --target=host --nostdlib --emit-llvm-text                          -o - | FileCheck %s -check-prefix=CHECK_PASS_0
+// RUN:     %{ispc} %s --target=host --nostdlib --emit-llvm-text                          -o - | FileCheck %s -check-prefix=CHECK_PASS_2
```
e3m3, line 3: Is the only way to merge these into two checks using the same output to use a file? `%-NEXT` has to match the following line, and `%-SAME` has to match the same line.

src/main.cpp
dbabokin: yes

src/main.cpp
dbabokin: Let's go with `--ignore-preprocessor-errors`.
My reasoning is the following: `cpp` stands for "C PreProcessor" only for very limited number of people. For the rest it's simply confusing. So any other naming will be confusing and not clear without help message.
And `-ignorecpp` is especially confusing. Could you please change internal var name to `ignoreCPPErrors` as well?

src/main.cpp
```diff
@@ -113,6 +115,7 @@ static void lPrintVersion() {
     printf("    [--dllexport]\t\t\tMake non-static functions DLL exported.  Windows target only\n");
     printf("    [--dwarf-version={2,3,4}]\t\tGenerate source-level debug information with given DWARF version "
            "(triggers -g).  Ignored for Windows target\n");
+    printf("    [-E]\t\t\t\tRun only the preprocessor (if enabled)\n");
```
dbabokin, line 118: `(if enabled)` is excessive, please remove. Same for the other switch.

src/module.cpp
dbabokin: In general yes, we care about leaks, we should avoid all the leaks. But there are things that fundamentally difficult to fix wrt leaks, so we didn't pay close attention to it.

tests/lit-tests/2280.ispc
dbabokin: Good point! Default behavior for C compilers (clang, gcc and others) is to dump the output to stdout, i.e. assume `-o -` in this case. We should do the same.

tests/lit-tests/2280.ispc
dbabokin: No, you just need to have to lines with corresponding prefix, that's it. I.e.:
```
// CHECK_PASS_0: ret i32 1337
// CHECK_PASS_0: ret i32 7331
```

src/main.cpp
dbabokin: It's resolved, the naming of the "ignore errors" switch is discussed in another thread.

tests/lit-tests/2280.ispc
dbabokin: As for checking that no "object file output produced, I think it's ok not to check that.

e3m3:
Issue:  Currently, `-E` mode lit tests will self-pass because the comments containing the match-text are not stripped out by the preprocessor. Would it be acceptable to remove these lines entirely from the output after preprocessing but before writing?

e3m3:
> Issue: Currently, `-E` mode lit tests will self-pass because the comments containing the match-text are not stripped out by the preprocessor. Would it be acceptable to remove these lines entirely from the output after preprocessing but before writing?

Or specifically, any line matching starting `^//`?

dbabokin:
> Issue: Currently, `-E` mode lit tests will self-pass because the comments containing the match-text are not stripped out by the preprocessor. Would it be acceptable to remove these lines entirely from the output after preprocessing but before writing?

You can use different files as source file and file with patterns to match.

src/main.cpp
```diff
@@ -134,6 +135,7 @@ static void lPrintVersion() {
     printf("    [--host-stub <filename>]\t\tEmit host-side offload stub functions to file\n");
     printf("    [-h <name>/--header-outfile=<name>]\tOutput filename for header\n");
     printf("    [-I <path>]\t\t\t\tAdd <path> to #include file search path\n");
+    printf("    [--ignore-preprocessor-errors]\t\t\tSuppress errors from the preprocessor\n");
```
dbabokin, line 138: Only one /t is needed to do proper alignment.

src/module.cpp
```diff
@@ -3029,10 +3081,22 @@ int Module::CompileAndOutput(const char *srcFile, Arch arch, const char *cpu, st
...
-            if ((outputType == Bitcode) || (outputType == BitcodeText))
+            switch (outputType) {
+            case Bitcode:
+            case BitcodeText:
                 writeBitcode(dispatchModule, outFileName, outputType);
-            else
+                break;
+
+            case CPPStub:
+                fprintf(stderr, "Entering CPPStub function\n");
```
dbabokin, line 3091: Looks like a debug print leftover.

src/module.cpp
```diff
@@ -1925,6 +1935,39 @@ bool Module::writeDevStub(const char *fn) {
...
+    int fd;
+    int flags = O_CREAT | O_WRONLY | O_TRUNC;
+
+    if (!outFileName) {
+        return false;
+    } else if (!strcmp("-", outFileName)) {
+        fd = 1;
+    } else {
+#ifdef ISPC_HOST_IS_WINDOWS
+        flags |= O_BINARY;
```
dbabokin, line 1954: It's not binary data, so probably we don't need it. As I understand it will affect the end of lines.

e3m3:
Ready to merge

src/module.cpp
```diff
@@ -2805,8 +2850,11 @@ int Module::CompileAndOutput(const char *srcFile, Arch arch, const char *cpu, st
             return 1;
 
         m = new Module(srcFile);
-        if (m->CompileFile() == 0) {
-            llvm::TimeTraceScope TimeScope("Backend");
+        const int compileResult = m->CompileFile();
+
+        llvm::TimeTraceScope TimeScope("Backend");
+
+        if (compileResult == 0) {
```
e3m3, line 2857: > why did you do this change?
> 
> const int compileResult = m->CompileFile(); llvm::TimeTraceScope TimeScope("Backend"); if (compileResult == 0) {
> 
> the difference that I see is that only the scope that TimeScope("Backend"); covers
> 
> and it now covers a couple delete operators. I'm not sure if it's intended
> 
> I.e. in covers the timing between its creation till destruction. By pulling it to another scope you extended its live time. So the question if this is intended or not
@dbabokin 

In the previous version of `CompileAndOutput`, when `! (targets.size() == 0 || targets.size() == 1)`, on every path processing `m`, `Timescope` begins after the call to `CompileFile` and captures the timing for the check of the result of that call.

The change for the `(targets.size() == 0 || targets.size() == 1)` path done here brings the measurement in-line with the other path. This was the easiest way to make the measurement match on both paths, but another structure could also work.

src/module.h
```diff
@@ -197,6 +198,9 @@ class Module {
     const char *filename;
     AST *ast;
 
+    std::string CPPBuffer;
+    llvm::raw_string_ostream CPPStream;
```
e3m3, line 202: > When we do regular compilation, the changes will make previously short-lived buffer (of preprocessed code), live till the Module is destroyed, i.e. during the whole compilation.
@dbabokin

e3m3: > I would probably make a llvm::raw_string_ostream not a member of Module, but a pointer to llvm::raw_string_ostream make a member of the Module and allocate it on the heap. This will require more sophisticated delete logic, though
@dbabokin

tests/lit-tests/2280.ispc
```diff
@@ -0,0 +1,38 @@
...
+// RUN: not %{ispc} %s      --target=host --nostdlib -E                                               -DDO_ELSE 2>&1 | grep -v -e '^// ' | FileCheck %s -check-prefix=CHECK_ERROR
+// RUN: not %{ispc} %s -o - --target=host --nostdlib    --emit-llvm-text                              -DDO_ELSE 2>&1                     | FileCheck %s -check-prefix=CHECK_ERROR
+
+// RUN:     %{ispc} %s      --target=host --nostdlib -E                  --ignore-preprocessor-errors                | grep -v -e '^// ' | FileCheck %s -check-prefix=CHECK_PASS_0
+// RUN:     %{ispc} %s -o - --target=host --nostdlib    --emit-llvm-text --ignore-preprocessor-errors                                    | FileCheck %s -check-prefix=CHECK_PASS_1
+// RUN:     %{ispc} %s      --target=host --nostdlib -E                  --ignore-preprocessor-errors -DDO_ELSE      | grep -v -e '^// ' | FileCheck %s -check-prefix=CHECK_PASS_2
+// RUN:     %{ispc} %s -o - --target=host --nostdlib    --emit-llvm-text --ignore-preprocessor-errors -DDO_ELSE                          | FileCheck %s -check-prefix=CHECK_PASS_3
+
+
```
e3m3, line 11: Additional testing:
> warning when the suffix of the output file is not *.i or *.ispi
> 
> behavior with multi-target. This will need to be a separate test, as it will be x86 specific (as you will need to specify specific targets). Here's the list of pre-defined macros: https://ispc.github.io/ispc.html#the-preprocessor you can use one of them (for example TARGET_WIDTH) in combination with targets of different width like --target=sse4-i32x4,avx2-i32x8 and check how the parsing was done in output files. Also as it's not just a bug fix, but a feature, would be good to give a meaningful name to the test, like preprocessor.ispc / preprocessor_multi_target.ispc
@dbabokin

e3m3:
Todo:
- [ ] Add documentation in [user guide](docs/ispc.rst).

src/module.cpp
dbabokin: That's a good catch, we need to unify it for single-target and multi-target compilations. I think the changes you did make sense. Thanks for explainaing.

e3m3:
Ready to for review.

src/module.cpp
```diff
@@ -272,10 +272,23 @@ int Module::CompileFile() {
...
+        bufferCPP.reset(new CPPBuffer{});
+
+        const int numErrors = execPreprocessor(!IsStdin(filename) ? filename : "-", bufferCPP->os.get());
+        errorCount += (g->ignoreCPPErrors) ? 0 : numErrors;
+
+        if (g->onlyCPP) {
+            return errorCount; // Return early
+        }
+
+        YY_BUFFER_STATE strbuf = yy_scan_string(bufferCPP->str.c_str());
```
dbabokin, line 291: Old behavior - `std::string buffer` is destroyed when the scope is closed. New behavior - `bufferCPP` still lives will the `Module` is destroyed. Am I missing something?

docs/ispc.rst
```diff
@@ -663,6 +663,16 @@ To generate a text assembly file, pass ``--emit-asm``:
...
+To generate a stub after running the preprocessor, use the ``-E`` flag.
+
+::
+
+    ispc foo.ispc -E -o foo.i  OR  ispc foo.ispc -E -o foo.ispi
+
+In this mode, the preprocessor will assume ``stdout`` if no output file is
+specified.
+Both ``.i`` or ``.ispi`` are valid suffixes.
+
```
dbabokin, line 675:
```suggestion
To run only the preprocessor, use ``-E`` flag. By default the output will be directed to ``stdout``, otherwise to the file specified by ``-o`` flag. The output file with preprocessed source is assumed to have ``.i`` or ``.ispi`` extension.

::

    ispc foo.ispc -E -o foo.i
    ispc foo.ispc -E -o foo.ispi

```

src/module.cpp
```diff
@@ -1925,6 +1944,45 @@ bool Module::writeDevStub(const char *fn) {
...
+            perror(outFileName);
+            return false;
+        }
+    }
+
+    // The CPP stream should have been initialized: print and clean it up.
+    if (module->bufferCPP) {
+        llvm::raw_fd_ostream fos(fd, (fd != 1), false);
+        fos << module->bufferCPP->str;
+        module->bufferCPP.reset();
```
e3m3, line 1977: `bufferCPP` deleted by `std::unique_ptr::reset` after emitting to output stream.

src/module.cpp
e3m3: See clean up [here](https://github.com/ispc/ispc/pull/2280/files#r833824859). I think this is sufficient to release the memory, but I may have misaccounted.

Basically, the issue is that we need to keep the string buffer around until the call to `writeOutput`/`writeCPPStub`. I allocated a new pointer for every call to `execPreprocessor`, which should be accompanied by exactly one call to `reset` in `writeCPPStub` once safe to do so.

src/module.cpp
```diff
@@ -160,7 +160,7 @@ static void lStripUnusedDebugInfo(llvm::Module *module) { return; }
 ///////////////////////////////////////////////////////////////////////////
 // Module
 
-Module::Module(const char *fn) {
+Module::Module(const char *fn) : bufferCPP{nullptr} {
```
e3m3, line 163: `bufferCPP` starts as a null `std::unique_ptr`.

src/module.cpp
dbabokin: Yes, this is what happens when `-E` is specified, the problem is when it's not specified.

e3m3: Gotcha. Good catch.

src/module.cpp
e3m3: See commit 006bf80.

dbabokin: This doesn't seem to solve the issue.

What seems to be ok'ish solution - not very elegant, but good enough, is the following:
- buffer is pointer to a wrapper class in the Module (as it is now) and it is initialized as nullptr
- create buffer before calling the preprocessor (as it is now)
- if we go regular compilation - destroy buffer after the parser (after yy_delete_buffer())
- if it's -E compilation - destroy the buffer after it's used in WriteCPPStub.
- assert in WriteCPPStub that buffer valid nullptr.

src/module.cpp
```diff
@@ -1925,6 +1944,45 @@ bool Module::writeDevStub(const char *fn) {
...
+
+    // The CPP stream should have been initialized: print and clean it up.
+    if (module->bufferCPP) {
+        llvm::raw_fd_ostream fos(fd, (fd != 1), false);
+        fos << module->bufferCPP->str;
+        module->bufferCPP.reset();
+    } else {
+        perror("Unexpected empty/null `CPPStream`");
+        return false;
+    }
```
e3m3, line 1981: Would this qualify as a validity assertion, or is a call to `assert` preferred so that the check is removed under `Release`?

src/module.cpp
e3m3: > * if it's -E compilation - destroy the buffer after it's used in WriteCPPStub.
> * assert in WriteCPPStub that buffer valid nullptr.

See [comment](https://github.com/ispc/ispc/pull/2280/files#r834598949).

src/module.cpp
dbabokin: I think it's better to have as an assert, as if nullptr comes here, it's really an internal error, which should never happen.

Note that `false` returned in multi-target compilation is currently ignored. This is probably need to be fixed, as well.

e3m3: Nice catch. Should be resolved now.

e3m3: > Nice catch. Should be resolved now.

Nevermind, there are some regressions.

src/module.cpp
```diff
@@ -3029,10 +3092,19 @@ int Module::CompileAndOutput(const char *srcFile, Arch arch, const char *cpu, st
         lEmitDispatchModule(dispatchModule, exportedFunctions);
 
         if (outFileName != NULL) {
-            if ((outputType == Bitcode) || (outputType == BitcodeText))
+            switch (outputType) {
+            case CPPStub:
```
dbabokin, line 3096: Why not writing multiple *.i files in case of multi-target? This worked fine in previous revisions?

src/module.cpp
```diff
@@ -3029,10 +3092,19 @@ int Module::CompileAndOutput(const char *srcFile, Arch arch, const char *cpu, st
...
+                break;
+
+            case Bitcode:
+            case BitcodeText:
                 writeBitcode(dispatchModule, outFileName, outputType);
-            else
+                break;
+
+            default:
+                Assert((outputType == Module::Object || outputType == Module::Asm) && "Unexpected `outputType`");
```
dbabokin, line 3105: Let's have a case for `Object` and `Asm` and assert in default case that it should not be reachable.

src/module.cpp
```diff
@@ -3029,10 +3092,23 @@ int Module::CompileAndOutput(const char *srcFile, Arch arch, const char *cpu, st
...
-            else
+                break;
+
+            case Asm:
+            case Object:
                 writeObjectFileOrAssembly(firstTargetMachine, dispatchModule, outputType, outFileName);
+                break;
+
+            default:
+                Assert(false && "Unexpected `outputType`");
```
dbabokin, line 3110:
```suggestion
                FATAL("Unexpected `outputType`");
```

src/module.h
```diff
@@ -197,6 +198,17 @@ class Module {
     const char *filename;
     AST *ast;
 
+    // Definition and member object capturing preprocessing stream during Module lifetime.
+    struct CPPBuffer {
+        CPPBuffer() : str{}, os{std::make_unique<llvm::raw_string_ostream>(str)} {}
+        CPPBuffer(CPPBuffer &&o) : str{std::move(o.str)}, os{std::move(o.os)} {}
```
dbabokin, line 204: Move constructor seems to be unused. Probably we should remove it.

src/module.cpp
dbabokin:
```suggestion
            case CPPStub:
                // No preprocessor output for dispatch module.
```

src/module.cpp
```diff
@@ -3029,10 +3092,23 @@ int Module::CompileAndOutput(const char *srcFile, Arch arch, const char *cpu, st
...
 
         if (outFileName != NULL) {
-            if ((outputType == Bitcode) || (outputType == BitcodeText))
+            switch (outputType) {
+            case CPPStub:
+                break;
+
+            case Bitcode:
+            case BitcodeText:
                 writeBitcode(dispatchModule, outFileName, outputType);
```
dbabokin, line 3101: Let's do proper error reporting here, i.e.
```c++
if (!writeBitcode(dispatchModule, outFileName, outputType))
    return 1;
```

Same for `writeObjectFileOrAssembly` below

PR #2299, opened 2022-04-15
--ast-dump switch

dbabokin:
This PR adds `--ast-dump` switch, which is similar to clang's one. May be used in the following form:
- `--ast-dump=user` or `--ast-dump` to print AST for user code.
- `--ast-dump=all` to print all the code including stdlib.

The format of AST dump is intentionally made similar to format of clang AST, so it would look familiar for those who has experience with clang.

Other changes worth to note:
- `ASTNode` (and hence `Stmt` and `Expr`) class, and `Function` got member function `void Print()`, which works in the debugger. The one that was there previously didn't work in the debugger out of the box, as didn't do flashing, so nothing was printed unless `fflush()` was invoked manually.
- Most of functionality for `Print()` functions were kept the same, in some cases it was extended. But the key focus was new format. So there are things to improve in terms of what is being printed.
- `MathLib` and `CodegenOptLevel` `enum`s were changed to be `enum class` for code style uniformity.

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -66,7 +66,8 @@ void Indent::pushList(int i) {
 
 void Indent::setNextLabel(std::string s) { label = s; }
 
-void Indent::Print() {
+// Print indent and an optional string
+void Indent::Print(const char *title) {
     printCalls++;
     Assert(!stack.empty());
     int &top = stack.back();
@@ -91,22 +92,24 @@ void Indent::Print() {
         printf("(%s) ", label.c_str());
         label.clear();
     }
-}
 
-void Indent::Print(const char *title) {
-    Print();
+    // An optional string
     if (title != nullptr) {
         printf("%s", title);
     }
 }
 
 void Indent::Print(const char *title, const SourcePos &pos) {
+    // Same as previous version
     Print(title);
+    // Plus source position info
     pos.Print();
 }
 
 void Indent::PrintLn(const char *title, const SourcePos &pos) {
+    // Same as previous version
     Print(title, pos);
+    // Plus end of line
     printf("\n");
 }
 
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -147,6 +147,36 @@ void AST::GenerateIR() {
         functions[i]->GenerateIR();
 }
 
+void AST::Print(Globals::ASTDumpKind printKind) const {
+    if (printKind == Globals::ASTDumpKind::None) {
+        return;
+    }
+
+    printf("AST\n");
+    Indent indent;
+
+    int funcsToPrint = 0;
+    if (printKind == Globals::ASTDumpKind::All) {
+        funcsToPrint = functions.size();
+    } else if (printKind == Globals::ASTDumpKind::User) {
+        for (unsigned int i = 0; i < functions.size(); ++i) {
+            if (!functions[i]->IsStdlibSymbol()) {
+                funcsToPrint++;
+            }
+        }
+    }
+
+    indent.pushList(funcsToPrint);
+    for (unsigned int i = 0; i < functions.size(); ++i) {
+        if (printKind == Globals::ASTDumpKind::All ||
+            (printKind == Globals::ASTDumpKind::User && !functions[i]->IsStdlibSymbol())) {
+            functions[i]->Print(indent);
+        }
+    }
+
+    fflush(stdout);
+}
+
 ///////////////////////////////////////////////////////////////////////////
 
 ASTNode *ispc::WalkAST(ASTNode *node, ASTPreCallBackFunc preFunc, ASTPostCallBackFunc postFunc, void *data) {
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -125,6 +125,13 @@ void Indent::Done() {
 
 ASTNode::~ASTNode() {}
 
+void ASTNode::Print() const {
+    Indent indent;
+    indent.pushSingle();
+    Print(indent);
+    fflush(stdout);
+}
+
 ///////////////////////////////////////////////////////////////////////////
 // AST
 
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -1,5 +1,5 @@
 /*
-  Copyright (c) 2011-2021, Intel Corporation
+  Copyright (c) 2011-2022, Intel Corporation
   All rights reserved.
 
   Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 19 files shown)


src/ast.cpp
```diff
@@ -48,11 +48,90 @@
...
+
+void Indent::Print(const char *title, const SourcePos &pos) {
+    Print(title);
+    pos.Print();
+}
+
+void Indent::PrintLn(const char *title, const SourcePos &pos) {
+    Print(title, pos);
+    printf("\n");
+}
```
e3m3, line 111: Why `printf()` instead of `Print()`?

dbabokin: `Indent::Print()` functions are implemented using `printf()`.

`PrintLn()` does the same as `Print()`, but prints `\n` in the on top of it.

src/decl.cpp
```diff
@@ -307,23 +314,31 @@ void Declarator::Print(int indent) const {
...
+        indent.setNextLabel("init");
+        initExpr->Print(indent);
     }
 
     if (functionParams.size() > 0) {
         for (unsigned int i = 0; i < functionParams.size(); ++i) {
-            printf("\n%*cfunc param %d:\n", indent, ' ', i);
-            functionParams[i]->Print(indent + 4);
+            char buffer[20];
+            snprintf(buffer, 20, "func param %d", i);
```
e3m3, line 330:
```suggestion
            static constexpr std::size_t BUFSIZE{20};
            char buffer[BUFSIZE];
            snprintf(buffer, BUFSIZE, "func param %d", i);
```

src/ast.h
```diff
@@ -38,10 +38,59 @@
...
+    Parent Node
+    `-Single Nested Node
+
+    Parent Node
+    |-List Node #1
+    |-List Node #2
+    `-List Node #3
+
+    If the nested node needs to be annotated, setNextLabel(string) should be called
+    before recusing to the nested node.
```
e3m3, line 62:
```suggestion
    before recursing to the nested node.
```

src/expr.cpp
```diff
@@ -1315,30 +1315,47 @@ int UnaryExpr::EstimateCost() const {
...
+        printf("prefix '++'");
+        break;
+    case PreDec: ///< Pre-decrement
+        printf("prefix '--'");
+        break;
+    case PostInc: ///< Post-increment
+        printf("postfix '++'");
+        break;
+    case PostDec: ///< Post-decrement
+        printf("postfix '++'");
```
e3m3, line 1339:
```suggestion
        printf("postfix '--'");
```

src/expr.cpp
```diff
@@ -2910,13 +2932,13 @@ std::pair<llvm::Constant *, bool> BinaryExpr::GetConstant(const Type *type) cons
...
     case AssignExpr::Assign:
-        return "assignment operator";
+        return "=";
     case AssignExpr::MulAssign:
         return "*=";
     case AssignExpr::DivAssign:
         return "/=";
     case AssignExpr::ModAssign:
-        return "%%=";
+        return "%=";
```
e3m3, line 2941: Very cool :)

src/expr.cpp
dbabokin: Good catch! Thanks!

src/ast.cpp
```diff
@@ -48,11 +48,90 @@
...
+        printf("%s", title);
+    }
+}
+
+void Indent::Print(const char *title, const SourcePos &pos) {
+    Print(title);
+    pos.Print();
+}
+
+void Indent::PrintLn(const char *title, const SourcePos &pos) {
```
aneshlya, line 108: Seems that `void Indent::PrintLn(const char *title)` method might be also useful. I see the potential uses for this, for example `indent.PrintLn("SymbolExpr: <NULL EXPR>");` instead of `indent.Print("SymbolExpr: <NULL EXPR>\n");`

src/func.cpp
```diff
@@ -71,6 +71,90 @@
...
+        printf(" \"%s\"\n", sym->name.c_str());
+    } else {
+        printf("<NULL>");
+    }
+
+    indent.pushList(args.size() + 1);
+    if (args.size() > 0) {
+        for (int i = 0; i < args.size(); i++) {
+            char buffer[15];
+            snprintf(buffer, 15, "param %d", i);
```
dbabokin, line 133:
```suggestion
            static constexpr std::size_t BUFSIZE{15};
            char buffer[BUFSIZE];
            snprintf(buffer, BUFSIZE, "param %d", i);
```

src/ast.cpp
dbabokin: I decided to keep it simple, as in this case it might be solved with extra `\n` in the end of the passed string parameter. In case of two parameters, it can be solved only be extra `printf("\n");` call, which is bulky.

aneshlya:
I think we should avoid mixing `indent.Print()` and `printf()` functions, like here:
```
indent.Print("AssertStmt", pos);
printf("Message: %s\n", message.c_str());
indent.pushSingle();
```
Without looking into `Indent` class implementation, it is not clear when to use `indent.Print` and when to use `printf`.

I suggest clearly describing the expected usage in the comments for `Indent` class and adding comments for each Indent::Print method.
Also I recommend creating an issue to implement a wrapper class which can be used for all AST print operations to avoid mixing `indent.Print` and `printf` and to hide implementation details about printing from users.

e3m3:
> I think we should avoid mixing `indent.Print()` and `printf()` functions, like here:
> 
> ```
> indent.Print("AssertStmt", pos);
> printf("Message: %s\n", message.c_str());
> indent.pushSingle();
> ```
> 
> Without looking into `Indent` class implementation, it is not clear when to use `indent.Print` and when to use `printf`.
> 
> I suggest clearly describing the expected usage in the comments for `Indent` class and adding comments for each Indent::Print method. Also I recommend creating an issue to implement a wrapper class which can be used for all AST print operations to avoid mixing `indent.Print` and `printf` and to hide implementation details about printing from users.

I think this was part of my confusion [here](https://github.com/ispc/ispc/pull/2299#discussion_r851554933) as well. I think for clarity, calling `Print(...)` should be `Indent::Print(this, ...)` or `this->Print(...)` in lines where the function is called without an object

dbabokin:
To address the confusion of mixing `indent.Print()` and `printf()` outside of `Indent` class, I added a comment in [this commit](https://github.com/ispc/ispc/pull/2299/commits/60146142505445121d44270deeff6f64a1091572).

As for mixing `Indent::Print()` and `printf()` in `Indent` implementation, I like neither of suggested improvement and addressed that in [this commit](https://github.com/ispc/ispc/pull/2299/commits/13add768e7008013682dfd900a5701b32a4a6a9e) through comments. Hope that works.

PR #2303, opened 2022-04-20
CI to test with LLVM 14.0

dbabokin:
- `alloy.py` to build 14.0.1 when building `14.0`
- Rebuild14 pipeline to build files named 14.0.1
- `nightly-14` pipeline renames to `nightly-15` and changed accordingly. It supposed to be enabled when 15.0 is branched
- Remove build directories on Windows before packing LLVM 14, as the worker may run out of space
- GitHub Actions to testing to run LLVM 14 jobs: LLVM 13.0 & 14.0 - full testing, LLVM 11.1 & 12.0 sanity. Was LLVM 13.0 & 12.0 full, LLVM 11.1 - sanity
- Appveyor to test LLVM 13.0 and 14.0, instead of 12.0 and 13.0
- Travis to test 14.0, instead of 13.0.

### DIFF
--- a/.github/workflows/rebuild-llvm14.yml
+++ b/.github/workflows/rebuild-llvm14.yml
@@ -175,6 +175,7 @@ jobs:
       shell: cmd
       run: |
         cd llvm
+        rmdir /s /q build-14.0
         set TAR_BASE_NAME=llvm-14.0.1-win.vs2022-Release+Asserts-x86.arm.wasm
         7z.exe a -ttar -snl %TAR_BASE_NAME%.tar bin-14.0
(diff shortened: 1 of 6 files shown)


.github/workflows/rebuild-llvm14.yml
```diff
@@ -175,6 +175,7 @@ jobs:
       shell: cmd
       run: |
         cd llvm
+        rmdir /s /q build-14.0
```
aneshlya, line 178: Would removing of build directory be beneficial for older llvm versions?  Or we do not build them anymore?

dbabokin: It doesn't matter as long as we don't run out of space. So I didn't do that for other builds.

PR #2309, opened 2022-05-03
Update deps in Dockerfile for xpu build

aneshlya:
Use LLVM 13 and latest dependencies versiions.


dbabokin:
I've push a couple more commits to the branch to use LLVM 13.0 by default across all the `Dockerfile`s and scripts.

PR #2328, opened 2022-05-20
WIP: Support of opaque pointers

aneshlya:


dbabokin (commented):
We should try avoiding working with LLVM types directly and try to have ISPC types in interfaces only.

Though if it appears that it's not practical, we can use the approach with working with LLVM types directly.

aneshlya:
I'm reworking this entirely and will create new PR.

PR #2339, opened 2022-07-07
Updates for Xe targets (8)

dbabokin:
Update for Xe targets with a bunch of features designed for ISPC/SYCL interop (experimental stage):
1. Support of `__regcall` attribute
2. Support of `extern "C"` functions definition
3. New language construct "invoke_sycl" which is used to call SYCL function from ISPC. The function must be declared on ISPC side with `extern "SYCL" __regcall` qualifiers.
4. Support of dynamic linking is ISPC Runtime

Also, the change contains a bit of refactoring and removed `xe` folder that had globals localization pass for Xe targets


aneshlya:
We have two lit tests failing on arm. They need `// REQUIRES: X86_ENABLED`.

aneshlya (approved):

Issue #2283, opened 2022-03-17
Failure building 1.17.0 and 1.18.0 with XE_ENABLED


ArchangeGabriel:
On ArchLinux, using the following components:
```
clang/llvm/openmp 13.0.1
spirv-llvm-translator 13.0.0@c9f1e369
level-zero 1.7.9
vc-intrinsics 0.1.0
```

Building command:
```
cmake -B build -S ispc-1.17.0 \
    -DCMAKE_INSTALL_PREFIX=/usr \
    -DISPC_INCLUDE_EXAMPLES=OFF \
    -DISPC_NO_DUMPS=ON \
    -DXE_ENABLED=ON \
    -DXE_DEPS_DIR=/usr
make -C build
```

The main (ispc) compilation itself goes fine, but during `ispc_xpu_examples` we get this:
```
make[5]: Entering directory '/build/ispc/src/build/xpu_examples/src/ispc_xpu_examples-build'
[ 50%] Building ISPC object /build/ispc/src/build/xpu_examples/src/ispc_xpu_examples-build/aobench/local_ispc/aobench.dev.o
/build/ispc/src/ispc-1.17.0/src/main.cpp(396): FATAL ERROR: Unhandled signal sent to process; terminating.
***
*** Please file a bug report at https://github.com/ispc/ispc/issues
*** (Including as much information as you can about how to reproduce this error).
*** You have apparently encountered a bug in the compiler that we'd like to fix!
***
make[5]: *** [aobench/CMakeFiles/xe_aobench.dir/build.make:75: aobench/local_ispc/aobench.dev.o] Aborted (core dumped)
make[5]: Leaving directory '/build/ispc/src/build/xpu_examples/src/ispc_xpu_examples-build'
make[4]: *** [CMakeFiles/Makefile2:387: aobench/CMakeFiles/xe_aobench.dir/all] Error 2
```

ArchangeGabriel:
The issue is a bit different now with 1.18.0, it has a more specific output:
```
/usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/12.1.0/../../../../include/c++/12.1.0/bits/stl_vector.h:1123: std::vector::reference std::vector<llvm::Constant *>::operator[](std::vector::size_type) [_Tp = llvm::Constant *, _Alloc = std::allocator<llvm::Constant *>]: Assertion '__n < this->size()' failed.
[...]

PR #2353, opened 2022-08-04
Replace usage of std::vector with array in LLVMShuffleVectors

aneshlya:
Fix for #2283 .
Usage of `&shufVec[shufSize]` when `shufVec` is `std::vector` seems invalid, so I changed it with to a standard array. It is the same approach as we use in other similar places of ISPC code.


dbabokin (approved):
Looks good!

PR #2357, opened 2022-08-22
[opaque pointers 1] Use int8 pointer instead of typed pointers in stdlib/builtins

aneshlya:
The change is required to support opaque pointers in ISPC.
I use cast to `uniform int8* uniform` in stdlib.ispc and not to `void * uniform` for example because when we match stdlib to builtins functions we restore ISPC type from LLVM types (used in builtins) https://github.com/ispc/ispc/blob/main/src/builtins.cpp#L127.
To keep the logic working we can only use `uniform int8* uniform`.

### DIFF
--- a/stdlib.ispc
+++ b/stdlib.ispc
@@ -57,6 +57,7 @@
 #error Unknown value of ISPC_MASK_BITS
 #endif
 
+typedef uniform int8 *uniform opaque_ptr_t;
 ///////////////////////////////////////////////////////////////////////////
 // Xe specific primitives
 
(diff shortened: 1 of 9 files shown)


stdlib.ispc
```diff
@@ -711,25 +711,31 @@ __declspec(safe) static inline int64 count_trailing_zeros(int64 v) {
 // AOS/SOA conversion
 
 static inline void aos_to_soa2(uniform float a[], varying float *uniform v0, varying float *uniform v1) {
-    __aos_to_soa2_float(a, v0, v1);
```
dbabokin, line 714: What about defining `opaque_ptr_t` and using it though the file? Otherwise it's too ugly.

dbabokin:
LGTM!

stdlib.ispc
aneshlya: Good idea, thanks!

dbabokin: The commit with `opaque_ptr_t` looks good.

PR #2358, opened 2022-08-23
Fixes to enable ISPC build with LLVM15

aneshlya:
- `alloy.py` changes to build LLVM15 rc2
- workflow to build LLVM 15
- changes to support LLVM 16 / trunk
- unused GitLab files are removed
- CMake `ISPC_OPAQUE_PTR_MODE` flag to govern if opaque pointers are enabled (off by default)
- disabling `llvm::createArgumentPromotionPass()` for LLVM15+, as it's not supported by old pass manager. To be reenabled when we migrate to the new pass manager.
- Tests to build with `--pic`.

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -490,6 +490,11 @@ def run_test(testname, host, target):
                 cc_cmd = "%s -O2 -I. %s test_static.cpp -DTEST_SIG=%d -DTEST_WIDTH=%d %s -o %s" % \
                     (options.compiler_exe, gcc_arch, match, width, obj_name, exe_name)
 
+                # Produce position independent code for both c++ and ispc compilations.
+                # The motivation for this is that Clang 15 changed default
+                # from "-mrelocation-model static" to "-mrelocation-model pic", so
+                # we enable PIC compilation to have it consistently regardless compiler version.
+                cc_cmd += ' -fPIE'
                 if should_fail:
                     cc_cmd += " -DEXPECT_FAILURE"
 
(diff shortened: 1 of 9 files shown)


src/opt.cpp
```diff
@@ -84,6 +84,9 @@
 #include <llvm/Transforms/Scalar.h>
 #include <llvm/Transforms/Scalar/GVN.h>
 #include <llvm/Transforms/Scalar/InstSimplifyPass.h>
+#if ISPC_LLVM_VERSION > ISPC_LLVM_14_0
```
dbabokin, line 87: Let's have >= 15.0. This way we are ready for theoretically possible 14.1.

run_tests.py
```diff
@@ -501,10 +501,11 @@ def run_test(testname, host, target):
                              match, width, exe_name)
                     exe_name = "./" + exe_name
                     cc_cmd += " -DTEST_ZEBIN" if options.ispc_output == "ze" else " -DTEST_SPV"
-
             ispc_cmd = ispc_exe_rel + " --woff %s -o %s --arch=%s --target=%s -DTEST_SIG=%d" % \
                         (filename, obj_name, options.arch, xe_target if target.is_xe() else options.target, match)
-
+            # Produce position independent code. Required when host compiler is clang 15.0+.
+            if host.is_linux():
+                ispc_cmd += " --pic"
```
dbabokin, line 508: Clang 15 changed default from `-mrelocation-model static` to `-mrelocation-model pic`. I think if we are going to use PIC consistently, we should add it to both c++ and ispc compilations, so it's PIC across the board (regardless compiler or compiler versions).

dbabokin:
I've added `alloy.py` changes to build LLVM15 rc2, workflow to build LLVM 15, changes to support LLVM 16 / trunk and removed unused GitLab files. Please review and if the changes looks good - please merge.

PR #2360, opened 2022-08-26
[opaque pointers 2] Removed usage of getElementPointerType from opt.cpp

aneshlya:
This change prepares ISPC code base to support opaque pointers mode.
1. removes usage of getElementPointerType from opt.cpp (also hidden in PTYPE macro)
NOTE: there is a bunch of functions related to the processing of our pseudo gathers/pseudo scatters intrinsics where the base pointer element type is always int8. There is no way to get it from the function signature in opaque mode, so I'm keeping it in the constructors of helper structures.
2. aligns common pointer calculation for CPU and Xe targets (before we calculated the base before the common pointer due to limitations in VC BE, now it's fine)


dbabokin:
LGTM!

PR #2362, opened 2022-09-01
ISPC 

5678890:
ISPC


dbabokin:
I think it's time to remove `master` branch. We kept it around for a while, as we have link in the documentation and didn't want to break the links from the released documents.

PR #2363, opened 2022-09-07
[opaque pointers 3] Introduce AddressInfo class to ISPC

aneshlya:
`AddressInfo` class is necessary to support opaque pointers. It contains the pointer itself, LLVM pointer element type and ISPC type (may be NULL).
This PR is the first set of changes switching ISPC to use `AddressInfo`:
1. `AllocaInst` starts returning `AddressInfo*` instead of `llvm::Value*`
2. `Symbol` class keeps `AddressInfo*` instead of simple `llvm::Value*`

Note: Many `getPointer()` calls will be removed in future PRs.

### DIFF
--- a/src/ctx.cpp
+++ b/src/ctx.cpp
@@ -2039,6 +2039,17 @@ llvm::Value *FunctionEmitContext::MakeSlicePointer(llvm::Value *ptr, llvm::Value
     return ret;
 }
 
+const PointerType *FunctionEmitContext::RegularizePointer(const Type *ptrRefType) {
+    const PointerType *ptrType;
+    if (CastType<ReferenceType>(ptrRefType) != NULL)
+        ptrType = PointerType::GetUniform(ptrRefType->GetReferenceTarget());
+    else {
+        ptrType = CastType<PointerType>(ptrRefType);
+    }
+    AssertPos(currentPos, ptrType != NULL);
+    return ptrType;
+}
+
 llvm::Value *FunctionEmitContext::GetElementPtrInst(llvm::Value *basePtr, llvm::Value *index, const Type *ptrRefType,
                                                     const llvm::Twine &name) {
     if (basePtr == NULL || index == NULL) {
@@ -2047,13 +2058,7 @@ llvm::Value *FunctionEmitContext::GetElementPtrInst(llvm::Value *basePtr, llvm::
     }
 
     // Regularize to a standard pointer type for basePtr's type
-    const PointerType *ptrType;
-    if (CastType<ReferenceType>(ptrRefType) != NULL)
-        ptrType = PointerType::GetUniform(ptrRefType->GetReferenceTarget());
-    else {
-        ptrType = CastType<PointerType>(ptrRefType);
-    }
-    AssertPos(currentPos, ptrType != NULL);
+    const PointerType *ptrType = RegularizePointer(ptrRefType);
 
     if (ptrType->IsSlice()) {
         AssertPos(currentPos, llvm::isa<llvm::StructType>(basePtr->getType()));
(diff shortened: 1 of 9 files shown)


src/ctx.h
```diff
@@ -51,6 +51,32 @@ namespace ispc {
 
 struct CFInfo;
 
+///////////////////////////////////////////////////////////////////////////
+/** AddressInfo is a helper class to work with pointers.
+    It keeps llvm pointer, llvm element type, and ISPC type.
+*/
+class AddressInfo {
+  public:
+    AddressInfo(){};
```
dbabokin, line 60: Do we really want to do have default constructor without zero-initializing the members?

src/ctx.cpp
```diff
@@ -198,26 +200,48 @@ CFInfo *CFInfo::GetForeach(bool isUniformEmulated, FunctionEmitContext::ForeachT
...
-                      savedSwitchFallThroughMaskPtr, savedDefaultBlock, savedCases, savedNext,
+    return new CFInfo(Switch, isUniform, isUniformEmulated, breakTarget, continueTarget, savedBreakLanesAddressInfo,
+                      savedContinueLanesAddressInfo, savedMask, savedBlockEntryMask, savedSwitchExpr,
+                      savedSwitchFallThroughMaskAddressInfo, savedDefaultBlock, savedCases, savedNext,
                       savedSwitchConditionUniform);
 }
 
 ///////////////////////////////////////////////////////////////////////////
 
+AddressInfo::AddressInfo(llvm::Value *p, llvm::Type *t) : pointer(p), elementType(t) {
```
dbabokin, line 223: Shouldn't we zero-initialize `ispcType`?

src/ctx.cpp
```diff
@@ -198,26 +200,48 @@ CFInfo *CFInfo::GetForeach(bool isUniformEmulated, FunctionEmitContext::ForeachT
...
+    Assert(pointer != nullptr && "Pointer cannot be null");
+    Assert(elementType != nullptr && "Element type cannot be null");
+}
+AddressInfo::AddressInfo(llvm::Value *p, const Type *t) : pointer(p), ispcType(t) {
+    Assert(pointer != nullptr && "Pointer cannot be null");
+    Assert(ispcType != nullptr && "ISPC type cannot be null");
+    // Get LLVM pointer element type based on ISPC type.
+    // TODO: need more testing
+    if (CastType<ReferenceType>(t) != NULL) {
```
dbabokin, line 232: `nullptr` vs `NULL` is bugging me. I understand why you used both of the in the same function, and I'm fine with it. But probably we should fix all our sources to use `nullptr` uniformly.

src/ctx.cpp
```diff
@@ -2017,6 +2046,17 @@ llvm::Value *FunctionEmitContext::MakeSlicePointer(llvm::Value *ptr, llvm::Value
     return ret;
 }
 
+const PointerType *FunctionEmitContext::RegularizePointer(const Type *ptrRefType) {
```
dbabokin, line 2049: A comment with a function description would not hurt here.

src/ctx.cpp
aneshlya: I changed it here to use `nullptr` only. I think going forward we should use `nullptr` in the newly added source code. And yes, switch our current source base to `nullptr`.

src/ctx.cpp
aneshlya: Agree.

src/ctx.h
aneshlya: We don't need it, agree.

dbabokin (approved):
LGTM

PR #2365, opened 2022-09-09
[opaque pointers 4] Use AddressInfo in StoreInst

aneshlya:
Depends on https://github.com/ispc/ispc/pull/2363.
This PR introduces more usages of `AddressInfo` class in particular in `StoreInst` and `InitSymbol`.
It also introduces a method to get LLVM types of function arguments `LLVMFunctionArgTypes` and simplifies `StoreInst` - we don't need to pass info about uniform data anymore.

During the testing, I also discovered the problem with bool swizzles: https://github.com/ispc/ispc/issues/2367.


dbabokin (approved):
Looks good!

+100 to your karma for the lit test.

PR #2366, opened 2022-09-12
Added ISPC_FP16_SUPPORTED macro

aneshlya:
Introduce `ISPC_FP16_SUPPORTED` macro indicating where float16 type is supported by the target.


src/ispc.cpp
```diff
@@ -793,10 +793,17 @@ Target::Target(Arch arch, const char *cpu, ISPCTarget ispc_target, bool pic, boo
               target_string.c_str());
         return;
     }
+#ifdef ISPC_XE_ENABLED
```
dbabokin, line 796: Let's have {} around the if statement body and you don't need ifdef, as `ISPCTargetIsGen()` function is generally available.

tests/lit-tests/fp16_support.ispc
```diff
@@ -0,0 +1,26 @@
+// This test checks that ISPC_FP16_SUPPORTED macro is ON for XE and ARM platforms only.
+
+// RUN: %{ispc} %s --target=gen9-x16 --arch=xe32 --emit-llvm-text -o - | FileCheck %s --check-prefixes=CHECK_FP_16,CHECK
+// RUN: %{ispc} %s --target=gen9-x16 --arch=xe64 --emit-llvm-text -o - | FileCheck %s --check-prefixes=CHECK_FP_16,CHECK
+// RUN: %{ispc} %s --target=neon-i8x16 --arch=arm --emit-llvm-text -o - | FileCheck %s --check-prefixes=CHECK_FP_16,CHECK
+// RUN: %{ispc} %s --target=neon-i32x4 --arch=aarch64 --emit-llvm-text -o - | FileCheck %s --check-prefixes=CHECK_FP_16,CHECK
+// RUN: %{ispc} %s --target=host --arch=x86 --emit-llvm-text -o - | FileCheck %s --check-prefixes=CHECK_FP_NO_16,CHECK
```
dbabokin, line 7: What about running the tests of ARM host? ;) I suggest using `avx2-i32x8` instead.

docs/ispc.rst
```diff
@@ -902,6 +902,9 @@ preprocessor runs:
   * - ISPC_UINT_IS_DEFINED
     - 1
     - The macro is defined if uint8/uint16/uint32/uint64 types are defined in the ``ispc`` (it's defined in 1.13.0 and later)
+  * - ISPC_FP16_SUPPORTED
+    - 1
+    - The macro is defined if float16 type is supported by the target
```
dbabokin, line 907: Let's be explicit about the semantics - this macros indicates if the type is fully supported on the target, whether through native hardware support or software emulation.

dbabokin (approved):

PR #2368, opened 2022-09-15
[opaque pointers 5] Use AddressInfo in LoadInst

aneshlya:
This PR switches `StoreInst` to use `AddressInfo` class.
Also eliminates usage of `getPointerElementType` in easy cases.

### DIFF
--- a/src/module.cpp
+++ b/src/module.cpp
@@ -2647,10 +2647,7 @@ static void lCreateDispatchFunction(llvm::Module *module, llvm::Function *setISA
 
     // Now we can load the system's ISA enumerant
 #if ISPC_LLVM_VERSION >= ISPC_LLVM_11_0
-    llvm::PointerType *ptr_type = llvm::dyn_cast<llvm::PointerType>(systemBestISAPtr->getType());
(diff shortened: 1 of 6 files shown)


src/module.cpp
```diff
@@ -2782,10 +2779,12 @@ static bool lCompatibleTypes(llvm::Type *Ty1, llvm::Type *Ty2) {
             break;
 
         case llvm::ArrayType::PointerTyID:
+// In opaque mode pointers are compatible
```
dbabokin, line 2782: Is it intended behavior to have fall-though in Opaque Mode?

dbabokin: I think it was assumes to `return true;` in else case of ifndef.
It also seems to affect how ISPC reports warnings about pointers to varying types in global variables.

aneshlya: The intention was to have `ifndef ` before `case llvm::ArrayType::PointerTyID`, not after. But anyway I reworked it in the right way.

dbabokin (approved):
Looks good now!

Issue #2239, opened 2022-01-21
Support foreach and varying for loops with pragma unroll


JeffRous:
#548 added support for #pragma unroll but was limited to uniform loop counters only. Would like support added for foreach and varying for loops.

Ex:
```
#pragma unroll 4
foreach(i = 0 ... Iterations)
{ ... }
```
and

```
#pragma unroll 4
for(varying int i = programIndex; i < Iterations; i+=programCount)
{ ... }
```

pbrubaker:
Closing as resolved.

PR #2373, opened 2022-09-20
Added Varying Index Support for Loop Unrolling

e3m3:
Closes Feature Request #2239 .
Changes:

*   Added varying index support for unrolling `for` and `foreach` loops.
    | `src/stmt.cpp`

*   Added benchmark for varying index loop unrolling.
    | `benchmarks/01_trivial/07_loop_unroll_varying-cpu.cpp`,
    | `benchmarks/01_trivial/07_loop_unroll_varying-cpu.ispc`,
    | `benchmarks/01_trivial/CMakeLists.txt`,
    | `benchmarks/01_trivial/README.md`

*   Added lit test for varying index loop unrolling.
    Currently expected to fail due to scalarization of loop body.
    | `tests/lit-tests/loop_unroll_varying.ispc`


src/stmt.cpp
```diff
@@ -1983,7 +1984,12 @@ Stmt *ForeachStmt::TypeCheck() {
 }
 
 void ForeachStmt::SetLoopAttribute(std::pair<Globals::pragmaUnrollType, int> lAttr) {
-    Warning(pos, "'#pragma unroll/nounroll' ignored - not supported for foreach loop.");
+    if (loopAttribute.first != Globals::pragmaUnrollType::none)
```
dbabokin, line 1987: Let's have {} around single-line then statement, this is less error prone.

tests/lit-tests/loop_unroll_varying.ispc
```diff
@@ -0,0 +1,158 @@
...
+
+
+// CHECK_ALL-LABEL: define void @foo_for___
+
+// CHECKO0: call void @goo_for___
+// CHECKO0: br label %{{[a-zA-Z_][a-zA-Z0-9_]*}}, !llvm.loop
+
+// CHECKO2: call void @goo_for___
+// CHECKO2: call void @goo_for___
+// CHECKO2: call void @goo_for___
```
dbabokin, line 20: If we'd like to check that we have exactly 3 calls, we should check that it's not match after that. Also, -COUNT-N suffix is more compact form of writing N same statements, i.e.:
```
// CHECKO2-COUNT-3: call void @goo_for___
// CHECKO2-NOT: call void @goo_for___
```

benchmarks/01_trivial/README.md
```diff
@@ -6,3 +6,4 @@
 - ``04_fastdiv`` - integer division by a constant is handled by an algorithm, which produces a code sequence without actual division operation. Current implementation relies on code generator to do the right thing on every specific platform. This benchmark tests performance integer division by a constant.
 - ``05_packed_load_store`` - test ``packed_[load|store]_active()`` stdlib functions performance.
 - ``06_math`` - test math functions performance.
+- ``07_loop_unroll_varying-cpu`` - test unrolling performance for varying index loop on CPU.
```
dbabokin, line 9: Any specific reason to mention that it's for CPU? Benchmarks are for CPU only at the moment. If they are ported to GPU, this should work for GPU as well, right? I.e. nothing specifically bound to CPU in these test. If you agree, let's remove "CPU" from here and from test names.

benchmarks/01_trivial/07_loop_unroll_varying-cpu.ispc
```diff
@@ -0,0 +1,412 @@
...
+    }
+
+
+
+// uint64
+
+UNROLL_VARYING_SIG_CPU(foreach, uint64, 1) {
+    dst[0] = src[0];
+    #pragma nounroll
+    UNROLL_VARYING_FOREACH_BODY
```
dbabokin, line 22: NIT: having `;` after a macro use is good style. It happened that current definition of a macro is a loop body, which doesn't need a `;` after it. But having `;` safeguards for any macro definition and look more native in C/C++ program.

benchmarks/01_trivial/07_loop_unroll_varying-cpu.ispc
```diff
@@ -0,0 +1,412 @@
...
+#define UNROLL_VARYING_SIG_CPU(T_FOR, T_ISPC, UNROLL_FACTOR)              \
+    export void unroll_varying_cpu_##T_FOR##_##T_ISPC##_##UNROLL_FACTOR(  \
+        uniform T_ISPC src[],                                             \
+        uniform T_ISPC dst[],                                             \
+        uniform uint64 num_elems                                          \
+    )
+
+#define UNROLL_VARYING_FOREACH_BODY   \
+    foreach (i = 1 ... num_elems) {   \
+        dst[i] = src[i] + src[i - 1]; \
```
dbabokin: As a note - alternatively we can have computations with loop carried dependency to observe good gains thanks to unrolling. But current implementation is good enough.

benchmarks/01_trivial/07_loop_unroll_varying.cpp
```diff
@@ -0,0 +1,143 @@
...
+            break;
+        }
+    }
+
+    aligned_free_helper(buf);
+}
+
+#define ARGS Arg(8192ull)->Arg(8192ull << 10ull)->Arg(8192ull << 15ull)
+
+#ifndef UNROLL_VARYING_BENCH_CPU
```
dbabokin, line 58: "CPU" here and in the rest of the file.

tests/lit-tests/loop_unroll_varying.ispc
```diff
@@ -0,0 +1,213 @@
+// Test to check '#pragma unroll' functionality for different loop statements
+//   with varying type induction variables.
+
+// RUN: %{ispc} %s --target=host --nowrap -O0 --emit-llvm-text --no-discard-value-names --nostdlib -o - | FileCheck %s -check-prefixes=CHECK_ALL,CHECKO0
+// RUN: %{ispc} %s --target=host --nowrap -O2 --emit-llvm-text --no-discard-value-names --nostdlib -o - | FileCheck %s -check-prefixes=CHECK_ALL,CHECKO2
+
+// XFAIL: LLVM_15_0+
```
dbabokin, line 7: Why is it failing with LLVM 15?

tests/lit-tests/loop_unroll_varying.ispc
```diff
@@ -0,0 +1,213 @@
...
+
+
+#define NUM 1024
+
+
+// CHECK_ALL-LABEL: define void @foo_for___
+
+// CHECKO0:         for_loop:
```
dbabokin, line 15: Looks like this will not work in Release build - I think it will not have named labels.

tests/lit-tests/loop_unroll_varying.ispc
```diff
@@ -0,0 +1,213 @@
...
+// CHECKO0:         call void @goo_for___
+// CHECKO0-NOT:     call void @goo_for___
+// CHECKO0:         br label %{{[a-zA-Z_][a-zA-Z0-9_]*}}
+// CHECKO0:         }
+
+// CHECKO2:         for_loop{{.*}}:
```
dbabokin, line 21: Is this an attempt to match basic block label?

Isn't it enough to just count the number of function calls in the function?

tests/lit-tests/loop_unroll_varying.ispc
e3m3: Not sure yet - investigating. I am assuming that the test is fragile due to some change in the code structure in the newer version.

tests/lit-tests/loop_unroll_varying.ispc
e3m3: Its not sufficient because the compiler generates different basic blocks depending on the mask or lane status. In those cases, counting may cause false passing tests because some instructions are repeated across basic blocks (not due to unrolling). In some edge cases, optimizations can cause entire sections of these duplicated basic blocks not to be present, making counting even trickier. There may be a better way to use the FileCheck syntax to accomplish this though.

tests/lit-tests/loop_unroll_varying.ispc
e3m3: That's unfortunate due to the reason [here](https://github.com/ispc/ispc/pull/2373#discussion_r998804382).

tests/lit-tests/loop_unroll_varying.ispc
dbabokin: What about defining the function `unmasked`? Would it help?

e3m3: For `for` varying loops, the unrolling is generating a basic block for each unrolled section (i.e. a branch separates each unrolled instruction from it's unrolled copies). This may have negative impacts on the performance benefits to be gained by the transformation.

For `foreach` loops, the unrolled instructions copies are generally (there are cases that break this, e.g. [odd number unroll factors](https://github.com/e3m3/ispc/blob/6a441e9b5fa3f2c344b4a1092e6d28639ce3b4ce/tests/lit-tests/loop_unroll_varying.ispc#L33)) in the same basic block.

e3m3: > What about defining the function `unmasked`? Would it help?

I'll try this.

e3m3: > What about defining the function `unmasked`? Would it help?

Neither of these versions helped.

```cpp
extern void goo_for(int);
void foo_for() {
    unmasked {
        #pragma unroll(3)
        for (int iter1 = 0; iter1 < NUM; iter1++) {
            goo_for(iter1);
        }
    }
}
```
```cpp
extern void goo_for(int);
void foo_for() {
    #pragma unroll(3)
    for (int iter1 = 0; iter1 < NUM; iter1++) {
        unmasked {
            goo_for(iter1);
        }
    }
}
```

dbabokin:
Is it known what CI failures are?

e3m3:
> Is it known what CI failures are?

CI/Tests are now passing. Ready for review.

tests/lit-tests/lit.cfg
```diff
@@ -42,6 +42,12 @@ if llvm_version >= LooseVersion("14.0.0"):
 else:
     print("LLVM_14_0+: NO")
 
+if llvm_version >= LooseVersion("15.0.0"):
```
dbabokin, line 45: This seems redundant now.

dbabokin:
Ok, we are done with NITs and minor stuff. But I have more general concern. The resulting code contains branches between unrolled iterations - it defeats the purpose of unrolling. The key benefit of unrolling is to improve static scheduling of the code and possibly enable other optimizations triggered by larger optimization window. The if conditions between unrolled iterations are the barriers for optimization.

For the scalar case unroller implementation does the following:
- in case of statically known number of iterations (and statically known that there is not need for the reminder loop), the code is unrolled without if checks.
- if the iteration bounds are not statically known or unroller fail to analyze them, the unroller introduces the if checks between iterations (possibly with intention for other optimizations to remove some or all of them later).

This is not what I would expect. I would expect the loop to be unrolled by the requested unroll factor with reminder loop on the exit. Could you please check if unroll phase has any logic to produce the code like that?

e3m3:
> Ok, we are done with NITs and minor stuff. But I have more general concern. The resulting code contains branches between unrolled iterations - it defeats the purpose of unrolling. The key benefit of unrolling is to improve static scheduling of the code and possibly enable other optimizations triggered by larger optimization window. The if conditions between unrolled iterations are the barriers for optimization.
> 
> For the scalar case unroller implementation does the following:
> 
> * in case of statically known number of iterations (and statically known that there is not need for the reminder loop), the code is unrolled without if checks.
> * if the iteration bounds are not statically known or unroller fail to analyze them, the unroller introduces the if checks between iterations (possibly with intention for other optimizations to remove some or all of them later).
> 
> This is not what I would expect. I would expect the loop to be unrolled by the requested unroll factor with reminder loop on the exit. Could you please check if unroll phase has any logic to produce the code like that?

This is true only for `for` loops with a varying induction variable. Unrolling `foreach` loops behaves as expected. It is not clear how to effectively ensure that the predicate for a `for` loop evaluates to true for all data elements without checking each lane independently. That is, unrolling a `for` loop with a `varying` type predicate *necessitates* code generation for cases where the predicate is not all-on, resulting in branching code that limits optimization. I think the intended use case for this form of unrolling is the `foreach` loop anyway - we would be able to extend the pattern matching for `for` loops such that they are semantically equivalent to a `foreach` loop, and then generate the code we expect. I would argue that at this point, we should only allow unrolling for the `foreach` construct and disable it for `for` loops as the intended use is clearer this way.

dbabokin:
> I would argue that at this point, we should only allow unrolling for the `foreach` construct and disable it for `for` loops as the intended use is clearer this way.

I would say we should emit a warning that unroll for `for` loop with varying index is not as efficient as for `foreach`. And we need to update docs.

Documentation for unroll happened to be in [Preprocessor](https://github.com/ispc/ispc/blob/main/docs/ispc.rst#the-preprocessor) section, which is obviously incorrect. it need to be in [Compiler Optimization Hints](https://github.com/ispc/ispc/blob/main/docs/ispc.rst#compiler-optimization-hints) section.

e3m3:
> > I would argue that at this point, we should only allow unrolling for the `foreach` construct and disable it for `for` loops as the intended use is clearer this way.
> 
> I would say we should emit a warning that unroll for `for` loop with varying index is not as efficient as for `foreach`. And we need to update docs.
> 
> Documentation for unroll happened to be in [Preprocessor](https://github.com/ispc/ispc/blob/main/docs/ispc.rst#the-preprocessor) section, which is obviously incorrect. it need to be in [Compiler Optimization Hints](https://github.com/ispc/ispc/blob/main/docs/ispc.rst#compiler-optimization-hints) section.

Addressed in commit [a5a1613](https://github.com/ispc/ispc/pull/2373/commits/a5a1613c3b79ae0015aad0190b5aea6bef176e75).

PR #2374, opened 2022-09-21
[opaque pointers 6] Final fixes

aneshlya:
The last piece of changes related to opaque pointers.
1. Eliminated remaining usages of getPointerElementType
2. Fixed signature of `llvm.memcpy` intrinsic in opaque mode
3. Fixed lit tests

### DIFF
--- a/src/ctx.h
+++ b/src/ctx.h
@@ -67,6 +67,9 @@ class AddressInfo {
     // Return the type of the values stored in this address.
     llvm::Type *getElementType() const { return elementType; }
 
+    // Return the ISPC type. May be NULL.
+    const Type *getISPCType() const { return ispcType; }
+
     // Return the address space that this address resides in.
     unsigned getAddressSpace() const { return getType()->getAddressSpace(); }
 
@@ -449,7 +452,8 @@ class FunctionEmitContext {
         and an integer offset to a slice within that type. */
     llvm::Value *MakeSlicePointer(llvm::Value *ptr, llvm::Value *offset);
 
-    /* Regularize to a standard pointer type. */
+    /* Regularize to a standard pointer type.
+       May return NULL if type is not PointerType or ReferenceType */
     const PointerType *RegularizePointer(const Type *ptrRefType);
 
     /** These GEP methods are generalizations of the standard ones in LLVM;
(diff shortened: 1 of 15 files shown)


src/expr.cpp
```diff
@@ -5073,22 +5075,19 @@ llvm::Value *VectorMemberExpr::GetValue(FunctionEmitContext *ctx) const {
...
         llvm::Value *elementMask = ctx->GetFullMask();
 
-        const Type *elementPtrType = NULL;
-        if (CastType<ReferenceType>(basePtrType) != NULL)
-            elementPtrType = PointerType::GetUniform(basePtrType->GetReferenceTarget());
-        else
-            elementPtrType = basePtrType->IsUniformType() ? PointerType::GetUniform(exprVectorType->GetElementType())
-                                                          : PointerType::GetVarying(exprVectorType->GetElementType());
-
+        const PointerType *ptrType = ctx->RegularizePointer(basePtrType);
```
aneshlya, line 5078: `elementPtrType` was incorrect before for `ReferenceType`.
It affects code generation in LoadInst. Now for reference types we correctly set alignment = 1 here https://github.com/ispc/ispc/blob/main/src/ctx.cpp#L2440. Before it used default LLVM alignment  = 32.

src/expr.cpp
```diff
@@ -759,10 +759,13 @@ void ispc::InitSymbol(AddressInfo *ptrInfo, const Type *symType, Expr *initExpr,
...
                 llvm::Value *ep;
                 if (CastType<StructType>(symType) != NULL)
-                    ep = ctx->AddElementOffset(ptrInfo->getPointer(), i, NULL, "element");
-                else
+                    ep = ctx->AddElementOffset(new AddressInfo(ptrInfo->getPointer(), CastType<StructType>(symType)), i,
+                                               "element");
+                else {
                     ep = ctx->GetElementPtrInst(ptrInfo->getPointer(), LLVMInt32(0), LLVMInt32(i),
-                                                PointerType::GetUniform(elementType), "gep");
+                                                /* Type of aggregate structure */ PointerType::GetUniform(symType),
```
aneshlya, line 766: There was a bug before. We need here type of pointer, not its element.

src/ctx.h
```diff
@@ -70,6 +70,8 @@ class AddressInfo {
     // Return the address space that this address resides in.
     unsigned getAddressSpace() const { return getType()->getAddressSpace(); }
 
+    static llvm::Type *GetAddressElementType(const PointerType* pt);
```
dbabokin: Function name is not descriptive and there are no comments. What about `GetPointeeLLVMType`?

src/expr.cpp
dbabokin: Is the a way to trigger it in the test?

src/expr.cpp
dbabokin: Would be good to check this in the test

aneshlya: Added load_unaligned.ispc lit test.

src/expr.cpp
aneshlya: I don't think so. We construct uniform pointer `PointerType::GetUniform(symType)` in `ctx->GetElementPtrInst` call. From here we always go to https://github.com/ispc/ispc/blob/main/src/ctx.cpp#L2145 and check that ptrType->IsUniformType(). It was uniform before and it is uniform now. `ptrType` is not used for anything else.

tests/lit-tests/2376.ispc
```diff
@@ -0,0 +1,13 @@
+// This test checks that compiler is not crashing and producing
+// error when varying index is used with short vectors
+// Currently fails (https://github.com/ispc/ispc/issues/2376).
+
+// RUN: %{ispc} %s --target=host --nostdlib
```
dbabokin, line 5: The test will need to be modified when the bug is fixed. I suggest assuming that it will fail (i.e. have "not" prefix: `RUN: not %{ispc} ...`) and checking that crash message is not dispalyed.

dbabokin (approved):
LGTM!

PR #2377, opened 2022-09-23
Enable ISPC build with opaque pointers in GA

aneshlya:
- Removed `-Wno-deprecated-declarations` option from ISPC build
- Enabled `ISPC_OPAQUE_PTR_MODE` for ISPC build with LLVM 15.0 in GA
- Fixed load_unaligned and bool_swizzle tests for opaque pointers


dbabokin (approved):
LGTM!

PR #2381, opened 2022-10-03
Enable Windows arm64 target

edorobek:
Changes:

- Added Microsoft SQ1 and SQ2 to list of supported CPUs
- Removed restriction on Windows + arm64 in GenerateBuiltins.cmake


src/ispc.cpp
```diff
@@ -1346,6 +1356,8 @@ Target::Target(Arch arch, const char *cpu, ISPCTarget ispc_target, bool pic, boo
 #else
                 CPUID = CPU_AppleA13;
 #endif
+            } else if (g->target_os == TargetOS::windows) {
+                CPUID = CPU_MicrosoftSQ1
```
dbabokin, line 1360:
```suggestion
                CPUID = CPU_MicrosoftSQ1;
```

src/ispc.cpp
```diff
@@ -480,6 +486,8 @@ class AllCPUs {
 #if ISPC_LLVM_VERSION >= ISPC_LLVM_12_0
         names[CPU_AppleA14].push_back("apple-a14");
 #endif
+        names[CPU_MicrosoftSQ1].push_back("microsoft-sq1");
```
dbabokin, line 489: Neither `microsoft-sq1`, nor `microsoft-sq2` are available as CPU definitions in open source LLVM. They should likely to be defined somewhere in [llvm/lib/Target/AArch64/AArch64.td](https://github.com/llvm/llvm-project/blob/main/llvm/lib/Target/AArch64/AArch64.td).

Would it be sufficient to use just default CPU definition for aarch64?

cmake/GenerateBuiltins.cmake
```diff
@@ -158,8 +158,16 @@ function(builtin_to_cpp bit os_name arch supported_archs supported_oses resultFi
     endif()
 
     # OS to arch constrains
-    if (${os_name} STREQUAL "windows" AND ${arch} STREQUAL "arm")
-        set(SKIP ON)
+    if (${os_name} STREQUAL "windows")
```
dbabokin, line 161: No need to set `target_arch` to `arm64`. `arm64` is Apple specific name for `aarch64`.

I suggest just adding to the original condition `AND "${bit}" STREQUAL "32"` This should be sufficient - this will skip ARM 32bit only.

src/ispc.cpp
```diff
@@ -1552,8 +1564,7 @@ std::string Target::GetTripleString() const {
             Error(SourcePos(), "Arm is not supported on Windows.");
             exit(1);
         } else if (m_arch == Arch::aarch64) {
-            Error(SourcePos(), "Aarch64 is not supported on Windows.");
-            exit(1);
+            triple.setArchName("arm64");
```
dbabokin, line 1567: `arm64` is Apple specific name of `aarch64`. So probably it should be `aarch64`, though clang seem to accept equally well `aarch64-none-windows-msvc19.20.0` and `arm64-none-windows-msvc19.20.0` as `--target=<val>` argument.

I would check what is the default target triple on Windows ARM to figure out what is the correct value is.

steplong: I think the triple we are using is aarch64-pc-windows-msvc

dbabokin: This makes sense and that's what I expected, thanks for conforming.

src/ispc.cpp
steplong: I get the following warning with this:
```
'microsoft-sq1' is not a recognized processor for this target (ignoring processor)
```
I think it should be fine to just remove this if block and have CPUID be CPU_CortexA35 like suggested above.

src/ispc.cpp
edorobek: Thanks for catching that. I committed changes that address this (and the cmake changes.) Let me know if any other issues need addressed.

dbabokin:
The changes look good!

I've tried generating aarch64 code on Windows and it looks ok, but I didn't have a chance to run it on Windows ARM hardware yet. If you find any issues on actual hardware, feel free to file and issue or submit a PR.

Thank you for the contribution!

PR #2384, opened 2022-10-19
Reset FTZ/DAZ flags on ISPC function entrance / restore on return

### LINKED ISSUE #2006: Reset FTZ/DAZ flags on ISPC function entrance / restore on return.
ISPC code may be affected by FTZ/DAZ flags not being set even if program doesn't deal with denormals explicitly, if they appear in the inactive lane as a garbage. The performance degradation may be on the scale of several times for the kernel. Possible mitigations are to (1) set these flags on C++ "host" side, or (2) set them on ISPC "kernel" entrance.

While it's possible to set these flags by calling `fastmath()` stdlib function now, this function just sets them, but not store or restores old the values.

The proposal is to have something like `--opt=reset-ftz-daz`, which would set these flags on the entrance and restores them upon return to "host" code.

aneshlya:
Closes #2006


builtins/target-avx-utils.ll
```diff
@@ -218,6 +218,28 @@ define void @__fastmath() nounwind alwaysinline {
   ret void
 }
 
+define i32 @__set_ftz_daz_flags() nounwind alwaysinline {
```
dbabokin, line 221: These functions seem to be candidates for `util.m4`

src/ctx.cpp
```diff
@@ -365,6 +365,18 @@ FunctionEmitContext::FunctionEmitContext(Function *func, Symbol *funSym, llvm::F
         SetInternalMaskAnd(LLVMMaskAllOn, allOnMask);
     }
 
+    // If reset of FTZ/DAZ flags is requested and we're inside external function,
+    // allocate memory for keeping the old FTZ/DAZ value
+    if ((g->opt.resetFTZ_DAZ) &&
+        (func->GetType()->isExported || func->GetType()->isExternC || func->GetType()->isExternSYCL ||
+         func->GetType()->IsISPCKernel()) &&
+        (lf->getFunctionType()->getNumParams() == func->GetType()->GetNumParameters())) {
```
dbabokin, line 373: I think this line needs an explanation comment

builtins/target-neon-common.ll
```diff
@@ -197,6 +197,20 @@ define void @__fastmath() nounwind alwaysinline {
   ret void
 }
 
+define i32 @__set_ftz_daz_flags() nounwind alwaysinline {
```
dbabokin, line 200: NEON 32 bit and NEON 64 bit seem to have different instructions / behavior. I suggest double checking that this works on both as expected.

dbabokin (commented):
We also need docs update.

dbabokin (approved):
Looks good!

PR #2387, opened 2022-10-21
Remove zlib dependency from LLVM, upgrade to 15.0.3

dbabokin:
ISPC doesn't use LLVM functionality, which requires zlib (or any other compression), so build LLVM without ZLIB support (`-DLLVM_ENABLE_ZLIB=OFF`) and drop this dependency for ISPC. Zlib was pretty problematic recently wrt security.

Also use `15.0.3` as a tag for LLVM 15 and use this version in CI.

LLVM 12-15 (the latest revision of each version) were updated in [ispc/llvm-project](https://github.com/ispc/llvm-project) releases to be built without `zlib`.


aneshlya:
Can we remove zlib from third-party-programs.txt and dockerfiles now?

dbabokin:
> Can we remove zlib from third-party-programs.txt and dockerfiles now?

I should have mentioned that, but forgot. Yes, these are two places where zlib remains. I wanted to do that in separate PR. But given that it seems that we have no problems with CI passing, we can add it to this PR.

mingodad:
For me to build with an standard `clang/llvm` installation I still need the `-lz` and also `-DISPC_INCLUDE_TESTS=NO` to not fail searching for `FileCheck` .

dbabokin:
@mingodad I suggest downloading LLVM distribution that we built specifically for using it for ISPC development from here: https://github.com/ispc/ispc.dependencies/releases - this way you don't need to spend time for LLVM build and it's configured the way it works best for ISPC.

The decision to drop `zlib` is motivated by two reasons - (1) zlib tends to have security flaws (CVEs) that give us hard time fixing them, while we don't really use `zlib` functionality (2) we don't want to have dependency that we don't really use.

PR #2389, opened 2022-11-04
Set of minor fixes

dbabokin:
Most of the changes are editorial, except these two:
- `Function` class constructor to run `TypeCheck()` and `Optimize()` after all the symbols are saved. This should not affect compiler behavior, but it's done, as it's more consistent.
- LLVM 15 to build in `alloy.py` is `15.0.4` instead of `15.0.3`. CI still uses `15.0.3`

Editorial changes:
- Reorder includes in `parse.yy` in alphabetical order.
- Add documentation for `Function` class constructor and outline typechecking & optimization calls.
- `Symbol` constructor to use initialization list.


aneshlya (approved):
LGTM

Issue #2385, opened 2022-10-19
Allow disabling of static ispcrt library


StefanBruens:
Currently, both static and dynamic libraries are always built.

As the static library is referenced in the exported targets (`ispcrt_Exports-<BuildType>.cmake`), e.g.:
```
list(APPEND _cmake_import_check_files_for_ispcrt::ispcrt_static "${_IMPORT_PREFIX}/lib64/libispcrt_static.a" )
```
it is not sufficient to just delete the library file, as the CMake config would become invalid.

aneshlya:
Fixed by https://github.com/ispc/ispc/pull/2391

PR #2391, opened 2022-11-07
ispcrt: Add ISPCRT_BUILD_STATIC option.

nurmukhametov:
Default value is ON. When OFF static library building is omitted.

### DIFF
--- a/ispcrt/CMakeLists.txt
+++ b/ispcrt/CMakeLists.txt
@@ -52,6 +52,8 @@ if (ISPCRT_BUILD_GPU)
   endif()
 endif()
 
+option(ISPCRT_BUILD_STATIC "Build ispcrt static library" ON)
+
 if (NOT ISPCRT_BUILD_CPU AND NOT ISPCRT_BUILD_GPU)
   message(FATAL_ERROR "You must enable either CPU or GPU support!")
 endif()
@@ -185,6 +187,7 @@ macro(build_ispcrt SHARED_OR_STATIC TARGET_NAME)
     ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
   )
 endmacro()
+
 if (WIN32)
     set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
 endif()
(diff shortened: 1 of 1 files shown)


nurmukhametov:
#2385

ispcrt/CMakeLists.txt
```diff
@@ -189,10 +191,13 @@ if (WIN32)
     set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
 endif()
 build_ispcrt(SHARED ${PROJECT_NAME})
-if (WIN32)
-    set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS OFF)
+
+if (ISPCRT_BUILD_STATIC)
+  if (WIN32)
+      set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS OFF)
+  endif()
```
aneshlya, line 198: It seems that `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` relates to building of shared library: we set it `ON` on line 191, build the shared library, and should restore it back to `OFF` after `build_ispcrt(SHARED ${PROJECT_NAME})`.
Please move it from `if (ISPCRT_BUILD_STATIC)` condition.

aneshlya (changes requested):

ispcrt/CMakeLists.txt
nurmukhametov: OK, fixed.

ispcrt/CMakeLists.txt
```diff
@@ -52,6 +52,8 @@ if (ISPCRT_BUILD_GPU)
   endif()
 endif()
 
+option(ISPCRT_BUILD_STATIC "Enable static library building" ON)
```
aneshlya, line 55: Let's add `ispcrt` to the option description so it'll be aligned with others. 
`Enable static library building` -> `Build ispcrt static library`

nurmukhametov: OK, reworded

aneshlya (approved):

PR #2392, opened 2022-11-10
Fix of ISPC Runtime memory leak

aneshlya:
Related to https://github.com/ispc/ispc/commit/fa3c06ef4055686b3c70151f991cb38e390c2cf4.
The fix was missed for `Array` constructor using `Context`.

Plus submodules update having required fix for build on Ubuntu 22.04: for https://github.com/google/googletest/pull/3024


dbabokin (approved):
LGTM

PR #2393, opened 2022-11-16
WIP: Function templates support

dbabokin:
This PR introduces function template support with syntax and semantic borrowed from C++. Specifically, keywords `template` and `typename` were added. They can be used to declare or define the function template:
```cpp
template <typename T> T foo(T t) { return t + 1; }
```

Template functions may be called through explicit type parameters specification syntax:
```cpp
void bar() {
    int i = foo<int>(1);
    uniform int ui = foo<uniform int>(1);
    float f = foo<float>(1.f);
}
```
While template argument deduction rules generally follow C++, there are some differences caused by existence of ``uniform``, ``varying`` and ``unbound`` types in ``ispc`` type system.  The template type parameter may resolve only to ``uniform`` and ``varying`` types, but not to ``unbound`` type. Consider the following example:
```cpp
template <typename T> T add(T a, T b) { return a + b; }

void foo() {
    // Note that these two lines call the same function:
    int i1 = add<int>(1, 2);                 // T = varying int
    varying int i2 = add<varying int>(1, 2); // T = varying int

    // And this call a uniform version:
    uniform int i3 = add<uniform int>(1, 2); // T = uniform int
}
```
The variability of template type parameter ``T`` may be overwritten by ``uniform`` and ``varying`` keywords, so ``uniform T`` and ``varying T``  are always valid types. But when ``uniform T`` and ``varying T`` are used to specify template function parameters it has an effect on template argument deduction process.  If variability keyword is specified and the type was successfully deduced, the default variability of type ``T`` is assumed to be the opposite of the variability keyword.  The logic behind it is that the keyword was specified "on purpose" to changed the variability of the type ``T``. Consider the following example:

```cpp
template <typename T> void foo1(T t);
template <typename T> void foo2(uniform T t);
template <typename T> void foo3(varying T t);

void bar() {
    uniform int ui;
    varying int vi;
    foo1(ui); // T is uniform int
    foo1(vi); // T is varying int
    foo2(ui); // T is varying int!
    foo2(vi); // error: varying type cannot be passed to uniform parameter
    foo3(ui); // T is uniform int!
    foo3(vi); // T is uniform int!
}
```
Template parameters type deduction is not yet implemented, but planned in this PR, i.e.:
```cpp
void bar() {
    int i = foo((varying int)1);
    uniform int ui = foo(1);
    float f = foo(programIndex + 1.f);
}
```

Also, this PR introduces the syntax for a set of function template features that are planned, but not yet implemented:
- default values for template type parameters
- non-type template type parameters (integer parameters)
- function specialization
The test demonstrating what is not supported is [here](https://github.com/dbabokin/ispc/blob/templates_new/tests/lit-tests/func_template_not_supported.ispc).

Here are the tests that demonstrate what is supported and working: [test1](https://github.com/dbabokin/ispc/blob/templates_new/tests/lit-tests/func_template_1.ispc), [test2](https://github.com/dbabokin/ispc/blob/templates_new/tests/lit-tests/func_template_2.ispc).

TODO for this PR:
- [x] type deduction support in the call expression to support `foo(arg1, arg2);` call syntax (partially completed)
- [x] support all types of expressions in template code (call expressions depended types are not supported yet)
- [x] user documentation update
- [x] design document
- [x] test coverage
- [ ] consider several refactorings and review memory allocations.

TODO for either this or follow up PRs:

- [ ] support for non-type template parameters - integer parameters.

**Feedback for this feature is very welcome.**
Linux and Windows builds of the latest version can be found in Appveyor artifacts. Current (as of writing this comment) Linux build is [here](https://ci.appveyor.com/api/buildjobs/j858ese3ok5ffrqa/artifacts/build%2Fispc-trunk-linux.tar.gz). Windows build is [here](https://ci.appveyor.com/api/buildjobs/s5qdooa3ps58aitt/artifacts/build%2Fispc-trunk-windows.zip).

### DIFF
--- a/src/ast.h
+++ b/src/ast.h
@@ -213,6 +213,8 @@ class AST {
         information and source code. */
     void AddFunction(Symbol *sym, Stmt *code);
 
+    void AddFunctionTemplate(TemplateSymbol *templ, Stmt *code);
+
     /** Generate LLVM IR for all of the functions into the current
         module. */
     void GenerateIR();
@@ -221,6 +223,7 @@ class AST {
 
   private:
     std::vector<Function *> functions;
+    std::vector<FunctionTemplate *> functionTemplates;
 };
 
 /** Callback function type for preorder traversial visiting function for
--- a/src/ast.h
+++ b/src/ast.h
@@ -134,6 +134,8 @@ class ASTNode : public Traceable {
         enumerant values defined in ispc.h. */
     virtual int EstimateCost() const = 0;
 
+    virtual ASTNode *Instantiate(TemplateInstantiation &templInst) const = 0;
+
     /** All AST nodes must track the file position where they are
         defined. */
     SourcePos pos;
--- a/src/ast.h
+++ b/src/ast.h
@@ -158,6 +158,7 @@ class ASTNode : public Traceable {
         IndexExprID,
         StructMemberExprID,
         VectorMemberExprID,
+        DependentMemberExprID,
         NewExprID,
         NullPointerExprID,
         ReferenceExprID,
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -8439,7 +8439,16 @@ FunctionSymbolExpr::getCandidateTemplateFunctions(const std::vector<const Type *
                     paramType = paramType->GetAsNonConstType();
 
                     // If P is not a reference:
-                    // TODO: - if A is an array type, apply array-to-pointer conversion.
+                    // - if A is an array type, apply array-to-pointer conversion.
+                    const ArrayType *at = CastType<ArrayType>(argType);
+                    if (at) {
+                        const Type *targetType = at->GetElementType();
+                        if (targetType == NULL) {
+                            deductionFailed = true;
+                            break;
+                        }
+                        argType = PointerType::GetUniform(targetType, at->IsSOAType());
+                    }
                     // TODO: - if A is a function type, apply function-to-pointer conversion.
                     // - if A is a cv-qualified type, remove the top level cv-qualifiers from A.
                     argType = argType->GetAsNonConstType();
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -1097,6 +1097,10 @@ const Type *UnaryExpr::GetType() const {
     if (type == NULL)
         return NULL;
 
+    if (type->IsDependentType()) {
+        return AtomicType::Dependent;
+    }
+
     // Unary expressions should be returning target types after updating
     // reference address.
     if (CastType<ReferenceType>(type) != NULL) {
@@ -1242,6 +1246,10 @@ Expr *UnaryExpr::TypeCheck() {
     if (expr == NULL || (type = expr->GetType()) == NULL)
         // something went wrong in type checking...
         return NULL;
+
+    if (type->IsDependentType()) {
+        return this;
+    }
 
     if (type->IsSOAType()) {
         Error(pos, "Can't apply unary operator to SOA type \"%s\".", type->GetString().c_str());
@@ -2061,6 +2069,10 @@ const Type *BinaryExpr::GetType() const {
     if (type0 == NULL || type1 == NULL)
         return NULL;
 
+    if (type0->IsDependentType() || type1->IsDependentType()) {
+        return AtomicType::Dependent;
+    }
+
     // If this hits, it means that our TypeCheck() method hasn't been
     // called before GetType() was called; adding two pointers is illegal
     // and will fail type checking and (int + ptr) should be canonicalized
@@ -2511,6 +2523,10 @@ Expr *BinaryExpr::TypeCheck() {
     const Type *type0 = arg0->GetType(), *type1 = arg1->GetType();
     if (type0 == NULL || type1 == NULL)
         return NULL;
+
+    if (type0->IsDependentType() || type1->IsDependentType()) {
+        return this;
+    }
 
     // If either operand is a reference, dereference it before we move
     // forward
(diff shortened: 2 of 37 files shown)


src/func.h
```diff
@@ -90,4 +90,12 @@ class TemplateParms {
     std::vector<const TemplateTypeParmType *> parms;
 };
 
+class TemplateArgs {
+  public:
+    TemplateArgs(const std::vector<std::pair<const Type *, SourcePos>> &args);
+    bool IsEqual(TemplateArgs &otherArgs) const;
+
+    std::vector<std::pair<const Type *, SourcePos>> args;
```
aneshlya, line 100: Should it be a private field?

src/func.h
```diff
@@ -90,4 +90,12 @@ class TemplateParms {
     std::vector<const TemplateTypeParmType *> parms;
 };
 
+class TemplateArgs {
```
aneshlya, line 95: A comment on the class description would be great here.

src/type.h
```diff
@@ -151,6 +151,9 @@ class Type {
         unbound. */
     bool HasUnboundVariability() const { return GetVariability() == Variability::Unbound; }
 
+    /** Resolves dependent type with mapping of template types to concrete type passed as argeument */
```
aneshlya, line 157: argeument-argument

src/ast.h
```diff
@@ -134,6 +134,8 @@ class ASTNode {
         enumerant values defined in ispc.h. */
     virtual int EstimateCost() const = 0;
 
+    virtual ASTNode *Instantiate(TemplateInstantiation &templInst) const = 0;
```
aneshlya, line 137: Comment with method description is needed here.

mattgodbolt:
This branch is now built and installed every night on Compiler Explorer: https://godbolt.org/z/5959M8o8c (thanks to @pbrubaker for poking me)

src/type.h
```diff
@@ -73,15 +73,16 @@ struct Variability {
...
+    ATOMIC_TYPE,              // 0
+    ENUM_TYPE,                // 1
+    POINTER_TYPE,             // 2
+    ARRAY_TYPE,               // 3
+    VECTOR_TYPE,              // 4
+    STRUCT_TYPE,              // 5
+    UNDEFINED_STRUCT_TYPE,    // 6
+    REFERENCE_TYPE,           // 7
+    FUNCTION_TYPE,            // 8
+    TEMPALATE_TYPE_PARM_TYPE, // 9
```
aneshlya, line 85: Typo TEMPALATE_TYPE_PARM_TYPE->TEMPLATE_TYPE_PARM_TYPE

src/stmt.cpp
```diff
@@ -344,6 +356,18 @@ void DeclStmt::Print(Indent &indent) const {
...
+DeclStmt *DeclStmt::Instantiate(TemplateInstantiation &templInst) const {
+    std::vector<VariableDeclaration> instVars;
+    for (auto var : vars) {
+        Expr *instInit = var.init ? var.init->Instantiate(templInst) : nullptr;
+        Symbol *instSym = templInst.InstantiateSymbol(var.sym);
+        VariableDeclaration instDecl(instSym, instInit);
+        instVars.push_back(instDecl);
+    }
+    return new DeclStmt(instVars, pos);
+    ;
```
aneshlya, line 368: Extra semicolon

src/stmt.cpp
```diff
@@ -344,6 +356,18 @@ void DeclStmt::Print(Indent &indent) const {
 
 int DeclStmt::EstimateCost() const { return 0; }
 
+DeclStmt *DeclStmt::Instantiate(TemplateInstantiation &templInst) const {
+    std::vector<VariableDeclaration> instVars;
+    for (auto var : vars) {
```
aneshlya, line 361: Is it more efficient to use `auto&`?

src/type.cpp
```diff
@@ -627,6 +629,23 @@ const Type *TemplateTypeParmType::GetAsSOAType(int width) const {
...
 }
 
+const Type *TemplateTypeParmType::ResolveDependence(TemplateInstantiation &templInst) const {
+    const Type *resolvedType = templInst.InstantiateType(GetName());
+    if (resolvedType == NULL) {
+        Assert(m->errorCount > 0);
+        return NULL;
+    }
+    Variability v = variability;
+    // TODO: review this.
```
aneshlya: Still needs review?

src/func.cpp
```diff
@@ -878,6 +878,24 @@ Symbol *TemplateInstantiation::InstantiateSymbol(Symbol *sym) {
     return instSym;
 }
 
+Symbol *TemplateInstantiation::InstantiateTemplateSymbol(TemplateSymbol *sym) {
+    // The function is assumed to be called once per instnatiation and
```
aneshlya, line 1047: instnatiation-> instantiation

src/expr.cpp
```diff
@@ -3855,6 +3932,8 @@ Expr *FunctionCallExpr::TypeCheck() {
     if (func == NULL || args == NULL)
         return NULL;
 
+    // TODO: ???
```
aneshlya, line 3935: ???

src/type.h
dbabokin: Nice catch, fixed.

src/stmt.cpp
dbabokin: Fixed.

src/stmt.cpp
dbabokin: Fixed.

src/type.cpp
dbabokin: It's gone by the final commit in the PR.

src/func.cpp
dbabokin: Fixed.

src/expr.cpp
dbabokin: It's not a todo anymore, removed.

PR #2394, opened 2022-11-16
Editorial and test fixes

dbabokin:
- Fix FTZ/DAZ test on macOS (which doesn't support 32 bit version of x86 and arm).
- Removed unused LLVM14/15 macros in lit testing, as LLVM13, which is used by one of the tests.
- Move LLVM15 build (not use in CI) to `15.0.5`
- Clean up in FunctionSymbolExpr::ResolveOverloads() - this changes error message for builtins, but the logic is not changed. This logic supposed to be different for builtins, but it's not true for very long time.
- Several editorial changes, which were split from Templates PR.

### DIFF
--- a/.github/workflows/rebuild-llvm15.yml
+++ b/.github/workflows/rebuild-llvm15.yml
@@ -20,7 +20,7 @@ jobs:
     uses: ./.github/workflows/reusable.rebuild.yml
     with:
       version: '15.0'
-      full_version: '15.0.4'
+      full_version: '15.0.5'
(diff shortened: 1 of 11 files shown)


aneshlya (approved):

dbabokin:
Benchmarks fail looks like a runner problem, the changes doesn't have anything that might potentially affect benchmarks. I verified that benchmark binary is identical before and after the changes.
CC: @Lectem

I proceed with the merge.

PR #2396, opened 2022-11-18
Benchmark: Loop Unroll Fixes

e3m3:
Changes:
*   Reduced the run time for the loop unroll benchmark.
*   ~Formatting.~

### DIFF
--- a/benchmarks/01_trivial/07_loop_unroll_varying.cpp
+++ b/benchmarks/01_trivial/07_loop_unroll_varying.cpp
@@ -53,7 +53,7 @@ template <typename T> void check(T *src, T *dst, size_t num_elems) {
     aligned_free_helper(buf);
 }
 
-#define ARGS Arg(8192ull)->Arg(8192ull << 10ull)->Arg(8192ull << 15ull)
+#define ARGS Arg(8192ull)->Arg(8192ull << 10ull)
 
 #ifndef UNROLL_VARYING_BENCH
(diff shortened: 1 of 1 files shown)


dbabokin:
I suggest doing the same as [here](https://github.com/ispc/ispc/blob/ee0bb874ccad26426765b8e1ccfbfd47a083356f/benchmarks/01_trivial/01_aossoa.cpp#L16).

And we need an explanation in the comments - so the next person changing the values understands what's going on and why these values were picked. Cache sizes are changing and ideally we should use different values based on actual cache sizes.

dbabokin:
@Lectem looks like we found the problem with crashing benchmarks, so the problem is on our side - we use too large arrays, so it runs out of memory on your runner. Sorry for the false alarm.

e3m3:
> I suggest doing the same as [here](https://github.com/ispc/ispc/blob/ee0bb874ccad26426765b8e1ccfbfd47a083356f/benchmarks/01_trivial/01_aossoa.cpp#L16).
> 
> And we need an explanation in the comments - so the next person changing the values understands what's going on and why these values were picked. Cache sizes are changing and ideally we should use different values based on actual cache sizes.

See commit https://github.com/ispc/ispc/pull/2396/commits/ca974615f5b281d2ce006408b0c1d6d9cec9f126.

dbabokin (approved):
This looks good. Note that it's not expected to pass benchmarks tests right now, because of the way these checks work - they use benchmarks from the `main` branch to validate the changes in the PR.

PR #2399, opened 2022-12-01
Fix CI and move to LLVM 15.0.6

dbabokin:
- Build LLVM `15.0.6` with `alloy.py`
- Use LLVM `15.0.6` in CI (instead of `15.0.3`)
- Build LLVM with explicitly turned off ZSTD compression. This should not matter on Linux and Windows, but ZSTD shows up on macOS. We don't use compression anyway.
- Fix git version in Ubuntu 18.04 Dockerfile.
- Explicitly install `libtinfo5` for ISPC build, as it's not available by default in `ubuntu-22.04` image from Github.
- Use `ubuntu-20.04` image for testing with LLVM 11 & 12, as new version of library GLIBCXX has deprecation attributes for some of the stuff used by LLVM 11 & 12 and it causes build errors (with `-Werror` that we have enabled in CI)
- Fix Github Actions warnings: avoid using `set-output` and update `microsoft/setup-msbuild` action by switching to `v1` tag.


aneshlya (approved):

dbabokin:
@aneshlya I've added two more commits to fix Github Actions warnings.

aneshlya:
LGTM

PR #2403, opened 2022-12-09
GH Actions: fix Windows & enable macOS build

dbabokin:
- one of dependencies downloads on Windows we randomly failing - the reason redirect routes to the server with new TLS version requirement. Fix by enabling TLS of all versions.
- add macOS build. macOS testing is not added yet, as `alloy.py` in testing mode fails on macOS in a strange way (this still needs debugging).


aneshlya (approved):
Looks good!

PR #2404, opened 2022-12-12
Make dispatch function dllexport if needed

aneshlya:
Make dispatch export function callable from DLLs.
Required when producing dynamic library from ISPC object for further usage from ISPC Runtime on CPU.


dbabokin (approved):
Looks good. I've added a commit with minor fixes for the test. Once the testing passed, it can be merged.

PR #2410, opened 2022-12-21
WIP: Provide ispc implementation of __gnu_*_ieee.

nurmukhametov:
During codegeneration, LLVM lowers single<->half precison float casts to these calls. It happens for platform without hardware support for such instructions. E.g., for x86 it means for CPU without FP16C extension.


aneshlya:
Closing the PR since it's not needed anymore.

PR #2411, opened 2022-12-21
Add check_copyright.sh script

nurmukhametov:
To use run with any git ref as argument, e.g.:
```bash
$ ./check_copyright.sh HEAD^
```

### DIFF
--- a/.github/workflows/copyright-check.yml
+++ b/.github/workflows/copyright-check.yml
@@ -1,0 +1,21 @@
+# Copyright 2022 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause
+
+name: Check copyright
+
+on: [pull_request]
+
+jobs:
+  check_copyright:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+        submodules: false
+
+    - name: Check copyright
+      run: |
+        ./check_copyright.sh "origin/$GITHUB_BASE_REF"
+
--- a/.github/workflows/copyright-check.yml
+++ b/.github/workflows/copyright-check.yml
@@ -1,4 +1,4 @@
-# Copyright 2022 Intel Corporation
+# Copyright 2022, Intel Corporation
 # SPDX-License-Identifier: BSD-3-Clause
 
 name: Check copyright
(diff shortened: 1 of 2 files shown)


aneshlya:
It would be good to add a Github Action job running this script on PR.

dbabokin:
If I understand it correctly, it checks only the last commit in the PR. Any ideas how to check the whole PR?

dbabokin:
> If I understand it correctly, it checks only the last commit in the PR. Any ideas how to check the whole PR?

Oh, looks like I misunderstood it. So does it check the whole PR?

nurmukhametov:
> > If I understand it correctly, it checks only the last commit in the PR. Any ideas how to check the whole PR?
> 
> Oh, looks like I misunderstood it. So does it check the whole PR?

Yeah, it should check the whole PR diffing current state with GITHUB_BASE_REF

aneshlya (approved):

check_copyright.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+# ##################################################
+#  Copyright (c) 2022, Intel Corporation
+#  All rights reserved.
+#
+#  Redistribution and use in source and binary forms, with or without
```
dbabokin, line 6: I suggest using SDPX version of the license, like here: https://github.com/ispc/ispc/blob/0387b0679f2951caeff2a32446790980ac92215b/.github/workflows/rebuild-llvm15.yml

nurmukhametov: fixed

.github/workflows/copyright-check.yml
```diff
@@ -0,0 +1,21 @@
+# Copyright 2022 Intel Corporation
```
dbabokin, line 1:
```suggestion
# Copyright 2022, Intel Corporation
```

check_copyright.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+# Copyright 2022 Intel Corporation
```
dbabokin, line 2:
```suggestion
# Copyright 2022, Intel Corporation
```

dbabokin (approved):
Changes looks good to me! Thanks for implementing this check.

As a general note - going forward we should probably move this and other scripts to `scripts` folder.

PR #2412, opened 2022-12-21
Remove ISPC_NO_DUMPS

nurmukhametov:


tests/CMakeLists.txt
```diff
@@ -53,12 +53,10 @@ list(APPEND LIT_ARGS "-Dxe_enabled=$<IF:$<BOOL:${XE_ENABLED}>,ON,OFF>")
 # ISPC enabled OS.
 list(APPEND LIT_ARGS "-Dwindows_enabled=$<IF:$<BOOL:${ISPC_WINDOWS_TARGET}>,ON,OFF>")
 list(APPEND LIT_ARGS "-Dmacos_arm_enabled=$<IF:$<BOOL:${ISPC_MACOS_ARM_TARGET}>,ON,OFF>")
-# Dumps enabled/disbaled
-list(APPEND LIT_ARGS "-Ddumps_enabled=$<IF:$<BOOL:${ISPC_NO_DUMPS}>,OFF,ON>")
 # ISPC opaque pointers
 list(APPEND LIT_ARGS "-Dopaque_mode=$<IF:$<BOOL:${ISPC_OPAQUE_PTR_MODE}>,ON,OFF>")
 add_custom_target(check-all DEPENDS ispc
-    COMMAND ${LIT_COMMAND} ${LIT_ARGS}
+    COMMAND ${LIT_COMMAND} ${LIT_ARGS} "--verbose"
```
dbabokin, line 59: Please commit this line as a separate commit.

Arina and I recently discussed if we need to have `--verbose` by default and concluded that "probably yes". So please keep it in the PR. It's helpful for investigating problems in CI.

dbabokin (approved):
The changes look good. I remember that I tried using LLVM functionality available in Release build when initially implemented NO_DUMPS, but something didn't work as expected. But now I don't see any problems, it works like a charm. Either I did something wrong or LLVM has changed.

I also checked that changes in binary size and looks - release binary increased by ~1.5Kb.

tests/CMakeLists.txt
nurmukhametov: Committed as separate. 

As I understand `--verbose` would dump detailed log only if some tests have failed.

dbabokin: Yep.

aneshlya (approved):

aneshlya:
@dbabokin , should we use `--llvm-disable-assertions` in alloy.py by default now? 
Also regarding testing in Github Actions, it seems that we can now use LLVM without assertions for all types of testing, right?

dbabokin:
> @dbabokin , should we use `--llvm-disable-assertions` in alloy.py by default now? Also regarding testing in Github Actions, it seems that we can now use LLVM without assertions for all types of testing, right?

No, assertions are orthogonal. Build with assertions is preferable for testing, as it may found more bugs. Release needs to be build without assertions, as it's faster.

PR #2413, opened 2022-12-22
Detect supported ISPC architectures in ispc.cmake

aneshlya:
Introduce CMake variables `ISPC_X86_ENABLED`, `ISPC_ARM_ENABLED`, `ISPC_XE_ENABLED` to ispc.cmake to detect what targets are supported. It's needed in is ispc.cmake to check if x86 targets are enabled, otherwise the build is failing. These variables may be also beneficial in user's apps.


nurmukhametov (approved):

dbabokin (approved):
LGTM

PR #2414, opened 2022-12-29
Add support for ARM64 MSVC binary targets

anthony-linaro:
The existing `__aarch64__` macro is not defined for MSVC ARM64 platforms, so an additional check for `_M_ARM64` needs to be made in order to compile a native ARM64 binary.


dbabokin (approved):
Thanks for the contribution! The changes looks good to me.

PR #2417, opened 2023-01-05
alloy.py: shallow clone repo by default, add --full-checkout option

nurmukhametov:
Add flag `--shallow-clone` to clone `llvm` repository with `--depth=1`. It would be faster and consume less disk space.

### DIFF
--- a/alloy.py
+++ b/alloy.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 #
-#  Copyright (c) 2013-2022, Intel Corporation
+#  Copyright (c) 2013-2023, Intel Corporation
 #  All rights reserved.
 #
 #  Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


dbabokin:
Seems like useful trick. Should we use it by default? What are we missing without shallow clone? I've check standard build and it works fine.

nurmukhametov:
We should but not all use cases are clear for me. I thought maybe CI would uncover something but it looks like that CI doesn't use `alloy.py` to build llvm, so I am not sure that it would not break something.

dbabokin:
LLVM rebuild happens in CI, but upload of resulting binaries is a manual step.

I think we are good to use shallow copy by default. And then have full checkout as an option.

nurmukhametov:
> LLVM rebuild happens in CI, but upload of resulting binaries is a manual step.
Can you get me a link to jobs that build LLVM?

> I think we are good to use shallow copy by default. And then have full checkout as an option.
Ok, then I will rework change to `--no-shallow-clone` option.

nurmukhametov:
> > I think we are good to use shallow copy by default. And then have full checkout as an option.
> 
> Ok, then I will rework change to `--no-shallow-clone` option.

Made it `--full-checkout`.

dbabokin (approved):
LGTM.

Issue #2325, opened 2022-05-15
Test failure of llvm_ident.ispc


ArchangeGabriel:
This is on Arch Linux, using system LLVM 13.0.1.

```
******************** TEST 'ispc :: llvm_ident.ispc' FAILED ********************
Script:
--
: 'RUN: at line 1';   ispc /build/ispc/src/ispc-1.18.0/tests/lit-tests/llvm_ident.ispc --emit-llvm-text -o - | FileCheck /build/ispc/src/ispc-1.18.0/tests/lit-tests/llvm_ident.ispc
: 'RUN: at line 2';   ispc /build/ispc/src/ispc-1.18.0/tests/lit-tests/llvm_ident.ispc --emit-llvm-text --nostdlib -o - | FileCheck /build/ispc/src/ispc-1.18.0/tests/lit-tests/llvm_ident.ispc
--
Exit Code: 1

Command Output (stdout):
--
$ ":" "RUN: at line 1"
$ "ispc" "/build/ispc/src/ispc-1.18.0/tests/lit-tests/llvm_ident.ispc" "--emit-llvm-text" "-o" "-"
# command stderr:
Warning: No --target specified on command-line. Using default system target 
        "avx2-i32x8". 

$ "FileCheck" "/build/ispc/src/ispc-1.18.0/tests/lit-tests/llvm_ident.ispc"
# command stderr:
/build/ispc/src/ispc-1.18.0/tests/lit-tests/llvm_ident.ispc:6:11: error: CHECK: expected string not found in input
// CHECK: Intel(r) Implicit SPMD Program Compiler (Intel(r) ISPC), {{1\.[[:digit:]]+\.[[:digit:]]+(dev)?}} (build commit {{[[:xdigit:]]+}} @ {{([[:digit:]]{8})}}, LLVM [[LLVM_VERSION:[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+]])
          ^
<stdin>:1:1: note: scanning from here
; ModuleID = '/build/ispc/src/ispc-1.18.0/tests/lit-tests/llvm_ident.ispc'
^

Input file: <stdin>
Check file: /build/ispc/src/ispc-1.18.0/tests/lit-tests/llvm_ident.ispc

-dump-input=help explains the following input dump.

Input was:
<<<<<<
         1: ; ModuleID = '/build/ispc/src/ispc-1.18.0/tests/lit-tests/llvm_ident.ispc' 
[...]

Issue #2104, opened 2021-06-13
Please include only clang-capp.so instead of several libraries


luyatshimbalanga:
Fedora 32 and up uses libclang-cpp.so library to build some packages including ispc. It will be great to use that only according to the [changes on Fedora 32](https://fedoraproject.org/wiki/Changes/Stop-Shipping-Individual-Component-Libraries-In-clang-lib-Package) with this highlight:
```
The libclang-cpp.so shared object was added in clang 9.0 and now gives us an alternative shared library option that is supported by the upstream project and makes it possible for Fedora packages to stop using the component libraries. 
```

Here is an extract from an applied patch
```
From 22bbf8efc0deeb62cd706f1c2fee97201eb7427a Mon Sep 17 00:00:00 2001
From: Tom Stellard <tstellar@redhat.com>
Date: Fri, 13 Dec 2019 00:39:17 +0000
Subject: [PATCH] Link against libclang-cpp.so

---
 CMakeLists.txt | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/CMakeLists.txt b/CMakeLists.txt
index bd3483ab..edc97376 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -281,7 +281,7 @@
     list(APPEND ISPC_TARGETS wasm-i32x4)
 endif()
 
-set(CLANG_LIBRARY_LIST clangFrontend clangDriver clangSerialization clangParse clangSema clangAnalysis clangAST clangBasic clangEdit clangLex)
+set(CLANG_LIBRARY_LIST clang-cpp)
 set(LLVM_COMPONENTS engine ipo bitreader bitwriter instrumentation linker option)
 
 if (${LLVM_VERSION_NUMBER} VERSION_GREATER_EQUAL "10.0.0")
-- 
2.20.1
```

dbabokin:
We still do not recommend building ISPC against "system" version of LLVM libs. ISPC needs a patched version of LLVM, which is built specifically for ISPC, to be a fully functional and bug free.

dbabokin:
Is there a way for us to provide a correct recipe for building ISPC for official Fedora package?

Starting this release we build releases with Intel GPU support by default. This means much more complicated dependencies and build process. Also custom version of LLVM backend becomes more important.

luyatshimbalanga:
> Is there a way for us to provide a correct recipe for building ISPC for official Fedora package?

[...]

PR #2422, opened 2023-01-09
CMakeLists.txt: change libclang linking logic

nurmukhametov:
Link with static libraries (libclangFrontend.a, ...) if they found. Otherwise, dynamically link with libclang-cpp.so.

### DIFF
--- a/docker/fedora/Dockerfile
+++ b/docker/fedora/Dockerfile
@@ -1,67 +1,41 @@
-FROM fedora:26
-LABEL maintainer="Dmitry Babokin <dmitry.y.babokin@intel.com>"
+FROM fedora:36
+LABEL maintainer="Nurmukhametov, Aleksei <aleksei.nurmukhametov@intel.com>"
 SHELL ["/bin/bash", "-c"]
 
 ARG REPO=ispc/ispc
 ARG SHA=main
 
-# !!! Make sure that your docker config provides enough memory to the container,
-# otherwise LLVM build may fail, as it will use all the cores available to container.
-
-# Packages required to build ISPC and Clang.
+# Packages required to build ISPC.
 # RUN echo "proxy=http://proxy.yourcompany.com:888" >> /etc/dnf/dnf.conf
-RUN dnf install -y vim-enhanced-2:8.0.1806 wget-1.19.5 yum-utils-1.1.31 gcc-7.3.1 gcc-c++-7.3.1 git-2.13.6 python3-3.6.4 && \
-    dnf install -y m4-1.4.18 bison-3.0.4 flex-2.6.1 patch-2.7.6 make-4.2.1 ncurses-devel-6.0 glibc-devel-2.25-13.fc26.x86_64 glibc-devel-2.25-13.fc26.i686 && \
+RUN dnf install -y git-2.39.0 cmake-3.25.1 && \
+    dnf install -y clang-devel-14.0.5 llvm-devel-14.0.5 glibc-devel-2.35-20.fc36.i686 && \
+    dnf install -y flex-2.6.4 bison-3.8.2 && \
     dnf clean -y all
-
-# These packages are required if you need to link IPSC with -static.
-RUN dnf install -y ncurses-static-6.0 libstdc++-static-7.3.1 glibc-static-2.25 && \
-    dnf clean -y all
-
-# Download and install required version of cmake (3.13) for ISPC build
-RUN wget -q --retry-connrefused --waitretry=5 --read-timeout=20 --timeout=15 -t 5 https://github.com/Kitware/CMake/releases/download/v3.13.5/cmake-3.13.5-Linux-x86_64.sh && mkdir /opt/cmake && sh cmake-3.13.5-Linux-x86_64.sh --prefix=/opt/cmake --skip-license && \
-    ln -s /opt/cmake/bin/cmake /usr/local/bin/cmake && rm cmake-3.13.5-Linux-x86_64.sh
 
 # If you are behind a proxy, you need to configure git and svn.
-#RUN git config --global --add http.proxy http://proxy.yourcompany.com:888
+# RUN git config --global --add http.proxy http://proxy.yourcompany.com:888
 
 WORKDIR /usr/local/src
 
 # Fork ispc on github and clone *your* fork.
-RUN git clone https://github.com/$REPO.git ispc
-
-# This is home for Clang builds
-RUN mkdir /usr/local/src/llvm
+RUN git clone --depth=1 -b $SHA https://github.com/$REPO.git ispc
 
 ENV ISPC_HOME=/usr/local/src/ispc
-ENV LLVM_HOME=/usr/local/src/llvm
 
 # If you are going to run test for future platforms, go to
 # http://www.intel.com/software/sde and download the latest version,
 # extract it, add to path and set SDE_HOME.
 
-WORKDIR /usr/local/src/ispc
-RUN git checkout $SHA
-
-# Build Clang with all required patches.
-# Pass required LLVM_VERSION with --build-arg LLVM_VERSION=<version>.
-# By default 8.0 is used.
-# Note self-build options, it's required to build clang and ispc with the same compiler,
-# i.e. if clang was built by gcc, you may need to use gcc to build ispc (i.e. run "make gcc"),
-# or better do clang selfbuild and use it for ispc build as well (i.e. just "make").
-# "rm" are just to keep docker image small.
-ARG LLVM_VERSION=8.0
-RUN ./alloy.py -b --version="$LLVM_VERSION" --selfbuild && \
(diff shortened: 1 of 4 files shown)


aneshlya:
The change looks good but I'd add some comment like:
"ISPC is built with static libraries by default however it can be built against shared libraries in some controlled environments, like in package managers for particular OSes (issue 2418)
The linking against clang-cpp instead of individual clang libraries is recommended way by LLVM: 
https://fedoraproject.org/wiki/Changes/Stop-Shipping-Individual-Component-Libraries-In-clang-lib-Package.".

dbabokin:
Closes #2104

dbabokin:
@nurmukhametov Thanks for looking into this.

I'm a bit concerned about testing this feature. Please make sure that locally it works as expected.

Also, I suggest updating Fedora [Dockerfile](https://github.com/ispc/ispc/blob/main/docker/fedora/Dockerfile) and making it using system LLVM instead of our own, so we would test that build works in Fedora environment with dynamic lib.

Also would be interesting to check if there is any difference in start up time between static lib version and dynamic lib version.

nurmukhametov:
> @nurmukhametov Thanks for looking into this.
> 
> I'm a bit concerned about testing this feature. Please make sure that locally it works as expected.
>
Except CI, I have tested it locally on Arch and it worked fine. 

> Also, I suggest updating Fedora [Dockerfile](https://github.com/ispc/ispc/blob/main/docker/fedora/Dockerfile) and making it using system LLVM instead of our own, so we would test that build works in Fedora environment with dynamic lib.
I can update this Dockerfile image up to Fedora 36 (contains LLVM 14). But I don't really understand where and how this image is used in CI.

> Also would be interesting to check if there is any difference in start up time between static lib version and dynamic lib version.
Probably, but I think if startup time changed (even drastically) it would be fraction of a second.

nurmukhametov:
> @nurmukhametov Thanks for looking into this.
> 
> I'm a bit concerned about testing this feature. Please make sure that locally it works as expected.
> 
> Also, I suggest updating Fedora [Dockerfile](https://github.com/ispc/ispc/blob/main/docker/fedora/Dockerfile) and making it using system LLVM instead of our own, so we would test that build works in Fedora environment with dynamic lib.
> 
> Also would be interesting to check if there is any difference in start up time between static lib version and dynamic lib version.

I have update Fedora Dockerfile up to Fedora 36 (contains LLVM 14), see changes in #2422 . But I don't really understand where and how this image is used in CI.

docker/fedora/Dockerfile
```diff
@@ -1,67 +1,39 @@
...
 
 # Packages required to build ISPC and Clang.
 # RUN echo "proxy=http://proxy.yourcompany.com:888" >> /etc/dnf/dnf.conf
-RUN dnf install -y vim-enhanced-2:8.0.1806 wget-1.19.5 yum-utils-1.1.31 gcc-7.3.1 gcc-c++-7.3.1 git-2.13.6 python3-3.6.4 && \
-    dnf install -y m4-1.4.18 bison-3.0.4 flex-2.6.1 patch-2.7.6 make-4.2.1 ncurses-devel-6.0 glibc-devel-2.25-13.fc26.x86_64 glibc-devel-2.25-13.fc26.i686 && \
-    dnf clean -y all
-
-# These packages are required if you need to link IPSC with -static.
-RUN dnf install -y ncurses-static-6.0 libstdc++-static-7.3.1 glibc-static-2.25 && \
+RUN dnf install -y vim-enhanced wget yum-utils git python3 cmake && \
```
dbabokin, line 11: We need to have specific package versions, otherwise we have problems with security scans. @aneshlya know more about it.

docker/fedora/Dockerfile
```diff
@@ -1,67 +1,39 @@
...
+RUN git clone --depth=1 -b $SHA https://github.com/$REPO.git ispc
 
 ENV ISPC_HOME=/usr/local/src/ispc
-ENV LLVM_HOME=/usr/local/src/llvm
 
 # If you are going to run test for future platforms, go to
 # http://www.intel.com/software/sde and download the latest version,
 # extract it, add to path and set SDE_HOME.
 
-WORKDIR /usr/local/src/ispc
```
dbabokin, line 43: I suggest adding a note here explaining why we don't build LLVM and mentioning that it links dynamically a system LLVM and that the way ISPC is build in Fedora package.

tests/lit-tests/llvm_ident.ispc
```diff
@@ -3,7 +3,7 @@
 
 // Check that version string embeded in !llvm.ident metadata is formed correctly.
 
-// CHECK: Intel(r) Implicit SPMD Program Compiler (Intel(r) ISPC), {{1\.[[:digit:]]+\.[[:digit:]]+(dev)?}} (build commit {{[[:xdigit:]]+}} @ {{([[:digit:]]{8})}}, LLVM [[LLVM_VERSION:[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+]])
```
dbabokin, line 6: Could you explain what is going on here and why?

dbabokin (commented):
Also need to change the description of Dockerfile in README:
https://github.com/ispc/ispc/blob/e88e49eb161aa5e75aecc843d479d5e757f5c43e/docker/README.md?plain=1#L29

Dockerfile for Fedora is not used in CI. It would be cool to build all Dockerfiles in CI from time to time, but in public CI we don't have enough compute resources to do that.

tests/lit-tests/llvm_ident.ispc
ArchangeGabriel: I don’t know because I can’t understand the code, but I wouldn’t be surprised that this is related to the failure I’ve been seeing for ages in https://github.com/ispc/ispc/issues/2325.

ArchangeGabriel: Confirmed that this fixes #2325.

ArchangeGabriel:
Seems to work fine for me.

tests/lit-tests/llvm_ident.ispc
nurmukhametov: It pattern matches versions of ISPC and LLVM embedded in bitcode file. LLVM version string is slightly different across different LLVM builds.
E.g., Fedora
```
!0 = !{!"Intel(r) Implicit SPMD Program Compiler (Intel(r) ISPC), 1.19.0dev (build commit 5cce243d430fc30c @ 20230114, LLVM 14.0.5)"}
!1 = !{!"LLVM version 14.0.5 (Fedora 14.0.5-2.fc36)"}
```
Arch
```
!0 = !{!"Intel(r) Implicit SPMD Program Compiler (Intel(r) ISPC), 1.18.1 (build  @ 20221112, LLVM 14.0.6)"}
!1 = !{!"LLVM version 14.0.6"}
```
Our build
```
!0 = !{!"Intel(r) Implicit SPMD Program Compiler (Intel(r) ISPC), 1.19.0dev (build commit 6efb053cacaa69e3 @ 20230111, LLVM 13.0.1)"}
!1 = !{!"LLVM version 13.0.1 (https://github.com/llvm/llvm-project.git 75e33f71c2dae584b13a7d1186ae0a038ba98838)"}
```

docker/fedora/Dockerfile
nurmukhametov: Security scans? Who or what performs them?

nurmukhametov:
> Also need to change the description of Dockerfile in README:
> 
> https://github.com/ispc/ispc/blob/e88e49eb161aa5e75aecc843d479d5e757f5c43e/docker/README.md?plain=1#L29
> 
> Dockerfile for Fedora is not used in CI. It would be cool to build all Dockerfiles in CI from time to time, but in public CI we don't have enough compute resources to do that.

It is not a problem to change that but I have lost the point why we need these all (Dockerfile, docker image, ISPC built in it) at all?

tests/lit-tests/llvm_ident.ispc
dbabokin: Changes in the regexp looks good.

docker/fedora/Dockerfile
nurmukhametov: Added versions. Fixed Readme.

docker/fedora/Dockerfile
nurmukhametov: Done

nurmukhametov:
> > > Also need to change the description of Dockerfile in README:
> > > https://github.com/ispc/ispc/blob/e88e49eb161aa5e75aecc843d479d5e757f5c43e/docker/README.md?plain=1#L29
> > > 
> > > Dockerfile for Fedora is not used in CI. It would be cool to build all Dockerfiles in CI from time to time, but in public CI we don't have enough compute resources to do that.
> > 
> > 
> > It is not a problem to change that but I have lost the point why we need these all (Dockerfile, docker image, ISPC built in it) at all?
> 
> So I saved old Dockerfile and just added a new one.

Replace an old with new one.

PR #2423, opened 2023-01-09
builtins: add argn macro

nurmukhametov:
Use it to portably select greater than ninth argument.

### DIFF
--- a/builtins/target-avx-utils.ll
+++ b/builtins/target-avx-utils.ll
@@ -1,4 +1,4 @@
-;;  Copyright (c) 2010-2020, Intel Corporation
+;;  Copyright (c) 2010-2023, Intel Corporation
 ;;  All rights reserved.
 ;;
 ;;  Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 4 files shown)


builtins/target-avx-utils.ll
```diff
@@ -294,25 +294,25 @@ define_avgs()
 define(`convert_scale_to_const', `
 
 
- switch i32 %$10, label %default_$1 [ i32 1, label %on_one_$1
+ switch i32 %argn(`10',$@), label %default_$1 [ i32 1, label %on_one_$1
                                       i32 2, label %on_two_$1
```
dbabokin, line 298: Here and in other instances below - need to change alignment of line following the changed line.

dbabokin:
Why do we need this change? I mean what is specific environment that has this problem?

nurmukhametov:
I tried to build in under FreeBSD and their m4 doesn't support two-digits arguments. 

Quote from here that two-digits is GNU extension that is incompatible with POSIX https://www.gnu.org/software/m4/manual/m4.html

> GNU m4 allows the number following the ‘$’ to consist of one or more digits, allowing macros to have any number of arguments. The extension of accepting multiple digits is incompatible with POSIX, and is different than traditional implementations of m4, which only recognize one digit. Therefore, future versions of GNU M4 will phase out this feature. To portably access beyond the ninth argument, you can use the argn macro documented later (see [Shift](https://www.gnu.org/software/m4/manual/m4.html#Shift)).

builtins/target-avx-utils.ll
nurmukhametov: Fixed

Issue #2419, opened 2023-01-08
Build failure with spirv-llvm-translator 15.x


ArchangeGabriel:
While trying to upgrade our stack to LLVM 15, we faced a build failure in ispc git head:
```cxx
/build/ispc/src/ispc-1.18.1.r285+ga77f4e2/src/opt/MangleOpenCLBuiltins.cpp:66:5: error: no matching function for call to 'mangleOpenClBuiltin'
    mangleOpenClBuiltin(funcName, ArgTy, mangledName);
    ^~~~~~~~~~~~~~~~~~~
/usr/include/LLVMSPIRVLib/LLVMSPIRVLib.h:166:6: note: candidate function not viable: requires 4 arguments, but 3 were provided
void mangleOpenClBuiltin(const std::string &UnmangledName,
     ^
/build/ispc/src/ispc-1.18.1.r285+ga77f4e2/src/opt/MangleOpenCLBuiltins.cpp:73:5: error: no matching function for call to 'mangleOpenClBuiltin'
    mangleOpenClBuiltin(func.getName().str(), func.getArg(0)->getType(), mangledName);
    ^~~~~~~~~~~~~~~~~~~
/usr/include/LLVMSPIRVLib/LLVMSPIRVLib.h:166:6: note: candidate function not viable: requires 4 arguments, but 3 were provided
void mangleOpenClBuiltin(const std::string &UnmangledName,
     ^
/build/ispc/src/ispc-1.18.1.r285+ga77f4e2/src/opt/MangleOpenCLBuiltins.cpp:87:5: error: no matching function for call to 'mangleOpenClBuiltin'
    mangleOpenClBuiltin(func.getName().str(), func.getArg(0)->getType(), mangledName);
    ^~~~~~~~~~~~~~~~~~~
/usr/include/LLVMSPIRVLib/LLVMSPIRVLib.h:166:6: note: candidate function not viable: requires 4 arguments, but 3 were provided
void mangleOpenClBuiltin(const std::string &UnmangledName,
     ^
```
This is due to a change in spirv-llvm-translator that is part of the release targeting LLVM 15: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/commit/0eb9a7d2937542e1f95a4e1f9aa9850e669dc45f.

foutrelis:
[...]

PR #2426, opened 2023-01-13
fix for build with spirv_translator v15.0

aneshlya:
This PR fixes ISPC build with spirv_translator v15.0 and removes usages of LLVM `freeze` instruction since it's not supported by LLVM translator v15.0.
Relates to #2419


nurmukhametov (approved):

dbabokin (approved):
Looks good!

PR #2431, opened 2023-01-20
Use LLVM 15.0.7 for build and CI

dbabokin:
- Build LLVM `15.0.7` in `alloy.py` instead of `15.0.6`
- Use LLVM `15.0.7` in CI
- Fix Ubuntu 18.04 Dockerfile to avoid certain pinned package versions - more general fix is needed.


aneshlya (approved):
LGTM

PR #2434, opened 2023-01-27
Add avx512spr targets and MeteorLake CPU definition

dbabokin:
This PR adds support for SPR targets: `-x4`, `-x8`, `-x16`, `-x32`, and `-x64`. The key difference vs SKX targets, beyond just different code scheduling, which might be enabled through `--device=spr` switch, is that SPR targets support `float16` natively and the targets enable full set of `float16` instructions that assumed to be reachable from ISPC language.

Note that the SPR support is enabled only starting form LLVM 14. Many patches were committed to LLVM 15, 16, and 17 to improve different aspects of `float16` code generation, some of them were backported ported to LLVM 14, 15, and 16. This causes initial tests failure in CI, I'll update pre-built LLVM 14/15 and return the tests when it's ready.

MTL CPU definition enables `--device=mtl` and `--device=meteorlake` for builds with LLVM 16 and later.


check_isa.cpp
```diff
@@ -115,36 +115,70 @@ static const char *lGetSystemISA() {
...
+    if (osxsave && avx2 && avx512_f && __os_has_avx512_support()) {
         // We need to verify that AVX2 is also available,
         // as well as AVX512, because our targets are supposed
         // to use both.
 
-        if ((info2[1] & (1 << 17)) != 0 && // AVX512 DQ
-            (info2[1] & (1 << 28)) != 0 && // AVX512 CDI
-            (info2[1] & (1 << 30)) != 0 && // AVX512 BW
-            (info2[1] & (1 << 31)) != 0) { // AVX512 VL
+        // Knights Landing:          KNL = F + PF + ER + CD
```
aneshlya, line 173: Seems that the comment is not right. Should it be KNL = PF + ER + CD?

check_isa.cpp
```diff
@@ -115,36 +115,70 @@ static const char *lGetSystemISA() {
...
+    bool avx512_vl =           (info2[1] & (1 << 31)) != 0;
+    bool avx512_vbmi2 =        (info2[2] & (1 << 6))  != 0;
+    bool avx512_gfni =         (info2[2] & (1 << 8))  != 0;
+    bool avx512_vaes =         (info2[2] & (1 << 9))  != 0;
+    bool avx512_vpclmulqdq =   (info2[2] & (1 << 10)) != 0;
+    bool avx512_vnni =         (info2[2] & (1 << 11)) != 0;
+    bool avx512_bitalg =       (info2[2] & (1 << 12)) != 0;
+    bool avx512_vpopcntdq =    (info2[2] & (1 << 14)) != 0;
+    bool avx512_bf16 =         (info3[0] & (1 << 5))  != 0;
+    bool avx512_vp2intersect = (info2[3] & (1 << 8))  != 0;
```
aneshlya, line 161: It should be on the line above before info3 starts.

dbabokin: It's sorted on the "left" part, which is going in the order it's defined in ISA docs.

check_isa.cpp
dbabokin: F is checked upfront

check_isa.cpp
```diff
@@ -115,36 +130,84 @@ static const char *lGetSystemISA() {
...
-            (info2[1] & (1 << 28)) != 0 && // AVX512 CDI
-            (info2[1] & (1 << 30)) != 0 && // AVX512 BW
-            (info2[1] & (1 << 31)) != 0) { // AVX512 VL
+        // Knights Landing:          KNL = F + PF + ER + CD
+        // Skylake server:           SKX = F + DQ + CD + BW + VL
+        // Cascade Lake server:      CLX = SKX + VNNI
+        // Cooper Lake server:       CPX = CLX + BF16
+        // Ice Lake client & server: ICL = CLX + VBMI2 + GFNI + VAES + VPCLMULQDQ + BITALG + VPOPCNTDQ
+        // Tiger Lake:               TGL = ICL + VP2INTERSECT
+        // Sapphire Rapids:          SPR = TGL + BF16 + AMX_BF16 + AMX_TILE + AMX_INT8 + AVX_VNNI
```
aneshlya, line 179: avx512_fp16 is missed in the comment

src/ispc.cpp
```diff
@@ -157,44 +157,88 @@ static ISPCTarget lGetSystemISA() {
...
-        if ((info2[1] & (1 << 17)) != 0 && // AVX512 DQ
-            (info2[1] & (1 << 28)) != 0 && // AVX512 CDI
-            (info2[1] & (1 << 30)) != 0 && // AVX512 BW
-            (info2[1] & (1 << 31)) != 0) { // AVX512 VL
+        // Knights Landing:          KNL = F + PF + ER + CD
+        // Skylake server:           SKX = F + DQ + CD + BW + VL
+        // Cascade Lake server:      CLX = SKX + VNNI
+        // Cooper Lake server:       CPX = CLX + BF16
+        // Ice Lake client & server: ICL = CLX + VBMI2 + GFNI + VAES + VPCLMULQDQ + BITALG + VPOPCNTDQ
+        // Tiger Lake:               TGL = ICL + VP2INTERSECT
```
aneshlya, line 205: Please add comment about Sapphire Rapids here

check_isa.cpp
dbabokin: Fixed. Thanks for catching it.

src/ispc.cpp
dbabokin: Fixed.

CMakeLists.txt
```diff
@@ -277,6 +277,7 @@ if (X86_ENABLED)
         avx2-i8x32 avx2-i16x16
         avx2-i32x4 avx2-i32x8 avx2-i32x16 avx2-i64x4
         avx512knl-x16
+        # avx512spr targets are alised to avx512skx targets with attributes enabling fp16 and other new instructions.
```
aneshlya, line 280: alised->aliased

tests/lit-tests/targets_x86_llvm14plus.ispc
```diff
@@ -0,0 +1,13 @@
+// The test checks that target definitions (including all synonyms) are successfully consumed by compiler.
+
+//; RUN: %{ispc} %s -o %t.o --nostdlib --target=avx512skx-x4
```
aneshlya, line 3: Is the intention really check *skx-x* targets or *spr-x* targets?

src/ispc.h
```diff
@@ -426,9 +429,12 @@ class Target {
...
 
-    /** Indicates whether the target has native support for float/half
-        conversions. */
-    bool m_hasHalf;
+    /** Indicates whether the target has native support for float/half conversions. */
+    bool m_hasHalfConverts;
+
+    /** Indicates whether the target has full native support for float16 type, i.e.
+        arithmetic operations, rsqrt, rcp, etc. */
+    bool m_hasHalfFullSupport;
```
aneshlya, line 437: What is difference between `m_hasHalfFullSupport` and `m_hasFp16Support` that we already have?

src/module.cpp
```diff
@@ -2417,7 +2417,7 @@ int Module::execPreprocessor(const char *infilename, llvm::raw_string_ostream *o
     else
         opts.addMacroDef("ISPC_POINTER_SIZE=64");
 
-    if (g->target->hasHalf())
+    if (g->target->hasHalfConverts())
         opts.addMacroDef("ISPC_TARGET_HAS_HALF");
```
aneshlya, line 2421: I would rename the macro `ISPC_TARGET_HAS_HALF` to `ISPC_TARGET_HAS_HALF_CONVERSION` to avoid possible confusions with `ISPC_FP16_SUPPORTED`

CMakeLists.txt
dbabokin: I've reimplemented the targets to make them non-alias, this line is gone now.

tests/lit-tests/targets_x86_llvm14plus.ispc
dbabokin: Good catch, fixed.

src/ispc.h
dbabokin: That's very good sighting! I've added a TODO to add merge them. Right now they are different entities because m_hasHalfFullSupport is not enabled on ARM, as we need to differentiate between platform that support and do not support fp16 on ARM

src/module.cpp
dbabokin: `ISPC_TARGET_HAS_HALF ` is not mentioned in our docs, so it's probably ok to remove it. But i don't want to change it for compatibility reasons.

PR #2440, opened 2023-02-05
Fix crash in main when no command argument is provided

Lectem:
#2438 introduced a new `link` command, but failed to check `argc` before reading `argv[1]`, which resulted in a crash.


aneshlya (approved):

aneshlya:
@Lectem , thanks a lot for noticing and fixing this!

PR #2441, opened 2023-02-09
LLVM build and usage: NODUMPs, discard-value-names

dbabokin:
- Remove `LLVM_ENABLE_DUMP` for optimized/release LLVM build. The functionality is enabled by default for debug build (`--debug` switch for `alloy.py`) anyway. For optimized build it costs us ~1.5Mb of binary size and slight slowdown for compile time.
- Change the default behavior for `--[no]discard-value-names`. By default it's only enabled for if the output is bitcode (text or binary). This reduces compile time **significantly**. For the test case I use for the measurement, it brought down compile time from 3:52 to 3:25 (LLVM build was with asserts).
- Add libclangASTMatchers for the list of linked libraries for LLVM 16, as it's required now.
- Upgrade SDE in CI `9.0.0` -> `9.14.0`.

### DIFF
--- a/tests/lit-tests/discard_value_names.ispc
+++ b/tests/lit-tests/discard_value_names.ispc
@@ -1,14 +1,22 @@
 
+// Check that --[no]discard-value-names has an effect
 // RUN: %{ispc} %s --target=host --nowrap -O2 --emit-llvm-text --discard-value-names -o - | FileCheck %s -check-prefix=CHECK_DISCARD
 // RUN: %{ispc} %s --target=host --nowrap -O2 --emit-llvm-text --no-discard-value-names -o - | FileCheck %s -check-prefix=CHECK_NODISCARD
 
-// CHECK_DISCARD-NOT: allocas:
-// CHECK_NODISCARD: allocas:
+// Check that by default values are generated for LLVM IR dump
+// RUN: %{ispc} %s --target=host --nowrap -O2 --emit-llvm-text -o - | FileCheck %s -check-prefix=CHECK_NODISCARD
+
+// Check that for assembler by default there no names and that we can still see them if commanded so.
+// RUN: %{ispc} %s --target=host --nowrap -O2 --emit-asm -o - | FileCheck %s -check-prefix=CHECK_DISCARD
+// RUN: %{ispc} %s --target=host --nowrap -O2 --emit-asm --no-discard-value-names -o - | FileCheck %s -check-prefix=CHECK_NODISCARD
+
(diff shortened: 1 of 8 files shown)


tests/lit-tests/discard_value_names.ispc
```diff
@@ -1,6 +1,7 @@
 
 // RUN: %{ispc} %s --target=host --nowrap -O2 --emit-llvm-text --discard-value-names -o - | FileCheck %s -check-prefix=CHECK_DISCARD
 // RUN: %{ispc} %s --target=host --nowrap -O2 --emit-llvm-text --no-discard-value-names -o - | FileCheck %s -check-prefix=CHECK_NODISCARD
+// RUN: %{ispc} %s --target=host --nowrap -O2 --emit-llvm-text -o - | FileCheck %s -check-prefix=CHECK_NODISCARD
```
nurmukhametov, line 4: I think we need to add a test checking that by default value names are really discarded. Something like that probably should work:
```
// RUN: %{ispc} %s --target=host --nowrap -O2 --emit-asm -o - | FileCheck %s -check-prefix=CHECK_DISCARD
```

nurmukhametov (changes requested):

tests/lit-tests/discard_value_names.ispc
dbabokin: There's no way to check that. These are purely LLVM IR values, they are not propagated anywhere further.

nurmukhametov: Are you sure?

dbabokin: I see no ways to check, unless to dump LLVM IR. My understanding is that these values are purely to make LLVM IR more human readable and are discard downstream.

nurmukhametov: I just checked a random assembler file from my build directory.

```asm
test_uint_f___UM_vyu:                   # @test_uint_f___UM_vyu
# %bb.0:                                # %allocas
```

Is not this enough?

dbabokin: Hmmm, I checked assembler and haven't noticed that. Now I see it. I'll modify the test. That's a good point.

dbabokin: I've updated the files for review

nurmukhametov: Looks good!

nurmukhametov (approved):

PR #2442, opened 2023-02-10
Redesign target-dependent performance warnings

dbabokin:
This PR add a generic mechanism to enable performance warnings only for selected targets. Such warnings are enumerated in `PerfWarningType` enum and are explicitly enabled for every target (or class of targets based on ISA).

Also, the set of warnings for inefficient converts were changed:
- warnings for `[float16|float|double]` -> `uint[8|16]` were removed, as they are not a problem on any of existing target. Looks like it was a copy-past error for these warnings from day-one (most of them were added really long time ago).
- warnings for `uint[32|64]` -> `[float16|float|double]` were added, because they are as problematic as converts in reverse direction (which we warn for).

Warnings related to `float16` converts were disabled for SSE2/SSE4/AVX1 targets, as these targets do not have any instruction for `float16` type support and anything related to `float16` is dead slow anyway.

Most of covered warnings were disabled for non-x86 targets (because they were not intended to be enabled there in the first place and we need a dedicated investigation of what needs to be covered for ARM and possibly Xe) and AVX512 targets (because they don't have these problems).

PS This redesign was triggered by enabling SPR targets, as I observed the warnings that are not relevant for them.

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -6218,6 +6218,7 @@ static llvm::Value *lTypeConvAtomic(FunctionEmitContext *ctx, llvm::Value *exprV
                 PerformanceWarning(pos, "Conversion from float16 to unsigned int is slow. Use \"int\" if possible");
             cast = ctx->CastInst(llvm::Instruction::FPToUI, // unsigned int
                                  exprVal, targetType, cOpName);
+            break;
         case AtomicType::TYPE_FLOAT:
             if (fromType->IsVaryingType())
                 PerformanceWarning(pos, "Conversion from float to unsigned int is slow. Use \"int\" if possible");
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -6214,20 +6214,14 @@ static llvm::Value *lTypeConvAtomic(FunctionEmitContext *ctx, llvm::Value *exprV
             cast = ctx->TruncInst(exprVal, targetType, cOpName);
             break;
         case AtomicType::TYPE_FLOAT16:
-            if (fromType->IsVaryingType())
-                PerformanceWarning(pos, "Conversion from float16 to uint8 is slow. Use \"int8\" if possible");
             cast = ctx->CastInst(llvm::Instruction::FPToUI, // unsigned int
                                  exprVal, targetType, cOpName);
             break;
         case AtomicType::TYPE_FLOAT:
-            if (fromType->IsVaryingType())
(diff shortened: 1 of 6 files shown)


src/ispc.h
```diff
@@ -327,7 +346,9 @@ class Target {
 
     bool hasFp64Support() const { return m_hasFp64Support; }
 
-    bool warnFtoU32IsExpensive() const { return m_warnFtoU32IsExpensive; }
+    void setWarning(PerfWarningType warningType) { m_warnings |= static_cast<unsigned int>(warningType); }
+
+    bool shouldWarn(PerfWarningType warningType) { return (m_warnings & static_cast<unsigned int>(warningType)) != 0; }
```
nurmukhametov, line 351: I am slightly worried about that because enum is actually signed int.

dbabokin: That should be an issue if the underlying type is picked smaller than `unsigned int`, so the conversion would involve sign extend. Default choice for the compiler in this case is `int`, but it's implementation defined. So you are right that it's a gray zone. I've pushed commit that fixes that.

nurmukhametov: Great!

nurmukhametov (approved):

aneshlya (approved):

src/expr.cpp
```diff
@@ -6058,8 +6058,20 @@ static llvm::Value *lTypeConvAtomic(FunctionEmitContext *ctx, llvm::Value *exprV
             break;
         case AtomicType::TYPE_UINT8:
         case AtomicType::TYPE_UINT16:
+            cast = ctx->CastInst(llvm::Instruction::UIToFP, // unsigned int to float16
+                                 exprVal, targetType, cOpName);
+            break;
         case AtomicType::TYPE_UINT32:
+            if (fromType->IsVaryingType() && g->target->shouldWarn(PerfWarningType::CVTUIntFloat16)) {
+                PerformanceWarning(pos, "Conversion from uint32 to float16 is slow. Use \"int32\" if possible");
```
aneshlya, line 6066: I think it would be good to use AtomicType::GetString() instead of writing types explicitly inside the message.
It would also allow to use break through in switch statement again.

dbabokin: Agree, though it would require to "sanitize" input AtomicType to drop possible `const` qualifier, I think. I'll leave it for another round of warnings review.

Issue #2248, opened 2022-01-30
WASM target build error. Undeclared symbol "__rcp_fast_varying_double"


DVLP:
When building with default target the code is working but when switching to `--target=wasm-i32x4`
I'm getting:

#6 0.404 stdlib.ispc:2152:16: Error: Undeclared symbol "__rcp_fast_varying_double".
#6 0.404 stdlib.ispc:2170:12: Error: Undeclared symbol "__rcp_varying_half". Did you
#6 0.404         mean "__exp_varying_half"?

`m_hasRcpd` is false for wasm-i32x4

Could it be related to this issue? Some placeholder declaration missing?
https://github.com/ispc/ispc/issues/553

DVLP:
Ok so building with  -DXE_ENABLED=FALSE seems to be a workaround but it's not good enough if someone is using a pre-built version. i.e. version 18dev has both WASM and XE enabled

Then there's another one in line std 1548 about non-void function that should return a value. I removed the condition and left just `return (float16)(rcp((float)v))` and it worked. Interesting. Could it be that this if/else statement is being interpreted incorrectly by static type checking due to lack of brackets?
```
if (__is_xe_target)
        return __rcp_varying_half(v);
    else
        return (float16)(rcp((float)v));
`

dbabokin:
Seems placeholder declarations are missing in `builtins/target-wasm-i32x4.ll`.

[rcph_decl()](https://github.com/ispc/ispc/blob/575929afa9e9ae0cc44e766b385bf1d70f7dc938/builtins/util.m4#L8104) macro call and some other should be in `builtins/target-wasm-i32x4.ll`.

Note that WASM is not fully functional target yet. @aschrein did great job bring this target in, but switched his focus before completing it. If there are any interest in this target, I would fully support contributions to make it work.

Lectem:
I tried to fix this earlier based on the main branch.
For me, only `__rcp_fast_varying_double` was missing. It seems there is an implementation in the ll file for `__rcp_varying_double` but not the fast one. (the float version just uses the same code for both).
I tried doing the same for the double version and could generate my library correctly. However it seems that the node.js version shipped with latest emscripten won't run it.
Apparently there is an unknown Opcode generated which I find surprising. (it is probably unrelated to this function).
Is there any way to check the output of ispc for errors, perhaps by using a disassembler? Any guidance would be helpful.

aneshlya:
Fixed by #2580

PR #2451, opened 2023-02-27
Fix wasm32 target

Lectem:
The ipsc stdlib was missing some builtins for the wasm32 target which caused the following issue: https://github.com/ispc/ispc/issues/2248

Running tests on windows was also not well supported (I don't think it ever worked ?) so I fixed it. (note I ran the x86 tests to check if everything was still working fine).

I bumped emscripten version to 3.1.31 (we could perhaps even bump to `3.1.32` or `3.1.33` ?).
The objective was to use a more recent version of emscripten to make sure we have as few issues as possible related to https://github.com/WebAssembly/simd/pull/209 (as mentionned in this old PR #1750  ).

Tests results:
```
1 / 1585 tests FAILED compilation
        .\tests\pragma-ignore-warning.ispc
167 / 1585 tests FAILED execution
        .\tests\atomics-1.ispc
        .\tests\atomics-13.ispc
        .\tests\atomics-2.ispc
        .\tests\atomics-9.ispc
        .\tests\atomics-uniform-7.ispc
        .\tests\clock.ispc
        .\tests\local-atomics-uniform-7.ispc
   ```

According to #1750 the clock test failing is expected.
However I'm not sure about the atomics. From what I saw this seems to be pretty generic code in `builtins\util.m4`, so maybe bad codegen from llvm if the version we are using predates the final renumbering of opcodes ? Those date back to 2020 though so I would expect LLVM13 to have the correct opcodes.

If @aschrein has any insight on what could be the issue it would be very helpful !

### DIFF
--- a/builtins/target-wasm-i32x4.ll
+++ b/builtins/target-wasm-i32x4.ll
@@ -640,10 +640,22 @@ entry:
   ret <4 x float> %div.i
 }
 
+define hidden <4 x double> @__rcp_varying_double(<4 x double> %v) local_unnamed_addr #0 {
+entry:
+  %div.i = fdiv <4 x double> <double 1.000000e+00, double 1.000000e+00, double 1.000000e+00, double 1.000000e+00>, %v
+  ret <4 x double> %div.i
+}
+
 define hidden <4 x float> @__rcp_fast_varying_float(<4 x float> %v) local_unnamed_addr #0 {
 entry:
   %div.i = fdiv <4 x float> <float 1.000000e+00, float 1.000000e+00, float 1.000000e+00, float 1.000000e+00>, %v
   ret <4 x float> %div.i
+}
+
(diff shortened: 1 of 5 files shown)


Lectem:
It seems the build on windows finds LLVM17 in the `PATH` instead of the local LLVM15 https://ci.appveyor.com/project/ispc/ispc/builds/46344947/job/wkfk2pa61mgh569x#L3681

I had this issue locally as I had an official release of clang in my `PATH` instead of ISPC's version. I fixed it by putting the local version at the beginning of my PATH so that it had priority over the system version. I'm not sure however how this should be fixed here for AppVeyor ?

aneshlya:
Hi @Lectem, there was another PR merged #2580 related to wasm support which highly intersects with this one. If you think that some changes are still missed, feel free to open another PR on top of main.

Lectem:
It seems the other PR fixes all the issues, and probably better than what I did, thanks !

PR #2463, opened 2023-03-07
Reduce complexity of ISPC opt phases

dbabokin:
Many of ISPC opt phases are unnecessary O(n^2) on the number of optimization triggers in the single basic block. This is because optimization restart on the basic block once it triggers, but there's not real need for that it iteration is done more carefully. The key is to maintain a valid iterator after the optimization removes "current" instruction.

The PR makes the complexity linear.

The PR reduce compile time for small examples approximately by 5%. For large examples the benefits are way smaller.

### DIFF
--- a/src/llvmutil.cpp
+++ b/src/llvmutil.cpp
@@ -1665,45 +1665,45 @@ bool LLVMIsValueUndef(llvm::Value *value) {
     return false;
 }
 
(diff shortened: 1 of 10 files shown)


nurmukhametov (approved):
I measured performance on ispc-corpus. Improvements are between 0.4% - 30%. It is great!

aneshlya (approved):
LGTM!

Issue #2402, opened 2022-12-05
ISPC uses SSE4.2 instructions on SSE4.1 platforms


UnityBartPieters:
**Summary**

It seems ISPC is generating SSE4.2 instructions that get used on a SSE4.1 platform, resulting in a crash on Penryn CPUs that support 4.1 but not 4.2. 

In ispc.cpp line 732 it is noted that ISPC does not output SSE 4.2 instructions: 

`// Penryn is here because ISPC does not use SSE 4.2`

Yet when running ISPC-generated code on the Intel SDE emulator, it detects the use of `pcmpgtq`. On a client machine with Penryn architecture, it causes a crash. 

Might I be doing something wrong?

**SDE Output** 
`sde-external-9.0.0-2021-11-07-win\sde.exe -pnr -- bc7etest.exe
TID 0 SDE-ERROR: Executed instruction not valid for specified chip (PENRYN): 0x7ff84b19009e: pcmpgtq xmm1, xmm5
Image: E:\BugFixes\bc7e\compress_bc7e.dll+0x1009e
Function: bc7e_compress_block_init___
Instruction bytes are: 66 0f 38 37 cd`

Tested with most recent ISPC and a couple of versions all the way back to v1.12.0, on Windows.

**Repro steps** 
- Compile [https://github.com/BinomialLLC/bc7e/blob/main/bc7e.ispc](https://github.com/BinomialLLC/bc7e/blob/main/bc7e.ispc) with target sse4.
- Use listed example program (below) that links to the library
- Emulate the executable using Intel SDE on Penryn architecture. 
- SDE outputs error

**Test file** 
```#include <stdlib.h>
#include <stdint.h>

#include "bc7e.h"

#pragma comment(lib, "compress_bc7e.dll.lib")

int main(const char* argv[], int argc)
{
	bc7elib_compress_block_init();

	bc7e_compress_block_params settingsBc7e{};
	bc7elib_compress_block_params_init_basic(&settingsBc7e, true);

	size_t blocksPerRow = 1024;

	size_t bufferSize = blocksPerRow * 4 * 4 * 4;
	char* buffer = (char*)malloc(bufferSize);

	unsigned long long* result = (unsigned long long*) malloc(blocksPerRow * 16);

	for (int i = 0; i < 10; ++i)
	{
		for (int j = 0; j < bufferSize; ++j)
			buffer[j] = rand();


[...]

PR #2464, opened 2023-03-09
Introduced sse4.1 target

aneshlya:
Resolves https://github.com/ispc/ispc/issues/2402


src/ispc.cpp
```diff
@@ -947,6 +948,46 @@ Target::Target(Arch arch, const char *cpu, ISPCTarget ispc_target, bool pic, boo
         this->m_maskBitCount = 32;
         CPUfromISA = CPU_Nehalem;
         break;
+    case ISPCTarget::sse41_i8x16:
```
dbabokin, line 951: That about bundling 4.1 and 4.2 cases and have conditional assignment for `CPUfromISA`

dbabokin (commented):
We also need to update docs.

tests/lit-tests/2402.ispc
```diff
@@ -0,0 +1,24 @@
+// The test checks that ISPC doesn't produce SSE42 instructions like "pcmpgtq" on SSE41 platforms.
+// It also checks that sse4.1 and sse4.2 targets can't be used together in multi-target compilation.
+
+// RUN: %{ispc} %s --target=sse4.1-i8x16 --emit-asm -o - | FileCheck %s --implicit-check-not "pcmpgtq"
```
dbabokin, line 4: I would add `--nostdlib` switch to make it run faster.

dbabokin (approved):
LGTM.

PR #2465, opened 2023-03-13
cmake: strip binaries during packaging

nurmukhametov:


dbabokin:
Well, this look like the correct way to do stripping, except than it doesn't work on macOS as intended :-(

I've created the post on CMake Discourse with macOS problem: https://discourse.cmake.org/t/cpack-strip-doesnt-work-with-llvm-strip-on-macos/7671

Let's see what they reply and go from there.

nurmukhametov:
We may try as workaround something like this, suggesting that macos strip (with '-u' flag in PATH):
```
if (APPLE)
  set(CMAKE_STRIP strip)
endif()
```

dbabokin:
> We may try as workaround something like this, suggesting that macos strip (with '-u' flag in PATH):
> 
> ```
> if (APPLE)
>   set(CMAKE_STRIP strip)
> endif()
> ```

That's the first thing I tried, but `CMAKE_STRIP` is internal option and setting it externally seem to have not effect.

nurmukhametov:
As I see, it does have effect. With `set(CMAKE_STRIP custom_strip_exe)`, I see that in verbose/trace/debug log of `cpack`:
```bash
$ /usr/bin/cpack --trace --debug --verbose --config ./CPackConfig.cmake 2>&1 | grep -i strip
/export/users/anurmukh/applications.compilers.ispc.core/build/CPackConfig.cmake(76):  set(CPACK_STRIP_FILES TRUE )
/export/users/anurmukh/applications.compilers.ispc.core/build/cmake_install.cmake(50):  if(CMAKE_INSTALL_DO_STRIP )
/export/users/anurmukh/applications.compilers.ispc.core/build/cmake_install.cmake(51):  execute_process(COMMAND custom_strip_exe $ENV{DESTDIR}${CMAKE_INSTALL_PREFIX}/bin/ispc )
/export/users/anurmukh/applications.compilers.ispc.core/build/ispcrt/cmake_install.cmake(66):  if(CMAKE_INSTALL_DO_STRIP )
/export/users/anurmukh/applications.compilers.ispc.core/build/ispcrt/cmake_install.cmake(67):  execute_process(COMMAND /usr/bin/strip ${file} )
```
But there is a second strip call for iscprt that is not redefined. You may try to set `CMAKE_STRIP` before `add_subdirectory(ispcrt)`. On my side, it forces both strip command to be redefined.
```bash
$ /usr/bin/cpack --trace --debug --verbose --config ./CPackConfig.cmake 2>&1 | grep -i strip
/export/users/anurmukh/applications.compilers.ispc.core/build/CPackConfig.cmake(76):  set(CPACK_STRIP_FILES TRUE )
/export/users/anurmukh/applications.compilers.ispc.core/build/cmake_install.cmake(50):  if(CMAKE_INSTALL_DO_STRIP )
/export/users/anurmukh/applications.compilers.ispc.core/build/cmake_install.cmake(51):  execute_process(COMMAND custom_strip_exe $ENV{DESTDIR}${CMAKE_INSTALL_PREFIX}/bin/ispc )
/export/users/anurmukh/applications.compilers.ispc.core/build/ispcrt/cmake_install.cmake(66):  if(CMAKE_INSTALL_DO_STRIP )
/export/users/anurmukh/applications.compilers.ispc.core/build/ispcrt/cmake_install.cmake(67):  execute_process(COMMAND custom_strip_exe ${file} )
```

nurmukhametov:
https://gitlab.kitware.com/cmake/cmake/-/blob/91fe2438b4b8839d8ab58ac76fbda2c3d5f46a05/Source/cmInstallTargetGenerator.cxx#L812-L850

According to these code, they add `-u` no matter which `strip` binary found or redefined on macOS.

Another workaround could be is to remove `llvm-strip`.

dbabokin:
It was my error, I tried `CPACK_STRIP`, instead of `CMAKE_STRIP`. The later does work. So please add it PR and we should be good to go. Make sure to mention/link the discussion on CMake Gitlab and the version it's expected to be fixed. This kind of background is valuable for understanding the workarounds to reevaluate it in the future.

nurmukhametov:
I hope it is OK now.

CMakeLists.txt
```diff
@@ -135,6 +135,13 @@ if (APPLE)
         )
         message(STATUS "Using macOS SDK path ${ISPC_MACOS_SDK_PATH}")
     endif()
+
+    # Workaround CMake/CPack issue https://gitlab.kitware.com/cmake/cmake/-/issues/24601
+    # When llvm toolchain is presented in PATH, llvm-strip is preferred over
+    # MacOS strip. However, CPack fails to correctly call llvm-strip because it
```
dbabokin, line 141: Correct spelling is "macOS" - first letter in not capital.

CMakeLists.txt
```diff
@@ -135,6 +135,13 @@ if (APPLE)
...
+    # adds flags like `-u` that are MacOS strip specific.
+    # It may be removed later when CMake fixes ambiguity.
```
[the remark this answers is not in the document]
nurmukhametov: done

CMakeLists.txt
```diff
@@ -135,6 +135,15 @@ if (APPLE)
         )
         message(STATUS "Using macOS SDK path ${ISPC_MACOS_SDK_PATH}")
     endif()
+
+    # Workaround CMake/CPack issue https://gitlab.kitware.com/cmake/cmake/-/issues/24601
+    # When llvm toolchain is presented in PATH, llvm-strip is preferred over
+    # macOS strip. However, CPack fails to correctly call llvm-strip because it
+    # adds flags like `-u` that are MacOS strip specific.
```
dbabokin, line 142: macOS

dbabokin (commented):
LGTM, but I need to carefully check that everything works as expected on may Mac. Will do that later today.

CMakeLists.txt
nurmukhametov: Why does it matter? Especially minding how many places in code already misspelled that.

```bash
$ grep -RnI MacOS ./
./benchmarks/vendor/google/benchmark/src/timers.cc:128:  // FIXME We want to use clock_gettime, but its not available in MacOS 10.11. See
./benchmarks/vendor/google/benchmark/src/timers.cc:152:  // FIXME We want to use clock_gettime, but its not available in MacOS 10.11. See
./builtins/util.m4:6378:;; They are different for Windows and Unix (Linux/MacOS),
./builtins/util-xe.m4:5118:;; They are different for Windows and Unix (Linux/MacOS),
./CMakeLists.txt:89:            message (FATAL_ERROR "Set ISPC_MACOS_SDK_PATH variable for cross compilation to MacOS e.g. C:/iusers/MacOSX10.14.sdk")
./CMakeLists.txt:113:            message (FATAL_ERROR "Set ISPC_MACOS_SDK_PATH variable for cross compilation to MacOS e.g. /iusers/MacOSX10.14.sdk")
./CMakeLists.txt:176:    message(STATUS "MacOS_SDK version: ${SDK_VER}")
./CMakeLists.txt:179:        message(WARNING "MacOS SDK version was not detected, assuming 11.0 or later, enabling ARM support")
./CMakeLists.txt:182:        message(STATUS "MacOS_SDK supports ARM (SDK ver >= 11.0)")
./CMakeLists.txt:185:        message(STATUS "MacOS_SDK does NOT supports ARM (SDK ver < 11.0)")
./docker/v1.14.0/centos7/Dockerfile:58:# MacOSX10.14.sdk to enable cross compilation to macOS.
./docker/v1.14.0/centos7/Dockerfile:59:COPY MacOSX10.14.sdk /usr/local/
./src/ispc.cpp:1982:        triple.setOS(llvm::Triple::OSType::MacOSX);
./docs/ReleaseNotes.txt:694:The release includes LLVM 3.6.1 binaries for Linux, MacOS, Windows and Windows based
./docs/ReleaseNotes.txt:699:for Linux, MacOS and Windows.
./docs/ReleaseNotes.txt:716:Similar to 1.8.0 all binaries are based on LLVM 3.5. MacOS binaries are built
./docs/ReleaseNotes.txt:717:for MacOS 10.9 Mavericks. Linux binaries are compatible with kernel 2.6.32
./docs/ReleaseNotes.txt:728:Note that MacOS binaries are build for MacOS 10.9 Mavericks. Linux binaries are
./docs/ReleaseNotes.txt:734:  on Linux only at the moment. MacOS and Windows users willing to experiment
./docs/ReleaseNotes.txt:802:stability. Linux and MacOS binaries are based on patched version of LLVM 3.3,
./docs/ReleaseNotes.txt:830:  (except MacOS 10.9, but we know about the problem and working on it).
./docs/ReleaseNotes.txt:919:Unix platforms (Linux and MacOS).
./docs/ReleaseNotes.txt:940:improvements on all supported platforms (Windows, Linux and MacOS).
./ispcrt/tests/vendor/google/googletest/googletest/include/gtest/internal/gtest-string.h:137:  // On MacOS X, it uses towlower, which also uses LC_CTYPE category of the
./ispcrt/tests/vendor/google/googletest/googletest/src/gtest.cc:2119:// On MacOS X, it uses towlower, which also uses LC_CTYPE category of the
./alloy.py:198:    # Starting from MacOS 10.9 Maverics, C and C++ library headers are part of the SDK, not the OS itself.
./alloy.py:216:            alloy_error("Can't find XCode (xcrun tool) - it's required on MacOS 10.9 and newer", 1)
./alloy.py:221:        # Starting with MacOS 10.9 Maverics, the system doesn't contain headers for standard C++ library and
```

dbabokin: Correct spelling is "macOS". And it's better to stick to it. Also some of the uses that you point to, are actually correct. They changes from "MacOS" to "macOS" at some point, so it's correct for old versions, not the new.

dbabokin: I've fixed that myself, I'll merge the PR now. Thanks!

PR #2469, opened 2023-03-22
cmake: look only for llvm-config not LLVM package

### LINKED ISSUE #2318: LLVM is found in suboptimal way
[Here](https://github.com/ispc/ispc/blob/main/cmake/FindLLVM.cmake#L36) the cmake script first looks for the LLVM package, and then for the llvm-config executable.

When multiple LLVM versions are installed the above 2 ways of finding LLVM could return different versions after which there would be unpredictable behavior.

I think LLVM is meant to be found with one of the 2 methods - either with find_package or with llvm-config executable, but not both.

Ideally the path to the ```llvm-config``` executable should be passed through a cmake variable.

nurmukhametov:
It may be useful when only minimal set of dependencies presented in PATHs not the whole LLVM installed.

Fix #2318

### DIFF
--- a/cmake/FindLLVM.cmake
+++ b/cmake/FindLLVM.cmake
@@ -33,20 +33,24 @@
 #
 # ispc FindLLVM.cmake
 #
-find_package(LLVM REQUIRED CONFIG)
-    if (NOT LLVM_FOUND )
-        message(FATAL_ERROR "LLVM package can't be found. \
-                Set CMAKE_PREFIX_PATH variable to LLVM's installation prefix.")
-    endif()
-    set(LLVM_VERSION "LLVM_${LLVM_VERSION_MAJOR}_${LLVM_VERSION_MINOR}")
-    message(STATUS "Found LLVM ${LLVM_VERSION}")
-
-find_program(LLVM_CONFIG_EXECUTABLE NAMES llvm-config
-    PATHS ${LLVM_TOOLS_BINARY_DIR} PATH_SUFFIXES bin NO_DEFAULT_PATH)
+if (NOT LLVM_CONFIG_EXECUTABLE)
+    find_program(LLVM_CONFIG_EXECUTABLE NAMES llvm-config)
     if (NOT LLVM_CONFIG_EXECUTABLE)
         message(FATAL_ERROR "Failed to find llvm-config")
     endif()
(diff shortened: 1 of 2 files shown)


aneshlya:
ChooseMSVCCRT.cmake is missing on Windows.

nurmukhametov:
> ChooseMSVCCRT.cmake is missing on Windows.

It should be fixed now. That also reveals some inconsistency, e.g., manually defined LLVM_DIR https://github.com/ispc/ispc/blob/602ff25f32ad69fdc0b2d86545c0f7c1b09e82bf/docker/v1.19.0/centos7/Dockerfile#L76

cmake/FindLLVM.cmake
```diff
@@ -33,20 +33,24 @@
...
+message(STATUS "LLVM_CONFIG_EXECUTABLE: ${LLVM_CONFIG_EXECUTABLE}")
+# It is better to use cmake_path here if we are OK to raise CMake version up to 3.20.
+# cmake_path(GET LLVM_CONFIG_EXECUTABLE PARENT_PATH LLVM_TOOLS_BINARY_DIR)
+get_filename_component(LLVM_TOOLS_BINARY_DIR "${LLVM_CONFIG_EXECUTABLE}" DIRECTORY)
+message(STATUS "LLVM_TOOLS_BINARY_DIR: ${LLVM_TOOLS_BINARY_DIR}")
+
+if (NOT LLVM_DIR)
+    # It is better to use cmake_path here if we are OK to raise CMake version up to 3.20.
+    # cmake_path(SET LLVM_DIR NORMALIZE "${LLVM_TOOLS_BINARY_DIR}/../lib/cmake/llvm")
+    set(LLVM_DIR "${LLVM_TOOLS_BINARY_DIR}/../lib/cmake/llvm")
```
aneshlya, line 51: Please add the check that this directory exists. I think that minimized LLVM installation may not have Cmake files and setting `LLVM_DIR` to non-existing directory may cause lots of problems.

nurmukhametov: I don't think it is really needed because we already indirectly check this directory existence on demand here: https://github.com/ispc/ispc/blob/602ff25f32ad69fdc0b2d86545c0f7c1b09e82bf/cmake/FindLLVM.cmake#L126

PR #2474, opened 2023-03-28
Use correct attribute to retain frame pointers

### LINKED ISSUE #2473: ISPC corrupt callstack vs '--no-omit-frame-pointer' option
Hi !
As reported here:
https://github.com/ispc/ispc/issues/2345

code generated by ISPC produces 'corrupt' callstacks which affect debugging and profiling. I was experimenting
with --no-omit-frame-pointer option and I found out, that it does not affect generated code at all.
After some googling, I found this: 
https://reviews.llvm.org/D71863

and it is mentioned there, that 

"function attributes ``no-frame-pointer-elim`` and  ``no-frame-pointer-elim-non-leaf`` have been replaced by ``frame-pointer``"

it seems to be the reason why there is no change in generated code and I hope that it has potential to fix callstacks
as well :-)

Would it be possible to use 'frame-pointer' instead of 'no-frame-pointer-elim' ? I wanted to try it, but I am struggling with
building ISPC from sources on Windows.

Thanks !

dbabokin:
Fixes #2473.

Note sure how to write good lit test to check that it actually works.

### DIFF
--- a/src/func.cpp
+++ b/src/func.cpp
@@ -283,7 +283,7 @@ void Function::emitCode(FunctionEmitContext *ctx, llvm::Function *function, Sour
     ctx->EmitVariableDebugInfo(maskSymbol);
 
     if (g->NoOmitFramePointer)
-        function->addFnAttr("no-frame-pointer-elim", "true");
+        function->addFnAttr("frame-pointer", "all");
     if (g->target->getArch() == Arch::wasm32)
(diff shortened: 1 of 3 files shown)


dbabokin:
The changes are correct changes, though they are not helping #2473. Need to come up with a way to write a test and merge.

nurmukhametov (approved):
LGTM!

PR #2477, opened 2023-04-04
Refactoring: NULL => nullptr

dbabokin:
`NULL` is a legacy constant from C that lacks proper typing, while `nullptr` is a typed alternative introduced in C++. Using `nullptr` is generally preferred over `NULL` in modern C++ code. Mixing both can create confusion for contributors and result in less readable code. This PR aims to standardize the use of `nullptr` by replacing all occurrences of `NULL` with `nullptr` in the `src`, `examples` (excluding C files), and `ispcrt` directories.

The changes in `src` folder is result of `sed` command followed by `clang-format` (except `parse.yy` and `lex.ll`). `examples` and `ispcrt` were changed manually.

Update:
In the source some of the changes were reverted with following policy and rational:
- all C++ `nullptr` remain - i.e. what is actually parsed by C++ as `nullptr`, but not comments or part of string literal;
- comments also use `nullptr` for the sake of consistency with the source code, even though in some cases it look a bit strange;
- string literals used for generating valid user errors keep using `NULL` - the intent is not to change use output
- string literals used for generating ICE, i.e. FATAL("...") message use `nullptr` - there's only one instance of such message and it looks good to me.
- string literals used in AST dump, i.e. `Print()` function keep using `NULL` as they indicate missing field and it's useful to have all-caps word to draw attention.

### DIFF
--- a/src/decl.cpp
+++ b/src/decl.cpp
@@ -48,8 +48,8 @@ static void lPrintTypeQualifiers(int typeQualifiers) {
     the type, returning the type that is the result.
 */
 static const Type *lApplyTypeQualifiers(int typeQualifiers, const Type *type, SourcePos pos) {
-    if (type == NULL)
-        return NULL;
+    if (type == nullptr)
+        return nullptr;
 
     if ((typeQualifiers & TYPEQUAL_CONST) != 0) {
         type = type->GetAsConstType();
@@ -80,7 +80,7 @@ static const Type *lApplyTypeQualifiers(int typeQualifiers, const Type *type, So
                        "qualifiers.");
 
         const Type *unsignedType = type->GetAsUnsignedType();
-        if (unsignedType != NULL)
+        if (unsignedType != nullptr)
             type = unsignedType;
         else {
             const Type *resolvedType = type->ResolveUnboundVariability(Variability::Varying);
@@ -108,7 +108,7 @@ DeclSpecs::DeclSpecs(const Type *t, StorageClass sc, int tq) {
     typeQualifiers = tq;
     soaWidth = 0;
     vectorSize = 0;
-    if (t != NULL) {
+    if (t != nullptr) {
         if (m->symbolTable->ContainsType(t)) {
             // Typedefs might have uniform/varying qualifiers inside.
             if (t->IsVaryingType()) {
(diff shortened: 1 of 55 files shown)


src/expr.cpp
```diff
@@ -1307,7 +1307,7 @@ UnaryExpr *UnaryExpr::Instantiate(TemplateInstantiation &templInst) const {
 
 void UnaryExpr::Print(Indent &indent) const {
     if (!expr || !GetType()) {
-        indent.Print("UnaryExpr: <NULL EXPR>\n");
```
nurmukhametov, line 1310: It looks like a reference to NULL value of ISPC language, isn't it?

src/decl.cpp
```diff
@@ -143,32 +143,32 @@ const Type *DeclSpecs::GetBaseType(SourcePos pos) const {
...
         const StructType *st = CastType<StructType>(retType);
 
-        if (st == NULL) {
+        if (st == nullptr) {
             Error(pos,
                   "Illegal to provide soa<%d> qualifier with non-struct "
                   "type \"%s\".",
-                  soaWidth, retType ? retType->GetString().c_str() : "NULL");
-            return NULL;
+                  soaWidth, retType ? retType->GetString().c_str() : "nullptr");
```
aneshlya, line 150: Here we should have NULL as a ISPC language specifier

src/stmt.cpp
```diff
@@ -2269,30 +2269,30 @@ void ForeachActiveStmt::Print(Indent &indent) const {
...
     indent.Print();
-    if (sym != NULL) {
+    if (sym != nullptr) {
         printf("%s", sym->name.c_str());
-        if (sym->type != NULL)
+        if (sym->type != nullptr)
             printf(" %s", sym->type->GetString().c_str());
     } else {
-        printf("NULL");
+        printf("nullptr");
```
aneshlya, line 2277: Here and in some other Print cases below, should we keep NULL as well?

src/expr.cpp
dbabokin: It refers to invalid expression (NULL pointer instead of an expression itself or invalid type).

I'm not sure if we should change it or not, seems that keeping at as is is good option.

src/decl.cpp
dbabokin: It doesn't refer to ISPC NULL, but rather a placeholder for a missing `retType`, which might be the case when recovering from an error. But I agree that we should not change error messages.

src/stmt.cpp
dbabokin: These are AST dump output / debug prints. Both are fine, but probably `NULL` is better as it's all capital word, which sticks out in the output drawing attention to it (and this is good, as it's an error typically).

src/expr.cpp
dbabokin: I've updated the PR and summarized what I did in "update" section of initial description.

src/decl.cpp
dbabokin: I've updated the PR and summarized what I did in "update" section of initial description.

nurmukhametov (approved):

aneshlya (approved):

PR #2478, opened 2023-04-04
WIP

nurmukhametov:


ispcrt/detail/gpu/GPUDevice.h
```diff
@@ -53,3 +53,11 @@ struct GPUDevice : public base::Device {
 };
 
 } // namespace ispcrt
+
+// Expose API of GPU device solib for dlsym.
+extern "C" {
```
aneshlya, line 58: Why not all methods which are defined as extern "C" in .cpp file are declared here?

ispcrt/ispcrt.cpp
```diff
@@ -78,6 +87,321 @@ static OBJECT_T &referenceFromHandle(HANDLE_T handle) {
...
+#define ISPCRT_DEVICE_CPU_SOLIB_FULL_VERSION_NAME ISPCRT_DEVICE_CPU_SOLIB_NAME "." ISPCRT_VERSION_FULL
+#define ISPCRT_DEVICE_GPU_SOLIB_FULL_VERSION_NAME ISPCRT_DEVICE_GPU_SOLIB_NAME "." ISPCRT_VERSION_FULL
+#endif
+
+// OS agnostic function to dynamically load a shared library.
+void *dyn_load_lib(const char *name, const char *name_major_version, const char *name_full_version) {
+#if defined(_WIN32) || defined(_WIN64)
+    return LoadLibrary(name);
+#else
+    void *handle = dlopen(name_full_version, RTLD_NOW | RTLD_LOCAL);
```
aneshlya, line 134: It's clear what you're doing here, but still would be good to add some comments explaining why do you search for libraries in this order.

aneshlya (changes requested):
The approach looks good!

ispcrt/CMakeLists.txt
```diff
@@ -52,158 +52,103 @@ if (ISPCRT_BUILD_GPU)
...
+    NAMELINK_ONLY
+  RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+)
 
-if (WIN32)
-    set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
-endif()
-build_ispcrt(SHARED ${PROJECT_NAME})
 if (WIN32)
```
dbabokin, line 148: Why spreading ON/OFF pair for `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` between this CMakeList and includes?

dbabokin:
I'm not a CMake expert, so I rely on Arina to review this.

ispcrt/detail/gpu/GPUDevice.h
nurmukhametov: This is a device header, two extra methods are in the context header. GPUDevice.cpp contains both context and device functions.

nurmukhametov:
> The approach looks good!

I forgot to mention it in the first place, but this change drops `ispcrt` static library building (`ISPCRT_BUILD_STATIC`) at the current state.

ispcrt/detail/gpu/GPUDevice.h
aneshlya: I see now, thanks

ispcrt/CMakeLists.txt
```diff
@@ -52,158 +53,179 @@ if (ISPCRT_BUILD_GPU)
...
+  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+    NAMELINK_ONLY
+  RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+)
 
-  if (ISPCRT_BUILD_CPU)
-    target_compile_definitions(${TARGET_NAME} PRIVATE ISPCRT_BUILD_CPU)
-  endif()
+# Static library target.
```
aneshlya, line 150: Could you combine duplicated logic for static and shared library to the macro as it was before?

nurmukhametov: Shared/static target logic become more different than before, e.g. list of files, include directories, an extra define, so I don't thinks the macro would be OK. I may set the common part in one place looping through ${PROJECT_NAME} ${PROJECT_NAME}_static.

aneshlya: I'm asking about combing common part only. Setting of C++ standard, security options, some compile definitions and part of install logic are 100% identical.

nurmukhametov: I have fixed that.

aneshlya (approved):

ispcrt/ispcrt.cpp
```diff
@@ -78,6 +87,408 @@ static OBJECT_T &referenceFromHandle(HANDLE_T handle) {
...
+        return ispc_sync_fptr(handle);
+    }
+    // TODO: this code can be called directly from ISPC code. Not sure how to exit error safely.
+    throw std::runtime_error("Missing ISPCSync symbol");
+}
+}
+#endif
+
+// Auxiliary function to load device CPU solib and initialize tasking API pointers if build with tasking support.
+void *handleCPUDevilceLib() {
```
e3m3, line 200: Typo?

ispcrt/ispcrt.cpp
```diff
@@ -78,6 +87,408 @@ static OBJECT_T &referenceFromHandle(HANDLE_T handle) {
...
+    if (!ispc_sync_fptr) {
+        throw std::runtime_error("Missing ISPCSync_cpu symbol");
+    }
+#endif
+
+    return handle;
+}
+
+// Auxiliary function to load device GPU solib.
+void *handleGPUDevilceLib() {
```
e3m3, line 234: Typo?

ispcrt/ispcrt.cpp
dbabokin: I guess you mean `handleCPUDevilceLib ` => `handleCPUDeviceLib `, right?

nurmukhametov: Yes, it is. I have fixed that.

PR #2481, opened 2023-04-11
Add copyrights to Dockerfiles

aneshlya:


docker/centos/7/xpu_ispc_build/Dockerfile
```diff
@@ -46,6 +46,11 @@ ENV XE_DEPS=/usr/local/deps
 # http://www.intel.com/software/sde and download the latest version,
 # extract it, add to path and set SDE_HOME.
 
+#
```
dbabokin, line 49: Supposed to be in the beginning of the file

docker/centos/8/cpu_ispc_build/Dockerfile
```diff
@@ -1,3 +1,9 @@
+
```
dbabokin, line 1: Empty first line here and many others is really not needed.

To my taste "Copyright" should be on the very first line, but one line with empty comment is ok as well.

docker/centos/7/xpu_ispc_build/Dockerfile
aneshlya: I have no idea how it appeared here :)

dbabokin (approved):
LGTM!

PR #2483, opened 2023-04-14
ispcrt: Expose device type

e3m3:
Changes:
*   Added `getType` to `Device` interface. 
    | `ispcrt/detail/Device.h`,
    | `ispcrt/detail/cpu/CPUDevice.cpp`,
    | `ispcrt/detail/cpu/CPUDevice.h`,
    | `ispcrt/detail/gpu/GPUDevice.cpp`,
    | `ispcrt/detail/gpu/GPUDevice.h`,
    | `ispcrt/ispcrt.cpp`,
    | `ispcrt/ispcrt.h`,
    | `ispcrt/ispcrt.hpp`,
    | `iscprt/tests/mock_tests/ispcrt_mock_main.cpp`

### DIFF
--- a/ispcrt/detail/Device.h
+++ b/ispcrt/detail/Device.h
@@ -33,6 +33,8 @@ struct Device : public RefCounted {
     virtual void *deviceNativeHandle() const = 0;
     virtual void *contextNativeHandle() const = 0;
 
+    virtual ISPCRTDeviceType getType() const = 0;
+
     virtual ISPCRTAllocationType getMemAllocType(void* appMemory) const = 0;
 };
 
--- a/ispcrt/ispcrt.hpp
+++ b/ispcrt/ispcrt.hpp
@@ -140,6 +140,7 @@ class Device : public GenericObject<ISPCRTDevice> {
     void* nativePlatformHandle() const;
     void* nativeDeviceHandle() const;
     void* nativeContextHandle() const;
+    ISPCRTDeviceType getType() const;
     // static methods to get information about available devices
     static uint32_t deviceCount(ISPCRTDeviceType type);
     static ISPCRTDeviceInfo deviceInformation(ISPCRTDeviceType type, uint32_t deviceIdx);
@@ -167,6 +168,10 @@ inline Device::Device(const Context &context, ISPCRTGenericHandle nativeDeviceHa
 inline void* Device::nativePlatformHandle() const { return ispcrtPlatformNativeHandle(handle()); }
 inline void* Device::nativeDeviceHandle() const { return ispcrtDeviceNativeHandle(handle()); }
 inline void* Device::nativeContextHandle() const { return ispcrtDeviceContextNativeHandle(handle()); }
+
+inline ISPCRTDeviceType Device::getType() const {
+    return ispcrtGetDeviceType(handle());
+}
 
 inline uint32_t Device::deviceCount(ISPCRTDeviceType type) {
     return ispcrtGetDeviceCount(type);
--- a/ispcrt/detail/cpu/CPUDevice.cpp
+++ b/ispcrt/detail/cpu/CPUDevice.cpp
@@ -294,6 +294,10 @@ void *CPUDevice::deviceNativeHandle() const { return nullptr; }
 
 void *CPUDevice::contextNativeHandle() const { return nullptr; }
 
+ISPCRTDeviceType CPUDevice::getType() const {
+    return ISPCRT_DEVICE_TYPE_CPU;
+}
+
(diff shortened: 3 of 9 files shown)


nurmukhametov:
1. I guess that C API method is needed to be added to `ispcrt.h`.
2. As for me, it is better to add some mock tests to `ispcrt_mock_main.cpp` as well.

e3m3:
> 1. I guess that C API method is needed to be added to `ispcrt.h`.
> 2. As for me, it is better to add some mock tests to `ispcrt_mock_main.cpp` as well.

Addressed in commit https://github.com/ispc/ispc/pull/2483/commits/2573efd9ac52801ae4bc1cd3ebd42e5e060cdd7f.

ispcrt/ispcrt.hpp
```diff
@@ -149,6 +149,7 @@ class Device : public GenericObject<ISPCRTDevice> {
     Module staticLinkModules(ISPCRTModule* modules, const uint32_t num);
     // check memory type
     ISPCRTAllocationType getMemoryAllocType(void *memBuffer);
+    ISPCRTDeviceType getDeviceType() const;
```
aneshlya, line 152: Here and in all other modified files: please move this method closer to other device-related methods. It should not be mixed with memory allocation methods.

dbabokin (changes requested):
Having the same name (`getDeviceType()`) for the function in public interface and in internal implementation seems to be a really bad idea.

We don't have practice for having any prefix or suffix for functions in `ispcrt/detail`, but we should definitely avoid having functions with the same name anyway.

e3m3:
> Having the same name (`getDeviceType()`) for the function in public interface and in internal implementation seems to be a really bad idea.
> 
> We don't have practice for having any prefix or suffix for functions in `ispcrt/detail`, but we should definitely avoid having functions with the same name anyway.

https://github.com/ispc/ispc/pull/2483/files#diff-faf5505f360d13e19e4cc71a7b260e4c0ef1cfd804b022abfaf178cc72f8d260R36

ispcrt/detail/Device.h
```diff
@@ -33,6 +33,8 @@ struct Device : public RefCounted {
     virtual void *deviceNativeHandle() const = 0;
     virtual void *contextNativeHandle() const = 0;
 
+    virtual ISPCRTDeviceType _getDeviceType() const = 0;
```
kraszkow, line 36: Why not just simply "getType()" since this method of Device?

dbabokin: I agree. This would be better much better than leading underscore. My second preference would be `Impl` suffix, but just `getType()` looks much better.

ispcrt/ispcrt.hpp
```diff
@@ -168,6 +169,10 @@ inline void* Device::nativePlatformHandle() const { return ispcrtPlatformNativeH
 inline void* Device::nativeDeviceHandle() const { return ispcrtDeviceNativeHandle(handle()); }
 inline void* Device::nativeContextHandle() const { return ispcrtDeviceContextNativeHandle(handle()); }
 
+inline ISPCRTDeviceType Device::getDeviceType() const {
```
kraszkow, line 172: In C API current name looks good (ispcrtGetDeviceType) but in C++ API you could also skip Device word in the middle:

Device::getType() looks more natural since it's method executed on Device.

kraszkow (changes requested):

ispcrt/ispcrt.hpp
e3m3: Sounds good. Same as before, shall we go with `Impl` to disambiguate with the internal interface in `ispcrt/detail/`?

kraszkow: I don't understand @dbabokin why we can't have getType() method (same name) in these 2 classes:
- ispcrt::base::Device (detail -> implementation)
- ispcrt::Device (API / wrapper)

From my perspective this approach is valid (separation is done using namespace and different classes, method name can be same in both classes).

But if we can't have same name for these 2 classes then I would keep "clean version" in API and add change getType() -> getTypeImpl() for ispcrt::base::Device (which looks awkward to me anyway) but @dbabokin - it's your call.

dbabokin: Sorry, I missed that they are class members. My bad! I was mislead by the name `getDeviceType()`, assuming that they belong to the `ispcrt` namespace, but not to the class. Then `getType()` in both cases are absolutely valid name and I think we should go with it.

e3m3:
Ready.

dbabokin (approved):
LGTM

nurmukhametov (approved):
LGTM!

Issue #2482, opened 2023-04-14
1.19.0 (and before): Test files are being installed


dvzrv:
Hi! I'm currently upgrading the package for ispc on Arch Linux to 1.19.0.

I noticed that we force remove a bunch of files after calling the install target and I believe they should not be installed at all.

```
build/
build/ispc/
build/ispc/src/
build/ispc/src/build/
build/ispc/src/build/tests/
build/ispc/src/build/tests/ispcrt_mock_tests
build/ispc/src/build/tests/libgtest.so
build/ispc/src/build/tests/libgtest.so.1.11.0
build/ispc/src/build/tests/libgtest_main.so
build/ispc/src/build/tests/libgtest_main.so.1.11.0
build/ispc/src/build/tests/libze_null.so
build/ispc/src/build/tests/libze_null.so.1
build/ispc/src/build/tests/libze_null.so.1.0.0
```

The installation of test files is not wanted (especially not in a non-FHS compliant directory structure).

dvzrv:
Trying to build ospray I just noticed that it requires that static lib (will amend the initial ticket).

nurmukhametov:
You may add `-DISPCRT_BUILD_TESTS=OFF` to avoid building and installing `ispcrt_mock_tests` and auxiliary libs.

As I see, the issue of incorrect directory structure caused by using `CMAKE_BINARY_DIR` in three places:
 1. https://github.com/ispc/ispc/blob/06ff7c4b5a078b5a1bd76ad08f28903a765022ab/ispcrt/tests/CMakeLists.txt#L58
 2. https://github.com/ispc/ispc/blob/06ff7c4b5a078b5a1bd76ad08f28903a765022ab/ispcrt/tests/mock_tests/CMakeLists.txt#L17
 3. https://github.com/ispc/ispc/blob/06ff7c4b5a078b5a1bd76ad08f28903a765022ab/ispcrt/tests/level_zero_mock/CMakeLists.txt#L25

PR #2484, opened 2023-04-14
Fix ispcrt tests installation

nurmukhametov:
Do not install ispcrt tests into build directory

Also add ispcrt-tests component target which install ispcrt_mock_tests with dependent libraries. This component install target is excluded from ALL target. It may be installed by command:

    cmake --install <dir> --component ispcrt-tests


nurmukhametov:
It fixes #2482

aneshlya:
If we enable tests as part of the build, we should install them.
The option to disable tests is `ISPCRT_BUILD_TESTS`. It can be OFF by default. But if it's ON, I prefer not to run additional commands to install tests which are already built.

pbrubaker:
Hey all, where do we stand on this issue?

nurmukhametov:
> Hey all, where do we stand on this issue?

My opinion is that there is no point in installing tests anywhere. Even if we do so, we need to install them at least under `CMAKE_INSTALL_PREFIX`, not inside build directory.

aneshlya (approved):

PR #2485, opened 2023-04-14
builtins: Refactor mangling utilities

e3m3:
Changes:
*   Move name mangling out of opt to common builtins header.
    | `src/builtins.cpp`,
    | `src/builtins.h`,
    | `src/opt/MangleOpenCLBuiltins.cpp`,
    | `src/opt/MangleOpenCLBuiltins.h`

Description:
>When adding functions declarations to ISPC modules via `AddFunctionDeclaration`, names are expected to be unique.
>If the the same function is registered more than once, the resulting declaration is renamed with a numeric suffix (e.g. `foo.1`).
>For the specialization constants feature, we need declarations for each `constspec` type we would like to initialize.
>However, the base name for the builtin for all types is `__spirv_specConstant`.
>Adding declarations with the appropriate signature does not prevent the mangling from attempting to disambiguate the function names.
>Moving the mangling utilities out of the optimization in which they are used allows for us to mangle the name manually so that the correct names can be used when emitting IR.

### DIFF
--- a/src/builtins.h
+++ b/src/builtins.h
@@ -38,4 +38,8 @@ void AddBitcodeToModule(const BitcodeLib *lib, llvm::Module *module, SymbolTable
  */
 Symbol *CreateISPCSymbolForLLVMIntrinsic(llvm::Function *func, SymbolTable *symbolTable);
 
+#ifdef ISPC_XE_ENABLED
+std::string mangleSPIRVBuiltin(const llvm::Function &func);
+#endif
+
 } // namespace ispc
--- a/src/opt/MangleOpenCLBuiltins.cpp
+++ b/src/opt/MangleOpenCLBuiltins.cpp
@@ -78,7 +78,11 @@ static std::string mangleOCLBuiltin(const llvm::Function &func) {
 std::string mangleSPIRVBuiltin(const llvm::Function &func) {
     Assert(func.getName().startswith("__spirv_") && "wrong argument: spirv builtin is expected");
     std::string mangledName;
-    mangleOpenClBuiltin(func.getName().str(), func.getArg(0)->getType(),
+    std::vector<llvm::Type *> tyArgs;
+    for (const auto &arg : func.args()) {
+        tyArgs.push_back(arg.getType());
+    }
+    mangleOpenClBuiltin(func.getName().str(), tyArgs,
 #if ISPC_LLVM_VERSION == ISPC_LLVM_15_0
                         // spirv builtins doesn't have pointer arguments
                         {},
--- a/src/opt/MangleOpenCLBuiltins.cpp
+++ b/src/opt/MangleOpenCLBuiltins.cpp
@@ -75,7 +75,7 @@ static std::string mangleOCLBuiltin(const llvm::Function &func) {
     return mangleMathOCLBuiltin(func);
 }
 
-static std::string mangleSPIRVBuiltin(const llvm::Function &func) {
+std::string mangleSPIRVBuiltin(const llvm::Function &func) {
     Assert(func.getName().startswith("__spirv_") && "wrong argument: spirv builtin is expected");
     std::string mangledName;
     mangleOpenClBuiltin(func.getName().str(), func.getArg(0)->getType(),


src/builtins.h
```diff
@@ -38,4 +38,14 @@ void AddBitcodeToModule(const BitcodeLib *lib, llvm::Module *module, SymbolTable
  */
 Symbol *CreateISPCSymbolForLLVMIntrinsic(llvm::Function *func, SymbolTable *symbolTable);
 
+#ifdef ISPC_XE_ENABLED
+std::string mangleMathOCLBuiltin(const llvm::Function &func);
+
+std::string manglePrintfOCLBuiltin(const llvm::Function &func);
+
+std::string mangleOCLBuiltin(const llvm::Function &func);
```
aneshlya, line 46: You only need `mangleOCLBuiltin` and `mangleSPIRVBuiltin` as external methods. `mangleMathOCLBuiltin` and `manglePrintfOCLBuiltin` can be left as static and removed from header file.

src/opt/MangleOpenCLBuiltins.cpp
```diff
@@ -99,9 +25,9 @@ bool MangleOpenCLBuiltins::mangleOpenCLBuiltins(llvm::BasicBlock &bb) {
             if (func->getName().startswith("__spirv_")) {
                 std::string mangledName;
                 if (func->getName().startswith("__spirv_ocl")) {
-                    mangledName = mangleOCLBuiltin(*func);
+                    mangledName = ispc::mangleOCLBuiltin(*func);
```
aneshlya, line 28: I think that specifying `ispc::` namespace is not needed here.

aneshlya:
Commits should be reorganized: first is "Move SPIR-V name mangling to builtins", second "Add mangling support for __spirv_SpecConstant"

src/builtins.cpp
```diff
@@ -1265,3 +1266,99 @@ void ispc::DefineStdlib(SymbolTable *symbolTable, llvm::LLVMContext *ctx, llvm::
...
+
+std::string ispc::mangleSPIRVBuiltin(const llvm::Function &func) {
+    Assert(func.getName().startswith("__spirv_") && "wrong argument: spirv builtin is expected");
+    const std::string fnName = func.getName().str();
+    std::string mangledName;
+
+    if (func.getName().startswith("__spirv_SpecConstant")) {
+        constexpr std::size_t MAX_ARGS{1024UL};
+        std::size_t i = 0;
+        llvm::SmallVector<llvm::Type *, MAX_ARGS> tyArgs;
```
aneshlya, line 1342: Why not to simply use `std::vector<llvm::Type *>tyArgs` instead of `llvm::SmallVector<llvm::Type *, MAX_ARGS>`?

src/builtins.cpp
```diff
@@ -1265,3 +1266,99 @@ void ispc::DefineStdlib(SymbolTable *symbolTable, llvm::LLVMContext *ctx, llvm::
...
+        return manglePrintfOCLBuiltin(func);
+    return mangleMathOCLBuiltin(func);
+}
+
+std::string ispc::mangleSPIRVBuiltin(const llvm::Function &func) {
+    Assert(func.getName().startswith("__spirv_") && "wrong argument: spirv builtin is expected");
+    const std::string fnName = func.getName().str();
+    std::string mangledName;
+
+    if (func.getName().startswith("__spirv_SpecConstant")) {
```
aneshlya, line 1339: Is it possible for __spirv_SpecConstant to have a pointer as parameter? If yes, take a look on this comment: https://github.com/ispc/ispc/pull/2485/files#diff-a53310a2317b61d1b121f39452c09f66d9a7ff0642ba1710f2c8f3fe4233a7deR1294-R1297

src/builtins.cpp
```diff
@@ -1265,3 +1266,99 @@ void ispc::DefineStdlib(SymbolTable *symbolTable, llvm::LLVMContext *ctx, llvm::
...
+        constexpr std::size_t MAX_ARGS{1024UL};
+        std::size_t i = 0;
+        llvm::SmallVector<llvm::Type *, MAX_ARGS> tyArgs;
+        for (const auto &arg : func.args()) {
+            tyArgs.push_back(arg.getType());
+            i++;
+        }
+        assert(i <= MAX_ARGS && "Unexpectedly large number of arguments to mangle __spirv_SpecConstant");
```
aneshlya, line 1347: Please use ispc `Assert` or `AssertPos` instead of generic `assert`.

src/builtins.cpp
e3m3: Not from what I've seen so far.

dbabokin:
What is the reason for moving this functionality around? Would be good to spell it out in PR description.

e3m3:
> What is the reason for moving this functionality around? Would be good to spell it out in PR description.

Updated: https://github.com/ispc/ispc/pull/2485#issue-1668583540

src/builtins.cpp
```diff
@@ -1265,3 +1266,32 @@ void ispc::DefineStdlib(SymbolTable *symbolTable, llvm::LLVMContext *ctx, llvm::
...
+            tyArgs.push_back(arg.getType());
+        }
+        mangleOpenClBuiltin(fnName, tyArgs,
+#if ISPC_LLVM_VERSION == ISPC_LLVM_15_0
+                            // spirv builtins doesn't have pointer arguments
+                            {},
+#endif
+                            mangledName);
+    } else {
+        mangleOpenClBuiltin(fnName, func.getArg(0)->getType(),
```
dbabokin, line 1288: Please figure out when this code triggers with more than one argument and why we pass only the first one. This code looks suspicious.

nurmukhametov (commented):
I think that it is better to leave `mangleSPIRVBuiltin` in `MangleOppenCLBuiltins.cpp` file. I suppose that removing `static` and exposing declaration in `builtins.h` is enough to use it in `module.cpp`.

src/builtins.cpp
e3m3: Failing lit tests with `Assert(func.arg_size() > 1 && "Too many arguments to SPIRV builtin");`:
```text
Failed Tests (19):
  ispc :: assume_uniform_xe.ispc
  ispc :: load_store_vectorizer_loopunroll.ispc
  ispc :: print_errors_xe.ispc
  ispc :: xe_assert.ispc
  ispc :: xe_barrier.ispc
  ispc :: xe_diarglist.ispc
  ispc :: xe_externC.ispc
  ispc :: xe_gather_coalescing_loop.ispc
  ispc :: xe_half_support.ispc
  ispc :: xe_kernels_spec.ispc
  ispc :: xe_kernels_spec_addrspace.ispc
  ispc :: xe_prefetchw.ispc
  ispc :: xe_regcall.ispc
  ispc :: xe_short_circuit.ispc
  ispc :: xe_spirv_ocl_mangle.ispc
  ispc :: xe_task_count.ispc
  ispc :: xe_yarpgen_3330730858.ispc
  ispc :: xe_yarpgen_4096674170.ispc
  ispc :: xehpc_vec_add.ispc
```

e3m3: All are failing at mangling function `__spirv_BuiltInWorkgroupId`.

e3m3: The only uses only have one argument, however:
```cpp
builtins/target-xe.ll:declare i64 @__spirv_BuiltInWorkgroupId(i32 %dim)
builtins/target-xe.ll:  %gr_id_x_64 = call i64 @__spirv_BuiltInWorkgroupId(i32 0)
builtins/target-xe.ll:  %gr_id_y_64 = call i64 @__spirv_BuiltInWorkgroupId(i32 1)
builtins/target-xe.ll:  %gr_id_z_64 = call i64 @__spirv_BuiltInWorkgroupId(i32 2)
builtins/target-xe.ll:  %gr_id_64 = call i64 @__spirv_BuiltInWorkgroupId(i32 $1)
```

dbabokin: I'm confused, what are the other arguments?

e3m3: Corrected assertion: `Assert(func.arg_size() <= 1 && "Too many arguments to SPIRV builtin");`

e3m3: No failures for lit tests now, so it looks like it would be safe to merge the cases.

e3m3:
Ready.

aneshlya (approved):

PR #2487, opened 2023-04-18
Enable Codespaces for ISPC

dbabokin:
PR includes a minimal Codespaces container for comfortable browsing, editing, and building ISPC project.

The configuration includes pre-built LLVM 15.0.7.

If it works well, we can add GPU-enabled env, though I'm not sure if we can have Intel GPU-enabled cloud instance.

### DIFF
--- a/.devcontainer/README.md
+++ b/.devcontainer/README.md
@@ -1,0 +1,20 @@
+Codespaces - ISPC developer's environment (CPU only)
+====================================================
+
+This is the minimal environment that enables CPU-only ISPC build (i.e. GPU backend is not enabled).
+
+To build ISPC and run the tests do the following:
+```bash
+# `build` folder already contains result of `cmake .. -DCMAKE_EXPORT_COMPILE_COMMANDS=ON`
+cd build
+# build ISPC and examples
+make -j4
+# run LIT tests
+make check-all
+```
+
+Note that `build` folder already contains `compile_commands.json`, which is needed for VSCode to be able correctly parse and browse C++ files.
+
+The container also contains pre-installed [C/C++ Extension Pack](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cpptools-extension-pack) for VSCode.
+
+Enjoy and please let us know if you use Codespaces to hack ISPC and have any suggestions or feedback: [Github Discussions](https://github.com/ispc/ispc/discussions)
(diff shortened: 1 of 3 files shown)


.devcontainer/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+To build ISPC and run the tests do the following:
+```bash
+# `build` folder already contains result of `cmake .. -DCMAKE_EXPORT_COMPILE_COMMANDS=ON`
+cd build
+# build ISPC and examples
+make -j4
+# run LIT tests
+make check-all
+```
```
e3m3, line 14: The contents of the root directory does not contain `build`:
```shell
@e3m3 ➜ /workspaces/ispc (main) $ ls
CMakeLists.txt   alloy.py            check_env.py     common.py    examples      perf.py       stdlib2cpp.py       tests_errors
CONTRIBUTING.md  benchmarks          check_format.sh  contrib      fail_db.txt   run_tests.py  test_static.cpp     third-party-programs.txt
LICENSE.txt      bitcode2cpp.py      check_isa.cpp    docker       ispcrt        scripts       test_static.isph    utils
README.md        builtins            cmake            docs         llvm_patches  src           test_static_l0.cpp
SECURITY.md      check_copyright.sh  common           doxygen.cfg  perf.ini      stdlib.ispc   tests
```

dbabokin: Check you build logs, container build probably failed. Before it's merged you need to manually set my repo and my branch as a source.

nurmukhametov (commented):
I have tested that it works well. Will it be more useful to ship (LLVM) and pre-configure debug build (ISPC)?

aneshlya (approved):

dbabokin:
> I have tested that it works well. Will it be more useful to ship (LLVM) and pre-configure debug build (ISPC)?

Hmm... This does make sense (when it comes to ISPC build). Shipping debug LLVM build is bad idea and as I understand you, you are not proposing that.

From another prospective, the key use case for Codespaces is to hack the code, test it and commit, not to debug. And building Release is better for that.

I suggest going forward "as is", see how it works and add more containers/configurations when we see that the base scenario works well.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,28 @@
+{
+  "__comment1__": "Copyright (c) 2023, Intel Corporation",
+  "__comment2__": "SPDX-License-Identifier: BSD-3-Clause",
+  "name": "ISPC developers environment (CPU only)",
+  "image": "mcr.microsoft.com/devcontainers/universal:2",
+  "features": {
+    "ghcr.io/devcontainers-contrib/features/apt-get-packages:1": { "packages" : "libc6-dev,libc6-dev-i386,flex,wget" },
```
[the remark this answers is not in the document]
dbabokin: Ok, as `libtbb-dev` is not settled yet, I'll let you add when it, when you are settled on the specific way of adding it.

PR #2492, opened 2023-04-21
Generate vector instructions for uniform vector conversions

aneshlya:
Currently for uniform short vector conversions ISPC generates scalar instructions.
For example for the following case:
```
uniform float<3> x = {1,2,3};
uniform double<3> z = x;
```
ISPC generates 3x`fpext float` instructions.

The PR enables generation of vector instructions for such cases. So instead of 3x`fpext float`, 1x`fpext<4xfloat>` will be produced.
It's used now for conversions only but planned to be extended for short vector initialization as well.

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -6127,9 +6127,11 @@ static llvm::Value *lTypeConvAtomic(FunctionEmitContext *ctx, llvm::Value *exprV
     }
     const char *cOpName = opName.c_str();
 
+    llvm::Type *targetType = fromType->IsUniformType() ? toType->GetAsUniformType()->LLVMType(g->ctx)
+                                                       : toType->GetAsVaryingType()->LLVMType(g->ctx);
+
     switch (toType->basicType) {
     case AtomicType::TYPE_FLOAT16: {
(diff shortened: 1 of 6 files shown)


src/expr.cpp
```diff
@@ -6710,9 +6738,9 @@ static llvm::Value *lUniformValueToVarying(FunctionEmitContext *ctx, llvm::Value
 
     // Otherwise we must have a uniform atomic or pointer type, so smear
     // its value across the vector lanes.
-    if (CastType<AtomicType>(type)) {
-        return lTypeConvAtomic(ctx, value, CastType<AtomicType>(type->GetAsVaryingType()), CastType<AtomicType>(type),
-                               pos);
+    if (CastType<AtomicType>(type) && CastType<AtomicType>(type->GetAsVaryingType())) {
```
dbabokin, line 6741: I understand that this change is to silence the static code checker, but this was unexpected to see.
I'm fine leaving it as is, but we should consider using asserts for such things.

dbabokin (approved):
Looks good!

PR #2493, opened 2023-04-21
gitignore: Add build directory to ignored files

e3m3:
Ignore the build directory. Useful for Codespaces with build included in the root.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,5 +19,5 @@ examples/*/ref
 examples/*/test
 *.swp
 check_isa.exe
-build
+build*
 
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -19,5 +19,5 @@ examples/*/ref
 examples/*/test
 *.swp
 check_isa.exe
-
+build
```
dbabokin, line 22: What about `build*`?

e3m3: Wasn't sure if there would ever be a script in the root starting with `build`. It's probably safe to do `build*`.

dbabokin: If we would add such script, we'll remove the `.gitignore`. Though I routinely do have more than build folder in the workspace.

e3m3:
```suggestion
build*
```

PR #2495, opened 2023-04-25
Enable OpenMP back in alloy.py

aneshlya:
Partial revert of 7b23d8243eb42ac6683d208fdc4108fd27b92d9a.
OpenMP is still needed in some jobs for testing of ISPC Runtime.


dbabokin (approved):
LGTM

PR #2497, opened 2023-04-25
Support custom installation of TBB

aneshlya:
Support custom installation of TBB through `TBB_ROOT` CMake variable


dbabokin (approved):
LGTM

PR #2499, opened 2023-04-26
Enable macOS build on the new OS supporting old OS

### LINKED ISSUE #2461: Enable release macOS builds on new mcOS versions
Right now release version of ISPC on macOS is built on old macOS to ensure compatibility with older macOS versions.

This approach is the most reliable, but it doesn't scale and bad for automation. The proposal is to enable the release build on any macOS version. I'm creating this issue to explore the options and implement solution in `alloy.py` and project CMake.

dbabokin:
Fixes #2461.

To enable the binary running on old macOS, `CMAKE_OSX_DEPLOYMENT_TARGET` needs to be properly set for all components of the binary, so it needs to be done for LLVM and ISPC itself. This PR does this defaulting the target to `10.12` (Sierra) for x86 and `11.0` (Big Sur). `10.12` is pretty old OS, but there seem to be no reasons not to support it. `11.0` is the minimum that supports macOS on ARM.

As a bonus, `CMAKE_EXPORT_COMPILE_COMMANDS` is always enabled now, it makes setting up VSCode browsing easier and it's handy for debugging compilation commands (figuring out what was really passed to the compiler).

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -15,7 +15,7 @@
   },
   "remoteEnv": { "PATH": "/llvm/bin-15.0/bin:${containerEnv:PATH}" },
   "onCreateCommand": {
-    "command1": "cd ${containerWorkspaceFolder} && mkdir build && cd build && cmake .. -DCMAKE_EXPORT_COMPILE_COMMANDS=ON"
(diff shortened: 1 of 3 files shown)


aneshlya (approved):
Looks good!

PR #2500, opened 2023-04-26
Build Universal Binaries on macOS

### LINKED ISSUE #2329: Ship ISPC for macOS as a fat binary for x86+arm64
It would be good to have a M1 native build in the release artifacts. It can be done right now, when building on M1 macOS. But more user friendly solution is to have a fat binary, which works on both platforms.

dbabokin:
Fixes #2329.

- add `--macos-universal-binary` to `alloy.py`, which triggers LLVM build as Universal Binary (x86_64 + arm64)
- add `ISPC_MACOS_UNIVERSAL_BINARIES` to CMake on macOS, which builds ISPC as Universal Binary.
- add arch suffix to package name on macOS, it's either `x64_64`, `arm64`, or `universal`.
- build LLVM as Universal Binary on macOS in CI.
- fix LLVM patch for `15.0`, so it properly applies on Windows in CI.
- run macOS rebuild jobs on `macos-12-xl` runners, which are "big" instances (12 cores) and allow finishing selfbuild jobs within the time limit.

The plan is to release 3 set of binaries `x64_64`, `arm64`, and `universal`.

### DIFF
--- a/.github/workflows/reusable.rebuild.yml
+++ b/.github/workflows/reusable.rebuild.yml
@@ -314,7 +314,7 @@ jobs:
         name: llvmrel_win_x86
         path: llvm/llvm-${{ inputs.full_version }}-win.${{ inputs.vs_version_str }}-Release-x86.arm.wasm.tar.7z
 
-  mac-build-1:
+  mac-build:
     runs-on: macos-11
 
     steps:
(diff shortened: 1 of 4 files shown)


nurmukhametov (approved):
LGTM!

Issue #2147, opened 2021-08-21
ISPC uses scalar SSE instructions to convert floats to doubles when compiling for AVX, AVX2 and AVX512


pbrubaker:
The title pretty much says it all.  ISPC is not doing wide conversions from float to double.  Padding doesn't seem to affect it either.

https://ispc.godbolt.org/z/Th6G5xxbn

JeffRous:
Fake varying can be abused to do this?

https://ispc.godbolt.org/z/j6cW8nfeh

pbrubaker:
Yeah, that does collapse the conversion instructions, but ideally I'd like to see a `vmovmaskps` instead of the inserts.

pbrubaker:
Hey @dbabokin is there any chance we can look at this one for the next release?  Some of the code I'm working on right now is relatively short and does promotion conversions from float to double often and I'm pretty sure this is affecting performance.

pbrubaker:
I ran into this one again with integers today.

```
    uniform float reduce_weight = reduce_add(weight);
    
    uniform float <4> reduce_col;
    reduce_col.x = reduce_add( col.x );
    reduce_col.y = reduce_add( col.y );
    reduce_col.z = reduce_add( col.z );
    reduce_col.w = reduce_add( col.w );

    reduce_col = (reduce_col / reduce_weight) * 255.0f;

    uniform uint32<4> ret = (uint32<4>) reduce_col;
    return ret;  //return value here is uint8<4>
```

```
        vdivps  xmm0, xmm0, xmm1
        vmulps  xmm0, xmm0, dword ptr [rip + .LCPI2_5]{1to4}
        vcvttss2si      eax, xmm0
        vmovshdup       xmm1, xmm0              # xmm1 = xmm0[1,1,3,3]
        vcvttss2si      ecx, xmm1
        vpermilpd       xmm1, xmm0, 1           # xmm1 = xmm0[1,0]
        vcvttss2si      edx, xmm1
        vpermilps       xmm0, xmm0, 255         # xmm0 = xmm0[3,3,3,3]
        vcvttss2si      esi, xmm0
        vmovd   xmm0, eax
        vpinsrb xmm0, xmm0, ecx, 1
        vpinsrb xmm0, xmm0, edx, 2
        vpinsrb xmm0, xmm0, esi, 3
```
[...]

PR #2501, opened 2023-04-27
Use vector conversion for uniform vector initialization

aneshlya:
Try to produce vector type conversion when initializing short vectors.
It's not a partial fix for https://github.com/ispc/ispc/issues/2147. The frontend produces the same code as in https://github.com/ispc/ispc/issues/2147#issuecomment-922027113 with fake varying trick.

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -576,6 +576,8 @@ bool ispc::PossiblyResolveFunctionOverloads(Expr *expr, const Type *type) {
     return true;
 }
 
+static llvm::Value *lTypeConvAtomicOrUniformVector(FunctionEmitContext *ctx, llvm::Value *exprVal, const Type *toType,
+                                                   const Type *fromType, SourcePos pos);
 /** Utility routine that emits code to initialize a symbol given an
     initializer expression.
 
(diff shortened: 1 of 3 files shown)


src/expr.cpp
```diff
@@ -726,14 +728,66 @@ void ispc::InitSymbol(AddressInfo *ptrInfo, const Type *symType, Expr *initExpr,
...
+            // instruction for type conversion if needed. We can generate one instruction for type conversion when all
+            // initializers have the same atomic type. For example: const uniform float<4> f = {1.f, 2.f, 3.f, 4.f};
+            // const uniform double<4> d = {f[1], f[3], f.y, 0.0f};
+            // If it's not the case, or if initializer expression is not trivial (not of AtomicType), initialize vector
+            // per element.
+            if (symType->IsVectorType() && symType->IsUniformType() && nInits > 0) {
+                int nVectorElements = CastType<VectorType>(symType)->getVectorMemoryCount();
+                Expr *firstInit = exprList->exprs[0];
+                // Construct vector instruction only if initializer expression has AtomicType
+                if (CastType<AtomicType>(firstInit->GetType())) {
```
dbabokin, line 741: Interesting observation, `CastType<>` appears to be preferable vs `isAtomic()` and `isVector()`, as it's safe for `nullptr` values.

src/expr.cpp
```diff
@@ -726,14 +728,66 @@ void ispc::InitSymbol(AddressInfo *ptrInfo, const Type *symType, Expr *initExpr,
...
+                    // Get the type of first initializer expression
+                    const AtomicType *firstInitType = CastType<AtomicType>(firstInit->GetType());
```
dbabokin, line 743: Do that before the `if` and just check in the `if` that it's not null?

src/expr.cpp
```diff
@@ -726,14 +728,66 @@ void ispc::InitSymbol(AddressInfo *ptrInfo, const Type *symType, Expr *initExpr,
...
+                    llvm::Type *llvmInitType = firstInitType->LLVMType(g->ctx);
+                    llvm::Value *initCast =
```
dbabokin, line 745: Looks like a memory leak if the vector initializer has failed to generate.

src/expr.cpp
```diff
@@ -726,14 +728,66 @@ void ispc::InitSymbol(AddressInfo *ptrInfo, const Type *symType, Expr *initExpr,
...
+                        // Check if initializer expression is trivial AtomicType
+                        if (CastType<AtomicType>(exprList->exprs[i]->GetType())) {
+                            llvm::Value *initializerValue = exprList->exprs[i]->GetValue(ctx);
+                            if (initializerValue == nullptr) {
+                                isVectorIntializer = false;
+                                break;
+                            }
+                            llvm::Type *initializerType = initializerValue->getType();
+                            // Check if initializer is of the same type as the first one
+                            if (initializerType != llvmInitType) {
```
dbabokin, line 764: `Type::EqualIgnoringConst(const Type *a, const Type *b)` should be a better alternative here.

src/expr.cpp
```diff
@@ -726,14 +728,66 @@ void ispc::InitSymbol(AddressInfo *ptrInfo, const Type *symType, Expr *initExpr,
...
+                                isVectorIntializer = false;
+                                break;
+                            }
+                            initCast = ctx->InsertInst(initCast, initializerValue, i);
```
dbabokin, line 768: If this fails before completing the sequence, we would have a dead code, which is not good. I suggest collecting `Expr`s in the vector before starting the transformation and verifying that we will success. And as a second step doing the actual transformation.

dbabokin (changes requested):
Need to do the transformation in two steps: analysis + code generation.

Also, we need to account for constants of different type. I.e.:
`double<4> d = {f.x, f.y, f.z, 0.0d};`

src/expr.cpp
```diff
@@ -726,14 +728,66 @@ void ispc::InitSymbol(AddressInfo *ptrInfo, const Type *symType, Expr *initExpr,
                 return;
             }
 
+            // When we initialize uniform vector try to construct a vector from initializers and use a vector
+            // instruction for type conversion if needed. We can generate one instruction for type conversion when all
+            // initializers have the same atomic type. For example: const uniform float<4> f = {1.f, 2.f, 3.f, 4.f};
+            // const uniform double<4> d = {f[1], f[3], f.y, 0.0f};
+            // If it's not the case, or if initializer expression is not trivial (not of AtomicType), initialize vector
+            // per element.
+            if (symType->IsVectorType() && symType->IsUniformType() && nInits > 0) {
```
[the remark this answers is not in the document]
aneshlya: No, `nInits` can be less than vector size. 
For example `float<3>` is represented as `<4 x float>` vector. We may have 3 initializers for it but we should construct a vector of width 4.

tests/lit-tests/vector_type_init-1.ispc
```diff
@@ -0,0 +1,51 @@
+// This test checks that uniform vectors conversion during initialization is perfomed using vector instructions.
+// RUN: %{ispc} %s --target=host -O0 --emit-llvm-text -o - | FileCheck %s
+
+// Vector initialization
+// CHECK: define <4 x double> @test1
```
dbabokin, line 5: Please add `CHECK-LABEL`s with function name, like this one:
`// CHECK-LABEL: define {{.*}} @bar1`

dbabokin (approved):
Looks good, but address the comment about the test.

PR #2503, opened 2023-05-01
Add snap package configuration file

nurmukhametov:
1. Add `snapcraft.yml` to package the binary release from `https://github.com/ispc/ispc/releases` to snap.
1. Add CI workflow to build it and publish to Snap Store.
1. Update `README.md` with installation options.


snap/snapcraft.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+grade: stable # must be 'stable' to release into candidate/stable channels
+confinement: strict
+
+parts:
+  ispc:
+    plugin: dump
+    source: https://github.com/ispc/ispc/releases/download/v$SNAPCRAFT_PROJECT_VERSION/ispc-v$SNAPCRAFT_PROJECT_VERSION-linux.tar.gz
+    stage-packages:
+      - libomp-14-dev
```
e3m3, line 31: I assume this means that the build included in snap is for LLVM 14 only?

e3m3 (approved):
LGTM

e3m3:
Btw, would it be useful to also take a look at a flatpak package?

e3m3:
Also, should be have an aarch64 build in addition to the amd64 build?

README.md
```diff
@@ -1,4 +1,5 @@
-[![Appveyor build status (Windows)](https://ci.appveyor.com/api/projects/status/xfllw9vkp3lj4l0v/branch/main?svg=true)](https://ci.appveyor.com/project/ispc/ispc/branch/main)
+[![Appveyor build status (Windows)](https://ci.appveyor.com/api/projects/status/xfllw9vkp3lj4l0v/branch/main?svg=true)](https://ci.appveyor.com/project/ispc/ispc/branch/main) [![ispc](https://snapcraft.io/ispc/badge.svg)](https://snapcraft.io/ispc)
```
dbabokin, line 1: Is the badge for `1.19.0`? If so, then it's confusing to have its badge next to main build status.

I guess we should have a link to it on the release page for 1.20.0 when we release.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,35 @@
+#  Copyright (c) 2023, Intel Corporation
+#  SPDX-License-Identifier: BSD-3-Clause
+
+name: ispc
+base: core22 # ubuntu 22.04
+version: '1.19.0'
+license: BSD-3-Clause
+summary: Compiler for high-performance SIMD programming on the CPU
```
dbabokin, line 8: Why n CPU only? Should we drop "on the CPU" part?

dbabokin (commented):
What is the process to update `snap` package when we do the release?

README.md
nurmukhametov: Yes, it is. OK.

snap/snapcraft.yaml
nurmukhametov: fixed

snap/snapcraft.yaml
nurmukhametov: This yaml file doesn't contain any build process. It just download prebuilt binary release from  `https://github.com/ispc/ispc/releases` and repack it into snap package. In case `1.19.0`, it means that it is `ISPC` prebuild with LLVM 14.

nurmukhametov:
> Also, should we have an aarch64 build in addition to the amd64 build?

If we had aarch64 prebuild binaries then we may probably pack them as well.

nurmukhametov:
> What is the process to update `snap` package when we do the release?

1. Increment version in `snap/snapcraft.yaml`.
2. Run `snapcraft` from `ispc` root directory inside environment suitable for building snapcraft packages (ubuntu).
3. Upload package to Snap Store via `snapcraft upload --release=latest/stable ispc_$VERSION_amd64.snap`

dbabokin:
> > What is the process to update `snap` package when we do the release?
> 
> 1. Increment version in `snap/snapcraft.yaml`.
> 2. Run `snapcraft` from `ispc` root directory inside environment suitable for building snapcraft packages (ubuntu).
> 3. Upload package to Snap Store via `snapcraft upload --release=latest/stable ispc_$VERSION_amd64.snap`

So, we need to increase version number during the release process. Who can upload the package, only the maintainer? Is it username/password auth?

We need to document this process somewhere - in a wiki, README (in `snap` folder), in the Notes on Teams (where we have some release recipes). I'm ok with either place.

e3m3:
> > > What is the process to update `snap` package when we do the release?
> > 
> > 
> > 
> > 1. Increment version in `snap/snapcraft.yaml`.
> > 2. Run `snapcraft` from `ispc` root directory inside environment suitable for building snapcraft packages (ubuntu).
> > 3. Upload package to Snap Store via `snapcraft upload --release=latest/stable ispc_$VERSION_amd64.snap`
> 
> So, we need to increase version number during the release process. Who can upload the package, only the maintainer? Is it username/password auth?
> 
> We need to document this process somewhere - in a wiki, README (in `snap` folder), in the Notes on Teams (where we have some release recipes). I'm ok with either place.

Can this be done as a CI task? Also, should we add a task to CI to pull down the snap package and run the test suite on it?

dbabokin:
> Can this be done as a CI task? Also, should we add a task to CI to pull down the snap package and run the test suite on it?

This not exactly CI, this CD. We don't have much automation for CD, but we may think about. Though it's clearly not the first priority right now.

nurmukhametov:
> Also, should we have an aarch64 build in addition to the amd64 build?

Quite a few Linux distros have aarch64 binaries in their repos, e.g., Alpine, Fedora, FreeBSD, openSUSE.

nurmukhametov:
> So, we need to increase version number during the release process. Who can upload the package, only the maintainer? Is it username/password auth?

Everyone with a token. 

> Can this be done as a CI task? Also, should we add a task to CI to pull down the snap package and run the test suite on it?

Yes. I implemented it here https://github.com/nurmukhametov/ispc/blob/snap/.github/workflows/cd.yml

> We need to document this process somewhere - in a wiki, README (in `snap` folder), in the Notes on Teams (where we have some release recipes). I'm ok with either place.

Let's consider CI job as documentation here. I have updates notes.

e3m3:
> > Btw, would it be useful to also take a look at a flatpak package?
> 
> I am not sure about flatpack. Isn't that for GUI applications?

It's similar to snap but is the default for distros like Fedora. [Docs](https://docs.flatpak.org/en/latest/first-build.html) and [FlatHub](https://flathub.org/).

README.md
```diff
@@ -86,6 +86,44 @@ Features
...
+
+Official release binaries can be downloaded from
+[the release page](https://github.com/ispc/ispc/releases).
+Linux users can install it from the Snap Store:
+
+```bash
+snap install ispc
+```
+
+ISPC is distributes as the pasrt of Intel® oneAPI. It can be installed from
```
dbabokin, line 100:
```suggestion
ISPC is also distributed as the part of Intel® oneAPI. It can be installed from
```

PR #2505, opened 2023-05-02
Remove GCC 4.x.x workaround in dispatch code.

### LINKED ISSUE #2504: check_isa.cpp crashes when compiled in -O0
On macOS, when compiled with `clang++` (for example, version 15.0, but I think it fails the same way with older versions as well) with `-O0`, the `check_isa` crashes:
```bash
> clang++ check_isa.cpp -O0 -o check_isa.exe
> ./check_isa.exe
ISA: AVX2 (codename Haswell)
Segmentation fault: 11
```

The crash happens after `main` is completed at the following instruction:
`movq   0x8(%rbx), %rax`
And it appears that `rbx` is indeed corrupted.

The investigation showed that we use the following code to check cpuid:
```c++
 /* Save %ebx in case it's the PIC register */
 static void __cpuidex(int info[4], int level, int count) {
     __asm__ __volatile__("xchg{l}\t{%%}ebx, %1\n\t"
                          "cpuid\n\t"
                          "xchg{l}\t{%%}ebx, %1\n\t"
                          : "=a"(info[0]), "=r"(info[1]), "=c"(info[2]), "=d"(info[3])
                          : "0"(level), "2"(count));
}
```

Note the comment about PIC register - output registers `eax`/`rax`, `ecx`/`rcx`, and `edx`/`rdx` are specified explicitly (with `a`, `c`, `d`), while `ebx`/`rbx` are not specified explicitly and `r` is used instead (any compiler assigned register). This is done intentionally to preserve `ebx` (note that I'm not saying `rbx`), as this register supposed to hold the pointer to GOT in case of PIC is used in 32 bit mode. Also, generally speaking, in case of 32 bit `cdecl` calling convention, "Registers EAX, ECX, and EDX are caller-saved, and the rest are callee-saved." ([link](https://en.wikipedia.org/wiki/X86_calling_conventions#cdecl)). So the trick with `xchgl` instruction does job and `ebx` is not affected.

In 64 bit mode, according to x86_64 ABI ([link](https://gitlab.com/x86-psABIs/x86-64-ABI/-/jobs/artifacts/master/raw/x86-64-ABI/abi.pdf?job=build)), GOT is stored in `r15`, but `rbx` still needs to be preserved by the callee. But the code breaks `rbx` by "storing" just lower 32 bits.

Note that this is compiler job to save and restore all caller-safe registers in case of inline-asm. And both modern clang and gcc do this correctly by pushing all required registers on stack and popping them after the inline-asm. So why is this hack was needed in the first place? Looks like this is due to a bug in GCC 4.x.x that was leading to compiler error inconsistent asm constrains. Compiler Explorer [link](https://godbolt.org/z/nqfKnE6e4).

The hack needs to be removed, as it's actually harmful, and we don't care about GCC 4.x.x anymore.

dbabokin:
Fix #2504.

This fix removes GCC 4.x.x workaround in dispatch code. It causes problems on 64 bit target, which showed up as `check_isa.cpp` fail when compiled with -O0. It could also potentially affect ISPC output produced with auto-dispatch.

For the complete description of the problem see #2504.

### DIFF
--- a/builtins/dispatch-macos.ll
+++ b/builtins/dispatch-macos.ll
@@ -33,13 +33,8 @@
 ;;                           : "0" (infoType));
 ;; }
 ;;
-;; // Save %ebx in case it's the PIC register.
 ;; static void __cpuid_count(int info[4], int level, int count) {
-;;   __asm__ __volatile__ ("xchg{l}\t{%%}ebx, %1\n\t"
-;;                         "cpuid\n\t"
-;;                         "xchg{l}\t{%%}ebx, %1\n\t"
-;;                         : "=a" (info[0]), "=r" (info[1]), "=c" (info[2]), "=d" (info[3])
-;;                         : "0" (level), "2" (count));
+;;     __asm__ __volatile__("cpuid" : "=a"(info[0]), "=b"(info[1]), "=c"(info[2]), "=d"(info[3]) : "0"(level), "2"(count));
 ;; }
 ;;
 ;; static int __os_has_avx_support() {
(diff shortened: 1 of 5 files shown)


nurmukhametov (approved):
LGTM!

PR #2506, opened 2023-05-04
Dispatcher to distinguish between SSE4.1 and SSE4.2

### LINKED ISSUE #2402: ISPC uses SSE4.2 instructions on SSE4.1 platforms

dbabokin:
Fixes #2402 

This PR extends the fix for #2402 and specifically addresses this [comment](https://github.com/ispc/ispc/issues/2402#issuecomment-1521737876). It enables dispatcher to distinguish between SSE4.1 and SSE4.2. The limitation to have only one of either SSE4.1 and SSE4.2 targets in multi-target compilation remains and it is motivated by desire to have backward compatibility with build systems which expect object files with `sse4` suffix (and it's used for both SSE4.1 and SSE4.2 targets). I also don't think anyone would really need both of these targets for anything other than experimenting.

The PR also introduces `dispatch.c`, which is a source for all 3 versions of dispatch LLVM IR modules that we have.


nurmukhametov:
Is it OK that dispatch code in `lGetSystemISA` and `__get_system_isa` is slightly different. The first one never returns `SPR` target otherwise the second one does it.

nurmukhametov (approved):
LGTM!

PR #2507, opened 2023-05-04
Updates for Github Actions CI

dbabokin:
- `alloy.py` to build LLVM `16.0.3` instead of `16.0.2`
- macOS job to use LLVM Universal Binaries.
- add test run on macOS, instead of just a build.
- limit the number of optsets to run with LLVM 13
- build ISPC as Universal Binary in GHA CI, as this is what we release. We need to verify that it build for both x86_64 and arm64


nurmukhametov (approved):
LGTM!

Issue #2514, opened 2023-05-08
ICE postfix expression has no member


nurmukhametov:
ICE with no error message:
```c
struct D {
    int h;
};
struct D d;
int foo() {
     return d.f;
}
```

```bash
/usr/local/src/ispc/src/main.cpp(412): FATAL ERROR: Unhandled signal sent to process; terminating.
***
*** Please file a bug report at https://github.com/ispc/ispc/issues
*** (Including as much information as you can about how to reproduce this error).
*** You have apparently encountered a bug in the compiler that we'd like to fix!
***
Compiler returned: 139
```

Compiler explorer link: https://godbolt.org/z/csr4qfWcP

dbabokin:
Fixed by #2515

PR #2515, opened 2023-05-08
Fix ICE when postfix expession fetches unknown member

nurmukhametov:
This fixes issue #2514 

```bash
./tests/lit-tests/2514.ispc:11:13: Error: 'varying struct D' has no member named "f"
     return d.f;
            ^^^

./tests/lit-tests/2514.ispc:11:6: Error: Must provide return value for return statement for non-void function.
     return d.f;
     ^^^^^^^^^^^
```


dbabokin (approved):
LGTM

Issue #2036, opened 2021-03-01
#pragma unroll produces different code than unrolling by hand


zzzoom:
After reading the latest release notes, I tried the code in issue #1320 again, hoping for it to get optimized:

```ispc
int32 gang_scan(int32 v) {
    #pragma unroll
    for (uniform int s = 1; s < programCount; s *= 2) {
        v += shift(v, -s);
    }
    return v;
}

int32 gang_scan_manual_unroll(int32 v) {
    v += shift(v, -1);
    v += shift(v, -2);
    v += shift(v, -4);
    return v;
}
```

The asm produced by `#pragma unroll` couldn't optimize the shift the way it did when unrolled by hand.

Output: https://godbolt.org/z/j9o7oq

dbabokin:
This is a tricky case, thanks for bringing it in.

What happens here is the following. `shift()` stdlib function is implemented differently depending whether the shift amount argument is a constant or not. In the manual unroll case it's obviously a constant and the `shift()` function call is got replaced by LLVM `shuffle` instruction (it's done by [ReplaceStdlibShiftPass](https://github.com/ispc/ispc/blob/main/src/opt/ReplaceStdlibShiftPass.h)). In case of loop unroll, `ReplaceStdlibShiftPass` is run before unrolling and hence it's unable to catch that it's a complete unroll and arguments are constant. So we have a problem.

So we have 3 phases that need to be properly ordered to this get optimized:
- unrolling (should happen before shift), and then
- shift optimization, and then
- inlining (it should not happen before shift, so "regular" stdlib version is not inlined).

Right now we do it in the following order:
- shift optimization
- inlining
- unrolling

Reshuffling this order will have significant impact on the optimizations. But it's not necessarily bad, I suggest experimenting with it. But we need to switch to the new pass manager first.

PR #2521, opened 2023-05-11
Support new pass manager in ISPC

aneshlya:
Primary motivation to move to the new PM is get rid of deprecated functionality and to be able to build with LLVM trunk. Nevertheless, using new PM is a good cause to review what we have in our optimization pipeline and if we have all needed analysis and up to date versions of passes in place.

Good reading on the background and implementation of new PM is here:
 - https://blog.llvm.org/posts/2021-03-26-the-new-pass-manager/
 - https://llvm.org/docs/NewPassManager.html


dbabokin:
I suggest adding `static bool isRequired() {return true;}` to those our passes that are "lowering" passes, i.e. may not be skipped. See details here: https://llvm.org/docs/WritingAnLLVMNewPMPass.html#required-passes

This affects the optimization for function with `optnone` attribute. I think we don't use it, but did try use it recently in some of our experiments.

src/opt/DebugPass.cpp
```diff
@@ -58,18 +52,9 @@ void DebugPassFile::run(llvm::Module &module, bool init) {
...
-
-bool DebugPassFile::doInitialization(llvm::Module &module) {
-    run(module, true);
-    return true;
-}
-
-llvm::Pass *CreateDebugPassFile(int number, llvm::StringRef name, std::string dir) {
-    return new DebugPassFile(number, name, dir);
+llvm::PreservedAnalyses DebugPassFile::run(llvm::Module &M, llvm::ModuleAnalysisManager &MAM) {
+    run(M, false);
```
dbabokin, line 56: Now `run` is invoked only with `false` as second parameter now, so the parameter needs to be removed.

dbabokin:
I spend some time with these changes and here's what I figured out:
- on large source code old PM spends considerable amount of time in "Loop unswitching" if LLVM compiled with asserts. The reason is dominator tree verification, which is not present in release builds. Here's [the line](https://github.com/llvm/llvm-project/blob/5ae9098e5c1564bdfe0ae0fdbc02e581138eeac7/llvm/lib/Transforms/Scalar/SimpleLoopUnswitch.cpp#L3804) where this costly verification runs. New PM doesn't have this verification.
- on large source code new PM spends way more time in "Loop Unroll". It does way more transformations in the Unroll. This is especially true for "old" targets, like SSE2/SSE4. And the reason for that is quite interesting - with old PM stdlib/builtin functions were optimized (specifically Unroll pass was run) before the inlining *for user code* has run. I.e. gather/scatter and masked load/stores code was unrolled before it was inlined. With new PM it is unrolled after the inlining, so this transformation needs to happen in every inline site (and there might be thousands on them).

Why optimization order has changed? Because old PM automatically did the right thing running optimizations "in-depth" as much as possible on optimization tree:
```
Module Passes
    GCSCC Passes << Note that Function inlining is GCSCC pass!!!!
        Function Passes
            Loop Passes
```

With new PM we force "breadth-first" approach by adding GCSCC/Function/Loop though adaptors, rather than explicitly adding them to appropriate low-level PM and adding it to higher-level PM.

I suggest that we:
1. try to fully mimic old PM with respect to phase ordering.
2. Add op-none instrumentation
3. Add trace/dumps which are not changing the order of optimizations and truly show how optimizations are applied.

If we cannot mimic old PM, we should make sure that Unroll is run on builtins before they are inlined.

Besides that we should think about getting rid of non-determinism / unpredictability caused by GCSCC inlining pass, which works on the whole Module, not just GCSCC region.

src/opt/DebugPass.cpp
aneshlya: DebugPass was removed.

aneshlya:
> I suggest adding `static bool isRequired() {return true;}` to those our passes that are "lowering" passes, i.e. may not be skipped. See details here: https://llvm.org/docs/WritingAnLLVMNewPMPass.html#required-passes
> 

`isRequired` API was used for old pass manager. It's not applicable to new pass manager anymore.

src/opt.cpp
```diff
@@ -61,122 +91,177 @@
...
 
     llvm::SimplifyCFGOptions simplifyCFGopt;
     simplifyCFGopt.HoistCommonInsts = true;
     if (optLevel == 0) {
-        // This is more or less the minimum set of optimizations that we
-        // need to do to generate code that will actually run.  (We can't
-        // run absolutely no optimizations, since the front-end needs us to
-        // take the various __pseudo_* functions it has emitted and turn
-        // them into something that can actually execute.
+        //  This is more or less the minimum set of optimizations that we
```
dbabokin, line 372: Looks like unintended formatting change.

dbabokin:
I've pushed the changes that introduce early LoopFullUnrollPass, which should fix the problem with increased compile time (due to inlining of gather/scatter before it's unrolled). It also fixes #2036.

But more work is needed on optimization of pass ordering after this PR is merged.

dbabokin (approved):
I think it's good to go, finally!

PR #2531, opened 2023-05-25
Support different types in short vector type initialization

aneshlya:
Support vector type conversions in initialization of short vectors from different types.
Additional fix for #2147


dbabokin (approved):
LGTM

PR #2537, opened 2023-06-09
Add signing to CMake

aneshlya:
To sign binaries and libraries, the one need to define `ISPC_SIGN_KEY`, when configuring CMake to build a package.

The signature assumed to work on Windows and macOS. Signing works when building `package` target and is executed after stripping.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -777,6 +777,13 @@ if (ISPC_PREPARE_PACKAGE)
     set(CPACK_PACKAGE_VERSION_PATCH ${ISPC_VERSION_PATCH}${ISPC_VERSION_SUFFIX})
     set(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_CURRENT_SOURCE_DIR}/LICENSE.txt")
     set(CPACK_RESOURCE_FILE_README "${CMAKE_CURRENT_SOURCE_DIR}/README.md")
+
+    # Sign files in the package if ISPC_SIGN_KEY is set
+    if (DEFINED ISPC_SIGN_KEY)
+        configure_file("${CMAKE_CURRENT_SOURCE_DIR}/cmake/Signing.cmake.in"
+                       "${CMAKE_CURRENT_BINARY_DIR}/Signing.cmake" @ONLY)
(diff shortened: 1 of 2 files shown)


cmake/Signing.cmake
```diff
@@ -0,0 +1,56 @@
...
+        get_property(ISPC_SIGN_TARGETS GLOBAL PROPERTY ISPCSignTargetsList)
+
+        # Construct the list of files to sign by retrieving the target file paths
+        set(ISPC_SIGN_FILES "")
+        foreach(sign_target ${ISPC_SIGN_TARGETS})
+            list(APPEND ISPC_SIGN_FILES $<TARGET_FILE:${sign_target}>)
+        endforeach()
+
+        add_custom_target(sign
+            COMMAND ${ISPC_SIGN_COMMAND} ${ISPC_SIGN_ARGS} ${ISPC_SIGN_FILES}
```
nurmukhametov, line 47: Should the sign command depends on files it signs, i.e., `DEPENDS ${ISPC_SIGN_FILES}`?

nurmukhametov: > Starting CMake 3.16 it's not needed anymore to set dependent targets explicitly. CMake will ensure that `${ISPC_SIGN_FILES}` will be available before `sign` target is built.

At the moment, we have `3.13`: https://github.com/ispc/ispc/blob/a8a1b8b3889807212ddb17bc967e36c0df154d67/CMakeLists.txt#L9

nurmukhametov: > Starting CMake 3.16 it's not needed anymore to set dependent targets explicitly. CMake will ensure that `${ISPC_SIGN_FILES}` will be available before `sign` target is built.

On my side this doesn't work (no `ispc` built triggered by `sign`), e.g.:

``` $ cmake ../
$ make sign
Build target sign
```

aneshlya: That's a valid point. As a separate change I think it's time to switch to CMkae 1.19 where ISPC support was introduced in CMake.

aneshlya: I fixed it.

dbabokin (approved):
Looks good! 🔥

PR #2542, opened 2023-06-21
ispcrt: add ispcrtSetTaskingCallbacks

### LINKED ISSUE #2540: Allow users to override ISPCLaunch/ISPCAlloc/ISPCSync
The new structure of ISPC Tasking and how the symbols for `ISPCLaunch`/`ISPCAlloc`/`ISPCSync` (https://github.com/ispc/ispc/blob/main/ispcrt/ispcrt.cpp#L173-L195) prevent users of ISPCRT from overriding these symbols to integrate into their own tasking systems and memory allocators (as we do in OSPRay).

It would be valuable to add APIs to allow users to set the function pointers that the entry points dispatch to call their their own function pointers instead of just loading them from the CPU module when tasking is enabled (https://github.com/ispc/ispc/blob/main/ispcrt/ispcrt.cpp#L214-L227).

Maybe some API like `ispcSetTaskingCallbacks(ISPCLaunch, ISPCAlloc, ISPCSync)` ? To enforce that the app must set all 3?

nurmukhametov:
The new structure of ISPCRT and the logic of loading symbols ISPCLaunch, ISPCAlloc, ISPCSync from ispcrt_device_cpu solib prevents users of ISPCRT (e.g., OSPRay) from overriding these symbols to integrate their own tasking systems and memory allocation with ISPC.

To address that, ispcrtSetTaskingCallbacks was added and exposed to ISPCRT API. It has three arguments to set ISPCLaunch, ISPCAlloc, ISPCSync implementations.

Fixes #2540.

### DIFF
--- a/ispcrt/ispcrt.cpp
+++ b/ispcrt/ispcrt.cpp
@@ -156,12 +156,17 @@ void *dyn_load_sym(void *handle, const char *symbol) {
 }
 
 // CPU device API.
-typedef void (*ISPCLaunchF)(void **, void *, void *, int, int, int);
-typedef void *(*ISPCAllocF)(void **, int64_t, int32_t);
-typedef void (*ISPCSyncF)(void *);
-static ISPCLaunchF ispc_launch_fptr = nullptr;
-static ISPCAllocF ispc_alloc_fptr = nullptr;
-static ISPCSyncF ispc_sync_fptr = nullptr;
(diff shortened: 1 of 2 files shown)


Twinklebear:
Thanks @nurmukhametov , this will work great for us in OSPRay. :+1:

aneshlya:
@nurmukhametov , please resolve a conflict and we can merge.

PR #2543, opened 2023-06-21
ispcrt: remove level-zero dependency search

nurmukhametov:
After the ispcrt split, it is not required anymore. An user application depends on ispcrt solib. It loads on-demand ispcrt_device_gpu that depends on level_zero loader. So, it should not be a compile time dependency for user application.


nurmukhametov:
This fix needs some internal CI synchronous change. This PR has been merged in the internal repo, so it would be propagated on the next promotion.

PR #2546, opened 2023-06-23
OneAPI version of v1.20.0 linux package

dbabokin:
The difference that it uses newer TBB for the build. For OpeAPI `libispcrt_device_cpu.so` depends on `libtbb.so.12`, while in regular distribution it depends on `libtbb.so.2`.


aneshlya (approved):
LGTM

Issue #2544, opened 2023-06-21
ISPC 1.20.0 ICE on invalid print() syntax


mikemccarty-vertex:
https://ispc.godbolt.org/z/3s985rnaW

The compilation issues the error message before the ICE but still doesn't seem like a good outcome.

nurmukhametov:
> I may suggest to check `yynerrs` in assert function. 

Here https://github.com/ispc/ispc/pull/2547.

dbabokin:
Fixed by #2547

PR #2547, opened 2023-06-26
Fix ICE on invalid print syntax

nurmukhametov:
It fixes the issue #2544 where user reported ICE on invalid syntax inside `print` arguments.

Parser encounters a recoverable syntax error and proceeds further. However, it breaks some assumption in an assert statement later during the code emitting of print statement. This assert is raised resulting in `abort` call to terminate program.

In this case, we may terminate the compiler with `exit(yynerrs)` adding message saying that the assert can be triggered by the reported syntax error.

```bash
$ ./bin/ispc  --arch=x86-64 --target=avx2 2544.ispc -o 2544.o
2544.ispc:3:33: Error: syntax error, unexpected ':', expecting ')' or ','.
    print("sum: %", reduce_add(x:1:0));
                                ^

./src/stmt.cpp:3748: Assertion failed: "!printImplArgs.empty() && "Haven't managed to produce __do_print args"".
***
*** Assertion may be caused by syntax error shown above,
*** otherwise follow instructions below.
***
***
$ echo $?
1
```

### DIFF
--- a/src/stmt.cpp
+++ b/src/stmt.cpp
@@ -3745,7 +3745,10 @@ void PrintStmt::EmitCode(FunctionEmitContext *ctx) const {
     }
 #endif /* ISPC_XE_ENABLED */
     auto printImplArgs = getDoPrintArgs(ctx);
-    Assert(!printImplArgs.empty() && "Haven't managed to produce __do_print args");
+    if (printImplArgs.empty()) {
+        AssertPos(pos, m->errorCount > 0);
+        return;
+    }
     auto printImplFunc = getPrintImplFunc();
     AssertPos(pos, printImplFunc);
     ctx->CallInst(printImplFunc, nullptr, printImplArgs, "");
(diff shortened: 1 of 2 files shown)


dbabokin (changes requested):
The idea for dealing with syntax error is the following - when syntax error is encountered, an invalid expression is created and an error is flagged through `errorCount`. Typically it's a `nullptr` returned instead of expression. And code gen should deal with it correctly.

Something like the following patch should do the trick.

```cpp
diff --git a/src/stmt.cpp b/src/stmt.cpp
index 1fb8094a0..2132a10a6 100644
--- a/src/stmt.cpp
+++ b/src/stmt.cpp
@@ -3745,7 +3745,10 @@ void PrintStmt::EmitCode(FunctionEmitContext *ctx) const {
     }
 #endif /* ISPC_XE_ENABLED */
     auto printImplArgs = getDoPrintArgs(ctx);
-    Assert(!printImplArgs.empty() && "Haven't managed to produce __do_print args");
+    if (printImplArgs.empty()) {
+        AssertPos(pos, m->errorCount > 0);
+        return;
+    }
     auto printImplFunc = getPrintImplFunc();
     AssertPos(pos, printImplFunc);
     ctx->CallInst(printImplFunc, nullptr, printImplArgs, "");
```

dbabokin (approved):
LGTM

Issue #2509, opened 2023-05-05
ICE when undeclared symbol in struct initializer


chloekek:
### ISPC version

Intel(r) Implicit SPMD Program Compiler (Intel(r) ISPC), 1.19.0 (build commit 147cc795073f50b6 @ 20230227, LLVM 14.0.6)

### Reproducible example

<https://godbolt.org/z/rexKdxcja>

```ispc
struct S
{
    float x;
};

void f()
{
    S s = { b };
}
```

### Compiler output

```
<source>:8:13: Error: Undeclared symbol "b". Did you mean "f"? 
    S s = { b };
            ^

ispc: /usr/local/src/llvm/bin-14.0/include/llvm/Support/Casting.h:104: static bool llvm::isa_impl_cl<ispc::ExprList, const ispc::Expr *>::doit(const From *) [To = ispc::ExprList, From = const ispc::Expr *]: Assertion `Val && "isa<> used on a null pointer"' failed.
/usr/local/src/ispc/src/main.cpp(412): FATAL ERROR: Unhandled signal sent to process; terminating.
***
*** Please file a bug report at https://github.com/ispc/ispc/issues
*** (Including as much information as you can about how to reproduce this error).
*** You have apparently encountered a bug in the compiler that we'd like to fix!
***
Compiler returned: 139
```

nurmukhametov:
> At least, it has the message about undeclared symbol.
> 
> There is probably the variation of same problem which is more confusing:
> 
> ```c
> struct D {
>     int h;
> };
> struct D d;
> int foo(varying int a) {
>      return a += d.f;
> }
> ```
> 
> ```shell
> /usr/local/src/ispc/src/main.cpp(412): FATAL ERROR: Unhandled signal sent to process; terminating.
> ***
> *** Please file a bug report at https://github.com/ispc/ispc/issues
> *** (Including as much information as you can about how to reproduce this error).
> *** You have apparently encountered a bug in the compiler that we'd like to fix!
> ***
> Compiler returned: 139
> ```
> 
> Compiler explorer link: https://godbolt.org/z/csr4qfWcP

It is a different issue, see #2514

mingodad:
Looking through the stack trace while debugging the change shown bellow seems to fix this issue:
```
--------------------------------- src/stmt.cpp ---------------------------------
index 1fb8094a..7db49716 100644
@@ -269,7 +269,7 @@ static bool checkInit(const Type *type, Expr **init) {
[...]

PR #2548, opened 2023-06-27
Fix ICE when undeclared symbol used in initializer expr

nurmukhametov:
This fixes ICE when an initializer expression is malformed, e.g., because of the usage of undeclared symbols, as reported in the issue #2509.

Thanks @mingodad for the suggested fix (https://github.com/ispc/ispc/issues/2509#issuecomment-1556850627)

### DIFF
--- a/src/stmt.cpp
+++ b/src/stmt.cpp
@@ -262,6 +262,11 @@ static bool checkInit(const Type *type, Expr **init) {
         return false;
     }
 
+    if (*init == nullptr) {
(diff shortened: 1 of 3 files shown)


dbabokin:
The test doesn't need to be for x86, so no need to fix the target (or otherwise it needs `// REQUIRES: X86_ENABLED`). I've fixed that for this test and for `2544.ispc` from previous commit.

dbabokin (approved):

Issue #2523, opened 2023-05-12
Double opening bracket causes assert


emoon:
Given the following code https://godbolt.org/z/dscqT4fx3

```C
void foobar() { {
}
```
Generates the following error
```
<source>:2:2: Error: Premature end of file: syntax error, unexpected 
        end of file. 
}

 ^

/usr/local/src/ispc/src/sym.cpp:72: Assertion failed: "variables.size() == 1".
***
*** Please file a bug report at https://github.com/ispc/ispc/issues
*** (Including as much information as you can about how to reproduce this error).
*** You have apparently encountered a bug in the compiler that we'd like to fix!
***
/usr/local/src/ispc/src/main.cpp(412): FATAL ERROR: Unhandled signal sent to process; terminating.
Compiler returned: 139
```
So the first error is fine, but I guess the assert shouldn't happen in this case. Also the problem happens with loops as well (where I encountered it first)

mingodad:
Looking through the code I came to this possible fix:
```
-------------------------------- src/module.cpp --------------------------------
index c1f1409a..7b4b56c2 100644
@@ -3158,6 +3158,8 @@ int Module::CompileAndOutput(const char *srcFile, Arch arch, const char *cpu, st
         }
 
         int errorCount = m->errorCount;
+        // cleanup symbol table to avoid failing assert on destructor
+        while(m->symbolTable->ScopeCount() > 1) m->symbolTable->PopScope();
         delete m;
         m = nullptr;
 
---------------------------------- src/sym.h ----------------------------------
index 0fadb64a..775e30f2 100644
@@ -134,6 +134,8 @@ class SymbolTable {
         that scope. */
     void PopScope();
 
+    size_t ScopeCount() const {return variables.size();}
+
     /** Adds the given variable symbol to the symbol table.
         @param symbol The symbol to be added

```

dbabokin:
@mingodad Feel free to open the PR with the fix - it's the most convenient way to discuss the fix.

Note, we recently added Codespaces to make it easer submitting small PRs and testing them before submitting for review.

aneshlya:
Fixed by https://github.com/ispc/ispc/pull/2551

PR #2551, opened 2023-06-29
Clean up symbol table to avoid assertion in destructor

aneshlya:
This is a fix for #2523. Thanks @mingodad for the idea of the fix.


dbabokin (approved):
LGTM

PR #2559, opened 2023-07-10
Deallocate memory allocated in lexer

nurmukhametov:
There are a few places in `lex.ll` that allocate memory on heap to pass some values to the parser. It should be deallocated. 

Two fields of `yylval` struct are used to pass heap allocated memory from lexer to parser: `std::string *stringVal` and `PragmaAttributes * pragmaAttributes`. The first one is filled in `TOKEN_IDENTIFIER`, `TOKEN_TEMPLATE_NAME`, `TOKEN_TYPE_NAME`, `TOKEN_FLOAT16_CONSTANT `, `TOKEN_INTRINSIC_CALL`, `TOKEN_STRING_LITERAL`. The second one is allocated in `TOKEN_PRAGMA`.

To make `stringVal` memory management identical for all tokens, `operator` rules are made to allocate string as well.

### DIFF
--- a/src/lex.ll
+++ b/src/lex.ll
@@ -453,13 +453,13 @@ while { RT; return TOKEN_WHILE; }
 \"SYCL\" { RT; return TOKEN_STRING_SYCL_LITERAL; }
 \.\.\. { RT; return TOKEN_DOTDOTDOT; }
 
-"operator*"  { return TOKEN_IDENTIFIER; }
-"operator+"  { return TOKEN_IDENTIFIER; }
-"operator-"  { return TOKEN_IDENTIFIER; }
-"operator<<" { return TOKEN_IDENTIFIER; }
-"operator>>" { return TOKEN_IDENTIFIER; }
-"operator/" { return TOKEN_IDENTIFIER; }
-"operator%" { return TOKEN_IDENTIFIER; }
+"operator*"  { yylval.stringVal = new std::string(yytext); return TOKEN_IDENTIFIER; }
+"operator+"  { yylval.stringVal = new std::string(yytext); return TOKEN_IDENTIFIER; }
+"operator-"  { yylval.stringVal = new std::string(yytext); return TOKEN_IDENTIFIER; }
+"operator<<" { yylval.stringVal = new std::string(yytext); return TOKEN_IDENTIFIER; }
+"operator>>" { yylval.stringVal = new std::string(yytext); return TOKEN_IDENTIFIER; }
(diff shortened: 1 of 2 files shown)


aneshlya (approved):

dbabokin (approved):

dbabokin:
How these bugs were discovered? Can we setup address sanitizer build and run on our tests to monitor new problems?

e3m3:
> How these bugs were discovered? Can we setup address sanitizer build and run on our tests to monitor new problems?

I had a similar question - the CI tasks for PRs in open source don't have sanitizers enabled last I checked. Should we look to add this?

dbabokin:
My another concern here is that it's not clear that we don't do double free on the same string after the changes. I assume we don't do that, but would be nice to verify this with some tool, i.e AddressSanitizer.

nurmukhametov:
> > How these bugs were discovered? Can we setup address sanitizer build and run on our tests to monitor new problems?
> 
> I had a similar question - the CI tasks for PRs in open source don't have sanitizers enabled last I checked. Should we look to add this?

Leak sanitizer discovered them (not all though). I tested on trivial example with stdlib, so most of leaks from lex.ll were from this allocation https://github.com/ispc/ispc/blob/85034d72d0324296898e42c88ced9e4464618d4d/src/lex.ll#L470

Then I looked through all of `new` allocations to fix similar issues.

Yes, we can setup sanitizers in CI to monitor problems but the leak sanitizer report has quite a lot of issues right now. The address sanitizer report is probably clean.

nurmukhametov:
> My another concern here is that it's not clear that we don't do double free on the same string after the changes. I assume we don't do that, but would be nice to verify this with some tool, i.e AddressSanitizer.

This is the biggest concern for me either. Yes, ASAN is good idea, let me check with it.

nurmukhametov:
> > > My another concern here is that it's not clear that we don't do double free on the same string after the changes. I assume we don't do that, but would be nice to verify this with some tool, i.e AddressSanitizer.
> > 
> > 
> > This is the biggest concern for me either. Yes, ASAN is good idea, let me check with it.
> 
> It looks OK (with make check-all). Moreover, I have created CI job to build and run make check-all with ASAN here #2564 . We may merge it first than proceed with the current PR.

I have rebased it on the current main with the merged ASAN job. It is [green](https://github.com/ispc/ispc/actions/runs/5542208907/jobs/10116594672?pr=2559).

e3m3 (approved):
LGTM

dbabokin (approved):

PR #2564, opened 2023-07-11
CI: add ASAN job (build and check-all)

nurmukhametov:
Add CI job to build ISPC with the address sanitizer (ASAN). 

According to [this](https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer), ASAN includes also the leak sanitizer (LSAN). At the moment, LSAN is turned off by `ASAN_OPTIONS=detect_leaks=0`, so only ASAN operates. It could find memory errors like double-free, use-after-free, stack/heap corruptions.


.github/workflows/ispc-ci-san.yml
```diff
@@ -0,0 +1,43 @@
...
+
+    - name: Install dependencies
+      run: |
+        .github/workflows/scripts/install-build-deps.sh
+
+    - name: Check environment
+      run: |
+        ./check_env.py
+        which -a clang
+        which -a gcc
```
dbabokin, line 31: Let's document that gcc is used because the clang is built without `compiler-rt` libraries.

nurmukhametov: I have documented that.

.github/workflows/ispc-ci-san.yml
```diff
@@ -0,0 +1,43 @@
...
+        .github/workflows/scripts/install-build-deps.sh
+
+    # We will use system gcc until we rebuild LLVM with compiler-rt.
+    - name: Check environment
+      run: |
+        ./check_env.py
+        which -a gcc
+        cat /proc/cpuinfo
+
+    - name: Build package
```
dbabokin: `Build package` is misleading name here, as only config is done. I suggest keeping the name, but doing package build here as well.

.github/workflows/ispc-ci-san.yml
```diff
@@ -0,0 +1,41 @@
...
+    - name: Check environment
+      run: |
+        ./check_env.py
+        which -a gcc
+        cat /proc/cpuinfo
+
+    # Address sanitizer includes leak sanitizer but we don't want it to report
+    # all leaks at the moment because there are quite a lot if them.
+    # So, we disable leaks detection to have clean report.
+    - name: Run ASAN
```
dbabokin, line 37: Let's have 3 different steps here for clarity / readability of the run:
- config and build ISPC with ASAN enabled
- run ASAN on LIT tests
- run ASAN on runtime tests

dbabokin (changes requested):
We also need to clean up GCC warnings - not as part of this PR, I would suggest having this as a separate job (with fixed GCC version and `-Werror` enabled)

.github/workflows/ispc-ci-san.yml
nurmukhametov: Fixed, test run is here https://github.com/nurmukhametov/ispc/actions/runs/5535409990/jobs/10101667972

nurmukhametov: Example of failed job on double-free error is here https://github.com/nurmukhametov/ispc/actions/runs/5535736851/jobs/10102438055

dbabokin (approved):
I've push the changes that I assumed in the review feedback. If the look good, please go forward with the merge.

nurmukhametov:
> I've push the changes that I assumed in the review feedback. If the look good, please go forward with the merge.

It looks good (test run is [here](https://github.com/nurmukhametov/ispc/actions/runs/5541689886/jobs/10115419740)).

PR #2566, opened 2023-07-13
Add CI jobs to build ISPC with GCC and Clang with -Werror

nurmukhametov:
As suggested here https://github.com/ispc/ispc/pull/2564#pullrequestreview-1527008252, I created CI jobs to control emitted compiler warnings.

CMake has been fixed to not pass clang-specific warning options to GCC.

There are two jobs to build ISPC with Clang and GCC. Test run of this pipeline is [here](https://github.com/nurmukhametov/ispc/actions/runs/5542662436/jobs/10117629900). At the moment, this pipeline is turned off for pull requests because there are some gcc emitted warnings.


.github/workflows/compiler-warnings.yml
```diff
@@ -0,0 +1,69 @@
+# Copyright 2023, Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause
+
+name: Compiler Warnings
+
+on:
+# Turned off until all jobs are green.
+#  pull_request:
```
e3m3, line 8: Good calll👍

dbabokin (approved):
LGTM

PR #2567, opened 2023-07-14
Disable generation of gathers/scatters by option

### LINKED ISSUE #2556: Gather/scatter disable switch
It's useful for experiments to be able to disable gather/scatter generation and use scalarized load/stores instead.

The proposal is to add `--opt=disable-gather` and `--opt=disable-scatter` switches:
```
disable-gather    Disable ‘gather’ instructions (on targets that support them)
disable-scatter   Disable ‘scatter’ instructions (on targets that support them)
```

The nuance of introducing such switch is that in current target files design, we'll have to add scalarized implementation of gather/scatter to every target that doesn't have them currently.

aneshlya:
This PR introduces `--opt=disable-gathers` and `--opt=disable-scatters` options which disable generation of gathers/scatters on platforms that support them (for performance experiments)
Fixes #2556.


dbabokin (approved):
Looks good.

One thing to note here though. Before the changes we lowered `__pseudo_gather*` to `__gather_*` assuming that it will have appropriate implementation (emulation on HW w/o gather support and intrinsics on HW with gather support). But now on HW with gather support we have two versions that we pick depending on the switches: `__gather_*` and `__gather*_generic_*` (which is exact copy of `__gather_*` for HW w/o gather support).

It would be more logical to have a single `__gather*_generic` version for all platforms and use it on platform w/o HW support.

We should do that change in the future, when redesigning the targets support.

PR #2568, opened 2023-07-14
ci: Superbuild switches

e3m3:
Changes:
*   Adds overall BUILD_TYPE switch and individual switches for each
    component built.
    | `superbuild/CMakeLists.txt`
*   Adds `JOBS` switch settable via environment or cmake -D option.
    Useful for thinly provisioned machines.
    | `superbuild/CMakeLists.txt`


superbuild/CMakeLists.txt
```diff
@@ -37,6 +37,13 @@ set(CCACHE_DIR "default" CACHE STRING "ccache cache directory")
 set(PREBUILT_STAGE2 "NO" CACHE STRING "Provide pre-built stage2 archive to skip pre ispc-stage2 jobs")
 set(PREBUILT_STAGE2_PATH "NO" CACHE STRING "Provide pre-built stage2 toolchain path")
 set(STAGE2_TOOLCHAIN_INSTALL_PREFIX "NO" CACHE STRING "Path to install stage2 Clang/LLVM to")
+set(JOBS "" CACHE STRING "Set the number of jobs for internal build commands (ninja)")
+set(BUILD_TYPE "" CACHE STRING "Set the build type for builds (overrides per-component build type switches)")
+set(BUILD_TYPE_ISPC "Release" CACHE STRING "Set the build type for ISPC build")
+set(BUILD_TYPE_LLVM "Release" CACHE STRING "Set the build type for LLVM build")
+set(BUILD_TYPE_SPIRV_TRANSLATOR "Release" CACHE STRING "Set the build type for SPIR-V Translator build")
+set(BUILD_TYPE_VC_INTRINSICS "Release" CACHE STRING "Set the build type for VC Intrinsics build")
+set(BUILD_TYPE_L0 "Release" CACHE STRING "Set the build type for L0 build")
```
nurmukhametov, line 46: All `JOBS` and `BUILD_*` vars are needed to be printed out.

superbuild/CMakeLists.txt
```diff
@@ -124,6 +131,28 @@ if (EXTERNAL_VERBOSE)
     set(EXTERNAL_VERBOSE_FLAG "-v")
 endif()
 
+if (JOBS)
+    set(NINJA_JOBS "-j ${JOBS}")
```
nurmukhametov, line 135: `NINJA_JOBS` is better be a list to avoid mistreating args.

superbuild/CMakeLists.txt
```diff
@@ -793,6 +824,7 @@ list(APPEND NINJA_COMMAND_WITH_ENV_STAGE3
     "PATH=${ENV_PATH_STAGE3_ESC_SEM}"
     ${NINJA_EXE}
     ${EXTERNAL_VERBOSE_FLAG}
+    ${NINJA_JOBS}
     )
```
nurmukhametov, line 828: Let's print out `NINJA_COMMAND_WITH_ENV_STAGE*` vars as well.

superbuild/CMakeLists.txt
```diff
@@ -37,6 +37,13 @@ set(CCACHE_DIR "default" CACHE STRING "ccache cache directory")
 set(PREBUILT_STAGE2 "NO" CACHE STRING "Provide pre-built stage2 archive to skip pre ispc-stage2 jobs")
 set(PREBUILT_STAGE2_PATH "NO" CACHE STRING "Provide pre-built stage2 toolchain path")
 set(STAGE2_TOOLCHAIN_INSTALL_PREFIX "NO" CACHE STRING "Path to install stage2 Clang/LLVM to")
+set(JOBS "" CACHE STRING "Set the number of jobs for internal build commands (ninja)")
+set(BUILD_TYPE "" CACHE STRING "Set the build type for builds (overrides per-component build type switches)")
```
nurmukhametov, line 41: I prefer to have default `Release` value for `BUILD_TYPE` to preserve old behavior.

superbuild/CMakeLists.txt
```diff
@@ -124,6 +131,28 @@ if (EXTERNAL_VERBOSE)
...
+    set(NINJA_JOBS "")
+endif()
+
+if (BUILD_TYPE)
+    set(BUILD_TYPE_ISPC ${BUILD_TYPE})
+    set(BUILD_TYPE_LLVM ${BUILD_TYPE})
+    set(BUILD_TYPE_SPIRV_TRANSLATOR ${BUILD_TYPE})
+    set(BUILD_TYPE_VC_INTRINSICS ${BUILD_TYPE})
+    set(BUILD_TYPE_L0 ${BUILD_TYPE})
+elseif($ENV{BUILD_TYPE})
```
nurmukhametov, line 148: I don't really like the idea of setting `BUILD_TYPE-*` from env vars values but if you need them it is OK. At least, add a message that says that the value was derived from environment.

nurmukhametov (changes requested):

superbuild/CMakeLists.txt
e3m3: Was `BUILD_TYPE` already in use before this patch? I set it to default to empty such that if it's defined to anything by the user, the subcomponent build types which default to Release are overridden. From what I understand the default behavior is preserved.

superbuild/CMakeLists.txt
e3m3: Don't need them, but wasn't sure if they might be useful for passing down configurations from CI tasks. I'll add a message.

nurmukhametov: `JOBS` in `BUILD_TYPE_*` message look like copy-paste leftover.

superbuild/CMakeLists.txt
```diff
@@ -124,6 +131,34 @@ if (EXTERNAL_VERBOSE)
     set(EXTERNAL_VERBOSE_FLAG "-v")
 endif()
 
+if (JOBS)
+    list(APPEND NINJA_JOBS "-j ${JOBS}")
```
nurmukhametov, line 135: I think it should be `list(APPEND NINJA_JOBS -j ${JOBS})` to not glue together `-j` and `jobs` into one argument.

superbuild/CMakeLists.txt
nurmukhametov: So, it was `Release` and now it is an empty. It is not clear for me why it is same.

superbuild/CMakeLists.txt
e3m3: Gotcha thanks

superbuild/CMakeLists.txt
e3m3: I think there's some confusion here. As far as I understand, there was no `BUILD_TYPE` variable in the cmake previously, so it would have been unset/empty before this. Are you referring to `CMAKE_BUILD_TYPE`?

e3m3: One idea would be to add `set(CMAKE_BUILD_TYPE $BULD_TYPE)` in the `if (BUILD_TYPE)` section to make sure they match.

dbabokin: I sounds like manually defined `BUILD_TYPE` tries to replace `CMAKE_BUILD_TYPE`. I.e. why `CMAKE_BUILD_TYPE` doesn't work for this purpose? It creates confusion and unnecessary complication of the interplay of the two. Do we define what the behavior supposed to be in case both are defined and contradict?

dbabokin:
Is defining different build types for every component is really needed? Is it a theoretical or practical need?

Keeping build files simple has its value and to me it has more priority than fulfilling theoretical needs. So if it's not a practical need, I would suggest not implementing this.

superbuild/CMakeLists.txt
```diff
@@ -37,6 +37,13 @@ set(CCACHE_DIR "default" CACHE STRING "ccache cache directory")
 set(PREBUILT_STAGE2 "NO" CACHE STRING "Provide pre-built stage2 archive to skip pre ispc-stage2 jobs")
 set(PREBUILT_STAGE2_PATH "NO" CACHE STRING "Provide pre-built stage2 toolchain path")
 set(STAGE2_TOOLCHAIN_INSTALL_PREFIX "NO" CACHE STRING "Path to install stage2 Clang/LLVM to")
+set(JOBS "" CACHE STRING "Set the number of jobs for internal build commands (ninja)")
```
dbabokin, line 40: Is it duplication of `CMAKE_BUILD_PARALLEL_LEVEL` env variable?

superbuild/CMakeLists.txt
e3m3: We can use `CMAKE_BUILD_TYPE` instead and remove `BUILD_TYPE`. The reason it didn't do what its designed to do is because we were explicitly setting it to Release for each sub-build. The per-component setters allow us to control them separately.

To clear up the implementation, we simply need to rename `BUILD_TYPE` in this patch to `CMAKE_BUILD_TYPE`. That way, if the user sets it, the entire build is of the type requested, but the user can specify type for individual components too.

superbuild/CMakeLists.txt
e3m3: Possibly, though using this variable does not specifically address the current issue where Ninja will spawn as many threads as there are logical cores available on the system.

e3m3:
> Is defining different build types for every component is really needed? Is it a theoretical or practical need?
> 
> Keeping build files simple has its value and to me it has more priority than fulfilling theoretical needs. So if it's not a practical need, I would suggest not implementing this.

The use case for this patch is not theoretical. Specifying debug builds for components allows for more detail debug info in gdb, and this allows us to get the info for components/dependencies. The ninja parallelization is a problem when running super build with a machine that is running out of ram/cores.

superbuild/CMakeLists.txt
dbabokin: It contradicts the description of this var: https://cmake.org/cmake/help/latest/envvar/CMAKE_BUILD_PARALLEL_LEVEL.html

Are you defining it as environment variable for the experiment?

e3m3: > Are you defining it as environment variable for the experiment?

Which experiment?

e3m3: > It contradicts the description of this var: https://cmake.org/cmake/help/latest/envvar/CMAKE_BUILD_PARALLEL_LEVEL.html

We can use this value as long as its passed along to ninja for the sub-builds. The sub-builds are not triggered by a call to `cmake --build`, so I'll have to test to see if setting this value obviates the need for `NINJA_JOBS`.

dbabokin:
```
cmake .. -G Ninja
CMAKE_BUILD_PARALLEL_LEVEL=1 cmake --build .
```
works for me.

superbuild/CMakeLists.txt
```diff
@@ -37,6 +37,13 @@ set(CCACHE_DIR "default" CACHE STRING "ccache cache directory")
...
+set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Set the build type for builds (overrides per-component build type switches)")
```
dbabokin, line 41: Have you tested it? As far as I understand how CMake works, redefining `CMAKE_*` variables work only when done before the `project(...)` directive, otherwise CMake silently discards this redeclaration. And it makes sense, as at the point of `project(...)` directive CMake needs to know about key parameters for the project, so hence changing them in the middle of the project is not supported.

That's another reason why redefining `CMAKE_*` variable or duplicating its functionality is typically a bad idea and may strike back in unexpected way.

dbabokin:
> Currently, any stack trace from while debugging in a docker image will have optimized out values. This patch provides an easy way to rebuild with debug on.

What are you trying to debug? In my experience, I need either ISPC being built with -O0, or ISPC+LLVM. Would having jsut a debug image with everything being build in debug mode solve the problem/

e3m3:
> Would having jsut a debug image with everything being build in debug mode solve the problem?

It would help, but this requires pushing and triggering a commit to CI in order for changes to propagate for testing. While making fast prototypes locally, it's helpful to commit and rebuild often (with debug on).

superbuild/CMakeLists.txt
e3m3: That's ok - we can delete the setter line. The logic controlling the type for sub-components is the main use for this change. I wasn't aware about that for Windows. If we want to prevent the user from mixing types, we should remove the sub-component variables, but we should still pass the build type along. Currently, it's hard-coded to Release.

e3m3:
> Also note that on Windows you can't mix object files with different build type. Linker will complain for some macro being defined differently in different object files.

That's ok - we can delete the setter line. The logic controlling the type for sub-components is the main use for this change. I wasn't aware about that for Windows. If we want to prevent the user from mixing types, we should remove the sub-component variables, but we should still pass the build type along. Currently, it's hard-coded to Release.

superbuild/CMakeLists.txt
e3m3: > ```
> cmake .. -G Ninja
> CMAKE_BUILD_PARALLEL_LEVEL=1 cmake --build .
> ```
> 
> works for me.

This is not working for me, including exporting the value.

superbuild/CMakeLists.txt
```diff
@@ -124,6 +129,20 @@ if (EXTERNAL_VERBOSE)
     set(EXTERNAL_VERBOSE_FLAG "-v")
 endif()
 
+if (CMAKE_BUILD_PARALLEL_LEVEL)
+    list(APPEND NINJA_JOBS -j ${CMAKE_BUILD_PARALLEL_LEVEL})
+else()
+    list(APPEND NINJA_JOBS "")
+endif()
```
e3m3, line 136: `NINJA_JOBS` is now set directly from the cmake variable.

superbuild/CMakeLists.txt
```diff
@@ -124,6 +129,20 @@ if (EXTERNAL_VERBOSE)
...
+
+if (CMAKE_BUILD_TYPE)
+    set(BUILD_TYPE_ISPC ${CMAKE_BUILD_TYPE})
+    set(BUILD_TYPE_LLVM ${CMAKE_BUILD_TYPE})
+    set(BUILD_TYPE_SPIRV_TRANSLATOR ${CMAKE_BUILD_TYPE})
+    set(BUILD_TYPE_VC_INTRINSICS ${CMAKE_BUILD_TYPE})
+    set(BUILD_TYPE_L0 ${CMAKE_BUILD_TYPE})
+endif()
```
e3m3, line 144: Sub-build types now set directly from the cmake variable.

superbuild/CMakeLists.txt
dbabokin: Try printing `CMAKE_BUILD_PARALLEL_LEVEL` inside CMake to see if it's appears set from CMake stand point.

If we can make it work with standard means, we should do that instead of adding duplicating functionality.

superbuild/CMakeLists.txt
```diff
@@ -146,6 +165,13 @@ message(STATUS "STAGE2_TOOLCHAIN_INSTALL_PREFIX is ${STAGE2_TOOLCHAIN_INSTALL_PR
 message(STATUS "LLVM_DISABLE_ASSERTIONS is ${LLVM_DISABLE_ASSERTIONS}")
 message(STATUS "LTO is ${LTO}")
 message(STATUS "PGO is ${PGO}")
+message(STATUS "CMAKE_BUILD_TYPE is ${CMAKE_BUILD_TYPE}")
+message(STATUS "CMAKE_BUILD_PARALLEL_LEVEL is ${CMAKE_BUILD_PARALLEL_LEVEL}")
+message(STATUS "BUILD_TYPE_ISPC is ${BUILD_TYPE_ISPC}")
+message(STATUS "BUILD_TYPE_LLVM is ${BUILD_TYPE_LLVM}")
+message(STATUS "BUILD_TYPE_SPIRV_TRANSLATOR is ${BUILD_TYPE_SPIRV_TRANSLATOR}")
+message(STATUS "BUILD_TYPE_VC_INTRINSICS is ${BUILD_TYPE_VC_INTRINSICS}")
+message(STATUS "BUILD_TYPE_L0 is ${BUILD_TYPE_L0}")
```
e3m3, line 174: Updated status messages.

superbuild/CMakeLists.txt
e3m3: The description for the variable is misleading, as setting it from the environment show it to be empty when printing. The value can be passed through the CLI though, which will set ninja's jobs [here](https://github.com/ispc/ispc/pull/2568#discussion_r1266021774).

dbabokin: Thats weird, it works for me out of the box when building ISPC without superbuild as I noted above. My CMake version is 3.21. Check if you have relatively new version.

superbuild/CMakeLists.txt
nurmukhametov: It looks good to me now.

aneshlya (approved):
I don't have any comments in addition to already provided.

e3m3:
Merging.

PR #2569, opened 2023-07-17
Deallocate memory allocated in parser

nurmukhametov:
After that change, no memory leaks left originated from `parse.yy` on trivial example with stdlib. The fixed allocation is cause of around 2MB in32k allocations. The left part after the fix is around 0.2MB in 2600 allocations.

`Symbol`, `Declaration`, `DeclSpec`, `Declarator`, `StructDeclaration` is freed via bookkeeping (`Tracable`). 

There are also quite a few auxiliary vectors created during parsing that deallocated directly after usage.


e3m3 (approved):
LGTM

aneshlya (approved):

dbabokin (approved):
The changes look good, but it's very easy to do a mistake handling these data structures in parser, I keep my fingers crossed that if there are any errors in these code, they will be caught by the sanitizers.

PR #2494, opened 2023-04-21
loops: Fix codegen for overflow in signed expressions

e3m3:
Changes:
*   Add test checking for performant loop codegen.
    | `tests/lit-tests/loop-codegen-acc-width.ispc`

May close #2460.

PR #2570, opened 2023-07-17
stdlib (`__fast_idiv`): Fix identity operation for fast integer division

e3m3:
Currently, the implementation for `__fast_idiv` fails for `1/1`. This patches in a fast path for the identity operation (`divisor == 1`).

Changes:
*   Add a fast path for the identity operation for each type.
    | `stdlib.ispc`
*   Add tests asserting division by 1 is 1 for each type.
    | `tests/idiv-int16-identity.ispc`,
    | `tests/idiv-int32-identity.ispc`,
    | `tests/idiv-int8-identity.ispc`,
    | `tests/idiv-uint16-identity.ispc`,
    | `tests/idiv-uint32-identity.ispc`,
    | `tests/idiv-uint8-identity.ispc`

### DIFF
--- a/stdlib.ispc
+++ b/stdlib.ispc
@@ -5729,7 +5729,9 @@ static const uniform int64 __idiv_table_s32[][3] = {
 
 __declspec(safe) static unmasked inline unsigned int8
     __fast_idiv(unsigned int8 numerator, uniform unsigned int8 divisor) {
-    if (__is_xe_target) {
+    if (divisor == 1) {
+        return numerator;
+    } else if (__is_xe_target) {
         return __idiv_uint8(numerator, divisor);
     }
     uniform int64 method = __idiv_table_u8[divisor - 2][0];
@@ -5751,7 +5753,9 @@ __declspec(safe) static unmasked inline unsigned int8
 }
 
 __declspec(safe) static unmasked inline int8 __fast_idiv(int8 numerator, uniform int8 divisor) {
-    if (__is_xe_target) {
+    if (divisor == 1) {
+        return numerator;
+    } else if (__is_xe_target) {
         return __idiv_int8(numerator, divisor);
     }
     uniform int8 method = __idiv_table_s8[divisor - 2][0];
@@ -5771,7 +5775,9 @@ __declspec(safe) static unmasked inline int8 __fast_idiv(int8 numerator, uniform
 
 __declspec(safe) static unmasked inline unsigned int16
     __fast_idiv(unsigned int16 numerator, uniform unsigned int16 divisor) {
-    if (__is_xe_target) {
+    if (divisor == 1) {
+        return numerator;
+    } else if (__is_xe_target) {
         return __idiv_uint16(numerator, divisor);
     }
     uniform int64 method = __idiv_table_u16[divisor - 2][0];
@@ -5793,7 +5799,9 @@ __declspec(safe) static unmasked inline unsigned int16
 }
 
 __declspec(safe) static unmasked inline int16 __fast_idiv(int16 numerator, uniform int16 divisor) {
-    if (__is_xe_target) {
+    if (divisor == 1) {
+        return numerator;
+    } else if (__is_xe_target) {
         return __idiv_int16(numerator, divisor);
     }
     uniform int64 method = __idiv_table_s16[divisor - 2][0];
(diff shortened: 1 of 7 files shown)


e3m3:
> What the paper says about this case? Is it a problem in the algorithm or in the implementation?

Here's a link to the [original paper](https://gmplib.org/~tege/divcnst-pldi94.pdf). For signed integer division, the fast path added here is a base case of the algorithm (Fig 5.2). For unsigned division, it looks like it is not needed (Fig 4.X), but our implementation doesn't handle 1/1 correctly as is.

Though we reference Halide and the paper, it looks like our implementation is not pulled directly from the Halide project, so there may be bugs introduced when translating the paper to our implementation.

e3m3:
> Though we reference Halide and the paper, it looks like our implementation is not pulled directly from the Halide project, so there may be bugs introduced when translating the paper to our implementation.

The [tabular implementation](https://github.com/halide/Halide/blob/601b5c539bf33d0bc6d4a6b31f52bf85b38a94ea/src/IntegerDivisionTable.cpp#L19) has evolved slightly compared to ours. It looks there are different/more entries in the tables, and each entry has an additional field.

dbabokin:
If we need to make `__fast_idiv()` handle division by 1, we should not introduce extra branch on the critical path for every invocation / every other divisor values. The simplest and probably the most efficient fix would be to index in the table by `divisor - 1`, not by `divisor - 2`, as it's done today. And for `divisor` equal to `1` do the shift by `0`.

The other thing is that `__is_xe_target` is compile time constant and on of it's branches will be eliminated in compile time. Adding unrelated logic to this `if` statement makes code less clear.

And the last and probably the most important point - `__fast_idiv()` is internal builtin, which is not supposed to be called by the user. Moreover, currently it's not used by the compiler as well. Originally it was used to do the fast division of a variable by a small constant and the constant was assumed to be `>= 2`, so `1` was not handled (it was introduced by https://github.com/ispc/ispc/commit/83e1630fbcfde4aa67b50245cd96e36cbe033660), but later on it was disabled by https://github.com/ispc/ispc/commit/cbeab3e095abe7226afc0ca0651efa74cab1a78a. `__fast_div` was left in stdlib as a rudiment.

I think it's safe not to remove `__fast_idiv` implementation, but I still would like to understand why these tests cause failures for `nuw` bit changes.

e3m3:
> If we need to make `__fast_idiv()` handle division by 1, we should not introduce extra branch on the critical path for every invocation / every other divisor values. The simplest and probably the most efficient fix would be to index in the table by `divisor - 1`, not by `divisor - 2`, as it's done today. And for `divisor` equal to `1` do the shift by `0`.
> 
> The other thing is that `__is_xe_target` is compile time constant and on of it's branches will be eliminated in compile time. Adding unrelated logic to this `if` statement makes code less clear.

The `is_xe_target` branch can be moved above the `divisor` check to maintain the compile-time optimization. I think having the base case as it is in the paper's figure is clearer, and isn't significantly less efficient: we already have branching for the `method` [check](https://github.com/ispc/ispc/blob/e76599a2669a91f51f676a61d1c420206c4da775/stdlib.ispc#L5741) on every invocation based on `divisor - 2` expression as you mention.

> And the last and probably the most important point - `__fast_idiv()` is internal builtin, which is not supposed to be called by the user. Moreover, currently it's not used by the compiler as well. Originally it was used to do the fast division of a variable by a small constant and the constant was assumed to be `>= 2`, so `1` was not handled (it was introduced by [83e1630](https://github.com/ispc/ispc/commit/83e1630fbcfde4aa67b50245cd96e36cbe033660)), but later on it was disabled by [cbeab3e](https://github.com/ispc/ispc/commit/cbeab3e095abe7226afc0ca0651efa74cab1a78a). `__fast_div` was left in stdlib as a rudiment.
> 
> I think it's safe not to remove `__fast_idiv` implementation, but I still would like to understand why these tests cause failures for `nuw` bit changes.

You mean, that it would be preferred to remove the function rather than patch it?

Regarding the `nsw` bit patch, the reason those tests fail is because of a corruption in the control flow which results in evaluating the inner loop with `div == 1`, even though this value is not in the original range of tests values (`[2,INT_MAX]`). Since this value was not handled, the tests failed. However, the real fix requires fixing the control flow, which at the moment I believe is caused by overflow in the inner induction variable (but only for `uint8` and `uint16`).

dbabokin:
> You mean, that it would be preferred to remove the function rather than patch it?

Yes. Because it's not used anywhere in the compiler.

e3m3:
CI is green.

e3m3:
> > ...
> > Regarding the `nsw` bit patch, the reason **those tests** fail is because of a corruption in the control flow which results in evaluating the inner loop with `div == 1`, even though this value is not in the original range of tests values (`[2,INT_MAX]`).
> > ....
> 
> I missed some context here. Which tests are you referring to?

`tests/idiv-uint8.ispc` and `tests/idiv-uint16.ispc` for PR #2494.

dbabokin:
Are you planning to remove `__fast_idiv()` or close this PR?

e3m3:
See https://github.com/ispc/ispc/pull/2574#issue-1812451252.

PR #2573, opened 2023-07-19
Types: Add IsSignedType and GetAsSignedType methods

e3m3:
Changes:
*   Add `IsSignedType` and `GetAsSignedType` counterparts to `IsUnsignedType` and `GetAsUnsignedType`.
    | `src/type.cpp`, `src/type.h`

### DIFF
--- a/src/type.cpp
+++ b/src/type.cpp
@@ -220,6 +220,10 @@ bool AtomicType::IsUnsignedType() const {
             basicType == TYPE_UINT64);
 }
 
+bool AtomicType::IsSignedType() const {
+    return (basicType == TYPE_INT8 || basicType == TYPE_INT16 || basicType == TYPE_INT32 || basicType == TYPE_INT64);
+}
+
 bool AtomicType::IsBoolType() const { return basicType == TYPE_BOOL; }
 
 bool AtomicType::IsConstType() const { return isConst; }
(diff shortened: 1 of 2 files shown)


src/type.cpp
```diff
@@ -1349,6 +1381,8 @@ bool ArrayType::IsIntType() const { return false; }
 
 bool ArrayType::IsUnsignedType() const { return false; }
 
+bool ArrayType::IsSignedType() const { return false; }
```
nurmukhametov, line 1384: I don't understand why `ArrayType::IsSignedType` is not derived from element type like it is done, e.g., for `VectorType`.

e3m3:
@dbabokin @aneshlya @nurmukhametov Please review this prerequisite PR for #2494. This PR only adds the getter functions necessary to mirror the preexisting `IsUnsignedType` utilities.

src/type.cpp
dbabokin: Array is an array, it's not signed and not unsigned, not fp and not int. It's an array. I.e. if the request if it's array of signed types, then appropriate request needs to be made. But the one cannot make conclusions like "it's signed int type, so I can do integer arithmetic on it, load or store it" out of this check, which would be misleading.

dbabokin (approved):
Looks good.

I don't really understand why we need `GetAsSignedType()`, as I think there are no plans to use it right now. But ok, let it be.

e3m3:
Thanks @dbabokin

PR #2574, opened 2023-07-19
stdlib: Remove unused fast integer division

e3m3:
Changes:
*   Remove unused function `__fast_idiv` and tests.
    | `fail_db.txt`,
    | `stdlib.ispc`,
    | `tests/idiv-int16.ispc`,
    | `tests/idiv-int32.ispc`,
    | `tests/idiv-int8.ispc`,
    | `tests/idiv-uint16.ispc`,
    | `tests/idiv-uint32.ispc`,
    | `tests/idiv-uint8.ispc`

See [discussion](https://github.com/ispc/ispc/pull/2570#issuecomment-1638973832) for PR #2570.


dbabokin (changes requested):
Generally looks good, but as we discussed, the tests are checking an interesting property - trying out for loops with different types of induction variables and boundaries around the upper bound of type range. I think we should add one test that checks that this kind of loops.

e3m3:
> Generally looks good, but as we discussed, the tests are checking an interesting property - trying out for loops with different types of induction variables and boundaries around the upper bound of type range. I think we should add one test that checks that this kind of loops.

We should add tests, but this should be done in a separate PR as it would be unrelated to integer fast division. This PR blocks #2494.

dbabokin:
I think it should be part of this PR, as it reduces test coverage.

And I don't see how it blocks #2494. These tests were problematic for that PR before, but not now. And these tests were a problem for a very good reason - they enabled finding a problem in the PR.

e3m3:
> I think it should be part of this PR, as it reduces test coverage.
> 
> And I don't see how it blocks #2494. These tests were problematic for that PR before, but not now. And these tests were a problem for a very good reason - they enabled finding a problem in the PR.

So we should keep the test and merge PR #2570, as the test is useful and the function may be shown to benefit some applications.

dbabokin:
`__fast_div()` should be removed, as it's not used.

The tests that check `__fast_div()` has useful property that is irrelevant to specific builtin - they contain loops with induction variables of different types, which boundaries at the maximum value of the type. This property has enabled finding a problem with your changes in PR #2494. It has nothing with `__fast_div()`.

e3m3:
> `__fast_div()` should be removed, as it's not used.
> 
> The tests that check `__fast_div()` has useful property that is irrelevant to specific builtin - they contain loops with induction variables of different types, which boundaries at the maximum value of the type. This property has enabled finding a problem with your changes in PR #2494. It has nothing with `__fast_div()`.

See new [test](https://github.com/ispc/ispc/pull/2574/files#diff-3e63ae690e161ac9a106d9f7e8d8740ca9a0cbaba392482ec9f1b69d2f993532).

tests/ind-var-bounds-and-types.ispc
```diff
@@ -0,0 +1,34 @@
+#include "../test_static.isph"
+task void f_v(uniform float RET[]) {
+    uniform int errorCount = 0;
+
+    for (unsigned int8 a = 0; a < 256; ++a) {
```
e3m3, line 5:
```suggestion
    for (unsigned int8 a = 0; a < 255; ++a) {
```

tests/ind-var-bounds-and-types.ispc
```diff
@@ -0,0 +1,40 @@
+#include "../test_static.isph"
+task void f_v(uniform float RET[]) {
+    uniform int errorCount = 0;
+
+    for (unsigned int8 a = 1; a < 255; ++a) {
+        for (uniform unsigned int8 b = 1; b < 64; b *= 2) {
```
dbabokin, line 6: 1. No need for inner loop. The property that we'd like to keep is just outer loop iteration space / induction variable type.
2. What'about keeping all the types? [u]int[8|16|32]?
3. The check inside must be preferably the cheapest possible. `pow(double, double)` is not cheap.
4. It's ok to remove any control flow from the loop, as we can speed up the test this way. I.e. verifying just the end result should ok.

dbabokin: Just in case, I don't thnik that https://github.com/ispc/ispc/pull/2574/commits/7a27f6bea740769aef105476a283a3899186bc1a addresses any of these concerns.

e3m3: Right - updating now. 7a27f6b was a fix for the computation in the previous version

e3m3: What do you think of the new test? 5ee16db55e7f23ed3f81eeb92daf836c828b6769

dbabokin: If adding 64 bit types, start not from 0, but close to the end of the range. Same for 32 bits to reduce the runtime.

Please report the runtime  for the test that you observe. Both `-O2` and `-O0`.

e3m3: Updated:
* O0: 0.013 s 
* O2: 0.006 s

aneshlya (approved):

dbabokin (approved):
Looks good.

PR #2575, opened 2023-07-20
Deallocate memory and enable ASAN for examples

nurmukhametov:
This change let us enable ASAN with leak sanitizer on examples. There are still a few deallocations detected by LSAN On LIT tests and runtime tests.

### DIFF
--- a/src/ctx.cpp
+++ b/src/ctx.cpp
@@ -500,7 +500,7 @@ void FunctionEmitContext::StartVaryingIf(llvm::Value *oldMask) {
 }
 
 void FunctionEmitContext::EndIf() {
-    CFInfo *ci = popCFState();
+    std::unique_ptr<CFInfo> ci(popCFState());
     // Make sure we match up with a Start{Uniform,Varying}If().
     AssertPos(currentPos, ci->IsIf());
 
@@ -583,7 +583,7 @@ void FunctionEmitContext::StartLoop(llvm::BasicBlock *bt, llvm::BasicBlock *ct, 
 }
 
 void FunctionEmitContext::EndLoop() {
-    CFInfo *ci = popCFState();
+    std::unique_ptr<CFInfo> ci(popCFState());
     AssertPos(currentPos, ci->IsLoop());
 
     if (!ci->IsUniform())
@@ -632,7 +632,7 @@ void FunctionEmitContext::StartForeach(ForeachType ft, bool isEmulatedUniform) {
 }
 
 void FunctionEmitContext::EndForeach() {
-    CFInfo *ci = popCFState();
+    std::unique_ptr<CFInfo> ci(popCFState());
     AssertPos(currentPos, ci->IsForeach());
 }
 
@@ -885,7 +885,7 @@ void FunctionEmitContext::StartSwitch(bool cfIsUniform, llvm::BasicBlock *bbBrea
 void FunctionEmitContext::EndSwitch() {
     AssertPos(currentPos, bblock != nullptr);
 
-    CFInfo *ci = popCFState();
+    std::unique_ptr<CFInfo> ci(popCFState());
     if (ci->IsVarying() && bblock != nullptr)
         restoreMaskGivenReturns(ci->savedMask);
 }
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -244,7 +244,9 @@ class ArgFactory {
   private:
     char *AllocateString(std::string string) {
         int len = string.length();
-        char *ptr = new char[len + 1];
+        // We use malloc here because of strdup in lAddSingleArg
+        char *ptr = (char *)malloc(len + 1);
+        memset(ptr, 0, len + 1);
         strncpy(ptr, string.c_str(), len);
         ptr[len] = '\0';
         return ptr;
@@ -328,7 +330,7 @@ class StringArgFactory : public ArgFactory {
 };
 
 // Forward reference
-static void lAddSingleArg(char *arg, std::vector<char *> &argv);
+static void lAddSingleArg(char *arg, std::vector<char *> &argv, bool duplicate);
 
 /** Add all args from a given factory to the argv passed as parameters, which could
  *  include recursing into another ArgFactory.
@@ -338,7 +340,7 @@ static void lAddArgsFromFactory(ArgFactory &Args, std::vector<char *> &argv) {
         char *NextArg = Args.GetNextArg();
         if (NextArg == nullptr)
             break;
-        lAddSingleArg(NextArg, argv);
+        lAddSingleArg(NextArg, argv, false);
     }
 }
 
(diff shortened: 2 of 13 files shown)


src/ctx.cpp
```diff
@@ -506,8 +506,10 @@ void FunctionEmitContext::EndIf() {
 
     // 'uniform' ifs don't change the mask so we only need to restore the
     // mask going into the if for 'varying' if statements
-    if (ci->IsUniform() || bblock == nullptr)
```
dbabokin, line 509: Would making `ci` a `unique_ptr` be a better / more elegant solution? I.e. declare it as `std::unique_ptr<CFInfo> ci(popCFState());`.

What's your opinion?

src/module.cpp
```diff
@@ -3186,6 +3187,8 @@ int Module::CompileAndOutput(const char *srcFile, Arch arch, const char *cpu, st
         delete g->target;
         g->target = nullptr;
 
+        delete targetMachine;
```
dbabokin, line 3199: A pointer to `llvm::TargetMachine` is owned by `Target`, why not deleting it in its destructor?

src/ctx.cpp
e3m3: I think the main advantage would be in the case the assertion is tripped, the destructor will still be called.

e3m3: Actually, I think this may be an issue for the path where the predicate is false, since the pointer should remain undeleted according to the current implementation. With unique_ptr, it will go out of scope later and be destructed. Unique_ptr assumes ownership.

src/main.cpp
```diff
@@ -244,7 +244,9 @@ class ArgFactory {
   private:
     char *AllocateString(std::string string) {
         int len = string.length();
-        char *ptr = new char[len + 1];
+        // We use malloc here because of strdup in lAddSingleArg
```
dbabokin, line 247: I'm not sure how is this related. Could you elaborate?

dbabokin: Ah, I see. You want to deallocate both of them in the same manner and it ha to be `free()`.

I think a better solution would be to use only C++ primitives to work with strings, but this that would require more changes. So this is ok.

src/ctx.cpp
dbabokin: I don't understand the last comment. Both explicit delete and `uniquie_ptr` will not change the current logic in this function.

e3m3: I mean that as it is now, if `! (ci->IsUniform() || bblock == nullptr)`, there's no call to delete.

e3m3: Ah never mind, I was looking at the left diff.

nurmukhametov: I am not sure about `unique_ptr`.
The problem here is that `popCFState` doesn't really create or allocate `CFInfo` object. The lifetime of `CFInfo` objects is the following (let's consider `if` only):
- Created via allocation on heap in `CFInfo::GetIf`: https://github.com/ispc/ispc/blob/f080d39c99decb9947e7fa5688a56773f841fe11/src/ctx.cpp#L145
- `CFInfo::GetIf` returns a pointer to a newly allocated object that stored in `controlFlowInfo` vector in two places:
https://github.com/ispc/ispc/blob/f080d39c99decb9947e7fa5688a56773f841fe11/src/ctx.cpp#L495
and
https://github.com/ispc/ispc/blob/f080d39c99decb9947e7fa5688a56773f841fe11/src/ctx.cpp#L499
- The pointer is stored there until being poped in `popCFState` https://github.com/ispc/ispc/blob/f080d39c99decb9947e7fa5688a56773f841fe11/src/ctx.cpp#L3771
- Then this pointer is returned by `popCFState` and accessed, e.g., in `EndIf` here.
So, we may bind lifetime of `CFInfo` object inside `CFInfo::GetIf` (and similar functions) to stack allocated `unique_ptr` but not the whole.

src/module.cpp
nurmukhametov: Unfortunately, `Target` doesn't really own `llvm::TargetMachine *m_targetMachine`, see comment here: https://github.com/ispc/ispc/blob/f080d39c99decb9947e7fa5688a56773f841fe11/src/ispc.h#L354

and example of usage after `Target` object deconstruction:
- `Target` created https://github.com/ispc/ispc/blob/f080d39c99decb9947e7fa5688a56773f841fe11/src/module.cpp#L3247
- Ownership of `m_targetMachine` is passed to the `targetMachines` https://github.com/ispc/ispc/blob/f080d39c99decb9947e7fa5688a56773f841fe11/src/module.cpp#L3260
- `Target` destroyed https://github.com/ispc/ispc/blob/f080d39c99decb9947e7fa5688a56773f841fe11/src/module.cpp#L3321
- `targetMachine` is used after that here https://github.com/ispc/ispc/blob/f080d39c99decb9947e7fa5688a56773f841fe11/src/module.cpp#L3340

src/main.cpp
nurmukhametov: Yes, exactly.
`argv` vector consisted of pointers to C string of two origins:
- real arguments of program from `Argv` of `main`
- heap allocated strings.
They can be mixed in order inside `argv` so I didn't find a easy way to distinguish them in order to deallocate the last ones.

I have ended up with the current fix that `strdup` the first ones and allocate the last ones with `malloc` to deallocate all of them with `free`.

src/ctx.cpp
nurmukhametov: Bearing that in mind, I have fixed that to use `unique_ptr` instead explicit `delete ci` that is better especially when multiple exit points exist.

dbabokin: `unique_ptr` doesn't need to manage the whole lifetime from creation to destruction, managing it from popping to destruction is ok from philosophical point of view.

src/module.cpp
dbabokin: Ok, that's caused by messy way of doing multi-target compilation.

aneshlya (approved):

src/module.cpp
dbabokin: I've had a deeper look and I think this needs to be cleaned up. Specifically:
- `TargetMachine` really belongs to the `Target` and needs to be destroyed in its destructor.
- The cases when it escapes `Target` lifetime are not legitimate:
  - `targetMachines[]` array should be a bool array with a name "CompiledTargets" or something like that.
  - the use of `firstTargetMachine` should be turned into `g->target->GetTargetMachine()`, as for the dispatch module we create a new `Target` anyway and it creates a `TargetMachine` as well - and this one needs to be used.

I'll proceed with merging this PR, and will create an issue to follow up with the clean up explained here.

dbabokin: I've created https://github.com/ispc/ispc/issues/2578 to follow up.

dbabokin (approved):
LGTM

PR #2576, opened 2023-07-20
Fix GCC warnings and MSVC /WX build into CI

nurmukhametov:
This fixes and suppresses GCC warnings, so it is green (https://github.com/nurmukhametov/ispc/actions/runs/5613763085/job/15210488452).

MSVC job with /WX (analogue of -Werror) flag was added also.

### DIFF
--- a/.github/workflows/compiler-warnings.yml
+++ b/.github/workflows/compiler-warnings.yml
@@ -67,3 +67,33 @@ jobs:
       run: |
           cmake -B build -DCMAKE_C_COMPILER=gcc-11 -DCMAKE_CXX_COMPILER=g++-11 -DCMAKE_C_FLAGS="-Werror" -DCMAKE_CXX_FLAGS="-Werror"
           cmake --build build --target ispc -j4
+
+  msvc-2022:
+    runs-on: windows-2022
+    env:
+      LLVM_VERSION: "15.0"
+      LLVM_TAR: llvm-15.0.7-win.vs2019-Release+Asserts-x86.arm.wasm.tar.7z
+      LLVM_HOME: "C:\\projects\\llvm"
+      CROSS_TOOLS_GNUWIN32: "C:\\projects\\cross\\gnuwin32"
+      BUILD_TYPE: "Release"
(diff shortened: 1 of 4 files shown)


.github/workflows/compiler-warnings.yml
```diff
@@ -67,3 +67,38 @@ jobs:
...
+      BUILD_TYPE: "Release"
+      ISPC_OPAQUE_PTR_MODE: "ON"
+
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: true
+
+    - name: Setup MSVC dev cmd
+      uses: ilammy/msvc-dev-cmd@v1
```
dbabokin, line 87: Would be good to avoid third-party actions from unverified developers.

Would this actions work instead https://github.com/microsoft/setup-msbuild ?

dbabokin:
Except the third party action the changes look good.

.github/workflows/compiler-warnings.yml
```diff
@@ -65,5 +65,40 @@ jobs:
...
+    - name: Check environment
+      shell: cmd
+      run: |
+        wmic cpu get caption, deviceid, name, numberofcores, maxclockspeed, status
+
+    - name: Build ISPC with MSVC in -Werror mode
+      shell: cmd
+      run: |
+        cmake -B build -DCMAKE_C_FLAGS="/WX" -DCMAKE_CXX_FLAGS="/WX" -DISPC_CROSS=ON -DCMAKE_BUILD_TYPE=Release -DISPC_GNUWIN32_PATH=%CROSS_TOOLS_GNUWIN32% -DISPC_OPAQUE_PTR_MODE=%ISPC_OPAQUE_PTR_MODE%
+        cmake --build build --target ispc -j4 --config Release
```
aneshlya, line 104: Is there any reason why only ispc build is checked in -Werror mode and not the whole package?

.github/workflows/compiler-warnings.yml
nurmukhametov: OK, moreover neither of them looks needed.

dbabokin (approved):
LGTM

PR #2577, opened 2023-07-21
LLVM 16 support in ISPC

aneshlya:
This PR enables build of ISPC with LLVM 16.0 for CPU and Xe targets.
Switch to LLVM 16.0 by default is not possible yet due to regressions to Xe which are under investigation.


.github/workflows/ispc-ci.yml
```diff
@@ -626,6 +703,48 @@ jobs:
         name: ispc_llvm15_win
         path: build/ispc-trunk-windows.zip
 
+  win-build-ispc-llvm16:
```
dbabokin, line 706: What about macOS with LLVM16?

dbabokin (approved):
Please address the comment and it's good to go!

PR #2579, opened 2023-07-21
Use self-hosted macOS runner for LLVM builds

dbabokin:


aneshlya (approved):

dbabokin:
Self-hosted macOS runner appears to be not ready: too old git and not installed XCode. I'll update the runner and will covert PR from draft to regular PR.

aneshlya (approved):
LGTM

PR #2580, opened 2023-07-22
wasm support

i-saint:
This PR fixes wasm32 support and adds support for wasm64.
It also addresses compile errors with LLVM 16.0. Compiling to wasm64 causes a crash if LLVM is 15.0 or older due to this bug: llvm/llvm-project#62443

I hope wasm target will be enabled in public releases.


i-saint:
I noticed https://github.com/ispc/ispc/pull/2577 right now. I will update this PR later.

dbabokin:
#2577 was merged, please rebase

cmake/GenerateBuiltins.cmake
```diff
@@ -400,7 +408,7 @@ function (generate_common_builtins resultList)
     message (STATUS "ISPC will be built with support of ${supported_oses} for ${supported_archs}")
     foreach (bit 32 64)
         foreach (os_name "windows" "linux" "freebsd" "macos" "android" "ios" "ps4" "web")
-            foreach (arch "x86" "arm" "wasm32")
+            foreach (arch "x86" "arm" "wasm32" "wasm64")
```
dbabokin, line 411: `wasm32` here needs to be changed to `wasm`. Though `target_arch` in `builtin_to_cpp()` needs to be still `wasm32` and `wasm64`.

I.e. it needs to be similar to `x86` (which triggers `x86` and `x86_64`) and `arm` (which triggers `armv7` and `aarch64`).

src/main.cpp
```diff
@@ -1205,8 +1205,9 @@ int main(int Argc, char *Argv[]) {
 
     for (auto target : targets) {
         if (target == ISPCTarget::wasm_i32x4) {
-            Assert(targets.size() == 1 && "wasm32 supports only one target: i32x4");
```
dbabokin, line 1208: Why removing the assert? The text needs to be modified, but still there only one target for wasm: `wasm_i32x4`.

dbabokin (changes requested):
Please address the comment in `cmake/GenerateBuiltins.cmake`, the rest looks good.

Note that WASM target is not complete and needs more work to be fully functional.

src/main.cpp
i-saint: you are right. that was a mistake.

src/main.cpp
```diff
@@ -1128,6 +1128,30 @@ int main(int Argc, char *Argv[]) {
...
     }
 
+#ifdef ISPC_WASM_ENABLED
+    // Default setting for wasm
+    if (arch == Arch::wasm32 || arch == Arch::wasm64) {
+        g->target_os = TargetOS::web;
+    }
+    for (auto target : targets) {
+        if (target == ISPCTarget::wasm_i32x4) {
+            Assert(targets.size() == 1 && "wasm supports only one target: i32x4");
```
i-saint, line 1138: Assert has moved here.

aneshlya:
@i-saint , have you tried to run tests for wasm64 using `run_tests.py`? I think some changes are missed in run_tests.py for `wasm64`, it has several code paths like ```if target.arch == "wasm32":```. `wasm64` should be added there as well.

aneshlya:
> @i-saint , have you tried to run tests for wasm64 using `run_tests.py`? I think some changes are missed in run_tests.py for `wasm64`, it has several code paths like `if target.arch == "wasm32":`. `wasm64` should be added there as well.

Feel free to cherry-pick https://github.com/aneshlya/ispc/commit/927a94c6b4336135f4d1e09afc9d9f00415a8329 from my fork aneshlya:wasm64

Please rebase and it's ready to be merged.

run_tests.py
```diff
@@ -465,9 +467,11 @@ def run_test(testname, host, target):
                 ispc_cmd += " -O2"
 
         exe_wd = "."
-        if target.arch == "wasm32":
+        if target.arch == "wasm32" or target.arch == "wasm64":
             cc_cmd += " -D__WASM__"
-            options.wrapexe = "v8 --experimental-wasm-simd"
+            options.wrapexe = "v8"
```
i-saint, line 472: on recent version of v8, simd is standardized and "--experimental-wasm-simd" option is no longer exists.

run_tests.py
```diff
@@ -465,9 +467,11 @@ def run_test(testname, host, target):
                 ispc_cmd += " -O2"
 
         exe_wd = "."
-        if target.arch == "wasm32":
+        if target.arch == "wasm32" or target.arch == "wasm64":
             cc_cmd += " -D__WASM__"
-            options.wrapexe = "v8 --experimental-wasm-simd"
+            options.wrapexe = os.environ["EMSDK_NODE"]
```
i-saint, line 472: emsdk includes node.js. we can use it for wasm tests. it is easier to maintain and more practical than using v8 directly.
also, simd is standardized in recent versions and "--experimental-wasm-simd" option is no longer exists.

src/ispc.cpp
```diff
@@ -1362,7 +1362,7 @@ Target::Target(Arch arch, const char *cpu, ISPCTarget ispc_target, bool pic, MCM
         this->m_dataTypeWidth = 32;
         this->m_vectorWidth = 4;
         this->m_hasHalfConverts = false;
-        this->m_hasHalfFullSupport = true;
+        this->m_hasHalfFullSupport = false;
```
i-saint, line 1365: wasm doesn't support half. so, this should be false. it had been causing link errors such as "undefined symbol: __rsqrt_varying_half".

i-saint:
now ```run_tests.py --arch=wasm32 --target=wasm-i32x4```  pass (or skip) all tests.
on wasm64 some tests fail, but it seems to be LLVM's problem (wasm64 is still in active development). I will continue to maintain.

aneshlya (approved):
Looks good!

aneshlya:
@i-saint, thanks a lot for your contribution! Please be aware that, from our perspective, WASM is not yet production-ready. It will not be enabled by default in the upcoming v1.21.0 release. However, we are committed to working towards enabling it in future releases.

PR #2581, opened 2023-07-24
Destroy TargetMachine in Target destructor.

### LINKED ISSUE #2578: Clean up ownership and lifetime of TargetMachine
This is a follow up on the [review](https://github.com/ispc/ispc/pull/2575#discussion_r1271114507) of [PR2575](https://github.com/ispc/ispc/pull/2575).

- `TargetMachine` really belongs to the `Target` and needs to be destroyed in its destructor.
- The cases when it escapes `Target` lifetime are not legitimate:
  - `targetMachines[]` array should be a bool array with a name "compiledTargets" or something like that.
  - the use of first `TargetMachine` should be turned into `->target->GetTargetMachine()`, as for the dispatch module we create a new `Target` anyway and it creates a `TargetMachine` as well - and this one needs to be used.

This should make the design a little cleaner and reduce memory footprint for multi-target compilation because of earlier release of `TargetMachine`.

nurmukhametov:
This is a follow up on the review of PR2575 and fixes #2578.

TargetMachine really belongs to the Target and needs to be destroyed in its destructor.

The cases when it escapes Target lifetime are not legitimate. For the dispatch module we create a new Target anyway and it creates a TargetMachine as well - and this one needs to be used.

This should make the design a little cleaner and reduce memory footprint for multi-target compilation because of earlier release of TargetMachine.

### DIFF
--- a/src/ispc.cpp
+++ b/src/ispc.cpp
@@ -1732,6 +1732,8 @@ Target::~Target() {
         delete m_dataLayout;
     if (m_tf_attributes)
         delete m_tf_attributes;
+    if (m_targetMachine)
+        delete m_targetMachine;
 }
 
 bool Target::checkIntrinsticSupport(llvm::StringRef name, SourcePos pos) {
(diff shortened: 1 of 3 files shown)


dbabokin (approved):
Looks good.

PR #2582, opened 2023-07-24
Support macOS in superbuild

nurmukhametov:
This PR supports macOS in superbuild CMake. It supports universal binaries generation with `-DMACOS_UNIVERSAL_BIN=ON`.

### DIFF
--- a/superbuild/CMakeLists.txt
+++ b/superbuild/CMakeLists.txt
@@ -32,27 +32,33 @@ check_cache_var_declaration(L0_TAG "Level Zero url")
 check_cache_var_declaration(ISPC_CORPUS_URL "ispc-corpus url")
 
 set(GNUWIN32 "c:/gnuwin32" CACHE STRING "Set path to gnuwin32 root dir (Windows only)")
-set(CCACHE_VERSION "4.7.4" CACHE STRING "ccache version ")
+set(CCACHE_VERSION "4.8.2" CACHE STRING "ccache version ")
 set(CCACHE_DIR "default" CACHE STRING "ccache cache directory")
+set(PREBUILT_STAGE1 "NO" CACHE STRING "Provide pre-built stage1 archive to skip stage1 jobs")
+set(PREBUILT_STAGE1_PATH "NO" CACHE STRING "Provide pre-built stage1 toolchain path")
 set(PREBUILT_STAGE2 "NO" CACHE STRING "Provide pre-built stage2 archive to skip pre ispc-stage2 jobs")
 set(PREBUILT_STAGE2_PATH "NO" CACHE STRING "Provide pre-built stage2 toolchain path")
-set(STAGE2_TOOLCHAIN_INSTALL_PREFIX "NO" CACHE STRING "Path to install stage2 Clang/LLVM to")
 set(BUILD_TYPE_ISPC "Release" CACHE STRING "Set the build type for ISPC build")
 set(BUILD_TYPE_LLVM "Release" CACHE STRING "Set the build type for LLVM build")
 set(BUILD_TYPE_SPIRV_TRANSLATOR "Release" CACHE STRING "Set the build type for SPIR-V Translator build")
 set(BUILD_TYPE_VC_INTRINSICS "Release" CACHE STRING "Set the build type for VC Intrinsics build")
 set(BUILD_TYPE_L0 "Release" CACHE STRING "Set the build type for L0 build")
+set(ISPC_ANDROID_NDK_PATH "NO" CACHE STRING "Path to Android NDK. It is needed to ISPC_CROSS=ON on macOS.")
 option(LLVM_DISABLE_ASSERTIONS "Enable LLVM build without assertions" ON)
 option(EXTERNAL_VERBOSE "Enable verbose for external project build and install commands" OFF)
 option(CCACHE "Enable ccache to speed up repetitive builds" OFF)
 option(FULL_STAGE2_LLVM "Enable full llvm stage2 build" OFF)
 option(LTO "Enable LTO" OFF)
 option(PGO "Enale PGO (implicitly enable LTO)" OFF)
+option(BUILD_STAGE1_TOOLCHAIN_ONLY "Enable build of stage1 toolchain only." OFF)
 option(BUILD_STAGE2_TOOLCHAIN_ONLY "Enable build of stage2 toolchain only, i.e., skip ispc building and stage3." OFF)
 option(BUILD_SPIRV_TRANSLATOR_ONLY "Enable only build of SPIRV-LLVM-Translator" OFF)
 option(BUILD_VC_INTRINSICS_ONLY "Enable only build of vc-intrinsics" OFF)
 option(BUILD_L0_LOADER_ONLY "Enable only build of level-zero loader" OFF)
-option(XE_DEPS "Enable build of XE dependencies (L0 loader, vc-intrinsics, SPIRV-LLVM-Translator" ON)
+option(INSTALL_ISPC "Enable installation of ISPC to CMAKE_INSTALL_PREFIX" OFF)
+option(INSTALL_WITH_XE_DEPS "Enable installation ISPC or stage2 toolchain with XE_DEPS to CMAKE_INSTALL_PREFIX" OFF)
+option(XE_DEPS "Enable build of all XE dependencies (L0 loader, vc-intrinsics, SPIRV-LLVM-Translator" ON)
+option(MACOS_UNIVERSAL_BIN "Enable build of universal binaries for macOS" ON)
 
 # Force LTO when PGO enabled to the sake of simplicity and less number of
 # possible build variants.
@@ -60,8 +66,19 @@ if (PGO)
     set(LTO ON)
 endif()
 
+set(SKIP_STAGE1 OFF)
+if (NOT PREBUILT_STAGE1 STREQUAL "NO" OR NOT PREBUILT_STAGE1_PATH STREQUAL "NO")
+    set(SKIP_STAGE1 ON)
+    set(FULL_STAGE2_LLVM ON)
+endif()
+
+if (NOT PREBUILT_STAGE1 STREQUAL "NO" AND NOT PREBUILT_STAGE1_PATH STREQUAL "NO")
+    message(FATAL_ERROR "Provide only one source of stage1 toolchain.")
+endif()
+
+set(SKIP_STAGE2 OFF)
 if (NOT PREBUILT_STAGE2 STREQUAL "NO" OR NOT PREBUILT_STAGE2_PATH STREQUAL "NO")
-    set(SKIP_STAGE2_DEPS ON)
+    set(SKIP_STAGE2 ON)
 endif()
 
 if (NOT PREBUILT_STAGE2 STREQUAL "NO" AND NOT PREBUILT_STAGE2_PATH STREQUAL "NO")
@@ -69,9 +86,57 @@ if (NOT PREBUILT_STAGE2 STREQUAL "NO" AND NOT PREBUILT_STAGE2_PATH STREQUAL "NO"
 endif()
 
 set(STAGE2_PATH "${CMAKE_BINARY_DIR}/stage2")
+if (NOT PREBUILT_STAGE1_PATH STREQUAL "NO")
+    set(STAGE2_PATH "${PREBUILT_STAGE1_PATH}")
+endif()
 if (NOT PREBUILT_STAGE2_PATH STREQUAL "NO")
     set(STAGE2_PATH "${PREBUILT_STAGE2_PATH}")
 endif()
+set(STAGE3_PATH "${CMAKE_BINARY_DIR}/stage3")
+
+if (NOT PREBUILT_STAGE1_PATH STREQUAL "NO" AND NOT PREBUILT_STAGE2_PATH STREQUAL "NO")
+    message(FATAL_ERROR "PREBUILT_STAGE1_PATH and PREBUILT_STAGE2_PATH is incompatible together.")
+endif()
+
+if (BUILD_STAGE1_TOOLCHAIN_ONLY OR
+    BUILD_SPIRV_TRANSLATOR_ONLY OR BUILD_VC_INTRINSICS_ONLY OR BUILD_L0_LOADER_ONLY)
+    set(XE_DEPS OFF)
+endif()
+
+set(BUILD_SPIRV_STANDALONE OFF)
+set(BUILD_VC_INTRINSICS_STANDALONE OFF)
+set(BUILD_L0_LOADER_STANDALONE OFF)
+if (INSTALL_WITH_XE_DEPS)
+    set(BUILD_SPIRV_STANDALONE ON)
+    set(BUILD_VC_INTRINSICS_STANDALONE ON)
+    set(BUILD_L0_LOADER_STANDALONE ON)
+endif()
+if (BUILD_SPIRV_TRANSLATOR_ONLY)
+    set(BUILD_SPIRV_STANDALONE ON)
+endif()
+if (BUILD_VC_INTRINSICS_ONLY)
+    set(BUILD_VC_INTRINSICS_STANDALONE ON)
+endif()
+if (BUILD_L0_LOADER_ONLY)
+    set(BUILD_L0_LOADER_STANDALONE ON)
+endif()
+
+if (APPLE)
+    # XE are not supported on macOS targets
+    set(XE_DEPS OFF)
+
+    set(FULL_STAGE2_LLVM ON)
+
+    set(MACOS_VERSION_MIN 10.12)
+    if (NOT MACOS_UNIVERSAL_BIN AND CMAKE_SYSTEM_PROCESSOR MATCHES "arm64")
+        set(MACOS_VERSION_MIN 11.0)
+    endif()
+
+    if (APPLE AND ISPC_ANDROID_NDK_PATH STREQUAL "NO")
+        message(FATAL_ERROR "You need provide path to ANDROID NDK on macOS via -DISPC_ANDROID_NDK_PATH=..")
+    endif()
+endif()
+
 
 string(APPEND LIT_TOOLS_DIR ${GNUWIN32} "/bin")
 
@@ -157,11 +222,13 @@ message(STATUS "ISPC_CORPUS_URL is ${ISPC_CORPUS_URL}")
 message(STATUS "CCACHE_VERSION is ${CCACHE_VERSION}")
 message(STATUS "CCACHE_DIR is ${CCACHE_DIR}")
 message(STATUS "CCACHE is ${CCACHE}")
+message(STATUS "PREBUILT_STAGE1 is ${PREBUILT_STAGE1}")
+message(STATUS "PREBUILT_STAGE1_PATH is ${PREBUILT_STAGE1_PATH}")
 message(STATUS "PREBUILT_STAGE2 is ${PREBUILT_STAGE2}")
 message(STATUS "PREBUILT_STAGE2_PATH is ${PREBUILT_STAGE2_PATH}")
 message(STATUS "EXTERNAL_VERBOSE is ${EXTERNAL_VERBOSE}")
 message(STATUS "FULL_STAGE2_LLVM is ${FULL_STAGE2_LLVM}")
-message(STATUS "STAGE2_TOOLCHAIN_INSTALL_PREFIX is ${STAGE2_TOOLCHAIN_INSTALL_PREFIX}")
+message(STATUS "ISPC_ANDROID_NDK_PATH is ${ISPC_ANDROID_NDK_PATH}")
 message(STATUS "LLVM_DISABLE_ASSERTIONS is ${LLVM_DISABLE_ASSERTIONS}")
 message(STATUS "LTO is ${LTO}")
 message(STATUS "PGO is ${PGO}")
@@ -176,12 +243,30 @@ message(STATUS "BUILD_STAGE2_TOOLCHAIN_ONLY is ${BUILD_STAGE2_TOOLCHAIN_ONLY}")
 message(STATUS "BUILD_SPIRV_TRANSLATOR_ONLY is ${BUILD_SPIRV_TRANSLATOR_ONLY}")
 message(STATUS "BUILD_VC_INTRINSICS_ONLY is ${BUILD_VC_INTRINSICS_ONLY}")
 message(STATUS "BUILD_L0_LOADER_ONLY is ${BUILD_L0_LOADER_ONLY}")
+message(STATUS "BUILD_SPIRV_STANDALONE is ${BUILD_SPIRV_STANDALONE}")
+message(STATUS "BUILD_VC_INTRINSICS_STANDALONE is ${BUILD_VC_INTRINSICS_STANDALONE}")
+message(STATUS "BUILD_L0_LOADER_STANDALONE is ${BUILD_L0_LOADER_STANDALONE}")
+message(STATUS "INSTALL_ISPC is ${INSTALL_ISPC}")
+message(STATUS "INSTALL_WITH_XE_DEPS is ${INSTALL_WITH_XE_DEPS}")
 message(STATUS "XE_DEPS is ${XE_DEPS}")
 message(STATUS "LLVM_TAG is ${LLVM_TAG}")
 message(STATUS "LLVM_VERSION_DOTTED is ${LLVM_VERSION_DOTTED}")
-message(STATUS "SKIP_STAGE2_DEPS is ${SKIP_STAGE2_DEPS}")
-message(STATUS "STAGE2_BIN is ${STAGE2_BIN}")
+message(STATUS "SKIP_STAGE1 is ${SKIP_STAGE1}")
+message(STATUS "SKIP_STAGE2 is ${SKIP_STAGE2}")
+message(STATUS "STAGE2_PATH is ${STAGE2_PATH}")
 message(STATUS "CMAKE_INSTALL_PREFIX is ${CMAKE_INSTALL_PREFIX}")
+message(STATUS "CMAKE_C_FLAGS is ${CMAKE_C_FLAGS}")
+message(STATUS "CMAKE_CXX_FLAGS is ${CMAKE_CXX_FLAGS}")
+message(STATUS "MACOS_VERSION_MIN is ${MACOS_VERSION_MIN}")
+message(STATUS "MACOS_UNIVERSAL_BIN is ${MACOS_UNIVERSAL_BIN}")
+
+
+if (NOT XE_DEPS)
+    # To suppress CMake warnings about unused user-provided variables.
+    set(ignoreMe "${VC_INTRINSICS_URL}${VC_INTRINSICS_SHA}")
+    set(ignoreMe "${SPIRV_TRANSLATOR_URL}${SPIRV_TRANSLATOR_SHA}${SPIRV_TRANSLATOR_BRANCH}")
+    set(ignoreMe "${L0_URL}${L0_TAG}")
+endif()
 
 
 # Ninja is hard-coded for two reasons:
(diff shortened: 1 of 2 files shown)


superbuild/CMakeLists.txt
```diff
@@ -1096,7 +1097,7 @@ if (NOT BUILD_STAGE2_TOOLCHAIN_ONLY)
     if (NOT STAGE2_ISPC_INSTALL_PREFIX STREQUAL "NO")
         add_custom_target(install-stage2-ispc
             COMMENT "Install stage2 to ${STAGE2_ISPC_INSTALL_PREFIX}"
-            COMMAND ${CMAKE_COMMAND} -E copy_directory ispc-stage2 ${STAGE2_ISPC_INSTALL_PREFIX}
+            COMMAND ${CMAKE_COMMAND} -E copy_directory ${ISPC_STAGE2_INSTALL_PREFIX} ${STAGE2_ISPC_INSTALL_PREFIX}
```
aneshlya, line 1248: What is the difference between `${ISPC_STAGE2_INSTALL_PREFIX}` and `${STAGE2_ISPC_INSTALL_PREFIX}`? When should be used one or another?

superbuild/CMakeLists.txt
```diff
@@ -868,25 +912,70 @@ endif()
...
+        # Note: We may rebuild instead stage2 clang/lld with stage1 toolchain to
+        # generate more performant stage2 compiler.
+
+        list(APPEND LLVM_STAGE2_DEPS stage1)
+
+        if (BUILD_STAGE1_TOOLCHAIN_ONLY)
+            # Custom target to install stage1 toolchain
+            # It may be slightly confusing because stage1 toolchain placed under stage2 folder
+            if (NOT STAGE1_TOOLCHAIN_INSTALL_PREFIX STREQUAL "NO")
+                add_custom_target(install-stage1-toolchain
```
aneshlya, line 1095: Should this target depend on `stage1` as `package-stage1` target?
I would also use `WORKING_DIRECTORY` here.

superbuild/CMakeLists.txt
```diff
@@ -32,29 +32,47 @@ check_cache_var_declaration(L0_TAG "Level Zero url")
...
-set(CCACHE_VERSION "4.7.4" CACHE STRING "ccache version ")
+set(CCACHE_VERSION "4.8.2" CACHE STRING "ccache version ")
 set(CCACHE_DIR "default" CACHE STRING "ccache cache directory")
+set(PREBUILT_STAGE1 "NO" CACHE STRING "Provide pre-built stage1 archive to skip stage1 jobs")
+set(PREBUILT_STAGE1_PATH "NO" CACHE STRING "Provide pre-built stage1 toolchain path")
 set(PREBUILT_STAGE2 "NO" CACHE STRING "Provide pre-built stage2 archive to skip pre ispc-stage2 jobs")
 set(PREBUILT_STAGE2_PATH "NO" CACHE STRING "Provide pre-built stage2 toolchain path")
+set(STAGE1_TOOLCHAIN_INSTALL_PREFIX "NO" CACHE STRING "Path to install stage1 Clang/LLVM to")
 set(STAGE2_TOOLCHAIN_INSTALL_PREFIX "NO" CACHE STRING "Path to install stage2 Clang/LLVM to")
+set(STAGE2_ISPC_INSTALL_PREFIX "NO" CACHE STRING "Path to install stage2 ISPC to")
```
e3m3, line 43: Should the default values here be `"NO"` if they are paths?

superbuild/CMakeLists.txt
```diff
@@ -556,27 +731,36 @@ else()
...
 endif()
-]])
+]] @ONLY)
 
 # Base toolchain file for building ISPC and LLVM dependencies.
 file(CONFIGURE OUTPUT stage2-toolchain.cmake CONTENT [[
 @CCACHE_LAUNCHER_ACTION@(CMAKE_C_COMPILER_LAUNCHER   @CCACHE_EXE@)
 @CCACHE_LAUNCHER_ACTION@(CMAKE_CXX_COMPILER_LAUNCHER @CCACHE_EXE@)
-set(ENV{PATH} "@ENV_PATH_STAGE2@")
+set(ENV{PATH}       "@ENV_PATH_STAGE2@")
```
e3m3, line 740: Should `PATH` be appended to instead of set?

superbuild/README.md
```diff
@@ -35,9 +35,21 @@ $ cmake ${ISPC}/superbuild --preset os -DPREBUILT_STAGE2=/path/to/stage2-archive
...
+# Build and install only dependencies
+$ cmake ${ISPC}/superbuild --preset os -DCMAKE_INSTALL_PREFIX=/opt/spirv-translator -DBUILD_SPIRV_TRANSLATOR_ONLY=ON
+$ cmake --build .
+$ cmake ${ISPC}/superbuild --preset os -DCMAKE_INSTALL_PREFIX=/opt/spirv-translator -DBUILD_VC_INTRINSICS_ONLY=ON
+$ cmake --build .
+$ cmake ${ISPC}/superbuild --preset os -DCMAKE_INSTALL_PREFIX=/opt/spirv-translator -DBUILD_L0_LOADER_ONLY=ON
+$ cmake --build .
+
+# Build and install ISPC with XE dependencies
+$ cmake ${ISPC}/superbuild --preset os -DPREBUILT_STAGE2_PATH=/tmp/stage2-path -DCMAKE_INSTALL_PREFIX=/opt/ispc-with-xe/ -DINSTALL_WITH_XE_DEPS=ON
```
e3m3, line 51: Is `/opt/ispc-with-xe` new?

superbuild/CMakeLists.txt
e3m3: Also is the additional whitespace accidental?

superbuild/CMakeLists.txt
nurmukhametov: Yes, naming here is quite misleading. So, [this patch](https://github.com/ispc/ispc/pull/2582/commits/c6cf7e1e272158df36220a87c2373abcc954d056) adds the support of installing the output of superbuild procedure to the outer folder. 
`ISPC_STAGE2_INSTALL_PREFIX` is internal variable and should not be used by user, its value either `ispc-stage2` folder inside `build` folder or value from `CMAKE_INSTALL_PREFIX` which user should use.

`STAGE2_ISPC_INSTALL_PREFIX` was a user provided var to generate custom target to install only `stage2-ispc`. Probably, this special target is obsolete now when the superbuild takes into consideration `CMAKE_INSTALL_PREFIX`. Let me check that and adjust if needed.

superbuild/README.md
nurmukhametov: Yes

superbuild/CMakeLists.txt
nurmukhametov: It shouldn't because ENV_PATH_STAGE2 is full env not the extra path to append to. It was appended earlier here https://github.com/ispc/ispc/blob/fce2cc3599023ff0c9a5797d8d91f60f597c6b38/superbuild/CMakeLists.txt#L527

superbuild/CMakeLists.txt
nurmukhametov: They are strings, and "NO" here means that no paths have been provided.

superbuild/CMakeLists.txt
nurmukhametov: Yes, it should. Let me reconsider that, I probably need to clean up all `STAGE*_PREFIX` in favor of using `CMAKE_INSTALL_PREFIX` that, I think, let me remove this extra custom target.

superbuild/15_0_apple_disable_ios.patch
```diff
@@ -0,0 +1,26 @@
+From a49d5f847d169fd79bf0f528dd8d4681089d2dbb Mon Sep 17 00:00:00 2001
+From: Alexey Nurmukhamedov <anurmukh@sctpimac1.sc.intel.com>
+Date: Fri, 14 Jul 2023 08:34:42 -0700
+Subject: [PATCH] Fix compiler-rt build issue
+
+see issue https://github.com/llvm/llvm-project/issues/57037
```
dbabokin, line 6: Do we need a patch for that? Can this be achieved by just passing `COMPILER_RT_ENABLE_IOS` value passed externally?

superbuild/CMakeLists.txt
```diff
@@ -373,9 +472,32 @@ if (APPLE AND CMAKE_SYSTEM_VERSION VERSION_GREATER_EQUAL 13)
...
+    # build LLVM for old macOS to enable ISPC builds targeting an old macOS.
     list(APPEND LLVM_COMMON_CMAKE_ARGS
-        -DLLVM_ENABLE_RUNTIMES=libcxx$<SEMICOLON>libcxxabi
+        -DCMAKE_OSX_DEPLOYMENT_TARGET=${MACOS_VERSION_MIN}
         )
+
+    if (MACOS_UNIVERSAL_BIN)
+        # macOS Universal Binaries is a fat binary for x86_64 and arm64.
+        list(APPEND LLVM_COMMON_CMAKE_ARGS
+            -DCMAKE_OSX_ARCHITECTURES=x86_64$<SEMICOLON>arm64
```
dbabokin, line 498: Do we need quotes around `x86_64$<SEMICOLON>arm64`? I'm not sure about it, just checking.

superbuild/CMakeLists.txt
```diff
@@ -64,9 +82,42 @@ if (NOT PREBUILT_STAGE2 STREQUAL "NO" AND NOT PREBUILT_STAGE2_PATH STREQUAL "NO"
...
+    # XE are not supported on macOS targets
+    set(XE_DEPS OFF)
+
+    if (CMAKE_SYSTEM_PROCESSOR MATCHES "(x86)|(X86)|(amd64)|(AMD64)")
+        set (MACOS_VERSION_MIN 10.12)
+    else ()
+        set (MACOS_VERSION_MIN 11.0)
+    endif()
+
+    if (APPLE AND ISPC_ANDROID_NDK_PATH STREQUAL "NO")
```
dbabokin, line 116: `ISPC_ANDROID_NDK_PATH` is needed only when building Linux based targets (Linux, Android).

nurmukhametov: Hm, that's really strange because I somehow concluded that NDK is needed on macOS when ISPC_CROSS is on.

superbuild/15_0_apple_disable_ios.patch
nurmukhametov: I need to check that.

superbuild/CMakeLists.txt
```diff
@@ -64,9 +82,42 @@ if (NOT PREBUILT_STAGE2 STREQUAL "NO" AND NOT PREBUILT_STAGE2_PATH STREQUAL "NO"
...
+if (BUILD_SPIRV_TRANSLATOR_ONLY OR BUILD_VC_INTRINSICS_ONLY OR BUILD_L0_LOADER_ONLY)
+    set(INSTALL_WITH_XE_DEPS ON)
+endif()
+
+if (APPLE)
+    # XE are not supported on macOS targets
+    set(XE_DEPS OFF)
+
+    if (CMAKE_SYSTEM_PROCESSOR MATCHES "(x86)|(X86)|(amd64)|(AMD64)")
+        set (MACOS_VERSION_MIN 10.12)
```
dbabokin, line 110: I need to double check this logic. It matches `alloy.py`, but I think the intention was to have `10.12` vs `11.0` by default for `x86_64` vs `arm64`, but do `10.12` in case of universal binary. AR for me - to check this.

superbuild/CMakeLists.txt
dbabokin: Cross triggers Android build.

nurmukhametov: The thing is that with `-DISPC_CROSS=ON` and without `ISPC_ANDROID_NDK_PATH` on `APPLE`, cmake reports this error:
https://github.com/ispc/ispc/blob/d3f561cc742752acbb4ff0c0a517c615800c049b/CMakeLists.txt#L98

aneshlya (approved):
Looks good to me. I rely on @dbabokin to check exact options needed for MacOS build.

superbuild/README.md
```diff
@@ -33,11 +33,31 @@ $ cmake ${ISPC}/superbuild --preset os -DCCACHE=ON -DCCACHE_DIR=/tmp/shared-ispc
...
+$ cmake --build .
+$ cmake ${ISPC}/superbuild --preset os -DCMAKE_INSTALL_PREFIX=/opt/l0-loader -DBUILD_L0_LOADER_ONLY=ON
+$ cmake --build .
+
+# Build and install ISPC with XE dependencies
+$ cmake ${ISPC}/superbuild --preset os -DPREBUILT_STAGE2_PATH=/tmp/stage2-path -DCMAKE_INSTALL_PREFIX=/opt/ispc-with-xe/ -DINSTALL_WITH_XE_DEPS=ON
+$ cmake --build .
 ```
 
 # Building Process.
```
dbabokin, line 63:
```suggestion
# Build Process.
```

superbuild/README.md
```diff
@@ -33,11 +33,31 @@ $ cmake ${ISPC}/superbuild --preset os -DCCACHE=ON -DCCACHE_DIR=/tmp/shared-ispc
 $ cmake ${ISPC}/superbuild --preset os -DPREBUILT_STAGE2=/path/to/stage2-archive.tgz
```
dbabokin, line 33: Here, in this `README` we use `${ISPC}`, while in other scripts we always use `${ISPC_HOME}`. I suggest using it consistently, i.e. use it here as well.

superbuild/CMakeLists.txt
dbabokin: Ok, so these are different default setting here and in ISPC Cmake, so here we just enabling CROSS and it triggers Android build, but here we are not managing that directly. It's confusing, but ok, let it be this way.

superbuild/CMakeLists.txt
```diff
@@ -32,41 +32,91 @@ check_cache_var_declaration(L0_TAG "Level Zero url")
...
+    else ()
+        set (MACOS_VERSION_MIN 11.0)
+    endif()
+
+    if (APPLE AND ISPC_ANDROID_NDK_PATH STREQUAL "NO")
+        message(FATAL_ERROR "You need provide path to ANDROID NDK on macOS.")
```
dbabokin, line 116: Would be good specify explicitly how to define it. i.e. `use -DISPC_ANDROID_NDK_PATH=...`

dbabokin:
I see a lot messages like that:
```
[949/2695] Building CXX object lib/Transforms/IPO/CMakeFiles/LLVMipo.dir/OpenMPOpt.cpp.o
clang-15: warning: argument unused during compilation: '--ld-path=/Users/dybaboki/ispc/dbabokin_ispc/wt/superbuild-support-macos/build/stage2/bin/ld64.lld' [-Wunused-command-line]
```

Is it supposed to be this way? I build the simple "os" preset + build.

dbabokin:
When doing initial config on macOS, I get the following warning:
```
CMake Warning:
  Manually-specified variables were not used by the project:

    SPIRV_TRANSLATOR_BRANCH

```

I assume Xe is not enabled, but `SPIRV_TRANSLATOR_BRANCH` is still passed somewhere where it's not used.

superbuild/CMakeLists.txt
dbabokin: Here's the logic for selecting macOS version target is here: https://github.com/ispc/ispc/blob/1c6074be25da008b6a5756868011b784b425de47/CMakeLists.txt#L38

I.e. if either:
- building non-universal (native) binaries on `x86_64`, or
- universal binaries on `arm64`
then macOS target version is `10.12`. Otherwise `11.0`.

Note that you need to check only `x86_64`, no need to check "(X86)|(amd64)|(AMD64)" - because Mac hardware doesn't have this variability.

superbuild/CMakeLists.txt
```diff
@@ -32,41 +32,91 @@ check_cache_var_declaration(L0_TAG "Level Zero url")
...
+endif()
+
+if (APPLE)
+    # XE are not supported on macOS targets
+    set(XE_DEPS OFF)
+
+    if (CMAKE_SYSTEM_PROCESSOR MATCHES "(x86)|(X86)|(amd64)|(AMD64)")
+        set (MACOS_VERSION_MIN 10.12)
+    else ()
+        set (MACOS_VERSION_MIN 11.0)
```
dbabokin, line 112: Somehow stage2 always default to `CMAKE_OSX_DEPLOYMENT_TARGET` equal to `10.12`. And this is not the `10.12` set above, it's coming from somewhere else. This needs to be fixed, otherwise there's no point.

Also it doesn't need to be set for stage1, as it's use only on the host where's built.

nurmukhametov:
> When doing initial config on macOS, I get the following warning:
> 
> ```
> CMake Warning:
>   Manually-specified variables were not used by the project:
> 
>     SPIRV_TRANSLATOR_BRANCH
> ```

This variable is passed from presets. I suppress that warnings for all variables of XE_DEPS from presets via `set(ignoreMe "${VAR}").

> I assume Xe is not enabled, but `SPIRV_TRANSLATOR_BRANCH` is still passed somewhere where it's not used.
> 
> Same during stage2 config:
> 
> ```
> CMake Warning:
>   Manually-specified variables were not used by the project:
> 
>     LEVEL_ZERO_ROOT
> ```

This one reported by `ISPC` cmake so I just skip passing `LEVEL_ZERO_ROOT` to `ISPC` for `APPLE` and for any target when `XE_DEPS` is `OFF`.

nurmukhametov:
> I see a lot messages like that:
> 
> ```
> [949/2695] Building CXX object lib/Transforms/IPO/CMakeFiles/LLVMipo.dir/OpenMPOpt.cpp.o
> clang-15: warning: argument unused during compilation: '--ld-path=/Users/dybaboki/ispc/dbabokin_ispc/wt/superbuild-support-macos/build/stage2/bin/ld64.lld' [-Wunused-command-line]
> ```
> 
> Is it supposed to be this way? I build the simple "os" preset + build.

I think it is harmless. It was because I put `--ld-path` to `CCFLAGS`. So, I moved them to "LDFLAGS" to avoid this.

superbuild/CMakeLists.txt
nurmukhametov: How do you check that?

dbabokin: Ideally, when building the final ISPC binary the whole binary should be built with either `11.0` or `10.12` OSX target. it can be verified by examining the binary. But if anything goes wrong, you'll get warnings. I noticed warnings that about `10.12` not being supported for Arm, so `11.0` is used. So I investigated. Grepping for `CMAKE_OSX_DEPLOYMENT_TARGET` in build directory points to stage2 CMake that is configured with `10.12` and I've played with it to see that it's some kind of default, which is not coming from parent CMake.

superbuild/15_0_apple_disable_ios.patch
nurmukhametov: It looks like it works. I have been misled by LLVM ticket and tried `-DBUILTINS_CMAKE_ARGS=-DCOMPILER_RT_ENABLE_IOS=OFF` that didn't work for me.

superbuild/CMakeLists.txt
nurmukhametov: So, I fixed that as follow: only arm non-universal is `11.0`, all other variants is `10.12`.

dbabokin:
I see quite a few messages like this:
```
-- Installing: /Users/dybaboki/ispc/dbabokin_ispc/wt/superbuild-support-macos/build/stage2/lib/libclangBasic.a
llvm-ranlib: error: unable to load '/Users/dybaboki/ispc/dbabokin_ispc/wt/superbuild-support-macos/build/stage2/lib/libclangBasic.a': file too small to be an archive
-- Installing: /Users/dybaboki/ispc/dbabokin_ispc/wt/superbuild-support-macos/build/stage2/lib/libclangAPINotes.a
llvm-ranlib: error: unable to load '/Users/dybaboki/ispc/dbabokin_ispc/wt/superbuild-support-macos/build/stage2/lib/libclangAPINotes.a': file too small to be an archive
```

While these seems to be valid universal binaries. Any idea why they show up?

nurmukhametov:
> I see quite a few messages like this:
> 
> ```
> -- Installing: /Users/dybaboki/ispc/dbabokin_ispc/wt/superbuild-support-macos/build/stage2/lib/libclangBasic.a
> llvm-ranlib: error: unable to load '/Users/dybaboki/ispc/dbabokin_ispc/wt/superbuild-support-macos/build/stage2/lib/libclangBasic.a': file too small to be an archive
> -- Installing: /Users/dybaboki/ispc/dbabokin_ispc/wt/superbuild-support-macos/build/stage2/lib/libclangAPINotes.a
> llvm-ranlib: error: unable to load '/Users/dybaboki/ispc/dbabokin_ispc/wt/superbuild-support-macos/build/stage2/lib/libclangAPINotes.a': file too small to be an archive
> ```
> 
> While these seems to be valid universal binaries. Any idea why they show up?

Unfortunately, I have no idea. At the same time, it looks like it doesn't cause any problem later. But I agree that this looks suspicious and probably need some investigation.

dbabokin (commented):
If superbuild is supposed to be used for release build, it needs to build packages and manage some other options of ISPC CMake, specifically for the release important options are SIGN key (`ISPC_SIGN_KEY`) and specification of target architecture (`CMAKE_OSX_ARCHITECTURES`), as 3 types of binaries are supposed to be released (`arm64`, `x86_64`, and `universal`).

Other important point about packages is that it triggers strip, so we should not forget about it.

And is there a way to run lit tests as part of superbuild. It would be oversight to not run lit tests for release binaries.

One other problem that I see, default settings build ISPC binary, which is Universal Binary, but targets `11.0`, while classic build yields `10.12`.

To check what OS is targeted, run `otool -l ispc | grep -E -A4 '(LC_VERSION_MIN_MACOSX|LC_BUILD_VERSION)'`

dbabokin:
> Unfortunately, I have no idea. At the same time, it looks like it doesn't cause any problem later. But I agree that this looks suspicious and probably need some investigation.

I can confirm that LLVM build through `alloy.py` produces the same messages on M1.

nurmukhametov:
> If superbuild is supposed to be used for release build, it needs to build packages and manage some other options of ISPC CMake, specifically for the release important options are SIGN key (`ISPC_SIGN_KEY`) and specification of target architecture (`CMAKE_OSX_ARCHITECTURES`), as 3 types of binaries are supposed to be released (`arm64`, `x86_64`, and `universal`).
> 

It builds package. I will add `ISPC_SIGN_KEY` and `CMAKE_OSX_ARCHITECTURES`.

> Other important point about packages is that it triggers strip, so we should not forget about it.
> 
> And is there a way to run lit tests as part of superbuild. It would be oversight to not run lit tests for release binaries.

There is a target `ispc-stage2-check-all` to run them.

> 
> One other problem that I see, default settings build ISPC binary, which is Universal Binary, but targets `11.0`, while classic build yields `10.12`.
> 
> To check what OS is targeted, run `otool -l ispc | grep -E -A4 '(LC_VERSION_MIN_MACOSX|LC_BUILD_VERSION)'`

Which classic do you mean? The last release from github reports `11.0` for universal binary:

```bash
ispc-bins % otool -l ./ispc-v1.20.0-macOS.universal/bin/ispc | grep -E -A4 '(LC_VERSION_MIN_MACOSX|LC_BUILD_VERSION)'
      cmd LC_BUILD_VERSION
  cmdsize 32
 platform 1
    minos 11.0
      sdk 10.17
```

> 
> Do we have a switch to govern is `Release` or `Release+Assert` LLVM build is done? What is the default with "os" preset? It seems that it's `Release+Assert`.

There is `-DLLVM_DISABLE_ASSERTIONS` with default `ON` that means `Release`, to build `Release+Assert` `-DLLVM_DISABLE_ASSERTIONS=OFF` is needed to be provided.

dbabokin:
> Which classic do you mean? The last release from github reports 11.0 for universal binary

Interesting, I just tried to repeat the experiment and it give the following for universal build done using LLVM from `alloy.py`:

```
> otool -l bin/ispc | grep -E -A4 '(LC_VERSION_MIN_MACOSX|LC_BUILD_VERSION)'
      cmd LC_VERSION_MIN_MACOSX
  cmdsize 16
  version 10.12
      sdk 13.3
Load command 10
--
      cmd LC_BUILD_VERSION
  cmdsize 32
 platform 1
    minos 11.0
      sdk 13.3
```

We need superbuild to yield the same.

nurmukhametov:
> > Which classic do you mean? The last release from github reports 11.0 for universal binary
> 
> Interesting, I just tried to repeat the experiment and it give the following for universal build done using LLVM from `alloy.py`:
> 
> ```
> > otool -l bin/ispc | grep -E -A4 '(LC_VERSION_MIN_MACOSX|LC_BUILD_VERSION)'
>       cmd LC_VERSION_MIN_MACOSX
>   cmdsize 16
>   version 10.12
>       sdk 13.3
> Load command 10
> --
>       cmd LC_BUILD_VERSION
>   cmdsize 32
>  platform 1
>     minos 11.0
>       sdk 13.3
> ```
> 
> We need superbuild to yield the same.

What type of experiment did you repeat? It is also interesting where you run `otool` (`arm64` or `x86_64` machine?). Can we trust the output of `otool` for universal binaries? Is it same for both cases? Is it reports minimal/maximal/common/both? It is really suspicious for me that this tool reports only one entry for the universal binaries that consists of two entities.

nurmukhametov:
> > > Which classic do you mean? The last release from github reports 11.0 for universal binary
> > 
> > 
> > Interesting, I just tried to repeat the experiment and it give the following for universal build done using LLVM from `alloy.py`:
> > ```
> > > otool -l bin/ispc | grep -E -A4 '(LC_VERSION_MIN_MACOSX|LC_BUILD_VERSION)'
> >       cmd LC_VERSION_MIN_MACOSX
> >   cmdsize 16
> >   version 10.12
> >       sdk 13.3
> > Load command 10
> > --
> >       cmd LC_BUILD_VERSION
> >   cmdsize 32
> >  platform 1
> >     minos 11.0
> >       sdk 13.3
> > ```
> >   
> > We need superbuild to yield the same.
> 
> What type of experiment did you repeat? It is also interesting where you run `otool` (`arm64` or `x86_64` machine?). Can we trust the output of `otool` for universal binaries? Is it same for both cases? Does it report minimal/maximal/common/both? It is really suspicious for me that this tool reports only one entry for the universal binaries that consists of two entities.

I repeated build from the scratch on the current state, and got same result. Otool reports two versions 10.12 and 11:
```bash
% otool -l ispc-stage2/bin/ispc | grep -E -A4 '(LC_VERSION_MIN_MACOSX|LC_BUILD_VERSION)'
      cmd LC_VERSION_MIN_MACOSX
  cmdsize 16
  version 10.12
      sdk 13.3
Load command 10
--
      cmd LC_BUILD_VERSION
  cmdsize 32
 platform 1
    minos 11.0
      sdk 13.3
```

dbabokin (approved):
I've did rebuild on x86 Mac of classic build and superbuild. Result seems to be good enough, i.e. binaries are not identical, but similar enough to be ok.

I still think that documentation can be done better for the superbuild, but I think a t can be addressed separately.

Though one thing I would prefer to be changed before merging - the name of the preset. I think it's should be something like "release", as it's optimized for the release build (Release LLVM + cross build options enabling targets that we enabled for the release).

And one other minor thing - should we enabled running lit tests as part of default preset? I think it would be logical.

aneshlya:
> I still think that documentation can be done better for the superbuild, but I think a t can be addressed separately.
> 
> Though one thing I would prefer to be changed before merging - the name of the preset. I think it's should be something like "release", as it's optimized for the release build (Release LLVM + cross build options enabling targets that we enabled for the release).
> 
> And one other minor thing - should we enabled running lit tests as part of default preset? I think it would be logical.

The renaming, documentation and further improvements of superbuild are not MacOS specific and will be addressed separately.

PR #2584, opened 2023-07-24
Fix memory leaks to enable ASAN for lit_tests and run_tests

nurmukhametov:
This change allows to enable ASAN for lit_tests and run_tests.


dbabokin (approved):
LGTM

PR #2586, opened 2023-07-25
Added 2468.ispc lit test

### LINKED ISSUE #2468: Unnecessary Gather Instruction Generation
Consider the following snippet:
```c
#define SOA_IX(a, v, nv) ((((a) / TARGET_WIDTH)*(nv) + (v))*TARGET_WIDTH + (a) % TARGET_WIDTH)

export void simple(uniform float vin[], uniform float vout[]) {
    foreach (index = 0 ... TARGET_WIDTH) {
        // Load the appropriate input value for this program instance.
        float v1 = vin[SOA_IX(index, 0, 3)];
        float v2 = vin[SOA_IX(index, 2, 3)];

        // And write the result to the output array.
        vout[index] = v1 + v2;
    }
}
```
which should resolve to a pair of loads, an add, and a store.  However, on ISPC 1.19 with x86 this translates into a pair of gather instructions and an associated performance warning.  The associated constants are:
```
.LCPI1_0:
        .long   0                               # 0x0
        .long   4                               # 0x4
        .long   8                               # 0x8
        .long   12                              # 0xc
        .long   16                              # 0x10
        .long   20                              # 0x14
        .long   24                              # 0x18
        .long   28                              # 0x1c
.LCPI1_1:
        .long   64                              # 0x40
        .long   68                              # 0x44
        .long   72                              # 0x48
        .long   76                              # 0x4c
        .long   80                              # 0x50
        .long   84                              # 0x54
        .long   88                              # 0x58
        .long   92 
```
which corresponds to contiguous loads.  There is also a performance warning about varying division being expensive.  Here, however, it is easy enough to see the division is by a power of two and hence inexpensive.

aneshlya:
Closes #2468.
The original issue was fixed by #2521.

### DIFF
--- a/tests/lit-tests/2468.ispc
+++ b/tests/lit-tests/2468.ispc
@@ -1,0 +1,24 @@
+// RUN: %{ispc} %s --target=avx2-i32x8 --arch=x86-64 --nostdlib --emit-asm -o - | FileCheck %s
+
+// REQUIRES: X86_ENABLED
+
+// The goal of this test is to check that code generation results into a pair of loads, an add, and a store.
+// In assembly we expect only one vmovups instruction for load since the second load is embedded into vaddps.
+
+// CHECK: vmovups
+// CHECK-NOT: vgather
+// CHECK-NEXT: vaddps
+// CHECK-NEXT: vmovups
+
+#define SOA_IX(a, v, nv) ((((a) / TARGET_WIDTH)*(nv) + (v))*TARGET_WIDTH + (a) % TARGET_WIDTH)
+
+void simple(uniform float vin[], uniform float vout[]) {
+    foreach (index = 0 ... TARGET_WIDTH) {
+        // Load the appropriate input value for this program instance.
+        float v1 = vin[SOA_IX(index, 0, 3)];
+        float v2 = vin[SOA_IX(index, 2, 3)];
+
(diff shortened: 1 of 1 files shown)


tests/lit-tests/2468.ispc
```diff
@@ -0,0 +1,23 @@
+// RUN: %{ispc} %s --target=avx2-i32x8 --arch=x86-64 --nostdlib --emit-asm -o - | FileCheck %s
+
+// REQUIRES: X86_ENABLED
+
+// The goal of this test is to check that code generation results into a pair of loads, an add, and a store
```
dbabokin, line 5: It's a pair of loads, technically, but only one `vmovups` is expected - because the second one is embedded in `vaddps` instruction. Would be good to mention it, otherwise it's not clear why one load is expected in the checks.

tests/lit-tests/2468.ispc
```diff
@@ -0,0 +1,23 @@
...
+
+// CHECK-COUNT-1: vmovups
```
dbabokin, line 7: `CHECK-COUNT-1` is equivalent to `CHECK`, as far as I understand. I think it's not checking that "no more than expected number of matches happen".

dbabokin (approved):
Looks good!

PR #2589, opened 2023-07-28
Updates for Xe targets (21)

dbabokin:
This PR introduces support of Intel Meteor Lake Xe-LPG graphics:
1. added two new ISPC targets: `xelpg-x16` and `xelpg-x8`
2. added two new device names: `mtl-m` and `mtl-p`

It also enables NewGVN pass on Xe targets which brings up to 5% performance improvement on ISPC workloads on Intel(R) Arc Graphics.

Level zero loader dependency updated to [1.31.1](https://github.com/oneapi-src/level-zero/releases/tag/v1.13.1).


aneshlya (approved):
Looks good.

PR #2590, opened 2023-07-31
Fix missing copy constructor warnings

nurmukhametov:
Some objects (`Target`, `Module`, `StructDeclaration`) own heap-allocated memory. The absence of user-defined copy constructor and assignment operator may lead to memory leak. At the moment, we do not copy them at all, so I think it is just better to delete them.


e3m3 (approved):
LGTM

dbabokin (approved):
LGTM

PR #2591, opened 2023-07-31
docs: remove javascript with analytics

nurmukhametov:


nurmukhametov:
It is better to have static htmls without any javascripts because we pack them into release tarballs and distribute to users.

dbabokin:
Note that it will be propagated to https://ispc.github.io/ only during the release.

PR #2593, opened 2023-08-03
LLVM 17.0 fixes and LLVM 18.0 definiton

dbabokin:
- Add `LLVM 17.0` support to `alloy.py`
- Add `LLVM 18.0` definition to ISPC sources
- Bump required CMake version in `ispcrt` to make newest CMake version (3.27) happy
- Fix compile and test problems with LLVM 17.0

This is not all the changes that are required to fully enable LLVM 17.0+. Starting 17.0 opaque pointers must be enabled.


aneshlya (approved):
Looks good!

PR #2594, opened 2023-08-07
ispcrt: make same hardening options as everywhere else

nurmukhametov:
This patch provides ispcrt shared library flags that are used for ispc binary. All of them are used for ispcrt_device_* shared libraries as well.


ispcrt/CMakeLists.txt
```diff
@@ -133,16 +133,23 @@ foreach(PRNAME ${LIB_LIST})
...
     else()
-      # Enable stack protector
-      # NOT to assume that null pointer deference does not exist
-      # Assume that signed overflow always wraps
-      target_compile_options(${PRNAME} PRIVATE -fstack-protector-strong -fno-delete-null-pointer-checks -fwrapv)
+        # Enable stack protector
+        # NOT to assume that null pointer deference does not exist
+        # Assume that signed overflow always wraps
+        target_compile_options(${PRNAME} PRIVATE -fstack-protector-strong -fno-delete-null-pointer-checks -fwrapv)
+        target_compile_options(${PRNAME} PRIVATE -fPIC)
```
aneshlya, line 145: I think we already have the libraries built with position independent code:
```
set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
https://github.com/ispc/ispc/pull/2594/files#diff-656ce737b6f5d8ae5ed47377acf495bbdc41d00cbb6df559df28fec0fd71d15cR20

aneshlya:
I don't like the idea of having the same security options in 3 (as minimum) different places. What do you think about creating an INTERFACE library with these security options and then linking it with ispcrt libraries (and potentially with ispc itself)?

dbabokin:
I've removed myself from reviewers list for this PR, I think @aneshlya is much more competent for reviewing CMake changes.

The changes looks generally good for me, but I think Arina raised a good point about single place for the security options.

ispcrt/detail/gpu/CMakeLists.txt
```diff
@@ -8,33 +8,21 @@ message(STATUS "${TARGET} LEVEL_ZERO_INCLUDE_DIR is ${LEVEL_ZERO_INCLUDE_DIR}")
...
-    # ... and the import lib into the devel package
-    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
-    )
-install(TARGETS ${TARGET}
-    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
-      NAMELINK_ONLY
-    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
-    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
-    )
+  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
```
aneshlya, line 25: `NAMELINK_SKIP` disappeared. Was it intentional?

aneshlya:
The changes look good now but I still want to understand the implication of removing of `NAMELINK_ONLY`/`NAMELINK_SKIP` options.

nurmukhametov:
> I don't like the idea of having the same security options in 3 (as minimum) different places. What do you think about creating an INTERFACE library with these security options and then linking it with ispcrt libraries (and potentially with ispc itself)?

I have made change to utilize INTERFACE library. Although, it required changes that I don't really understand, e.g., INTERFACE lib is propagated into exports.

ispcrt/detail/gpu/CMakeLists.txt
nurmukhametov: Yes, see explanation below

aneshlya (approved):
Looks good to me!

PR #2595, opened 2023-08-08
Function template specialization support

aneshlya:
Added support of function template specializations with explicit template arguments:
```cpp
// Primary template
template <typename T, typename C> noinline int goo(T argGooOne, C argGooTwo);

// Specialization with explicit template arguments
template <> noinline int goo<int, float>(int argGooOne, float argGooTwo);

// Not supported yet: specialization with implicit template arguments (requires template arguments type deduction)
template <> noinline int goo(int argGooOne, float argGooTwo);
```

### DIFF
--- a/docs/ispc.rst
+++ b/docs/ispc.rst
@@ -3702,12 +3702,14 @@ What is currently supported:
   function call syntax (i.e. ``add(1, 2);``).
 * Explicit template function instantiations (i.e.
   ``template int add<int>(int a, int b);``).
+* Explicit template function specializations (i.e.
+  ``template<> int add<int>(int a, int b) { return a - b;}``).
 
 What is currently not supported, but is planned to be supported:
 
 * Non-type template parameters.
 * Default values for template parameters.
-* Template function specializations.
+* Template arguments deduction in template function specializations.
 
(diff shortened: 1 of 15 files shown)


tests/lit-tests/func_template_spec_2.ispc
```diff
@@ -0,0 +1,33 @@
+// Check that function specialization can be declred, then used, then defined..
```
dbabokin, line 1: declred -> declared

dbabokin (commented):
The code looks good, but I need to play with the implementation.

dbabokin (approved):
The changes look good! Please go forward with squashing commits.

We'll also will need to make a note in docs that function template specializations are supported.

docs/ispc.rst
```diff
@@ -3758,6 +3760,48 @@ following example:
 Note, to get the insight of the results of template argument deduction, it might
 be useful to specify ``--ast-dump`` flag to ``ispc`` compiler.
 
+ISPC supports template function specializations that can be used to provide alternative
+implementations for a specific set of template parameters. To define template function
```
dbabokin, line 3764: template argument?

dbabokin (approved):

Issue #2472, opened 2023-03-23
Template name can clash with struct members


rkaric:
Given the following code, this will generate the error: "syntax error, unexpected TOKEN_TYPE_NAME"
 https://godbolt.org/z/rK1Y7zGTa

PR #2597, opened 2023-08-10
Treat TOKEN_TYPE_NAME as TOKEN_IDENTIFIER in postfix_expression

aneshlya:
Fixes the problem from #2472 that RHS of "." or "->" expression cannot be a typename and should be treated as field name.


dbabokin (approved):
Looks good.

PR #2598, opened 2023-08-10
Fix dockers

nurmukhametov:
This PR fixes two minor hadolint warnings. It also fixes the logic of installing level zero loader of the ubuntu 20.04 docker.  The link to download deb packages of level zero 1.10 for Ubuntu 18.04 doesn't exist, because level zero provides only deb packages for ubuntu 22.04 since around this version.


dbabokin (approved):
LGTM

aneshlya (approved):

PR #2600, opened 2023-08-11
Enable more compiler warnings

nurmukhametov:
Enables `-Wextra` that helps to uncover the real bug that has been introduced in https://github.com/nurmukhametov/ispc/commit/e4ea6d7e285d93a690a1bb36ad15c14999f14a61 and fixed in this PR in the commit https://github.com/ispc/ispc/pull/2600/commits/a73bc2b4cd1f62c66bac69542e2210ca60933467

It also contains some fixes to suppress compiler warnings. Most of them look harmless but clean report and enabling `-Wextra` may help fix some problems earlier in the future.


aneshlya (approved):
Looks good to me

dbabokin (approved):
Looks good.

PR #2601, opened 2023-08-11
Fix deallocation on parser error

### LINKED ISSUE #2599: Missing semicolon leads to SIGSEGV
Here's an examples ([Godbolt](https://godbolt.org/z/xvqxxaqsf)):
```
int foo(int i) { return 1;}

int bar() {
    int d = foo(2) // missing semicolon
    return d;
}
```

It crashes in `lCleanUpyylvalStringVal()` in `parse.yy`.

nurmukhametov:
It fixes #2599 

Intially, it was deallocated in `yyerror` function in https://github.com/nurmukhametov/ispc/commit/eaa5d755403d070ffec93356b9ebda8017136499.
It was wrong because `yylval` is union, so `stringVal` is not null for many different types, e.g., in 2599.ispc it is small integer value on the moment of calling `yyerror`.

Bison has the `%destructor` directive to define code that is called when a symbol is automatically discarded, e.g., during the error recovery. See manual for details:
https://www.gnu.org/software/bison/manual/html_node/Destructor-Decl.html

This change tag all tokens that semantically returns heap-allocated string as `<stringVal>`. The destructor directive is also added for `<stringVal>`. It results in `yydestruct` function filled with code to deallocate corresponding to `<stringVal>` tokens and types. It happened inside the body of `yydestruct` function in switch like this:
```c
  switch (yytype)
    {
    case 15: /* TOKEN_FLOAT16_CONSTANT  */
            { lCleanUpyylvalStringVal(); }
        break;

    case 20: /* TOKEN_IDENTIFIER  */
            { lCleanUpyylvalStringVal(); }
        break;
    ...
    }
```


tests/lit-tests/discard-float16.ispc
```diff
@@ -0,0 +1,7 @@
+// RUN: not %{ispc} --target=host --nostdlib --nowrap -o - %s 2>&1 | FileCheck %s
+
+// CHECK: Error: syntax error
+// CHECK-NOT: detected memory leaks
```
dbabokin, line 4: It's just compiled with ISPC, but not executed with ASAN, so where "detected memory leaks" would come from?

dbabokin (approved):
Looks good, but I'm not sure that "detected memory leaks" needs to be checked.

tests/lit-tests/discard-float16.ispc
nurmukhametov: Yeah, Nowhere. This test is specifically target asan build with `CHECK-NOT: detected memory leak`. Otherwise, it is quite useless.

I actually need to distinguish two states:
1. `ispc` exited with non-zero because of syntax errors,
2. `ispc` exited with non-zero because of memory leak.

The first one need to be `PASS`, the second one needs to be `FAIL` for tests in this PR. Maybe, the better approach is required.

aneshlya: You can try to detect `ASAN_OPTIONS` environment variable in `lit.cfg`, add it to available lit features `config.available_features.add("ASAN_BUILD")` and use it from the lit tests to enable them only for ASAN build.

aneshlya (approved):

tests/lit-tests/discard-float16.ispc
dbabokin: I don't think that the tests are useless without ASAN, they specifically checking that syntax error is generated. So possibly it's fine running without detecting ASAN, I'm good with either way of running them.

nurmukhametov: > You can try to detect `ASAN_OPTIONS` environment variable in `lit.cfg`, add it to available lit features `config.available_features.add("ASAN_BUILD")` and use it from the lit tests to enable them only for ASAN build.

I have added `ASAN_RUN` and `REQUIRES: ASAN_RUN` to tests.  But these tests are `UNSUPPORTED` without `ASAN_RUN`. Is there some way to run them always and does `CHECK-NOT` only when `ASAN_RUN`?

aneshlya: I think the only way is to split test into two. Always check that syntax error is generated in one test, and check for memory leaks in another.

nurmukhametov: OK, I have duplicated them.

aneshlya (approved):
LGTM

Issue #2605, opened 2023-08-13
vector_type_convert.ispc LIT test fails on ARM


dbabokin:
Running lit test on arm64 (macOS), I see that `vector_type_convert.ispc` fails. You can emulate it with `--target=neon-i32x4`.

It fails because the generated code doesn't contain `fpext`. Either the compiler needs to be fixed, or the test needs to run on x86 only. Or the test needs to be x86 only until the compiler is fixed.

I'm not sure what is the best way to generate LLVM IR for this code sequence for `neon-i32x4`

PR #2606, opened 2023-08-14
Fixed vector_type_convert.ispc lit test

aneshlya:
Fix for #2605 
The test was too aggressively optimized on some targets.


dbabokin (approved):
LGTM

PR #2608, opened 2023-08-15
Remove CWD from the search path for dbghelp.dll

aneshlya:
See comments in the code

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -19,6 +19,7 @@
 #include <stdlib.h>
 #ifdef ISPC_HOST_IS_WINDOWS
 #include <time.h>
+#include <windows.h>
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -18,6 +18,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #ifdef ISPC_HOST_IS_WINDOWS
+#include "windows.h"
```
dbabokin, line 21: Seems this need to be `<windows.h>`.

dbabokin (approved):
LGTM, if the comment about `<>` is addressed.

nurmukhametov (approved):

PR #2610, opened 2023-08-16
Updates for Xe targets (22)

aneshlya:
This PR fixes couple of static analysis issues and updates Level zero loader dependency to [1.13.5](https://github.com/oneapi-src/level-zero/releases/tag/v1.13.5).

### DIFF
--- a/src/ctx.cpp
+++ b/src/ctx.cpp
@@ -3868,7 +3868,7 @@ llvm::Value *FunctionEmitContext::XeSimdCFAny(llvm::Value *value) {
 }
 
 llvm::Value *FunctionEmitContext::XeSimdCFPredicate(llvm::Value *value, llvm::Value *defaults) {
-    AssertPos(currentPos, llvm::isa<llvm::VectorType>(value->getType()));
+    AssertPos(currentPos, llvm::isa<llvm::FixedVectorType>(value->getType()));
     llvm::FixedVectorType *vt = llvm::dyn_cast<llvm::FixedVectorType>(value->getType());
     if (defaults == nullptr) {
         defaults = llvm::ConstantVector::getSplat(
--- a/src/opt/MangleOpenCLBuiltins.cpp
+++ b/src/opt/MangleOpenCLBuiltins.cpp
@@ -18,7 +18,8 @@ static std::string mangleMathOCLBuiltin(const llvm::Function &func) {
     std::string mangledName;
     llvm::Type *retType = func.getReturnType();
     // spirv OpenCL builtins are used for double types only
-    Assert(retType->isVectorTy() && llvm::dyn_cast<llvm::FixedVectorType>(retType)->getElementType()->isDoubleTy() ||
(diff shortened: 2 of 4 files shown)


dbabokin (approved):

src/opt/MangleOpenCLBuiltins.cpp
```diff
@@ -18,7 +18,8 @@ static std::string mangleMathOCLBuiltin(const llvm::Function &func) {
     std::string mangledName;
     llvm::Type *retType = func.getReturnType();
     // spirv OpenCL builtins are used for double types only
-    Assert(retType->isVectorTy() && llvm::dyn_cast<llvm::FixedVectorType>(retType)->getElementType()->isDoubleTy() ||
+    Assert(retType->isVectorTy() && (llvm::isa<llvm::FixedVectorType>(retType) &&
+                                     llvm::dyn_cast<llvm::FixedVectorType>(retType)->getElementType()->isDoubleTy()) ||
```
e3m3, line 22: Unfortunately, `isa` does is not understood by the static analyzer to mean that `dyn_cast` will not return `nullptr`. An explicit check for `nullptr` is required here.
```suggestion
#ifdef debug
    Assert(llvm::isa<llvm::FixedVectorType>(retType));
    const auto fvt = llvm::dyn_cast<llvm::FixedVectorType>(retType);
    Assert(retType->isVectorTy() && (fvt != nullptr && fvt->getElementType()->isDoubleTy()) ||
// ...
#endif
```
The additional assert for `isa` is optional here. `#ifdef DEBUG` is used to guard the stack allocation of fvt. Without this guard, the allocation will also be done on a `Release` build, where `fvt` remains unused and will issue a warning/error.

src/ctx.cpp
```diff
@@ -3868,7 +3868,7 @@ llvm::Value *FunctionEmitContext::XeSimdCFAny(llvm::Value *value) {
 }
 
 llvm::Value *FunctionEmitContext::XeSimdCFPredicate(llvm::Value *value, llvm::Value *defaults) {
-    AssertPos(currentPos, llvm::isa<llvm::VectorType>(value->getType()));
+    AssertPos(currentPos, llvm::isa<llvm::FixedVectorType>(value->getType()));
     llvm::FixedVectorType *vt = llvm::dyn_cast<llvm::FixedVectorType>(value->getType());
     if (defaults == nullptr) {
```
e3m3, line 3873: Same issue as below: https://github.com/ispc/ispc/pull/2610/files#r1296438132, but here, `vt` is not only used during during a `Debug` build, so do the check in the guard to satisfy the analyzer.

```suggestion
    AssertPos(currentPos, llvm::isa<llvm::FixedVectorType>(value->getType()));
    llvm::FixedVectorType *vt = llvm::dyn_cast<llvm::FixedVectorType>(value->getType());
    if (defaults == nullptr && vt != nullptr) {
```

e3m3 (approved):
Issues mentioned in comments above are propagated from branch promotion. Will be addressed separately.

PR #2612, opened 2023-08-17
make a way for apps to give their device data to ispcrt

demarle:


demarle:
@aneshlya Please take a look when you get a chance. 

The new name better reflects that the flag has two effects which we had concerns about before.
1) MANAGED meaning app is responsible for deleting the data behind the ptr,
2) DEVICE the ptr is destined for device_ptr. 

I thought about adding the symmetric APPLICATION_MANAGED_HOST flag but that seems to be the default behavior so it isn't useful.

ispcrt/detail/gpu/GPUDevice.cpp
```diff
@@ -830,13 +830,19 @@ class ChunkedPool {
...
         }
 
+        if (flags->smHint == ISPCRT_SM_APPLICATION_MANAGED_DEVICE) {
+            m_devicePtr = appMem;
+        } else {
+            m_hostPtr = appMem;
+        }
+
         // Use MemPool only when it is explicitly enabled with env var and memory hint is not device/host read/write
         m_useMemPool = get_bool_envvar(ISPCRT_MEM_POOL) && (m_smhint != ISPCRT_SM_HOST_DEVICE_READ_WRITE);
```
aneshlya, line 847: ISPCRT_SM_APPLICATION_MANAGED_DEVICE memory is not supposed to be used in memory pools, right? 
```suggestion
        m_useMemPool = get_bool_envvar(ISPCRT_MEM_POOL) && (m_smhint != ISPCRT_SM_HOST_DEVICE_READ_WRITE) && (m_smhint != ISPCRT_SM_APPLICATION_MANAGED_DEVICE);
```

aneshlya:
I think we should add some ispcrt mock tests for the new flag. @nurmukhametov, could you take care of it?

ispcrt/detail/gpu/GPUDevice.cpp
demarle: correct.

nurmukhametov:
> I think we should add some ispcrt mock tests for the new flag. @nurmukhametov, could you take care of it?

Probably something like that should be enough for mock tests of `C API`:
```
TEST_F(MockTest, C_API_ApplicationManagedDevMem) {
    ISPCRTContext context = ispcrtNewContext(ISPCRT_DEVICE_TYPE_GPU);
    ISPCRTNewMemoryViewFlags flags = {ISPCRT_ALLOC_TYPE_SHARED, ISPCRT_SM_APPLICATION_MANAGED_DEVICE};
    char appMem[100] = {};
    ISPCRTMemoryView view = ispcrtNewMemoryViewForContext(context, appMem, 10, &flags);
    void *devPtr = ispcrtDevicePtr(view);
    ASSERT_EQ(devPtr, appMem);
    void *hostPtr = ispcrtHostPtr(view);
    ASSERT_EQ(hostPtr, appMem);
}
```
Regarding `C++ API`, I don't really understand how `ApplicationManagedDevice` hint is supposed to be used. I don't see a method/class that takes both pointer (like `appMem` for `ispcrtNewMemoryViewForContext`) and `SharedMemoryUsageHint`.

demarle:
Thanks @nurmukhametov.

I think that second assert in the mock test will fail because hostPtr should be null. I want to assign the device side only. In my intended use case I immediately call ispcrtMemoryView to copy it to a more standard memory view with a shared buffer.

Regarding the C++ api, I was coming in at the C api level. I _think_ you are right for C++. The Array constructor I see that takes the appMemory pointer hardcodes the flags. I'm still puzzling through what the "EnableForDeviceAllocation" argument does in practice though. ...

nurmukhametov:
> I think that second assert in the mock test will fail because hostPtr should be null. I want to assign the device side only. In my intended use case I immediately call ispcrtMemoryView to copy it to a more standard memory view with a shared buffer.
> 

In this test, memview is shared so `hostPtr` and `devicePtr` are really same. 

For allocation of device memory view, I can suggest an another test where `hostPtr` is really `null` :
```c++
TEST_F(MockTest, C_API_ApplicationManagedDevMemNotShared) {
    ISPCRTDevice device = ispcrtGetDevice(ISPCRT_DEVICE_TYPE_GPU, 0);
    ISPCRTNewMemoryViewFlags flags = {ISPCRT_ALLOC_TYPE_DEVICE, ISPCRT_SM_APPLICATION_MANAGED_DEVICE};
    char appMem[100] = {};
    ISPCRTMemoryView view = ispcrtNewMemoryView(device, appMem, 10, &flags);
    void *devPtr = ispcrtDevicePtr(view);
    ASSERT_EQ(devPtr, appMem);
    void *hostPtr = ispcrtHostPtr(view);
    ASSERT_EQ(hostPtr, nullptr);
}
```

But I am not sure about the second test because we apply shared memory hint to device memory allocation there. Isn't that confusing?

demarle:
Yes it is confusing. We were thinking a bit about the ispcrt's memory model and how a MemoryView has some structural issues and assumptions about the presence of host memory that we might want to revisit eventually. What I've done is kind of a stop gap fix. It lets my app use iscprt to do what most would use sycl::queue::memcpy to do, but we will probably want deeper changes later when we have a chunk of time to think about and implement them thoroughly.

For now I think the test that verifies that in the CPUDevice the hostptr is the destination but with the GPUDevice the deviceptr is the destination is correct.

aneshlya:
@demarle, we can merge your patch to unblock your development. Please cherry-pick ispcrt tests to your branch: https://github.com/aneshlya/ispc/commit/2493884f66c4bcaff7cd714f0ad0da0a96c55065

For the final solution please create a github issue with examples of what you need to do with "application managed memory", and let's brainstorm on it together. It would be good to finalize it before our next release.

PR #2618, opened 2023-08-17
Update Release Notes and docs for v1.21.0

aneshlya:


docs/ReleaseNotes.txt
```diff
@@ -1,3 +1,83 @@
...
+   SPIRV-Translator). It can generate an archive with dependencies or consume a pre-built archive to build ISPC only.
+   It also enables generating LTO or LTO+PGO enabled builds of LLVM and ISPC.
+ - Supported building ISPC with LLVM 16.
+
+New compiler switches:
+
+ - `--mcmodel` switch, which accepts `small` and `large` values. The definition is similar to gcc/clang. When `large`
+   model is used, it enables programs larger than 2Gb.
+ - `--opt=disable-gathers` and `--opt=disable-scatters` options, which disable generation of gathers and scatters
+   instructions on platforms that support them (for performance experiments).
```
e3m3, line 39:
```suggestion
New compiler switches:

 - `--mcmodel` switch, which accepts `small` and `large` values. The definition is similar to gcc/clang. When `large`
   model is used, it enables programs larger than 2Gb.
 - `--opt=disable-gathers` and `--opt=disable-scatters` options, which disable generation of gathers and scatters
   instructions on platforms that support them (for performance experiments).
 - `--[no-]wrap-signed-int` switches, which [does not] preserve(s) wrap-around behavior on signed integer overflow.
```

dbabokin (approved):
Looks good.

e3m3 (approved):

PR #2624, opened 2023-08-18
Sanitize MSVC builds and enable compiler-warnings job in CI

nurmukhametov:
This change enforces `c++17` standard for all projects including `check_isa.cpp`. It also specifies `/EHsc` for MSVC builds to properly enable exception handling. It fixes macro catching invocation in `ispcrt.cpp` and some minor warnings.

This change makes [compiler-warnings](https://github.com/ispc/ispc/actions/workflows/compiler-warnings.yml) green and enables it for pull requests.

### DIFF
--- a/alloy.py
+++ b/alloy.py
@@ -340,18 +340,20 @@ def unsupported_llvm_targets(LLVM_VERSION):
 # native - native targets run natively on current hardware.
 # sde - native target, which need to be emulated on current hardware.
 def check_targets():
+    from os.path import join
+
     result = []
     result_sde = []
-    # check what native targets do we have
-    if current_OS != "Windows":
-        if options.ispc_build_compiler == "clang":
-            cisa_compiler = "clang"
-        elif options.ispc_build_compiler == "gcc":
-            cisa_compiler = "g++"
 
-        try_do_LLVM("build check_ISA", cisa_compiler + " check_isa.cpp -o check_isa.exe", True)
-    else:
-        try_do_LLVM("build check_ISA", "cl check_isa.cpp", True)
+    cmd_conf = "cmake . -B check_isa_build -DISPC_INCLUDE_UTILS_ONLY=ON"
+    cmd_build = "cmake --build check_isa_build --target check_isa --config Release"
+    check_isa = "check_isa"
+    if current_OS == "Windows":
+        check_isa = join("Release", "check_isa.exe")
+    check_isa = join("check_isa_build", "bin", check_isa)
+
+    try_do_LLVM("build_check_isa configure", cmd_conf, True, True)
+    try_do_LLVM("build_check_isa build", cmd_build, True, True)
 
     # Dictionary mapping hardware architecture to its targets.
     # The value in the dictionary is:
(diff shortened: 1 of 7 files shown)


ispcrt/ispcrt.cpp
```diff
@@ -182,23 +182,23 @@ void ISPCLaunch(void **handlePtr, void *f, void *data, int countx, int county, i
...
+    abort();
 }
 
 void ISPCSync(void *handle) {
     if (ispc_sync_fptr) {
         return ispc_sync_fptr(handle);
     }
     // TODO: this code can be called directly from ISPC code. Not sure how to exit error safely.
-    throw std::runtime_error("Missing ISPCSync symbol");
+    abort();
```
e3m3, line 201: Should we write the missing symbol message that was removed to stderr before aborting?

nurmukhametov: Yes, we should.

e3m3 (approved):

alloy.py
```diff
@@ -340,18 +340,20 @@ def unsupported_llvm_targets(LLVM_VERSION):
...
-        if options.ispc_build_compiler == "clang":
-            cisa_compiler = "clang"
-        elif options.ispc_build_compiler == "gcc":
-            cisa_compiler = "g++"
 
-        try_do_LLVM("build check_ISA", cisa_compiler + " check_isa.cpp -o check_isa.exe", True)
-    else:
-        try_do_LLVM("build check_ISA", "cl check_isa.cpp", True)
+    cmd_conf = "cmake . -B check_isa_build -DISPC_INCLUDE_UTILS_ONLY=ON"
+    cmd_build = "cmake --build check_isa_build --target check_isa"
```
aneshlya, line 349: Minor note: I think CMAKE_BUILD_TYPE is Release by default but `cmake --build` builds Debug by default. Let's specify configuration explicitly: 
```suggestion
    cmd_build = "cmake --build check_isa_build --target check_isa --config Release"
```

alloy.py
```diff
@@ -340,18 +340,20 @@ def unsupported_llvm_targets(LLVM_VERSION):
...
+    check_isa = "check_isa"
+    if current_OS == "Windows":
+        check_isa = join("Debug", "check_isa.exe")
```
aneshlya, line 352:
```suggestion
        check_isa = join("Release", "check_isa.exe")
```

alloy.py
e3m3: > but `cmake --build` builds Debug by default

Interesting, I didn't know this. Is using `--config Release` in the build line different from specifying `-DCMAKE_BUILD_TYPE=Release` in the `cmd_conf` line?

aneshlya: I think it is. Our default build type (Release) for CMake configuration is set in CMakeLists.txt: https://github.com/ispc/ispc/blob/main/CMakeLists.txt#L233 but `cmake --build` without specified configuration will do Debug build (at least on Windows).

aneshlya (approved):
Looks good!

Issue #1318, opened 2018-04-25
ISPC doesn't comply with Win64 ABI


Myriachan:
ISPC does not fully comply with the Windows x86-64 ABI.  In all Windows ABIs except x86-32, every function that modifies the stack pointer is required to have an entry in the exception tables (`.pdata` and `.xdata` sections).  This allows the stack to be unwound properly in the event of an exception.  Only x86-32 Windows is exempt from this rule: x86-64, ARMv7-Thumb2 and ARM64 Windows all use exception tables.

Without these exception tables, if an exception occurs, `__try` blocks will not be respected, and the process will simply crash.  In addition, call stacks will not be generated properly in a crash dump or when debugging into ISPC-generated code.

There are additional strict rules about the format of prolog and epilog code, but I didn't see any issues there.

https://msdn.microsoft.com/en-us/library/ms235286.aspx

nurmukhametov:
PR #2630 adds `uwtable` to `ISPC` functions. It leads to generation of unwind info, so `.pdata` is not empty.

Consider the following example:
```ispc
export noinline uniform int BugFunction(uniform float x, uniform float y)
{
	uniform int a[100];
	uniform int sum = 0;
	for (uniform int i = 0; i < 100; i++)
		sum += a[i];
	return sum;
}
extern "C" uniform int ispc() {
  uniform int result = BugFunction(1.0f, 2.0f);
  return result;
}
```

It is compiled by the following command:
```bash
> ispc.exe --target=host -O0 -o .\uwtable\ex.ispc.obj --emit-obj .\ex.ispc
```

Without `uwtable` attributes, there is no `.pdata` section and unwind info is empty:
```cmd
> llvm-objdump.exe --section-headers .\ex.ispc.obj

.\ex.ispc.obj:  file format coff-x86-64

Sections:
Idx Name          Size     VMA              Type
  0 .text         000001d6 0000000000000000 TEXT
  1 .data         00000000 0000000000000000 DATA
  2 .bss          00000000 0000000000000000 BSS
  3 .rdata        00000004 0000000000000000 DATA
  4 .rdata        00000004 0000000000000000 DATA
  5 .drectve      00000033 0000000000000000

> llvm-objdump.exe --unwind-info .\ex.ispc.obj

.\ex.ispc.obj:  file format coff-x86-64
Unwind info:

```

With `uwtable` attribute, there is `.pdata` section and corresponding unwind info for 2 functions:
```cmd
[...]

PR #2630, opened 2023-08-21
Add uwtable attr to functions when targeting Windows

### LINKED ISSUE #2345: Corrupt callstack while inside ISPC functions
I'm doing some optimisations by replacing C++ with ISPC, and am finding it very difficult to use sampling profilers like Visual Studio's built-in CPU profiler, or third-party products such as Superluminal.

The timings surface in the wrong place in the call tree, and I believe it's because the callstack is corrupt while inside the ISPC function, resulting in samples being stored with a bad hierarchy behind them.

I'm using MSVC 16.11, ISPC 1.18.0.

You can see this when stepping through a simple contrived function such as:

```
// Bug.ispc
export uniform float BugFunction(uniform float x, uniform float y)
{
	uniform float xCos = cos(x);
	uniform float yCos = cos(y);
	return xCos + yCos;
}
```

```
// main.cpp
#include "bug.ispc.generated.h"

int main(int argc, char** argv)
{
	float result = ispc::BugFunction(1.0f, 2.0f);
	return (int)result;
}
```

In this gif, I am simply stepping through the program with F11. You can see that the callstack view on the right loses most entries once inside BugFunction.

ISPCBacktraceBug

I've not yet found any combination of ISPC parameters that fixes this. The gif above was produced using:
`"C:\Program Files\ISPC\ispc-v1.18.0-windows\bin\ispc.exe" %(FullPath) -o $(IntermediateOutputPath)%(Identity).obj -h $(ProjectDir)%(Identity).generated.h -g -O1 --target=avx2-i32x8 --no-omit-frame-pointer`

In the disassembly view, you can see a bit more of what's going on - the callstack corrupts on the very first instruction - `sub rsp 88h`, and is restored on the final `add rsp 88h` before the `ret`.

image

A more elaborate contrived function to show up the problem in the VS profiler (as the previous example is too quick for a sampling profiler):

```
// Bug.ispc

export uniform float BugFunction(uniform float x, uniform float y)
{
	uniform float total = 0;
	for (int64 i = 0; i < 10000000; ++i)
		total += cos(x);
	return total;
}
```

```
// main.cpp
#include <cmath>
#include <ctime>
#include <cstdint>
#include "bug.ispc.generated.h"

// C++ version of ispc::BugFunction
float BugFunction(float x, float y)
{
	float total = 0;
	for (int64_t i = 0; i < 1000000; ++i)
		total += cos(x);
	return total;
}

float DoBugFunction()
{
	float total = (float)rand();
	for (int i = 0; i < 1000; ++i)
		total = cos(total);
	return ispc::BugFunction(1.0f, total); // Swap for just BugFunction(1.0f, total) to see the "working" profile
}

int main(int argc, char** argv)
{
	std::srand((unsigned int)std::time(nullptr));
	float result = 0;
	for (int i = 0; i < 1000; ++i)
		result += DoBugFunction();
	return (int)result;
}
```
Note that ispc::BugFunction shows up under a "[Unwalkable]" heading in the wrong place in the call tree
image

This is how I would expect it to look (as if it were written in pure C++):
image

nurmukhametov:
Add `uwtable::Default` (which is async) to every function declatarion in the generated module when target OS is Windows. It enables generating unwind table during codegen. The unwind table is needed to generate correct backtrace during debugging and profiling.

It fixes #2345.

### DIFF
--- a/src/module.cpp
+++ b/src/module.cpp
@@ -925,6 +925,16 @@ void Module::AddFunctionDeclaration(const std::string &name, const FunctionType 
         function->addFnAttr(llvm::Attribute::NoInline);
     }
 
+    if (g->target_os == TargetOS::windows) {
+        // Enable generation an unwind table during codegen.
+        // It is needed to generate backtraces during debugging and to unwind callstack.
+#if ISPC_LLVM_VERSION <= ISPC_LLVM_14_0
+        function->setHasUWTable();
+#else
+        function->setUWTableKind(llvm::UWTableKind::Default);
+#endif
+    }
+
     if (functionType->isTask) {
         if (!g->target->isXeTarget()) {
             // This also applies transitively to members I think?


e3m3 (approved):
LGTM

dbabokin:
Do I understand it correctly that unwind table is a thing which is sitting aside from the code, so it doesn't affect performance directly, but increases object size? Does MSVC and clang generate them by default? Should we generate it without `-g`?

src/module.cpp
```diff
@@ -925,6 +925,24 @@ void Module::AddFunctionDeclaration(const std::string &name, const FunctionType
...
 
+    if (g->target_os == TargetOS::windows) {
+        // Enable generation an unwind table during codegen.
+        // It is needed to generate backtraces during debugging and to unwind callstack.
+#if ISPC_LLVM_VERSION <= ISPC_LLVM_13_0
+        llvm::AttrBuilder B;
+        B.addAttribute(llvm::Attribute::UWTable);
+        function->addAttributes(llvm::AttributeList::FunctionIndex, B);
+#elif ISPC_LLVM_VERSION <= ISPC_LLVM_14_0
+        llvm::AttrBuilder B(function->getContext());
```
aneshlya, line 936: Is `AttrBuilder` really needed? Does simple `function->addFnAttr(llvm::Attribute::UWTable);` work?

nurmukhametov:
> Do I understand it correctly that unwind table is a thing which is sitting aside from the code, so it doesn't affect performance directly, but increases object size? 

Yes, see here for details: https://github.com/ispc/ispc/issues/1318#issuecomment-1688075186

> Does MSVC and clang generate them by default? 

If the module contains functions that needs this entry then yes.

> Should we generate it without `-g`?

I think we should. This issue https://github.com/ispc/ispc/issues/1318 states that this tables are required by Windows x86-64 ABI. As I understand, stack unwinding may happen outside debugging during just normal program execution, e.g., when exception is raised.

aneshlya (approved):

dbabokin (approved):

Issue #2129, opened 2021-07-25
Feature Request: output dwarf debug info on Windows


memcache0:
I'm using MinGW with ispc on Windows 10. It is a strange combination but works fine for me. However, the Codeview debug information produced by ISPC cannot be recognized by MinGW. Thus it is difficult to debug it with GDB. So could you please support to output dwarf debug info on Windows?

dbabokin:
That's really not a typical use case, but seems that we can do that without much efforts, as LLVM should support that.

I would hate creating new switches for such specific use case, but we can trigger DWARF debug info if `--dwarf-version` is used on Windows. Right now that would be ignored.

As a side note - using WSL on Windows might be a good alternative. I tried it myself recently and it worked surprisingly well.

memcache0:
It's great if --dwarf-version can be supported on Windows. Thanks!

nurmukhametov:
> It's great if --dwarf-version can be supported on Windows. Thanks!

#2631 triggers the DWARF usage on Windows when `--dwarf-version=<n>` provided.

PR #2631, opened 2023-08-22
Force DWARF generation on windows with --dwarf-version

nurmukhametov:
This change lets `--dwarf-version=<n>` to force DWARF format debug info generation on Windows. It may be useful when ISPC code is linked with MINGW generated code. In that case, usual DWARF debug section (_debug_info, _debug_line, and etc.) are generated in PE files. This information is used by MINGW GDB during debug on Windows.

It fixes/implements #2129

### DIFF
--- a/src/ispc.h
+++ b/src/ispc.h
@@ -780,7 +780,7 @@ struct Globals {
     enum class DebugInfoType { None = 0, DWARF, CodeView };
     DebugInfoType debugInfoType;
 
-    /** Require generation of DWARF of certain version (2, 3, 4). For
+    /** Require generation of DWARF of certain version (2, 3, 4, 5). For
         default version, this field is set to 3. */
     // Hint: to verify dwarf version in the object file, run on Linux:
     // readelf --debug-dump=info object.o | grep -A 2 'Compilation Unit @'
--- a/src/ispc.h
+++ b/src/ispc.h
@@ -776,8 +776,12 @@ struct Globals {
         program in its output. */
     bool generateDebuggingSymbols;
 
+    /** Debug info type to generate. */
+    enum class DebugInfoType { None = 0, DWARF, CodeView };
+    DebugInfoType debugInfoType;
+
     /** Require generation of DWARF of certain version (2, 3, 4). For
-        default version, this field is set to 0. */
+        default version, this field is set to 3. */
     // Hint: to verify dwarf version in the object file, run on Linux:
     // readelf --debug-dump=info object.o | grep -A 2 'Compilation Unit @'
     // on Mac:
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -76,7 +76,7 @@ static void lPrintVersion() {
     PrintWithWordBreaks(cpuHelp, 16, TerminalWidth(), stdout);
     printf("    [--dllexport]\t\t\tMake non-static functions DLL exported.  Windows target only\n");
     printf("    [--dwarf-version={2,3,4}]\t\tGenerate source-level debug information with given DWARF version "
-           "(triggers -g).  Ignored for Windows target\n");
+           "(triggers -g).  It forces the usage of DWARF debug info on Windows target\n");
     printf("    [-E]\t\t\t\tRun only the preprocessor\n");
     printf("    [--emit-asm]\t\t\tGenerate assembly language file as output\n");
     printf("    [--emit-llvm]\t\t\tEmit LLVM bitcode file as output\n");
@@ -782,6 +782,7 @@ int main(int Argc, char *Argv[]) {
             if (2 <= val && val <= 4) {
                 g->generateDebuggingSymbols = true;
                 g->generateDWARFVersion = val;
+                g->debugInfoType = Globals::DebugInfoType::DWARF;
             } else {
                 errorHandler.AddError("Invalid value for DWARF version: \"%s\" -- "
                                       "only 2, 3 and 4 are allowed.",
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -75,7 +75,7 @@ static void lPrintVersion() {
              Target::SupportedCPUs().c_str());
     PrintWithWordBreaks(cpuHelp, 16, TerminalWidth(), stdout);
     printf("    [--dllexport]\t\t\tMake non-static functions DLL exported.  Windows target only\n");
-    printf("    [--dwarf-version={2,3,4}]\t\tGenerate source-level debug information with given DWARF version "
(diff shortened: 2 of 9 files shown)


src/ispc.h
```diff
@@ -777,13 +777,16 @@ struct Globals {
...
-        default version, this field is set to 0. */
+        default version, this field is set by deafaultDwarfVersion. */
     // Hint: to verify dwarf version in the object file, run on Linux:
     // readelf --debug-dump=info object.o | grep -A 2 'Compilation Unit @'
     // on Mac:
     // xcrun dwarfdump -r0 object.o
     int generateDWARFVersion;
 
+    /** Default DWARF version to be used. */
+    int defaultDWARFVersion;
```
e3m3, line 788: I think this default version should be implemented as a `constexpr`. Is there a use case where the default will change value?

dbabokin (commented):
The logic of the fix is really confusing. I failed to understand it.

We are going to have 3 different types of debug info (now we have 2, though) - `DWARF`, `CodeView` and `PDB` (not yet implemented), right? Should we have a variable `debugInfoType`, which would simplify and streamline the logic?

Also would be good to have lit tests for this.

nurmukhametov:
> The logic of the fix is really confusing. I failed to understand it.
> 
> We are going to have 3 different types of debug info (now we have 2, though) - `DWARF`, `CodeView` and `PDB` (not yet implemented), right? 

No, only `DWARF` and `CodeView`. I don't think that on `LLVM-IR` level there is `PDB` debug info type. Moreover, you can generate `PDB` file for project with `ISPC` code even now.

nurmukhametov:
>  Should we have a variable `debugInfoType`, which would simplify and streamline the logic?
> 
> Also would be good to have lit tests for this.

I have introduced `debugInfoType` variable and added lit tests.

tests/lit-tests/debug_type_dwarf.ispc
```diff
@@ -0,0 +1,12 @@
+// RUN: %{ispc} --target-os=windows --target=host --nostdlib --nowrap --emit-llvm-text -o - %s 2>&1 | FileCheck %s --implicit-check-not "llvm.dbg.cu"
+// RUN: %{ispc} -g --target-os=windows --target=host --nostdlib --nowrap --emit-llvm-text -o - %s 2>&1 | FileCheck %s -check-prefix=CHECK_CODEVIEW
+// RUN: %{ispc} -g --target=host --nostdlib --nowrap --emit-llvm-text -o - %s 2>&1 | FileCheck %s -check-prefix=CHECK_CODEVIEW
+// RUN: %{ispc} --dwarf-version=3 --target-os=windows --target=host --nostdlib --nowrap --emit-llvm-text -o - %s 2>&1 | FileCheck %s -check-prefix=CHECK_DWARF
+// RUN: %{ispc} -g --dwarf-version=3 --target-os=windows --target=host --nostdlib --nowrap --emit-llvm-text -o - %s 2>&1 | FileCheck %s -check-prefix=CHECK_DWARF
+// RUN: %{ispc} --dwarf-version=3 -g --target-os=windows --target=host --nostdlib --nowrap --emit-llvm-text -o - %s 2>&1 | FileCheck %s -check-prefix=CHECK_DWARF
```
aneshlya, line 6: The checks on line 5 and 6 are the same with different order of `-g`/`-dwarf-version` options. Is it intended?

tests/lit-tests/debug_type_dwarf.ispc
```diff
@@ -0,0 +1,12 @@
+// RUN: %{ispc} --target-os=windows --target=host --nostdlib --nowrap --emit-llvm-text -o - %s 2>&1 | FileCheck %s --implicit-check-not "llvm.dbg.cu"
+// RUN: %{ispc} -g --target-os=windows --target=host --nostdlib --nowrap --emit-llvm-text -o - %s 2>&1 | FileCheck %s -check-prefix=CHECK_CODEVIEW
+// RUN: %{ispc} -g --target=host --nostdlib --nowrap --emit-llvm-text -o - %s 2>&1 | FileCheck %s -check-prefix=CHECK_CODEVIEW
```
aneshlya, line 3: The test is supposed to be run on Windows host only, right? Please use `WINDOWS_HOST` in REQUIRES labels.

src/main.cpp
```diff
@@ -782,6 +782,7 @@ int main(int Argc, char *Argv[]) {
             if (2 <= val && val <= 4) {
```
dbabokin, line 782: The latest DWARF version is 5. I'm not sure if LLVM supports it (and what version of LLVM supports it). I suggest checking it and if it's supported changing the range of supported versions.

dbabokin (approved):
Looks good to me, just minor remarks below.

tests/lit-tests/debug_type_dwarf.ispc
nurmukhametov: Yes, it is intended to check that `-g` logic doesn't redefine debug info type.

tests/lit-tests/debug_type_dwarf.ispc
nurmukhametov: The test is supposed to be run on any host that supports windows as target-os. I thought that `WINDOWS_ENABLED` means exactly that.

src/main.cpp
nurmukhametov: It looks like LLVM started to somehow support DWARF v5 since 7 ([link](https://releases.llvm.org/7.0.0/docs/ReleaseNotes.html)). I have added 5 as option as the separate commit.

src/ispc.h
```diff
@@ -776,8 +776,12 @@ struct Globals {
         program in its output. */
     bool generateDebuggingSymbols;
 
+    /** Debug info type to generate. */
+    enum class DebugInfoType { None = 0, DWARF, CodeView };
+    DebugInfoType debugInfoType;
+
     /** Require generation of DWARF of certain version (2, 3, 4). For
-        default version, this field is set to 0. */
+        default version, this field is set by 3. */
```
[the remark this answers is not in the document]
nurmukhametov: fixed

tests/lit-tests/debug_type_dwarf.ispc
nurmukhametov: Oh, yeah. I have fixed that.

aneshlya: Now line 2 and 3 are the same :) please remove one of them

dbabokin (approved):

Issue #1248, opened 2016-12-10
Various compiler crashes/internal errors when using soa


Ashalah:
I've been running into various compiler crashes when trying to use the soa<n> feature.

For example:
```
struct Quaternion
{
	float x;
	float y;
	float z;
	float w;
};

export void QuaternionTestSOA(soa<32> Quaternion a[], uniform unsigned int count)
{
	foreach (i = 0 ... count)
	{
		a[i].x += 1.0;
	}
}
```

This crashes the compiler with the following output
```
ispc simple.ispc --target=sse2 --emit-asm -o out.s -h out.h -O2
simple.ispc:67:3: Performance Warning: Gather required to load value.
                a[i].x += 1.0;
  ^^^^^^

simple.ispc:67:3: Performance Warning: Scatter required to store value.
                a[i].x += 1.0;
  ^^^^^^

type.cpp:1162: Assertion failed: "m->errorCount > 0".
***
*** Please file a bug report at https://github.com/ispc/ispc/issues
*** (Including as much information as you can about how to reproduce this error).
*** You have apparently encountered a bug in the compiler that we'd like to fix!
***
```

The partially generated header ends in the following code before the compiler crashes:
```
#ifndef __ISPC_STRUCT_Quaternion__
#define __ISPC_STRUCT_Quaternion__
struct Quaternion_SOA32 {
    float x[32];
    float y[32];
    float z[32];
    float w[32];
};
#endif


///////////////////////////////////////////////////////////////////////////
// Functions exported from ispc code
///////////////////////////////////////////////////////////////////////////
#if defined(__cplusplus) && (! defined(__ISPC_NO_EXTERN_C) || !__ISPC_NO_EXTERN_C )
extern "C" {
#endif // __cplusplus
```

This empty export function also crashes the compiler as above (in type.cpp:1162), just without the performance warnings.

```
export void QuaternionTestSOA(soa<32> Quaternion a[], uniform unsigned int count)
{
}
```

The following version, where I try to copy data to a local soa array also crashes the compiler:
```
export void QuaternionTestSOA(void* uniform src, uniform unsigned int count)
{
	soa<32> Quaternion a[100];
	memcpy(a, src, 512);
	foreach (i = 0 ... count)
	{
		a[i].x += 1.0;
	}
}
```

Compiler output:
```
ispc simple.ispc --target=sse2 --emit-asm -o out.s -h out.h -O2
Invalid bitcast
[...]

PR #2636, opened 2023-08-23
tests: add regression tests for issues fixed in the past

nurmukhametov:
This PR adds several tests for issues that had been fixed in previous releases but issues have been opened until now.

### DIFF
--- a/tests/lit-tests/1238.ispc
+++ b/tests/lit-tests/1238.ispc
@@ -1,0 +1,34 @@
+// RUN: %{ispc} --arch=x86-64 --pic -O3 -DINLINE=inline --target=sse4-i32x8,avx1-i32x8,avx2-i32x8 --nostdlib --nowrap -o %t.ll --emit-llvm-text %s 2>&1 | FileCheck %s
+
+// REQUIRES: X86_ENABLED
+
+// CHECK-NOT: FATAL ERROR:
+struct Point3s {float x, y, z;};
+typedef struct Point3s Point3;
+
+#define NOISE_PERM_SIZE 8
+
+static uniform int NoisePerm[2 * NOISE_PERM_SIZE] = {
+    151, 160, 137, 91, 90, 15, 131, 13,
+    201, 95, 96, 53, 194, 233, 7, 225,
+};
+
+inline float Grad(int x, int y, int z) {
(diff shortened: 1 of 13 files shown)


tests/lit-tests/1191.ispc
```diff
@@ -0,0 +1,6 @@
+// RUN: not %{ispc} --target=host --nostdlib --nowrap -o - --emit-llvm-text %s 2>&1 | FileCheck %s
```
aneshlya, line 1: The issue number was 1197.

tests/lit-tests/1238.ispc
```diff
@@ -0,0 +1,52 @@
+// RUN: %{ispc} --arch=x86-64 --pic -O3 -DINLINE=inline --target=sse4-i32x8,avx1-i32x8,avx2-i32x8 --nostdlib --nowrap -o %t.ll --emit-llvm-text %s 2>&1 | FileCheck %s
```
aneshlya, line 1: I'd simplify this test. `-DINLINE` can be replaced with explicit `inline` specifier. Also NoisePerm can be reduced in size.

aneshlya (commented):
Please close #1248 as well.

nurmukhametov:
> Please close #1248 as well.

I am not sure that it should be closed because it consists of two parts. The first one is about compiler crash on some code. The second one is about the compiler considering some code as wrong. The first one is fixed, the second one is not.

tests/lit-tests/1191.ispc
nurmukhametov: Indeed, fixed.

tests/lit-tests/905.ispc
```diff
@@ -0,0 +1,4 @@
+// RUN: not %{ispc} --target=host --nostdlib --nowrap -o - --emit-llvm-text %s 2>&1 | FileCheck %s
```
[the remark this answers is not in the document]
nurmukhametov: fixed

tests/lit-tests/1238.ispc
nurmukhametov: I have reduced it.

aneshlya (approved):
Looks good!

PR #2640, opened 2023-08-25
centos/7/xpu_docker: support both tbb and oneapi-tbb

nurmukhametov:
Support via `build-arg TBB=<default|oneapi>` parameter building `ISPC` both with system tbb and oneapi-tbb.


aneshlya (approved):
LGTM!

docker/centos/7/xpu_ispc_build/Dockerfile
```diff
@@ -13,6 +13,7 @@ ARG LLVM_VERSION=15.0
 ARG L0L_VER=1.13.5
 ARG VC_INTRINSICS_COMMIT_SHA="910db4801d4a029834606e3e42a8d60358e74fdf"
 ARG SPIRV_TRANSLATOR_COMMIT_SHA="e82ecc2bd7295604fcf1824e47c95fa6a09c6e63"
+ARG TBB=default
```
dbabokin, line 16: Would be good to document here possible values and the reason we need two of them (when either of them should be used).

PR #2641, opened 2023-08-25
Fix function specifiers for templates

aneshlya:
- Produce error when template is used with `extern "C"`/ `extern "SYCL"`
- Fixed storage class for templates, so it's now extern by default as for regular ISPC functions
- Fixed linkage for different types of template instantiations (aligned with Clang behavior)

### DIFF
--- a/src/module.cpp
+++ b/src/module.cpp
@@ -1114,10 +1114,12 @@ void Module::AddFunctionTemplateDeclaration(const TemplateParms *templateParmLis
 
     // TODO: inline / noinline check.
 
-    // TODO: vectorcall checks
-
     // ...
 
+    // We don't need to worry here about passing vectorcall/regcall specifier since we support these
+    // conventions for extern "C"/extern "SYCL" functions only. And we don't support extern "C"/extern "SYCL" functions
+    // for templates. In case when --vectorcall option is passed on Windows, we will generate vectorcall functions for
+    // all functions.
     TemplateSymbol *funcTemplSym = new TemplateSymbol(templateParmList, name, ftype, sc, pos, isInline, isNoInline);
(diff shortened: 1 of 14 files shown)


nurmukhametov (approved):
LGTM!

src/parse.yy
```diff
@@ -2829,26 +2829,33 @@ lAddTemplateDeclaration(TemplateParms *templateParmList, DeclSpecs *ds, Declarat
...
-            Assert(m->errorCount > 0);
-            return;
-        }
+    }
+    // We can't support extern "C"/extern "SYCL" for templates because
+    // we need mangling information.
+    if (ds->storageClass == SC_EXTERN_C || ds->storageClass == SC_EXTERN_SYCL) {
+        Error(decl->pos, "Illegal linkage provided with function template.");
+        return;
+    }
```
dbabokin, line 2840: I think it make sense for clarity and readability to assert on possible `storageClass` values at this point. But if you prefer not to, it's not a problem.

src/module.cpp
```diff
@@ -1114,10 +1114,12 @@ void Module::AddFunctionTemplateDeclaration(const TemplateParms *templateParmLis
 
     // TODO: inline / noinline check.
 
-    // TODO: vectorcall checks
-
     // ...
 
+    // We don't need to worry here about passing vectorcall/regcall specifier since we support these
```
dbabokin, line 1119: Doesn't this mean that we should check that these specifiers are not present and generate the error otherwise?

dbabokin (approved):
Look good, but check the remarks below.

src/module.cpp
aneshlya: This check is in parse.yy. I've added it to lit test as well.

PR #2642, opened 2023-08-28
Support of template operators

### LINKED ISSUE #2634: Templated operator support
ISPC will fail to compile the following code. Having the ability to templatize operators is a useful feature for reducing code duplication.

https://ispc.godbolt.org/z/TjMfcMr1q

aneshlya:
Depends on https://github.com/ispc/ispc/pull/2641.
Introduces basic template support for operators.
Explicit instantiations and specializations are not supported yet.
Fixes #2634 .

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -1737,17 +1737,36 @@ bool lCreateBinaryOperatorCall(const BinaryExpr::Op bop, Expr *a0, Expr *a1, Exp
         std::string opName = std::string("operator") + lOpString(bop);
         std::vector<Symbol *> funs;
         m->symbolTable->LookupFunction(opName.c_str(), &funs);
-        if (funs.size() == 0) {
+        if (funs.size() > 0) {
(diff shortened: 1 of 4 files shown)


src/expr.cpp
```diff
@@ -1737,17 +1737,37 @@ bool lCreateBinaryOperatorCall(const BinaryExpr::Op bop, Expr *a0, Expr *a1, Exp
...
+            args->exprs.push_back(arg1);
+            op = new FunctionCallExpr(func, args, sp);
+            return abort;
+        }
+
+        // templates
+        std::vector<TemplateSymbol *> funcTempls;
+        m->symbolTable->LookupFunctionTemplate(opName.c_str(), &funcTempls);
+        if (funcTempls.size() > 0) {
+            std::vector<std::pair<const Type *, SourcePos>> *vec = new std::vector<std::pair<const Type *, SourcePos>>;
```
dbabokin, line 1753: Seems that it needs to be a vector, not a pointer to a vector. This way no need to dereference when passing it and no need to delete.

Issue #2635, opened 2023-08-22
Unmasked template triggers assert


rkaric:
https://ispc.godbolt.org/z/asG7G3jf4

aneshlya:
As a WA, it'll work if you have `unmasked` in both template definition and template specialization: https://ispc.godbolt.org/z/1nejajqTj

PR #2648, opened 2023-09-01
Function specifiers for template specializations and instantiations 

aneshlya:
This PR fixes and aligns usage of different function specifiers (storage class, inline hints, unmasked keyword etc) for template specializations and instantiations.  
* The keywords ``export``, ``task``, ``typedef``, ``extern "C"`` and ``extern "SYCL"``
  are not allowed.
* Calling conventions such as ``__vectorcall`` and ``__regcall`` must be used in conjunction
  with ``extern "C"`` or ``extern "SYCL"``, so they are not allowed as well.
* Performance hints like ``inline`` and ``noinline`` are allowed. Primary template, template
  specializations and explicit instantiations may have different ``inline`` hints.
* Storage types ``extern`` and ``static`` are allowed. Template specializations and explicit
  instantiations must share the same storage type as the primary template.
  If not specified, the storage type will be inherited from the primary template.
* ``unmasked`` specifier is allowed. Template specializations and explicit instantiations
  must maintain consistency with the primary template regarding the ``unmasked`` specifier.
  You cannot specify ``unmasked`` for a template specialization if it was not previously
  specified for the primary template. If unspecified, it will be inherited from the
  primary template.
Fixes https://github.com/ispc/ispc/issues/2635

### DIFF
--- a/src/parse.yy
+++ b/src/parse.yy
@@ -102,6 +102,10 @@ static void lAddThreadIndexCountToSymbolTable(SourcePos pos);
 static std::string lGetAlternates(std::vector<std::string> &alternates);
 static const char *lGetStorageClassString(StorageClass sc);
 static bool lGetConstantInt(Expr *expr, int *value, SourcePos pos, const char *usage);
+
+enum class TemplateType { Template, Instantiation, Specialization };
+static void lCheckTemplateDeclSpecs(DeclSpecs *ds, SourcePos pos, TemplateType type, const char* name);
+
 static EnumType *lCreateEnumType(const char *name, std::vector<Symbol *> *enums,
                                  SourcePos pos);
 static void lFinalizeEnumeratorSymbols(std::vector<Symbol *> &enums,
@@ -2566,6 +2570,7 @@ template_function_instantiation
           }
 
           d->InitFromDeclSpecs($2);
+          lCheckTemplateDeclSpecs($2, d->pos, TemplateType::Instantiation, $3->first->name.c_str());
           const FunctionType *ftype = CastType<FunctionType>(d->type);
 
           m->AddFunctionTemplateInstantiation($3->first->name, *$3->second, ftype, Union(@1, @6));
@@ -2582,6 +2587,7 @@ template_function_instantiation
           d->child = simpleTemplID->first;
 
           d->InitFromDeclSpecs($2);
+          lCheckTemplateDeclSpecs($2, d->pos, TemplateType::Instantiation, $3->first->name.c_str());
           const FunctionType *ftype = CastType<FunctionType>(d->type);
 
           m->AddFunctionTemplateInstantiation($3->first->name, *$3->second, ftype, Union(@1, @5));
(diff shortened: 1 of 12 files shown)


src/parse.yy
```diff
@@ -2813,53 +2819,70 @@ lAddDeclaration(DeclSpecs *ds, Declarator *decl) {
...
+            templateTypeStr = "function template";
+            break;
+        case TemplateType::Instantiation:
+            templateTypeStr = "template instantiation";
+            break;
+        case TemplateType::Specialization:
+            templateTypeStr = "template specialization";
+            break;
+        default:
+            templateTypeStr = "function template";
```
dbabokin, line 2837: I think it should be an error, should it?

src/parse.yy
```diff
@@ -2813,53 +2819,70 @@ lAddDeclaration(DeclSpecs *ds, Declarator *decl) {
...
+        Error(pos, "'task' not supported for %s.", templateTypeStr.c_str());
         return;
     }
     if (ds->typeQualifiers & TYPEQUAL_EXPORT) {
-        Error(decl->pos, "'export' not supported for templates.");
+        Error(pos, "'export' not supported for %s.", templateTypeStr.c_str());
+        return;
+    }
+    if (ds->storageClass == SC_TYPEDEF) {
+        // FIXME: source position
```
dbabokin, line 2849: What's wrong with source position here? Elaborate or remove.

I see that's coming from existing code, but nevertheless.

src/parse.yy
```diff
@@ -2813,53 +2819,70 @@ lAddDeclaration(DeclSpecs *ds, Declarator *decl) {
...
+    }
+    bool isRegCall = (ds->typeQualifiers & TYPEQUAL_REGCALL);
+    if (isRegCall) {
+        Error(pos, "Illegal to use \"__regcall\" qualifier on non-extern function \"%s\".", name);
+    }
+}
+
+static void
+lAddTemplateDeclaration(TemplateParms *templateParmList, DeclSpecs *ds, Declarator *decl) {
+    if (ds == nullptr || decl == nullptr)
```
dbabokin, line 2872: Add `{}`. It's already more than one line with the comment, easy to add something and do no notice missing `{}`.

I see that's coming from existing code, but nevertheless.

src/parse.yy
```diff
@@ -2876,34 +2899,26 @@ lAddTemplateSpecialization(const std::vector<std::pair<const Type *, SourcePos>>
...
-        const FunctionType *ftype = CastType<FunctionType>(decl->type);
-        if (ftype != nullptr) {
-            m->AddFunctionTemplateSpecializationDeclaration(decl->name, ftype, types, decl->pos);
-        }
-        else {
-            Error(decl->pos, "Only function template specializations are supported.");
-        }
+    const FunctionType *ftype = CastType<FunctionType>(decl->type);
+    if (ftype != nullptr) {
+        //bool isInline = (ds->typeQualifiers & TYPEQUAL_INLINE);
```
dbabokin: Do you need these commented out lines?

src/type.cpp
```diff
@@ -2727,6 +2727,22 @@ const Type *FunctionType::GetAsConstType() const { return this; }
 
 const Type *FunctionType::GetAsNonConstType() const { return this; }
 
+const Type *FunctionType::GetAsUnmaskedType() const {
```
dbabokin, line 2730: Two things here.

1. Check if the function is already masked/unmasked, possibly no need to create a new one.
2. It's better cache the result (i.e. add the fields `asUnmaskedType`, `asMaskedType`) and ensure that `ftype-> GetAsUnmaskedType()->GetAsMaskedType() == ftype`.

It's not necessary with current usage strictly speaking, but it could be assumed, as this is how `GetAsUniformType()`/`GetAsVaryingType()` are implemented.

PR #2651, opened 2023-09-06
Emit struct keyword for vectors to C/C++ header

### LINKED ISSUE #2650: Missing `struct` tags in header when referring to builtin vector types
Compile the following file to a C header with `ispc repro.ispc -h repro.h` (on ISPC v1.21):

`repro.ispc`:
```cpp
struct Test {
    uint<2> size;
};

export void use(uniform Test &test) {
}
```

Results in the following header file (preprocessor noise omitted for brevity):

```c
#ifdef _MSC_VER
__declspec( align(16) ) struct uint32_t2 { uint32_t v[2]; };
#else
struct uint32_t2 { uint32_t v[2]; } __attribute__ ((aligned(16)));
#endif

struct Test {
    uint32_t2  size; // error: must use 'struct' tag to refer to type 'uint32_t2'
};

extern void use(struct Test &test);
```

It seems ISPC went "out of its way" to flatten the generic `uint` vector down to a type that C could understand (i.e. no C++ `template` for the element count), but a C compiler will complain here that uses of `uint32_t2` have to be prefixed with the `struct` tag (unless it were defined with a `typedef`). Note that the `struct` tag _is added_ for the use of `struct Test` within the exported function.

Since the documentation at https://ispc.github.io/ispc.html references C/C++ a lot, is it expected that this header is parsable by a C compiler, and how should this be fixed? It is currently hampering us from generating proper Rust bindings with `bindgen` which requires "pure C": https://github.com/Twinklebear/ispc-rs/issues/27

nurmukhametov:
This fixes #2650

### DIFF
--- a/tests/lit-tests/2650.ispc
+++ b/tests/lit-tests/2650.ispc
@@ -1,0 +1,12 @@
+// RUN: %{ispc} --target=host --nostdlib --nowrap -h %t.h %s 2>&1
+// RUN: FileCheck %s --input-file=%t.h
+// RUN: %{cc} %t.h
+
+// CHECK: struct uint32_t2 size;
+
+struct Test {
+    uint<2> size;
+};
+
+export void use(uniform Test * uniform test) {
+}
--- a/src/module.cpp
+++ b/src/module.cpp
@@ -1732,6 +1732,8 @@ static void lEmitStructDecl(const StructType *st, std::vector<const StructType *
         if (!ftype->IsArrayType() && !ftype->IsPointerType() && ftype->IsVaryingType() &&
             (CastType<StructType>(ftype) == nullptr)) {
             fprintf(file, "%s[%d];\n", d.c_str(), g->target->getVectorWidth());
+        } else if (CastType<VectorType>(ftype) != nullptr) {
+            fprintf(file, "struct %s;\n", d.c_str());
         } else {
             fprintf(file, "%s;\n", d.c_str());
         }


tests/lit-tests/2650.ispc
```diff
@@ -0,0 +1,9 @@
+// RUN: %{ispc} --target=host --nostdlib --nowrap -h %t.h %s 2>&1
+// RUN: %{cc} %t.h
+
```
e3m3, line 3:
```suggestion
// RUN: %{ispc} --target=host --nostdlib --nowrap -h %t.h %s 2>&1
// RUN: FileCheck %s --input-file=%t.h
// RUN: %{cc} %t.h

// CHECK: struct uint32_t2

```

nurmukhametov: Done

e3m3 (approved):
LGTM

tests/lit-tests/2650.ispc
```diff
@@ -0,0 +1,12 @@
+// RUN: %{ispc} --target=host --nostdlib --nowrap -h %t.h %s 2>&1
+// RUN: FileCheck %s --input-file=%t.h
+// RUN: %{cc} %t.h
+
+// CHECK: struct uint32_t2
```
aneshlya, line 5: Please add `// CHECK-LABEL: struct Test` before this check, otherwise it will be found in `struct uint32_t2` declaration itself.

MarijnS95: @aneshlya theoretically `struct Test` could _also_ be found in `extern void use(struct Test &test);` a few lines later, but I assume:

1. `CHECK-LABEL` will match on the first `struct Test` it finds (which will be the declaration here);
2. After `extern void use(struct Test &test);` `CHECK: struct uint32_t2` won't be found and would falsely fail rather than falsely succeed.


Perhaps it is more clear to write out `// CHECK struct uint32_t2 size;` fully?

aneshlya: `CHECK-LABEL` should match the first occurrence. Anyway I agree that checking `// CHECK struct uint32_t2 size` would be even better.

tests/lit-tests/2650.ispc
```diff
@@ -0,0 +1,12 @@
+// RUN: %{ispc} --target=host --nostdlib --nowrap -h %t.h %s 2>&1
```
dbabokin, line 1: Why not using pipe here? I.e. `... 2>&1 | FileCheck ...`

dbabokin: Ah, I missed that output is not streamed as well. I suggest doing `%{ispc} --target=host --nostdlib --nowrap %s -h - 2>&1 | FileCheck %s`.

This would also check that `-` properly works for header output. Not a big deal, but would not harm.

MarijnS95: @dbabokin the output appears to be used _twice_, once in `FileCheck` and once in `cc`, so that might not work?

nurmukhametov: > Ah, I missed that output is not streamed as well. I suggest doing `%{ispc} --target=host --nostdlib --nowrap %s -h - 2>&1 | FileCheck %s`.

It doesn't work as expected. It just creates the header file names `-`.

> This would also check that `-` properly works for header output. Not a big deal, but would not harm.

It doesn't look like working at the moment.

tests/lit-tests/2650.ispc
nurmukhametov: I check `struct uint32_t2 size` now.

PR #2654, opened 2023-09-08
Fix C compatibility for headers generated by ISPC

### LINKED ISSUE #2652: Transform reference types to pointer types when generating C header
The header generated by `ispc -h` is supposed to be C-compliant (https://github.com/ispc/ispc/issues/2650#issuecomment-1708859970), but when using references in ISPC (which should be ABI-compatible with raw pointers) they are emitted as C++ references in the header. Is this expected?

```cpp
struct Test {
  uint &my_data;
};

export void use(uniform Test &test) {}
```

Run `ispc repro.ispc -h repro.h` to generate:

```cpp
struct Test {
    uint32_t &my_data;
};

extern void use(struct Test &test);
```

Transforming them does make for a degraded experience in C++ I think, but a (verbose!) middle ground can be found with `#if defined(__cplusplus)` which is already used around `extern "C"` blocks for example.

nurmukhametov:
In some cases, `ispc` generates headers that are not C compatible. This PR fixes two aspects:
1. Generate a raw pointer instead of a reference for function declarations. 
2. Same for member declarations inside structs.
1. Provide bool type definition either with stdbool.h for C99 (and newer) or via typedef.

`*Type::GetCDeclration` is changed to `*Type::GetDeclaration` that returns C or C++ function declaration syntax depending on the second parameter. `lPrintDeclaration` collects both definitions. If they are different then it emits both of them under corresponding ifdefs.

It fixes #2652


src/type.h
```diff
@@ -205,9 +205,10 @@ class Type : public Traceable {
         not to clash. */
     virtual std::string Mangle() const = 0;
 
-    /** Returns a string that is the declaration of the same type in C
-        syntax. */
-    virtual std::string GetCDeclaration(const std::string &name) const = 0;
+    /** Returns a string that is the declaration of the same type in C or C++
+        syntax. The second argument controls which syntax C (false) or
```
aneshlya, line 209: `C (false) or C++ (true)` -> `C (true) or C++ (false)`.

src/type.h
```diff
@@ -205,9 +205,10 @@ class Type : public Traceable {
...
+        C++ (true) is generated. */
+    virtual std::string GetDeclaration(const std::string &name, bool cpp = false) const = 0;
```
aneshlya, line 211: It's up to you, but I would prefer to have 
```
enum DeclarationType {
C,
Cpp,
}
```
and then use it instead of bool variable.
I think it would improve readability of the code and I prefer to avoid default parameters when possible because it's a potential source of future errors.

aneshlya:
For the lit tests which doesn't have issue number in the name, it would be good to add a description at the beginning explaining what the test is checking.

src/type.h
dbabokin: I totally agree with Arina, please do this change.

nurmukhametov: I have introduced enum and removed default parameter.

tests/lit-tests/header-bool-1.ispc
```diff
@@ -0,0 +1,37 @@
+// RUN: %{ispc} -DISPC --pic --target=host -h %t.h %s -o %t.o
+// RUN: %{cc} -x c -c %s -o %t.c.o --include %t.h
+// RUN: %{cc} %t.o %t.c.o -o %t.c.bin
+// RUN: %t.c.bin | FileCheck %s
+// RUN: %{cc} -x c++ -c %s -o %t.cpp.o --include %t.h
+// RUN: %{cc} %t.o %t.cpp.o -o %t.cpp.bin
+// RUN: %t.cpp.bin | FileCheck %s
+
+// CHECK-LABEL: bool x=false
+// CHECK: bool x=true
```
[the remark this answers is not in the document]
nurmukhametov: Fixed

aneshlya (approved):
Looks good to me!

PR #2658, opened 2023-09-18
cmake: add switch to opt out examples from package

nurmukhametov:
Switch ISPC_PACKAGE_EXAMPLES allows to enable/disable packaging examples into ISPC archive. Default is ON.
Add custom target `package-examples` to package them into separate archive from ISPC itself.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,6 +89,7 @@ option(ISPC_INCLUDE_RT "Generate build targets for ISPC runtime." ON)
 option(ISPC_INCLUDE_UTILS "Generate build targets for the utils." ON)
 option(ISPC_INCLUDE_UTILS_ONLY "Generate build targets only for the utils." OFF)
 option(ISPC_PREPARE_PACKAGE "Generate build targets for ispc package" OFF)
+option(ISPC_PACKAGE_EXAMPLES "Pack examples into the ISPC package" ON)
 
 option(ISPC_OPAQUE_PTR_MODE "Build ISPC with usage of opaque pointers" OFF)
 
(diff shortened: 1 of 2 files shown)


aneshlya (commented):
Could you add a CI job which would run `package-examples` target?

CMakeLists.txt
```diff
@@ -763,10 +764,12 @@ endif()
...
 if (ISPC_PREPARE_PACKAGE)
-    if (XE_ENABLED)
-        install (DIRECTORY "${PROJECT_SOURCE_DIR}/examples/" DESTINATION examples)
-    else()
-        install (DIRECTORY "${PROJECT_SOURCE_DIR}/examples/" DESTINATION examples PATTERN "xpu" EXCLUDE)
+    if (ISPC_PACKAGE_EXAMPLES)
+        if (XE_ENABLED)
+            install (DIRECTORY "${PROJECT_SOURCE_DIR}/examples/" DESTINATION examples)
+        else()
+            install (DIRECTORY "${PROJECT_SOURCE_DIR}/examples/" DESTINATION examples PATTERN "xpu" EXCLUDE)
```
e3m3, line 771: One option to remove redundancy is to set the value of the pattern in this branch and then call `install` once afterwards (with empty pattern for Xe).

CMakeLists.txt
```diff
@@ -878,4 +885,20 @@ if (ISPC_PREPARE_PACKAGE)
     set(CPACK_STRIP_FILES TRUE)
 
     include(CPack)
+
+    # Custom target to pack only examples
+    if (WIN32)
+        add_custom_target(package-examples
+            COMMENT "Package examples directory to ${EXAMPLES_PACKAGE_FILE_NAME}.zip"
+            COMMAND ${CMAKE_COMMAND} -E tar cf "${CMAKE_BINARY_DIR}/${EXAMPLES_PACKAGE_FILE_NAME}.zip" --format=zip examples
+            WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
```
e3m3, line 894: Didn't know tar handled zips! 👍

e3m3 (approved):
LGTM

CMakeLists.txt
nurmukhametov: This is cmake tool: https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-E-arg-tar

nurmukhametov:
> Could you add a CI job which would run `package-examples` target?

I have added two jobs: [win-package-examples](https://github.com/ispc/ispc/actions/runs/6262182690/job/17003793345?pr=2658) and [linux-package-examples](https://github.com/ispc/ispc/actions/runs/6262182690/job/17003790854?pr=2658)

PR #2667, opened 2023-09-28
Provide own definition of __libcpp_verbose_abort

nurmukhametov:
This fixes build error on macOS like this [here](https://github.com/ispc/ispc/actions/runs/6333187132/job/17200927555)
```bash
Undefined symbols for architecture x86_64:
  "std::__1::__libcpp_verbose_abort(char const*, ...)", referenced from:

```

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -40,6 +40,23 @@
 
 #include <llvm/IR/DataLayout.h>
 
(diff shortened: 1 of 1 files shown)


aneshlya (approved):
Looks good!

PR #2669, opened 2023-09-28
snap: add home plugs

### LINKED ISSUE #2665: Snap ispc for ubuntu can't compile: permission denied
I installed on ubuntu the ispc compiler with the command
`snap install ispc`

Then I tried compiling this example

```
  // filename: sum_vector.ispc
  export uniform double sum_vector(uniform double A[], uniform int n){
      double sum = 0;
      foreach (i = 0 ... n)
	          sum += A[i];
  
      return reduce_add(sum);
  }
```

With the command
`ispc sum_vector.ispc -o sum_vector.obj -h sum_vector.h`

But I get the error
```
Warning: No --target specified on command-line. Using default system target "avx2-i32x8". 
sum_vector.ispc: Permission denied
```

If I download the binaries from https://github.com/ispc/ispc/releases/download/v1.18.0/ispc-v1.18.0-linux.tar.gz the same commands work fine

nurmukhametov:
It allows snap packaged ISPC to access files in the user home directory.
It also adds step to upload the snap package as artifact in CI snap job .
It also add an option to choose the channel to publish snap to.

This fixes #2665

### DIFF
--- a/.github/workflows/snap.yml
+++ b/.github/workflows/snap.yml
@@ -23,6 +23,16 @@ on:
         required: true
         defaule: false
         type: boolean
+      channel:
+        description: 'Choose the channel to publish to'
+        required: true
+        default: 'stable'
+        type: choice
+        options:
+          - 'stable'
+          - 'candidate'
+          - 'beta'
+          - 'edge'
 
 jobs:
   snap-arm64:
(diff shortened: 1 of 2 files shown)


aneshlya (approved):
You may want to split the change to several commits but other than that looks good!

PR #2671, opened 2023-10-03
Updates for Xe targets (23)

nurmukhametov:
This PR updates dependencies:
- LLVM to 16 version,
- Level zero loader dependency to [v1.14.0](https://github.com/oneapi-src/level-zero/releases/tag/v1.14.0),
- SPIRV-LLVM Translator,
- vc-intrinsics.

It also fixes genx intrinsics declarations attributes that might have been resulted in performance regression on some workloads.

The PR also have some Xe-related minor changes:
- a couple of static analysis issues,
- superbuild related fixes,
- fix examples and cmake files to align with the latest sycl version.


aneshlya (approved):
Looks good!

PR #2673, opened 2023-10-04
CI: add job to build XE_ENABLED build

nurmukhametov:
This PR add jobs to build XE_ENABLED ISPC on Linux and Windows.

### DIFF
--- a/.github/workflows/ispc-ci.yml
+++ b/.github/workflows/ispc-ci.yml
@@ -254,6 +254,59 @@ jobs:
         name: ispc_llvm17_linux
         path: build/ispc-trunk-linux.tar.gz
 
+  linux-build-ispc-xe-llvm16-release:
+    needs: [define-flow]
+    runs-on: ubuntu-latest
+    env:
+      LLVM_VERSION: "16.0"
+      LLVM_TAR: llvm-16.0.6-ubuntu18.04-Release-x86.arm.wasm.tar.xz
+      ISPC_OPAQUE_PTR_MODE: "OFF"
+      INSTALL_COMPUTE_RUNTIME: 1
+
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: true
+
+    - name: Install dependencies
+      run: |
+        .github/workflows/scripts/install-build-deps.sh
+
+    - name: Check environment
+      run: |
+        ./check_env.py
+        which -a clang
+        cat /proc/cpuinfo
+
+    - name: Build package
+      run: |
+        cmake superbuild \
+            -B build \
+            --preset os \
+            -DINSTALL_WITH_XE_DEPS=ON \
+            -DISPC_INCLUDE_BENCHMARKS=ON \
+            -DPREBUILT_STAGE2_PATH=${GITHUB_WORKSPACE}/bin-${LLVM_VERSION} \
+            -DCMAKE_INSTALL_PREFIX=${GITHUB_WORKSPACE}/ispc-xe
+        cmake --build build
+
+    - name: Sanity testing (make check-all, make test)
+      run: |
+        find build -name check_isa -exec {} \;
+        find build -name ispc -exec {} --support-matrix \;
+        cmake --build build --target ispc-stage2-check
+
+    - name: Upload XE enabled package
+      uses: actions/upload-artifact@v3
+      with:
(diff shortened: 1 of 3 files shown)


.github/workflows/ispc-ci.yml
```diff
@@ -1103,6 +1256,52 @@ jobs:
...
+    - name: Check
+      run: |
+        # Print fails to the log.
+        git diff --exit-code
+
+    - name: Upload fail_db.txt
+      uses: actions/upload-artifact@v3
+      if: failure()
+      with:
+        name: fail_db.xe.llvm15rel.${{matrix.arch}}.${{matrix.target}}.txt
```
aneshlya, line 1302: fail_db.xe.llvm15rel.${{matrix.arch}}.${{matrix.target}}.txt -> fail_db.xe.llvm16rel.${{matrix.arch}}.${{matrix.target}}.txt

.github/workflows/ispc-ci.yml
```diff
@@ -496,6 +549,48 @@ jobs:
...
+    - name: Check
+      run: |
+        # Print fails to the log.
+        git diff --exit-code
+
+    - name: Upload fail_db.txt
+      uses: actions/upload-artifact@v3
+      if: failure()
+      with:
+        name: fail_db.xe.llvm15rel.${{matrix.target}}.txt
```
aneshlya, line 591: fail_db.xe.llvm15rel.${{matrix.target}}.txt -> fail_db.xe.llvm16rel.${{matrix.target}}.txt

.github/workflows/scripts/install-build-deps.sh
```diff
@@ -22,6 +22,22 @@ do
   fi
 done
 
+if [ -v INSTALL_COMPUTE_RUNTIME ]; then
+    echo "install Compute Runtime"
+    wget -U "$USER_AGENT" --retry-connrefused --waitretry=5 --read-timeout=20 --timeout=15 -t 5 \
+        https://github.com/intel/intel-graphics-compiler/releases/download/igc-1.0.14828.8/intel-igc-core_1.0.14828.8_amd64.deb
```
aneshlya, line 28: I'd prefer not to keep/update CR version in two more files in addition to what we already have in Dockerfiles unless we really need to use a particular version of CR runtime here. Could you try to install here version-independent agama packages like we do in https://github.com/ispc/ispc/blob/main/docker/ubuntu/20.04/xpu_ispc_build/Dockerfile#L46?
Docs: https://dgpu-docs.intel.com/driver/installation.html

.github/workflows/ispc-ci.yml
```diff
@@ -496,6 +549,48 @@ jobs:
...
 
+  # Test xe release version
+  linux-test-xe-llvm16-release:
+    needs: [define-flow, linux-build-ispc-xe-llvm16-release]
+    runs-on: ubuntu-latest
+    continue-on-error: false
+    strategy:
+      fail-fast: false
+    env:
+      INSTALL_COMPUTE_RUNTIME: 1
```
aneshlya, line 560: Since you run only CPU testing you don't need to install Compute Runtime here. But if it is done "to not forget in the future when we might have runner with GPU", I'm OK with it, just leave the comment.

nurmukhametov: If I remember correctly, `ocloc` compiler is needed for `lit-tests`.

aneshlya: Yes, so for build tasks which include lit testing, CR is needed, but for testing jobs - it's not needed IMO.

.github/workflows/ispc-ci.yml
nurmukhametov: done

.github/workflows/ispc-ci.yml
nurmukhametov: done

.github/workflows/scripts/install-build-deps.sh
nurmukhametov: Fixed

.github/workflows/ispc-ci.yml
nurmukhametov: Disabled `INSTALL_COMPUTE_RUNTIME` for test-xe.

aneshlya (approved):
Looks good!

PR #2675, opened 2023-10-09
interop: fix find dpcpp_compiler logic

nurmukhametov:
The first thing, oneAPI changed the layout of llvm-utilities so make the corresponding change by providing extra subfolders as hints where to find them.
The second one, on Windows oneAPI provides sycl libraries with version in name, e.g., sycl6.dll, sycl7.dll and etc.. Add versioned name in find_labrary to be able to find them.

### DIFF
--- a/ispcrt/cmake/Finddpcpp_compiler.cmake
+++ b/ispcrt/cmake/Finddpcpp_compiler.cmake
@@ -24,7 +24,8 @@ if (DPCPP_COMPILER)
     endif()
 
     # oneAPI distributions ship the llvm utility binaries under bin-llvm
-    set(DPCPP_LLVM_BIN_HINT "bin-llvm")
+    # or under bin/compiler (since oneAPI release 2023.3?)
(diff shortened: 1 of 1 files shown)


ispcrt/cmake/Finddpcpp_compiler.cmake
```diff
@@ -25,6 +25,8 @@ if (DPCPP_COMPILER)
 
     # oneAPI distributions ship the llvm utility binaries under bin-llvm
     set(DPCPP_LLVM_BIN_HINT "bin-llvm")
+    # or under bin/compiler (since oneAPI release 2023.3?)
+    set(DPCPP_LLVM_BIN_HINT2 "bin/compiler")
```
aneshlya, line 29: I suggest to combine both hints together: `set(DPCPP_LLVM_BIN_HINT "bin-llvm" "bin/compiler")`

nurmukhametov: fixed

aneshlya (approved):
LGTM!

PR #2680, opened 2023-10-12
Fix spelling errors.

mzf-guest:
This PR fixes small spelling error in messages which appears in the ispc binary.


nurmukhametov (changes requested):
Thanks for submitting PR!

You need to fix spelling of this message in tests as well to make CI green.
Here: https://github.com/ispc/ispc/blob/015f8fc1fcc8fdc7973116d2e658be6faad12abf/tests/lit-tests/intrinsic_target_specific.ispc#L7

mzf-guest:
@nurmukhametov Thanks for the guidance! I've updated the test file according to your suggestion. Let me know if this fixes the testing process.

Best,
François

nurmukhametov (approved):

PR #2684, opened 2023-10-16
Enable nontrivial unswitch in Simple Loop Unswitch

nurmukhametov:
Before v1.20.0, ISPC used LoopUnswitch that was removed and changed by SimpleLoopUnswitch in LLVM 15. However, it stops to optimize by default loops like this:
```
export void foo(const uniform uint32 input[], uniform uint32 output[],
                const uniform int index)
{
    foreach (i = 0 ... 32) {
        if (index == 0) {
                output[i] = input[i];
        } else {
                output[i] |= input[i];
        }
    }
}
```
To turn on unswithcing for this loop with SimpleLoopUnswitch, we need to enable unswitching of untrivial loops with --enable-nontrivial-unswitch.

### DIFF
--- a/src/opt.cpp
+++ b/src/opt.cpp
@@ -641,7 +641,8 @@ void ispc::Optimize(llvm::Module *module, int optLevel) {
             // not efficient for Xe targets. Moreover when this pass is used
             // some integer division tests are failing on TGLLP Windows.
             // Disable this pass on Xe until the problem is fixed on BE side.
-            optPM.addLoopPass(llvm::SimpleLoopUnswitchPass(false), 293);
+            // Note: enable both trivial and non-trivial loop unswitching.
+            optPM.addLoopPass(llvm::SimpleLoopUnswitchPass(true /* NonTrivial */, true /* Trivial */), 293);
         }
         optPM.commitLoopToFunctionPassManager(true, true);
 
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -1233,16 +1233,17 @@ int main(int Argc, char *Argv[]) {
     if (targets.size() > 1)
         g->isMultiTargetCompilation = true;
 
+    std::vector<const char *> Args;
+    Args.push_back("ispc (LLVM option parsing)");
+    Args.push_back("--enable-nontrivial-unswitch");
```
aneshlya, line 1238: Is `--enable-nontrivial-unswitch` equivalent of passing `true` to llvm::SimpleLoopUnswitchPass constructor?

nurmukhametov: Yes

aneshlya: Then I suggest pass it through constructor. Using of `llvm::cl` options is unintuitive and hidden, we use it only where there is no other way of passing options.

aneshlya:
Please add a lit test.

src/main.cpp
nurmukhametov: Done

nurmukhametov:
> Please add a lit test.

I thought about it but I don't understand what to check there.

nurmukhametov:
> > Please add a lit test.
> 
> I thought about it but I don't understand what to check there.

Done

aneshlya (approved):

PR #2686, opened 2023-10-20
Fix stability runs

nurmukhametov:
This PR fixes two issues with the full Tests pipeline:
1. Full Test pipeline has strange failures and hangs for win-tests on x86 avx1 targets. It is caused by misconfiguration in alloy.py code that resulted in same tests had been run twice: the first time natively on hardware, the second time in SDE. Targets after AVX1.1 have to include both AVX and AVX1.1 otherwise one of them will run under SDE.
2. LLVM14 tests jobs are matrix and have to be run with matrix arguments.

### DIFF
--- a/.github/workflows/ispc-ci.yml
+++ b/.github/workflows/ispc-ci.yml
@@ -333,8 +333,7 @@ jobs:
     - name: Running tests
       run: |
         echo PATH=$PATH
-        # Just sanity check for 14.0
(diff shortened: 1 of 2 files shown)


nurmukhametov:
Full tests pipeline run (https://github.com/ispc/ispc/actions/runs/6586806133/job/17923638652)

PR #2690, opened 2023-10-24
Fix casting of bool short vectors to numeric ones

### LINKED ISSUE #2689: Error/crash "Type too small for SExt"
Hi,
This code does not compile:
```
struct InvertInvertedSignAndIndexes
{
	double<3> InvVec_;
	uint8<3> Sign_;
};

uniform InvertInvertedSignAndIndexes getInvertInvertedSignAndIndexes()
{
	uniform InvertInvertedSignAndIndexes res;

	res.Sign_ = res.InvVec_ < 0.0;

	return res;
}
```
with error:
> Type too small for SExt
>   %19 = sext <4 x i1> %18 to <8 x i1>, !dbg !24061
> LLVM ERROR: Broken module found, compilation aborted!
> /usr/local/src/ispc/src/main.cpp(404): FATAL ERROR: Unhandled signal sent to process; terminating.
> Program terminated with signal: SIGSEGV
> Compiler returned: 139

See for example : https://godbolt.org/z/4rx9r5dv6

Can I workaround this one before you get it fixed ?
Regards,
Colin Chargy

nurmukhametov:
This PR fixes #2689.

ISPC generated a cast like this:
```c
%19 = sext <4 x i1> %18 to <8 x i1>, !dbg !24061
```
by this https://github.com/ispc/ispc/blob/5725fb8d495980af52b417ae0c7a89e2a258d2bd/src/expr.cpp#L6494

`LLVMTypes::Int1VectorType` always has target width whereas short vector length can be different.

### DIFF
--- a/tests/lit-tests/2689.ispc
+++ b/tests/lit-tests/2689.ispc
@@ -1,0 +1,24 @@
+// RUN: %{ispc} --target=host --nowrap --nostdlib -O2 --emit-llvm-text %s -o - | FileCheck %s --check-prefix=COMMON
+// RUN: %{ispc} --target=avx512skx-x32 --nowrap --nostdlib -O2 --emit-llvm-text %s -o - | FileCheck %s --check-prefix=AVX512_X32
+// RUN: %{ispc} --target=avx512skx-x16 --nowrap --nostdlib -O2 --emit-llvm-text %s -o - | FileCheck %s --check-prefix=COMMON
+// RUN: %{ispc} --target=avx2-i32x4 --nowrap --nostdlib -O2 --emit-llvm-text %s -o - | FileCheck %s --check-prefix=COMMON
+
+// COMMON-LABEL: @foo___unT_3C_4_3E_unT_3C_4_3E_(
+// COMMON-NEXT: allocas:
+// COMMON-NEXT: [[CMP:%.*]] = icmp ult <4 x i8> %a, %b
+// COMMON-NEXT: [[CAST:%.*]] = zext <4 x i1> [[CMP]] to <4 x i8>
+// COMMON-NEXT: ret <4 x i8> [[CAST]]
+
+// By some reason, avx512-x32 has 8 wide vector for uint8<4>, x64 has 16
+// AVX512_X32-LABEL: @foo___unT_3C_4_3E_unT_3C_4_3E_(
+// AVX512_X32-NEXT: allocas:
+// AVX512_X32-NEXT: [[CMP:%.*]] = icmp ult <8 x i8> %a, %b
+// AVX512_X32-NEXT: [[CAST:%.*]] = zext <8 x i1> [[CMP]] to <8 x i8>
+// AVX512_X32-NEXT: ret <8 x i8> [[CAST]]
+
+// REQUIRES: X86_ENABLED
+
+uniform uint8<4> foo(uniform uint8<4> a, uniform uint8<4> b)
+{
+	return a < b;
+}
--- a/tests/lit-tests/bool-casts-varying.ispc
+++ b/tests/lit-tests/bool-casts-varying.ispc
@@ -1,0 +1,91 @@
+// RUN: %{ispc} --target=host --nowrap --nostdlib -O2 --emit-llvm-text %s -o - | FileCheck %s
+
+// CHECK-LABEL: @to_float16___vyhvyh(
+// CHECK-NEXT: allocas:
+// CHECK-NEXT: [[CMP:%.*]] = fcmp olt <[[WIDTH:.*]] x [[TYPE:.*]]> %a, %b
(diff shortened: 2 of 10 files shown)


tests/lit-tests/bool-store-varying.ispc
```diff
@@ -0,0 +1,25 @@
...
+// CHECK-NEXT: store <[[WIDTH]] x i8> [[CAST]], {{.*}} @y
+// CHECK-NEXT: ret void
+
+void foo(uint8 a, uint8 b)
+{
+    y = a < b;
+}
+
+// Note:
+// It is quite suspicous that bool is stored in memory as <WIDTH x i8> but
```
dbabokin, line 21: How it's stored is governed by compatibility with C++. How it's passed is governed by platform ABI.

And yes, `bool` is a unique type in the sense that it's the only type that has different storage representation and register representation.

dbabokin: I would suggest taking out this comment and explaining why it's different.

tests/lit-tests/bool-store-varying.ispc
```diff
@@ -0,0 +1,25 @@
+// RUN: %{ispc} --target=host --nowrap --nostdlib -O2 --emit-llvm-text %s -o - | FileCheck %s
```
dbabokin, line 1: `bool` type is tricky - it's representation differs on different platforms. Check how this works on AVX512 targets, this test may be failing.

tests/lit-tests/2689.ispc
```diff
@@ -0,0 +1,12 @@
+// RUN: %{ispc} --target=host --nowrap --nostdlib -O2 --emit-llvm-text %s -o - | FileCheck %s
```
dbabokin, line 1: Host is ok as a target here, but make sure it works on AVX512 targets as well.

dbabokin (approved):
LGTM in general, but look for the comments below.

tests/lit-tests/2689.ispc
nurmukhametov: I have added explicit AVX512 runs.

tests/lit-tests/bool-store-varying.ispc
nurmukhametov: I have added explicit AVX512 runs.

tests/lit-tests/bool-store-varying.ispc
nurmukhametov: I've changed it.

nurmukhametov: One more thing I have noticed, that on `avx512-x32` target `uniform uint8<4>` vector has `<8 x i8>` type and `avx512-x64` has `<16 x i8>`. It looks intentional from code here: https://github.com/ispc/ispc/blob/5725fb8d495980af52b417ae0c7a89e2a258d2bd/src/type.cpp#L1823

but it is not clear for me why.

tests/lit-tests/bool-casts-varying.ispc
```diff
@@ -0,0 +1,94 @@
...
+// CHECK-LABEL: @to_uint64___vyUvyU(
+// CHECK-NEXT: allocas:
+// CHECK-NEXT: [[CMP:%.*]] = icmp ugt <[[WIDTH:.*]] x [[TYPE:.*]]> %a, %b
+// CHECK-NEXT: [[CAST:%.*]] = zext <[[WIDTH]] x i1> [[CMP]] to <[[WIDTH]] x [[TYPE]]>
+// CHECK-NEXT: ret <[[WIDTH]] x [[TYPE]]> [[CAST]]
+
+uint64 to_uint64(uint64 a, uint64 b) { return a > b; }
+
+// Note:
+// It is quite suspicous that bool is stored in memory as <WIDTH x i8> but
```
dbabokin, line 92: Same

tests/lit-tests/bool-store-varying.ispc
dbabokin: The logic seems good, but I have a question why the next power of two is computed differently, while it should be the same.

tests/lit-tests/bool-store-varying.ispc
dbabokin: Now the test `REQUIRES: X86_ENABLED`

tests/lit-tests/2689.ispc
dbabokin: Also `REQUIRES: X86_ENABLED`

tests/lit-tests/bool-store-varying.ispc
nurmukhametov: `nextPow2` is 4, but for all `g->target->getDataTypeWidth()` less than 32 (e.g., `avx512skx-x32` - 16, `avx512skx-x32` - 8) this returns not nextPow2 but `128 / g->target->getDataTypeWidth()`, i.e., 128/16 = 8 and 128/8 = 16 correspondingly

PR #2691, opened 2023-10-26
Fix static const declaration of bool types

### LINKED ISSUE #2333: Static initialization of varying bool crashes the compiler
```
void StoreEveryOther( uniform float dst[], const uniform uint32 baseIndex, const varying float src )
{
    //static const varying bool mask = { 1, 0, 1, 0, 1, 0, 1, 0 };  // crashes
    //static const varying bool mask = { true, false, true, false, true, false, true, false }; // crashes
    static const varying int32 mask = { 1, 0, 1, 0, 1, 0, 1, 0 };
    if ( mask != 0 )
    {
        dst[baseIndex+programIndex] = src;
    }
}
```

```
Global variable initializer type does not match global variable type!
<8 x i32>* @static.4.mask.1
/usr/local/src/ispc/src/main.cpp(404): FATAL ERROR: Unhandled signal sent to process; terminating.
```

CE Link
https://ispc.godbolt.org/z/16E3b388K

nurmukhametov:
This PR fixes #2333 and similar crashes with initalization of bool vectors and arrays. 

Code generating static const declaration should match an InitSymbol behaviour that was changed in the commit https://github.com/ispc/ispc/commit/c889cdee2b5c7ebe. The generated constants stored in memory so we should generate storage type bools and bool vectors.

Add also tests for different initialization cases (varying, vector, arrays).

### DIFF
--- a/src/stmt.cpp
+++ b/src/stmt.cpp
@@ -160,7 +160,7 @@ void DeclStmt::EmitCode(FunctionEmitContext *ctx) const {
             }
         }
 
-        llvm::Type *llvmType = sym->type->LLVMType(g->ctx);
+        llvm::Type *llvmType = sym->type->LLVMStorageType(g->ctx);
         if (llvmType == nullptr) {
             AssertPos(pos, m->errorCount > 0);
             return;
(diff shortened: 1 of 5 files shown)


dbabokin (approved):
LGTM

PR #2698, opened 2023-11-06
tests: do not run some tests on macOS

nurmukhametov:
It may have universal binaries that may cause failing due to arm64/x86_64 mess.

### DIFF
--- a/tests/lit-tests/2652_3.ispc
+++ b/tests/lit-tests/2652_3.ispc
@@ -6,6 +6,8 @@
 // RUN: %{cc} %t.o %t.cpp.o -o %t.cpp.bin
 // RUN: %t.cpp.bin | FileCheck %s
 
+// REQUIRES: !MACOS_HOST
+
 // CHECK: my_data = 10
 
 #ifdef ISPC
--- a/tests/lit-tests/header-bool-1.ispc
+++ b/tests/lit-tests/header-bool-1.ispc
@@ -9,9 +9,12 @@
 // RUN: %{cc} %t.o %t.cpp.o -o %t.cpp.bin
 // RUN: %t.cpp.bin | FileCheck %s
 
+// REQUIRES: !MACOS_HOST
+
 // CHECK-LABEL: bool x=false
 // CHECK-COUNT-2: bool x=true
 // CHECK: bool x=false
+
 
 #ifdef ISPC
 export void use(uniform bool x) {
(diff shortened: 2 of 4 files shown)


dbabokin:
What was the issue exactly? I think it should target host ISA by default. Most likely binary translation (Rosetta) creates some confusion, but still I don't understand how it can cause problems. Better option would be to disable these tests in cases Rosetta is enabled.

nurmukhametov:
The problem is following:
```bash
% file clang
clang: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64]
clang (for architecture x86_64):     Mach-O 64-bit executable x86_64
clang (for architecture arm64):      Mach-O 64-bit executable arm64


% ./ispc-v1.21.0-macOS.arm64/bin/ispc -DISPC --pic --target=host -h t.h -o t.ispc.o tests/lit-tests/2652_3.ispc
% clang -x c++ -c tests/lit-tests/2652_3.ispc -o t.c.o --include t.h
% clang -x c++ -c tests/lit-tests/2652_3.ispc -o t.cpp.o --include t.h
% clang t.ispc.o t.cpp.o -o t.bin
% ./t.bin
my_data = 10


% ./ispc-v1.21.0-macOS.universal/bin/ispc -DISPC --pic --target=host -h t.h -o t.ispc.o tests/lit-tests/2652_3.ispc
% clang -x c++ -c tests/lit-tests/2652_3.ispc -o t.cpp.o --include t.h
% clang t.ispc.o t.cpp.o -o t.bin
% ./t.bin
my_data = 10


% ./ispc-v1.21.0-macOS.x86_64/bin/ispc -DISPC --pic --target=host -h t.h -o t.ispc.o tests/lit-tests/2652_3.ispc
% clang -x c++ -c tests/lit-tests/2652_3.ispc -o t.cpp.o --include t.h
% clang t.ispc.o t.cpp.o -o t.bin
ld: warning: ignoring file t.ispc.o, building for macOS-arm64 but attempting to link with file built for unknown-x86_64
Undefined symbols for architecture arm64:
  "_use", referenced from:
      _main in t.cpp.o
ld: symbol(s) not found for architecture arm64

% file t.ispc.o
t.ispc.o: Mach-O 64-bit object x86_64
% file t.cpp.o
t.cpp.o: Mach-O 64-bit object arm64
```

dbabokin:
Seems to be very logical. I assume you run on ARM64 hardware. So what you've demonstrated:

- ispc (arm64) with `--target=host` produces `arm64` binary. Everything works. Everything works natively.
- ispc (universal) with `--target=host` produces `arm64` binary. Everything works. Everything works natively.
- ispc (x86_64) with `--target=host` produces `x86_64` binary. And it doesn't work, as it's attempted to be linked in `arm64` mode by `clang`.

So the proper solution would not to run these tests, when we build purely cross-`ispc`. Though I don't know on top of my mind how to detect that in `lit.cfg`.

PR #2710, opened 2023-11-29
ispcrt: support multiple drivers

### LINKED ISSUE #2708: ISPCRT: support multiple Intel drivers (e.g. iGPU and DG2)
see https://github.com/ospray/ospray_studio/issues/26

Even if we want to select the DG2 to run on, ISPCRT throws earlier when there is more than a single Intel driver at [ispcrt/detail/gpu/GPUDevice.cpp:1754](https://github.com/ispc/ispc/blob/main/ispcrt/detail/gpu/GPUDevice.cpp#L1754). One possible way could be to [change `g_deviceList`](https://github.com/ispc/ispc/blob/main/ispcrt/detail/gpu/GPUDevice.cpp#L1695) to store pairs of <driver, device>.

nurmukhametov:
This PR should fix #2708

### DIFF
--- a/ispcrt/detail/gpu/GPUDevice.cpp
+++ b/ispcrt/detail/gpu/GPUDevice.cpp
@@ -1,4 +1,4 @@
-// Copyright 2020-2023 Intel Corporation
+// Copyright 2020-2024 Intel Corporation
 // SPDX-License-Identifier: BSD-3-Clause
 
 #include "GPUDevice.h"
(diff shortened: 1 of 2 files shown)


aneshlya (commented):
Looks good! 
Please add a mock test checking that when `ISPCRT_GPU_DRIVER` is set, driver with corresponding number is selected.

ispcrt/detail/gpu/GPUDevice.cpp
```diff
@@ -1835,7 +1859,7 @@ GPUDevice::GPUDevice() : GPUDevice(nullptr, nullptr, 0) {}
 
 GPUDevice::GPUDevice(void *nativeContext, void *nativeDevice, uint32_t deviceIdx) {
     // Perform GPU discovery
-    m_driver = gpu::deviceDiscovery(&m_is_mock);
```
johguenther, line 1838: When a `nativeDevice` is passed then `m_driver` keeps being `nullptr`, which is problematic for callers of `platformNativeHandle()`. Can in this case the `gpu::g_deviceList` be searched for the `nativeDevice` to set `m_driver` as well?

PR #2716, opened 2023-12-08
stdlib: fix compilation errors in stdlib

nurmukhametov:
All of them caused by unbound casting diagnostic.
Under `--debug` there are quite a few errors reported: E.g., on atomic swaps, ispc reported:
```bash
Error: Unable to find any matching overload for call to function (/*unbound*/ int64 * uniform, varying int64)
```

This PR allows to use `--debug` for compilation with enabled stdlib.

### DIFF
--- a/stdlib.ispc
+++ b/stdlib.ispc
@@ -399,7 +399,7 @@ static inline void memcpy(void *uniform dst, void *uniform src, uniform int32 co
             ((int8 * uniform) dst)[j] = ((int8 * uniform) src)[j];
         }
     } else {
-        __memcpy32((int8 * uniform) dst, (int8 * uniform) src, count);
+        __memcpy32((uniform int8 * uniform) dst, (uniform int8 * uniform) src, count);
     }
 }
 
@@ -409,7 +409,7 @@ static inline void memcpy64(void *uniform dst, void *uniform src, uniform int64 
             ((int8 * uniform) dst)[j] = ((int8 * uniform) src)[j];
         }
     } else {
-        __memcpy64((int8 * uniform) dst, (int8 * uniform) src, count);
+        __memcpy64((uniform int8 * uniform) dst, (uniform int8 * uniform) src, count);
     }
 }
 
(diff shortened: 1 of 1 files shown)


aneshlya (commented):
Seems like it stopped working after v1.18.0 but I don't remember any user-visible changes about unbound variability. I recommend finding the commit after which it stopped working. It's possible that the change wasn't implemented correctly. The same problem can be easily reproduced in the user's code, even without the stdlib.

nurmukhametov:
Yes, indeed. Consider the following reproducer:

```ispc
void foo(void *uniform dst, void *uniform src, uniform int32 count) {
    __memcpy32((int8 * uniform) dst, (int8 * uniform) src, count);
}
```

Run with `--debug` and `--nostdlib`:

```bash
$ ./bin/ispc t2.ispc -o t.o -O2 --target=avx2-i32x8 --debug --nostdlib
Initial AST
`-Function <t2.ispc:2.5 - 2.67>  "foo"
  `-(body) StmtList <t2.ispc:2.5 - 2.67>
    `-ExprStmt <t2.ispc:2.5 - 2.66>
t2.ispc:2:5: Error: Unable to find any matching overload for call to function "__memcpy32" only considering exact matches.
        Passed types: (/*unbound*/ int8 * uniform, /*unbound*/ int8 * uniform, uniform int32)

    __memcpy32((int8 * uniform) dst, (int8 * uniform) src, count);
    ^^^^^^^^^^

```

on `v1.18.0` fails but good for `v1.17.0`.

Bisecting between these revisions shows that the guilty PR is https://github.com/ispc/ispc/pull/2299. 

The underlying cause is not clear for me, because somehow AST dumping functionality interferes with type checking and function overloads resolving, e.g. commenting out this line https://github.com/ispc/ispc/blob/9be6bd66d7b70e3f70002c545e7d3e0f100dfe93/src/func.cpp#L201 made error gone.

dbabokin (commented):
So, is it a problem of `--debug` implementation? Does it cause incorrect type resolution or incorrect function overloading behavior?

The PR that caused this problem introduced AST dumps, but it was streamlining the logic of type checks to enable further changes for template implementation. It's impossible to roll back that changes at this point - but would be good to understand the root cause, so we can fix it.

nurmukhametov:
> So, is it a problem of `--debug` implementation? Does it cause incorrect type resolution or incorrect function overloading behavior?
> 
> The PR that caused this problem introduced AST dumps, but it was streamlining the logic of type checks to enable further changes for template implementation. It's impossible to roll back that changes at this point - but would be good to understand the root cause, so we can fix it.

From my point of view, it is both. On the on hand, `--debug` should not influence compiler behavior but it does resulting in `error` produced. On the other, `error` reports about `stdlib.ispc` code look legitimate for me.

dbabokin (approved):

PR #2725, opened 2023-12-26
Use UTC with timestamp to help reproducible build.

mzf-guest:
The version 1.22.0 is not [reproducible](https://reproducible-builds.org/) because the binary prints the build time which depends on time zone (see [here](https://tests.reproducible-builds.org/debian/rb-pkg/unstable/amd64/diffoscope-results/ispc.html) for Debian package version 1.22.0-3 for example).

Adding the `UTC` keyword in the CMake command seems to fix the issue. Let me know if this is an acceptable solution.

Thanks!
François

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -485,7 +485,7 @@ if (WIN32)
 endif()
 
 # Build definitions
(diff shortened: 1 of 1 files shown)


nurmukhametov:
Thank you!

PR #2728, opened 2024-01-08
tests: move tests from tests_errors to lit-tests

nurmukhametov:
`tests_errors` folder contains tests checking ISPC error reports. It doesn't contain any tests needed to check in runtime. The proper place for them is `lit` tests, so move them there.


nurmukhametov:
> My key concern is that this creates too many lit tests increasing time to run lit tests. I think batching the tests by ~10 in one is a good way to reduce running time significantly. Challenging thing here is that it cannot be done automatically - not every test can be merged with any other automatically. So the question is if it worth spending time on this kind of optimization or not. What do you think?

I have merged some tests together. There are ~40 files at the moment.

> Also, tests naming doesn't hint that it's checking error messages, which makes it not systematic. I.e. this test corpus assumed to check user error messages - but it's very hard to verify that all the messages are covered. And some new messages are not covered. Some have duplicated tests, as we already do check some of error messages in lit corpus. Ideally, we need to check with on some systematic basis - i.e. classified by the source code file where it comes from, types of the error or something else. Any ideas here?

I have added 'err-' prefix for all tests moved from `tests_errors`. No competence checked. Collecting code coverage may probably help here.

dbabokin (approved):
Thanks for implementing these bulky changes. Everything looks good now!

PR #2739, opened 2024-01-25
Improve const int varying initialization

### LINKED ISSUE #2738: Improve const int varying initialization
It will be useful to allow to initialize const int varying with arithmetic expressions with other const int varying, e.g.:
```ispc
static const int x = { 0, 1, 2, 3 };
static const int y = x + 1;
int foo() {
    return y;
}
```

Now, it reports an error:  Initializer for global variable must be a constant.
```bash
$ ./bin/ispc --target=avx1-i32x4 test.ispc --emit-llvm-text -o -
test.ispc:2:22: Error: Initializer for global variable "y" must be a constant.
static const int y = x + 1;
                     ^^^^^
```

nurmukhametov:
Allow to use arithmetic expression with const int varying initialized by list initializer with constants.

This fixes #2738

### DIFF
--- a/src/module.cpp
+++ b/src/module.cpp
@@ -1,5 +1,5 @@
 /*
-  Copyright (c) 2010-2023, Intel Corporation
+  Copyright (c) 2010-2024, Intel Corporation
 
   SPDX-License-Identifier: BSD-3-Clause
 */
@@ -471,6 +471,107 @@ void Module::AddTypeDef(const std::string &name, const Type *type, SourcePos pos
     symbolTable->AddType(name.c_str(), type, pos);
 }
 
+// Construct ConstExpr as initializer for varying const value from given expression list if possible
+template <class T>
+Expr *lCreateConstExpr(ExprList *exprList, const AtomicType::BasicType basicType, const Type *type,
+                       const std::string &name, SourcePos pos) {
+    const int N = g->target->getVectorWidth();
+    bool canConstructConstExpr = true;
+    T vals;
+    if constexpr (std::is_same_v<T, std::vector<llvm::APFloat>>) {
+        switch (basicType) {
+        case AtomicType::TYPE_FLOAT16:
+            vals.resize(N, llvm::APFloat::getZero(llvm::APFloat::IEEEhalf()));
+            break;
+        case AtomicType::TYPE_FLOAT:
+            vals.resize(N, llvm::APFloat::getZero(llvm::APFloat::IEEEsingle()));
+            break;
+        case AtomicType::TYPE_DOUBLE:
+            vals.resize(N, llvm::APFloat::getZero(llvm::APFloat::IEEEdouble()));
+            break;
+        default:
+            return nullptr;
+        }
+    } else {
+        // T equals int8_t* and etc.
+        using PointToType = typename std::remove_pointer<T>::type;
+        vals = (T)alloca(N * sizeof(PointToType));
+        memset(vals, 0, N * sizeof(PointToType));
+    }
+
+    int i = 0;
+    for (Expr *expr : exprList->exprs) {
+        const ConstExpr *ce = llvm::dyn_cast<ConstExpr>(Optimize(expr));
+        // ConstExpr of length 1 implies that it contains uniform value
+        if (!ce || ce->Count() != 1) {
+            canConstructConstExpr = false;
+            break;
+        }
+
+        if constexpr (std::is_same_v<T, std::vector<llvm::APFloat>>) {
+            std::vector<llvm::APFloat> ce_vals;
+            llvm::Type *to_type = type->GetAsUniformType()->LLVMType(g->ctx);
+            ce->GetValues(ce_vals, to_type);
+            vals[i++] = ce_vals[0];
+        } else {
+            ce->GetValues(&vals[i++]);
+        }
+    }
+
+    if (i == 1) {
+        // In case when, e.g., { 1, }, we need to set all values as equaled to
+        // the first one to match the behaviour with other initializations.
+        while (i < N) {
+            vals[i++] = vals[0];
+        }
+    } else if (i != N) {
+        // In other cases when number of values in initializer don't match the
+        // vector width, we report error to match the other behaviour.
+        Error(pos, "Initializer list for %s \"%s\" must have %d elements (has %d).", name.c_str(),
+              type->GetString().c_str(), N, (int)exprList->exprs.size());
+    }
+
+    if (canConstructConstExpr) {
+        return new ConstExpr(type, vals, pos);
+    }
+    return nullptr;
+}
+
+Expr *lConvertExprListToConstExpr(Expr *initExpr, const Type *type, const std::string &name, SourcePos pos) {
+    ExprList *exprList = llvm::dyn_cast<ExprList>(initExpr);
+    if (type->IsConstType() && type->IsVaryingType() && type->IsAtomicType()) {
+        const AtomicType::BasicType basicType = CastType<AtomicType>(type)->basicType;
+        switch (basicType) {
+        case AtomicType::TYPE_BOOL:
+            return lCreateConstExpr<bool *>(exprList, basicType, type, name, pos);
+        case AtomicType::TYPE_INT8:
+            return lCreateConstExpr<int8_t *>(exprList, basicType, type, name, pos);
+        case AtomicType::TYPE_UINT8:
+            return lCreateConstExpr<uint8_t *>(exprList, basicType, type, name, pos);
+        case AtomicType::TYPE_INT16:
+            return lCreateConstExpr<int16_t *>(exprList, basicType, type, name, pos);
+        case AtomicType::TYPE_UINT16:
+            return lCreateConstExpr<uint16_t *>(exprList, basicType, type, name, pos);
+        case AtomicType::TYPE_INT32:
+            return lCreateConstExpr<int32_t *>(exprList, basicType, type, name, pos);
+        case AtomicType::TYPE_UINT32:
+            return lCreateConstExpr<uint32_t *>(exprList, basicType, type, name, pos);
+        case AtomicType::TYPE_INT64:
+            return lCreateConstExpr<int64_t *>(exprList, basicType, type, name, pos);
+        case AtomicType::TYPE_UINT64:
+            return lCreateConstExpr<uint64_t *>(exprList, basicType, type, name, pos);
+        case AtomicType::TYPE_FLOAT16:
+        case AtomicType::TYPE_FLOAT:
+        case AtomicType::TYPE_DOUBLE:
+            return lCreateConstExpr<std::vector<llvm::APFloat>>(exprList, basicType, type, name, pos);
+        default:
+            // Unsupported types.
+            break;
+        }
+    }
+    return nullptr;
+}
+
 void Module::AddGlobalVariable(const std::string &name, const Type *type, Expr *initExpr, bool isConst,
                                StorageClass storageClass, SourcePos pos) {
     // These may be nullptr due to errors in parsing; just gracefully return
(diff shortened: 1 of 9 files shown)


src/module.cpp
```diff
@@ -536,8 +536,43 @@ void Module::AddGlobalVariable(const std::string &name, const Type *type, Expr *
                 // the same type as the global.  (But not if it's an
                 // ExprList; they don't have types per se / can't type
                 // convert themselves anyway.)
```
aneshlya, line 538: Please remove this comment, it contradicts with newly added code.

src/module.cpp
```diff
@@ -536,8 +536,43 @@ void Module::AddGlobalVariable(const std::string &name, const Type *type, Expr *
...
+                    ExprList *exprList = llvm::dyn_cast<ExprList>(initExpr);
+
+                    // If we have exprList that initalizes const varying int, e.g.:
+                    // static const int x = { 0, 1, 2, 3 };
+                    // then we need to convert initExpr to constExpr that can
+                    // be utilized later in arithmetic expressions that can be
+                    // calculated in compile time (Optimize function call below),
+                    if (type->IsConstType() && type->IsVaryingType()) {
+                        const int N = g->target->getVectorWidth();
+                        int32_t *vals = (int32_t *)alloca(N * sizeof(int32_t));
```
aneshlya, line 551: Why not to use `std::vector` instead of manually managing memory?
`std::vector<int32_t> vals(N, 0);`

src/module.cpp
```diff
@@ -536,8 +536,43 @@ void Module::AddGlobalVariable(const std::string &name, const Type *type, Expr *
...
+                                break;
+                            }
+                            const Type *t = ce->GetType();
+                            if (!t || !t->IsUniformType() || !t->IsAtomicType() ||
+                                ce->getBasicType() != AtomicType::TYPE_INT32) {
+
+                                canInitVaryingConst = false;
+                                break;
+                            }
+                            // TODO! support other types.
```
aneshlya, line 568: It's very strange to have support only for one type. Why not to add others as well?

aneshlya (changes requested):

src/module.cpp
nurmukhametov: Done

src/module.cpp
nurmukhametov: Because I need only this type for coming patches.

aneshlya: I understand that. But it is a user visible change. Someone may try to use it and it will be hard to explain why we support this for `int` type and not for other atomic types.
If you still want to proceed without adding support for other types, please at least create a Github issue (+ ideally lit test which will fail but it's up to you).

nurmukhametov: OK, I have supported other basic types.

src/module.cpp
```diff
@@ -533,11 +533,129 @@ void Module::AddGlobalVariable(const std::string &name, const Type *type, Expr *
...
+                    // be utilized later in arithmetic expressions that can be
+                    // calculated in compile time (Optimize function call below),
+                    ExprList *exprList = llvm::dyn_cast<ExprList>(initExpr);
+                    if (type->IsConstType() && type->IsVaryingType() && type->IsAtomicType()) {
+                        const AtomicType::BasicType basicType = CastType<AtomicType>(type)->basicType;
+                        const int N = g->target->getVectorWidth();
+                        bool *vals_bool = (bool *)alloca(N * sizeof(bool));
+                        for (int j = 0; j < N; j++) {
+                            vals_bool[j] = false;
+                        }
```
aneshlya, line 554:
```suggestion
                        std::vector<bool> vals_bool(N, false);
```

aneshlya (approved):

src/module.cpp
```diff
@@ -533,11 +533,129 @@ void Module::AddGlobalVariable(const std::string &name, const Type *type, Expr *
...
+                    // The alternative is to create initializer expression with
+                    // correct type and value from ExprList.
+                    // If we have exprList that initalizes const varying int, e.g.:
+                    // static const int x = { 0, 1, 2, 3 };
+                    // then we need to convert initExpr to constExpr that can
+                    // be utilized later in arithmetic expressions that can be
+                    // calculated in compile time (Optimize function call below),
+                    ExprList *exprList = llvm::dyn_cast<ExprList>(initExpr);
+                    if (type->IsConstType() && type->IsVaryingType() && type->IsAtomicType()) {
+                        const AtomicType::BasicType basicType = CastType<AtomicType>(type)->basicType;
```
dbabokin, line 549: Why going directly to `basicType` instead of calling `getBasicType()`?

src/module.cpp
```diff
@@ -533,11 +533,129 @@ void Module::AddGlobalVariable(const std::string &name, const Type *type, Expr *
...
+                        const int N = g->target->getVectorWidth();
+                        bool *vals_bool = (bool *)alloca(N * sizeof(bool));
```
dbabokin, line 551: What's the problem to create vector of `bool`s? As it's not a stack value and there's no deallocation, then there's a memory leak.

src/module.cpp
```diff
@@ -533,11 +533,129 @@ void Module::AddGlobalVariable(const std::string &name, const Type *type, Expr *
...
+                                ce->GetValues(ce_vals);
+                                fpVal.push_back(ce_vals[0]);
+                                break;
+                            }
+                            default:
+                                // Unsupported types.
+                                break;
+                            }
+                        }
+                        if (canInitVaryingConst) {
```
dbabokin, line 618: What would happen if expr list is too short for the varying? Can it happen?
What would happen if the types in expr list are different types? Would the convert handled correctly?

tests/lit-tests/2738.ispc
```diff
@@ -0,0 +1,122 @@
...
+// RUN: %{ispc} -DTYPE_UINT16 --target=avx2-i32x4 --nostdlib --nowrap --emit-llvm-text -o - %s 2>&1 | FileCheck %s --check-prefix=INT16
+// RUN: %{ispc} -DTYPE_INT --target=avx2-i32x4 --nostdlib --nowrap --emit-llvm-text -o - %s 2>&1 | FileCheck %s --check-prefix=INT
+// RUN: %{ispc} -DTYPE_UINT --target=avx2-i32x4 --nostdlib --nowrap --emit-llvm-text -o - %s 2>&1 | FileCheck %s --check-prefix=INT
+// RUN: %{ispc} -DTYPE_INT64 --target=avx2-i32x4 --nostdlib --nowrap --emit-llvm-text -o - %s 2>&1 | FileCheck %s --check-prefix=INT64
+// RUN: %{ispc} -DTYPE_UINT64 --target=avx2-i32x4 --nostdlib --nowrap --emit-llvm-text -o - %s 2>&1 | FileCheck %s --check-prefix=INT64
+// RUN: %{ispc} -DTYPE_FLOAT16 --target=avx2-i32x4 --nostdlib --nowrap --emit-llvm-text -o - %s 2>&1 | FileCheck %s --check-prefix=FLOAT16
+// RUN: %{ispc} -DTYPE_FLOAT --target=avx2-i32x4 --nostdlib --nowrap --emit-llvm-text -o - %s 2>&1 | FileCheck %s --check-prefix=FLOAT
+// RUN: %{ispc} -DTYPE_DOUBLE --target=avx2-i32x4 --nostdlib --nowrap --emit-llvm-text -o - %s 2>&1 | FileCheck %s --check-prefix=DOUBLE
+
```
dbabokin, line 13: The test requires x86 target support

dbabokin (commented):
I agree with Arina that this is effectively implementation of const-expr language feature in a limited way without declaring it const-expr on the language level. So there should be either consistent coverage of functionality (like all arithmetic ops for basic types, or something, ideally it needs to be explainable in one sentence without exceptions), or it should be documented what works and what's not.

Also make sure that you document language changes in the docs and Release Notes.

src/expr.h
```diff
@@ -518,7 +518,6 @@ class ConstExpr : public Expr {
         varying. */
     int Count() const;
 
-  private:
     AtomicType::BasicType getBasicType() const;
 
     const Type *type;
```
dbabokin, line 523: I'm also not sure why the `union` below needs to be not private, I think it's not used outside of the class with these changes.

src/module.cpp
nurmukhametov: No, it won't work well with `ConstExpr::ConstExpr(const Type *t, bool *b, SourcePos p)`. Because, `std::vector<bool>` is a specialization of vector that does not necessarily store its elements as a contiguous array ([see here](https://en.cppreference.com/w/cpp/container/vector_bool)), so `data` is deleted (or returns undereferenceable `void`) ([godbolt example](https://godbolt.org/z/7qhes35Ej)).

src/module.cpp
nurmukhametov: `alloca` allocates memory on the stack ([man page](https://www.man7.org/linux/man-pages/man3/alloca.3.html)) so no memory leaks is happened. `std::vector<bool>` is discussed in the comment to Arina's change [here](https://github.com/ispc/ispc/pull/2739#discussion_r1471150928).

src/module.cpp
nurmukhametov: There is no such method in `AtomicType`.

tests/lit-tests/2738.ispc
nurmukhametov: Fixed

src/expr.h
nurmukhametov: Oops, it is an artifact of the initial version that accessed its members directly.

src/module.cpp
nurmukhametov: > What would happen if expr list is too short for the varying?

The other elements are initialized with defaults (zeros, false).

> Can it happen?

Yes, see `tests/lit-tests/2738_partial_init.ispc`

> What would happen if the types in expr list are different types?

Const values would be converted in `GetValues` methods of `ConstExpr`.

> Would the convert handled correctly?

See tests: 
- tests/lit-tests/2738_mixed_types.ispc
- tests/lit-tests/2738_cast_types.ispc

aneshlya:
>The only thing that was supported is the case where initalization of varying variable happened with initalization list.

A good place to add example of how initialization list can be used now is here: https://github.com/ispc/ispc/blob/main/docs/ispc.rst#L2650

src/module.cpp
nurmukhametov: > > What would happen if expr list is too short for the varying?
> 
> The other elements are initialized with defaults (zeros, false).

I changed that in the last revision to match the behavior of other initializations, i.e., if too short but not 1 then emit error otherwise initialize the whole vector with the same value:

src/module.cpp
```diff
@@ -471,6 +471,108 @@ void Module::AddTypeDef(const std::string &name, const Type *type, SourcePos pos
...
+Expr *lCreateConstExpr(ExprList *exprList, const AtomicType::BasicType basicType, const Type *type,
+                       const std::string &name, SourcePos pos) {
+    const int N = g->target->getVectorWidth();
+    bool canConstructConstExpr = true;
+    T vals;
+    if constexpr (std::is_same_v<T, std::vector<llvm::APFloat>>) {
+        for (size_t j = 0; j < N; j++) {
+            switch (basicType) {
+            case AtomicType::TYPE_FLOAT16:
+                vals.push_back(llvm::APFloat::getZero(llvm::APFloat::IEEEhalf()));
```
aneshlya, line 485: Are there any reasons why not to use `std::fill` here instead and remove `for` loop?
```
std::fill(vals.begin(), vals.end(), llvm::APFloat::getZero(llvm::APFloat::IEEEhalf()));
```

src/module.cpp
```diff
@@ -471,6 +471,108 @@ void Module::AddTypeDef(const std::string &name, const Type *type, SourcePos pos
...
+        if (!ce || ce->Count() != 1) {
+            canConstructConstExpr = false;
+            break;
+        }
+
+        if constexpr (std::is_same_v<T, std::vector<llvm::APFloat>>) {
+            std::vector<llvm::APFloat> ce_vals;
+            llvm::Type *to_type = type->GetAsUniformType()->LLVMType(g->ctx);
+            ce->GetValues(ce_vals, to_type);
+            vals[i++] = ce_vals[0];
```
aneshlya, line 516: Not clear why we always need `ce_vals[0];`

src/module.cpp
dbabokin: Right, I mixed it up.

src/module.cpp
dbabokin: > 

> I changed that in the last revision to match the behavior of other initializations, i.e., if too short but not 1 then emit error otherwise initialize the whole vector with the same value

This definitely needs to be documented in the language reference.

src/module.cpp
nurmukhametov: I have removed the `for` loop using `vals.resize(N, llvm::APFloat::getZero(llvm::APFloat::IEEEhalf()))`.

src/module.cpp
```diff
@@ -471,6 +471,108 @@ void Module::AddTypeDef(const std::string &name, const Type *type, SourcePos pos
...
+                break;
+            case AtomicType::TYPE_DOUBLE:
+                vals.push_back(llvm::APFloat::getZero(llvm::APFloat::IEEEdouble()));
+                break;
+            default:
+                return nullptr;
+            }
+        }
+    } else {
+        using BT = typename std::remove_pointer<T>::type;
```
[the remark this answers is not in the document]
nurmukhametov: No, it is not `basicType`. It is a type that `type` points to. I have renamed it to exclude that ambiguity.

src/module.cpp
nurmukhametov: I have documented this.

aneshlya (approved):

Issue #2707, opened 2023-11-23
Improved short vector types support


ColinChargyBentley:
Hi,
1. Std lib does not seems to have a large support to manipulate short vector types. For example, in my case, I need to get the max value of each item of a double<4> short vector and I have to define it by hand. See [https://ispc.godbolt.org/z/P8jcx6hhh](https://ispc.godbolt.org/z/P8jcx6hhh). It would be nice to have the help of the std lib to easy our life while manipulating short vector types.
2. More importantly, when (wrongly?) defining theses function by hand, a suboptimal code is generated. In my case, on AVX or better targets, I was hoping that VMAXPD/VMINPD instruction would be used but it does not seems to be the case. It would be nice if ISPC either was able to optimize the naive min or max code into the single correct instruction or if the std lib would provide theses fct (cf 1) AND the correct single instruction code to generate with those.

In the meantime, do I have a workaround to indicate which instruction to use to compute double<4> max (in plain C++, I could use _mm256_max_ps) ?

Best regards,
Colin Chargy

PR #2743, opened 2024-01-26
Redesign targets

nurmukhametov:
This PR follows #2823 in implementing the target redesign. The design document for the target redesign has been moved [here](https://github.com/ispc/ispc/blob/9ba441e50ec33d27116c320c555ab8d48b3a5a33/stdlib/TargetRedesign.md).

The ISPC build process has been updated to code-generate (to bitcode) the standard library with the slim ISPC binary. These files are then embedded into the composite ISPC binary.

This PR introduces two new user-visible headers: `core.isph` and `stdlib.isph`. It also introduces a few internal headers: `builtins.isph`, `target.isph`, and `svml.isph`.

This PR delivers notable performance improvements, particularly on Linux, with the new approach reducing time spent on parsing and optimization. It also opens the door to using templates for implementing standard library functionality (e.g., #2707) and further simplifying the built-in library.

Hierarchical linking of target-dependent built-in functions is to be implemented in subsequent PRs.


dbabokin:
I'll be looking at it over the weekend, but it wouldn't hurt to clean up the commit history - there are commits that just fix build or formatting.

But the bigger thing - would be good to have a design document for that. It's not just for making the review easier, but also to ensure that ideas behind the specific design are crisp and solid, and all trade offs and compromises are well documented.

aneshlya (commented):
In general the PR looks good! Before merging I'd like to see:
1. Design document
2. all TODOs implemented. 
3. lit test checking the order and expected content after each `Link*` function - `LinkDispatcher`, `LinkCommonBuiltins`, `LinkTargetBuiltins`
4. I believe that `lSetAsInternal` should disappear

dbabokin:
Would you mind rebasing this branch?

CMakeLists.txt
```diff
@@ -473,8 +472,26 @@ target_sources(${PROJECT_NAME}
         "src/parse.yy"
         "src/lex.ll"
         "common/version.h"
+        "src/target_libs.cpp.in"
 )
 
+set(BITCODE_FOLDER ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/${CMAKE_CFG_INTDIR}/../share/ispc)
```
dbabokin, line 478: I looked up `CMAKE_CFG_INTDIR ` and it's `Deprecated since version 3.21`

stdlib.isph
```diff
@@ -0,0 +1,1746 @@
...
+__declspec(safe, cost2) int64 clamp(int64 v, int64 low, int64 high);
+
+__declspec(safe, cost2) uniform int64 clamp(uniform int64 v, uniform int64 low, uniform int64 high);
+
+///////////////////////////////////////////////////////////////////////////
+// Global atomics and memory barriers
+
+void memory_barrier();
+
+#define DEFINE_ATOMIC_OP(TA, TB, OPA, OPB, MASKTYPE, TC)                                                               \
```
dbabokin, line 1053: Looks like it's not covered by clang-format.

dbabokin (commented):
A couple observations while I'm still looking for general picture.

nurmukhametov:
> But the bigger thing - would be good to have a design document for that. It's not just for making the review easier, but also to ensure that ideas behind the specific design are crisp and solid, and all trade offs and compromises are well documented.

See the design document added in the PR description.

aneshlya:
I was thinking more about introducing of bitcode files to ISPC distribution and I see the following drawbacks:

- Multiple versions mess up: It increases potential issues when people have several ISPC version installed on the system.
- Security Risk: Exposed bitcode files might be more easily tampered with, posing a security risk.
- Clutter: Can make the directory structure appear cluttered, and users might accidentally modify or delete necessary files.

But all of these can be more or less solved by wrapping them into static library.

But then I decided to look how people download and extract ISPC.
I reviewed the top 10 GitHub search results for "https://github.com/ispc/ispc/releases/download" and observed that users, including the OpenMoonray project, sometimes just copy `bin/ispc` directory in their installation scripts.

- https://github.com/dreamworksanimation/openmoonray/blob/release/building/Rocky9/CMakeLists.txt#L72
- https://github.com/supertuxkart/stk-code/blob/master/tools/linux_builder.sh#L169

So changing of ISPC distribution model and adding a bunch of LLVM bitcode files in a `share` folder will break their scripts.

I suggest to list all pros/cons of having LLVM bitcode files separately vs embedded into the ISPC binary and see if the pros overcomes all the issues listed above.

aneshlya:
I extracted the first commit to separate PR: https://github.com/ispc/ispc/pull/2953, it's not related to target redesign.

cmake/Stdlib.cmake
```diff
@@ -1,33 +1,229 @@
...
+        endforeach()
+    endif()
+
+    # ARM targets
+    if (ARM_ENABLED)
+        foreach (os ${os_list})
+            foreach (target ${ARM_TARGETS})
+                if (${os} STREQUAL "windows")
+                    continue()
+                endif()
```
aneshlya, line 147: Should this condition be one level up under `foreach (os ${os_list})`?

cmake/Stdlib.cmake
```diff
@@ -1,33 +1,229 @@
...
+    elseif ("${target}" MATCHES "gen9|xe")
+        set(arch "xe64")
+    endif()
+    if ("${arch}" STREQUAL "error")
+        message(FATAL_ERROR "Incorrect target or bit passed to write_stdlib_bitcode_lib ${target} ${os} ${bit}")
+    endif()
+
+    if ("${os}" STREQUAL "unix")
+        set(os "linux")
+    endif()
```
aneshlya, line 45: This part is a copy from GenerateBuiltins.cmake. I'd like to avoid logic duplication about ISPC's matrix support as much as possible. Let's create a separata Cmake module that will keep this logic and extract this part there into a separate function:
```
function(detect_architecture target bit os arch)
...
 set(${arch} ${${arch}} PARENT_SCOPE)
    set(${os} ${${os}} PARENT_SCOPE)
endfunction()

function(write_target_bitcode_lib name target os bit)
    detect_architecture(${target} ${bit} ${os} arch)
...
endfunction()
```

cmake/Stdlib.cmake
```diff
@@ -1,33 +1,229 @@
...
+
+    # WASM targets
+    if (WASM_ENABLED)
+        foreach (target ${WASM_TARGETS})
+            foreach (bit 32 64)
+                stdlib_to_cpp(${ispc_name} ${target} ${bit} web)
+            endforeach()
+        endforeach()
+    endif()
```
aneshlya, line 163: OK, here were have a code duplication again that will be super hard to maintain. It would be cool to move it to separate CMake module and pass "stdlib_to_cpp" or "target_ll_to_cpp" as a function pointer :) which is of course not possible in CMake. But we can do something like this:
```
function(process_targets cpp_function_name ispc_name_or_target targets_list bits_list os_list)
                if (${function_name} STREQUAL "stdlib_to_cpp")
                    ${function_name}(${ispc_name_or_target} ${target} ${bit} ${os})
                elseif (${function_name} STREQUAL "target_ll_to_cpp")
                    ${function_name}(${target} ${bit} ${os})
                else()
endfunction()
```
What do you think?

aneshlya: If you decide to eliminate code duplication, please do this in a separate PR and it'll hopefully reduce number of changes in CMake in this PR.

builtins.isph
```diff
@@ -0,0 +1,685 @@
...
+
+/** @file builtins.isph
+
+    @brief This file contains the declarations of the built-in functions that
+    are defined in builtins*.ll files but used in the ISPC code of standard
+    library. These functions should not be used outside of the standard library
+    code.
+*/
+
+extern "C" varying int32 __padds_vi32(varying int32, varying int32);
```
aneshlya, line 16: Please sort them alphabetically.

builtins/util-xe.m4
```diff
@@ -4160,723 +4160,6 @@ declare void @__pseudo_prefetch_write_varying_3(<WIDTH x i64>, <WIDTH x MASK>) n
 declare void
 @__pseudo_prefetch_write_varying_3_native(i8 *, i32, <WIDTH x i32>,
```
aneshlya, line 4161: Please double check this file. Looks like some pseudo_gathers/pseudo_scatters declarations were not removed from it.

cmake/Stdlib.cmake
nurmukhametov: No, there is two more entries after the innermost loop.

cmake/Stdlib.cmake
nurmukhametov: I created a file with common piece of code between builtins and stdlib cmakes.

cmake/Stdlib.cmake
nurmukhametov: > OK, here were have a code duplication again that will be super hard to maintain. It would be cool to move it to separate CMake module and pass "stdlib_to_cpp" or "target_ll_to_cpp" as a function pointer :) which is of course not possible in CMake. But we can do something like this:
> 
> ```
> function(process_targets cpp_function_name ispc_name_or_target targets_list bits_list os_list)
>                 if (${function_name} STREQUAL "stdlib_to_cpp")
>                     ${function_name}(${ispc_name_or_target} ${target} ${bit} ${os})
>                 elseif (${function_name} STREQUAL "target_ll_to_cpp")
>                     ${function_name}(${target} ${bit} ${os})
>                 else()
> endfunction()
> ```
> 
> What do you think?

Done.

builtins/util-xe.m4
nurmukhametov: Done

examples/cpu/cmake/AddISPCExampleModern.cmake
```diff
@@ -33,6 +33,7 @@ function(add_ispc_example)
         CXX_STANDARD_REQUIRED YES)
 
     set_property(TARGET ${example_NAME} PROPERTY POSITION_INDEPENDENT_CODE ON)
+    message(STATUS "ISPC_TARGETS: ${ISPC_TARGETS}")
```
aneshlya, line 36: Is it here for some debug purposes?  If not, please add "The example ${example_NAME} is built for ISPC_TARGETS: ${ISPC_TARGETS}". Otherwise it looks weird in the console.

core.isph
```diff
@@ -0,0 +1,773 @@
...
+*/
+
+// This whole file is included in the compilation of user code and stdlib.ispc
+// always and unconditionally.
+
+// It is also included to compile stdlib.ispc but under
+// ISPC_INTERNAL_STDLIB_COMPILATION. This allows us to differentiate these
+// situations if needed.
+
+// All macros that used here without defintions are defined in module.cpp
```
aneshlya, line 20:
```suggestion
// All macros that used here without definitions are defined in module.cpp
```

core.isph
```diff
@@ -0,0 +1,773 @@
...
+// The ImproveMemoryOps optimization pass finds these calls and then
+// tries to convert them to be calls to gather functions that take a uniform
+// base pointer and then a varying integer offset, when possible.
+//
+// For targets without a native gather instruction, it is best to factor the
+// integer offsets like "{1/2/4/8} * varying_offset + constant_offset",
+// where varying_offset includes non-compile time constant values, and
+// constant_offset includes compile-time constant values.  (The scalar loads
+// generated in turn can then take advantage of the free offsetting and scale by
+// 1/2/4/8 that is offered by the x86 addresisng modes.)
```
aneshlya, line 277:
```suggestion
// 1/2/4/8 that is offered by the x86 addressing modes.)
```

core.isph
```diff
@@ -0,0 +1,773 @@
...
+// GEN target specific
+// 4 bytes by default
+#ifndef PREFETCH_DATASIZE_DEFAULT
+#define PREFETCH_DATASIZE_DEFAULT 4
+#endif
+
+// Macros for attributes
+#define NOESCAPE __attribute__((noescape))
+#define ADDRSPACE(N) __attribute__((address_space(N)))
+// #define READONLY __attribute__((memory("read")))
```
aneshlya, line 205: Is READONLY and READNONE are commented intentionally?

aneshlya:
How about adding all new `.isph` files to `include` folder? Currently, they're scattered across utilities and test scripts, making them difficult to locate.

builtins.isph
```diff
@@ -0,0 +1,740 @@
...
+#define NOESCAPE __attribute__((noescape))
+#define ADDRSPACE(N) __attribute__((address_space(N)))
+// #define READONLY __attribute__((memory("read")))
+#define READONLY
+// #define READNONE __attribute__((memory("none")))
+#define READNONE
+// #define NOALIAS __attribute__((noalias))
+#define NOALIAS
+
+extern "C" varying int32 __padds_vi32(varying int32, varying int32);
```
aneshlya, line 26: I think all builtin functions should be `inline`.
Have you figure out the effect of lost attributes like `unwind`?

nurmukhametov: > I think all builtin functions should be `inline`.

It is not so at the moment. 

Moreover, I think that forcing inlining (always and unconditionally) can actually be harmful. LLVM should decide better whether the inlining of the particular function is profitable. Although this consideration is unrelated to the current change.

> Have you figure out the effect of lost attributes like `unwind`?

It is not lost. It is populated on all functions.

core.isph
nurmukhametov: As discussed, I went through more or less all function in builtins to check that I am not losing attributes when lifting declarations from *.ll files to *.isph. It also helped to understand which attributes are still to be implemented because of that they are commented.

aneshlya: OK, I was a little bit confused. 

- You've supported `noescape` and `address_space` attributes already and they are used in this PR - great.
- You've created issue https://github.com/ispc/ispc/issues/2941 to support "deprecated" but I don't see it anywhere in this PR. Is this attribute essential for the target redesign, or is it just a nice-to-have feature? 
- I don't see an issue created for supporting memory attributes. Please create it.

builtins.isph
aneshlya: > this consideration is unrelated to the current change.

Agree.

> It is not lost. It is populated on all functions.

I'm referring to the functions that were moved from `utils.m4` to `core.isph`, such as the `pseudo*`... functions. I’ve noticed that these functions are marked as nounwind in the generated LLVM IR. However, I'm unable to locate where this marking is applied—whether it's explicitly set somewhere or if LLVM is automatically applying it based on its own logic.

src/opt/ReplaceStdlibShiftPass.cpp
```diff
@@ -29,12 +29,16 @@ bool ReplaceStdlibShiftPass::replaceStdlibShiftBuiltin(llvm::BasicBlock &bb) {
...
-    shifts[0] = M->getFunction("shift___vytuni");
-    shifts[1] = M->getFunction("shift___vysuni");
-    shifts[2] = M->getFunction("shift___vyiuni");
-    shifts[3] = M->getFunction("shift___vyIuni");
-    shifts[4] = M->getFunction("shift___vyfuni");
-    shifts[5] = M->getFunction("shift___vyduni");
+    std::string targetSuffix = std::string("_") + g->target->GetISAString();
+    if (g->singleTargetCompilation) {
+        targetSuffix = "";
+    }
```
aneshlya, line 35: This is very tiny code but I think it makes sense to extract it as separate function in builtins.h/builtins.cpp. We may reuse it later in multiple places where we determine the target-related suffix (headers generation, function mangling ....).

tests/lit-tests/math-lib-value.ispc
```diff
@@ -0,0 +1,40 @@
+// RUN: %{ispc} %s --target=host --nowrap 2>&1 --debug-phase=2:2 -o t.o | FileCheck %s --check-prefix=CHECK-ISPC
```
aneshlya, line 1: Please add some description what this test is doing and where are defined libs values.

src/ispc.h
```diff
@@ -811,6 +823,9 @@ struct Globals {
         vector width to them. */
     bool mangleFunctionsWithTarget;
 
+    /** If enabled, it means that we generate code only for one target. */
+    bool singleTargetCompilation;
```
aneshlya, line 827: Not sure if static analysis will be OK with not-initializing this property in `Globals` constructor.

src/main.cpp
```diff
@@ -211,6 +211,7 @@ static void lPrintVersion() {
     printf("    [--[no-]discard-value-names]\tDo not discard/Discard value names when generating LLVM IR\n");
     printf("    [--dump-file[=<path>]]\t\tDump module IR to file(s) in "
            "current directory, or to <path> if specified\n");
+    printf("    [--gen-stdlib]\t\tEnable special compilation mode to generate LLVM IR for stdlib.ispc.");
```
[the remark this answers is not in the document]
aneshlya: This is still relevant. It would be good to issue a warning and stop compilation when `gen-stdlib` is used with input ispc file. Now this results in assertion: `ispc --gen-stdlib test.ispc`

src/main.cpp
```diff
@@ -199,19 +200,20 @@ static void lPrintVersion() {
...
     printf("    [--debug-pm-time-trace]\t\tPrint time tracing information from ispc pass manager\n");
-    printf("    [--debug-phase=<value>]\t\tSet optimization phases to dump. "
-           "--debug-phase=first,210:220,300,305,310:last\n");
+    printf("    [--debug-phase=<value>]\t\tSet optimization or construction phases to dump. "
+           "--debug-phase=pre:first,210:220,300,305,310:last\n");
     printf("    [--[no-]discard-value-names]\tDo not discard/Discard value names when generating LLVM IR\n");
     printf("    [--dump-file[=<path>]]\t\tDump module IR to file(s) in "
            "current directory, or to <path> if specified\n");
-    printf("    [--off-phase=<value>]\t\tSwitch off optimization phases. --off-phase=first,210:220,300,305,310:last\n");
+    printf("    [--gen-stdlib]\t\tEnable special compilation mode to generate LLVM IR for stdlib.ispc.");
```
aneshlya, line 214:
```suggestion
    printf("    [--gen-stdlib]\t\tEnable special compilation mode to generate LLVM IR for stdlib.ispc.\n");
```

core.isph
nurmukhametov: I have created issue for memory attributes https://github.com/ispc/ispc/issues/2956

src/main.cpp
nurmukhametov: done

src/ispc.h
nurmukhametov: Actually, I realized that there is `isMultiTargetCompilation` and removed `singleTargetCompilation`.

core.isph
nurmukhametov: done

core.isph
nurmukhametov: done

examples/cpu/cmake/AddISPCExampleModern.cmake
nurmukhametov: Yes, it is for debug. I have removed it.

CMakeLists.txt
```diff
@@ -821,6 +825,24 @@ function(configure_ispc_exe TARGET)
...
+    generate_stdlibs(${NAME})
+
+    # stdlib, stdlibs-bc, stdlibs-cpp targets are created in Stdlib.cmake
+    configure_ispc_obj(stdlib)
+
+    add_executable(${NAME} src/main.cpp src/binary_slim.cpp)
+    configure_ispc_exe(${NAME})
+    target_link_libraries(${NAME} ${LINK_LIBRARIES} optimization common frontend)
+    add_dependencies(${NAME} builtins-bc)
+    # add_dependencies(${NAME} stdlibs-bc)
```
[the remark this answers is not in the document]
nurmukhametov: done

nurmukhametov:
> How about adding all new `.isph` files to `include` folder? Currently, they're scattered across utilities and test scripts, making them difficult to locate.

I created `stdlib` directory and moved there `stdlib.ispc` and all headers under `include` subdirectory.

src/type.cpp
```diff
@@ -2625,6 +2625,17 @@ llvm::DIType *ReferenceType::GetDIType(llvm::DIScope *scope) const {
                : m->diBuilder->createReferenceType(llvm::dwarf::DW_TAG_reference_type, diTargetType);
 }
 
+bool lIsInsideStdlib(const SourcePos &pos) {
+    std::string filename = pos.name;
+    std::string files[] = {"core.isph", "builtins.isph", "svml.isph", "stdlib,isph", "stdlib.ispc"};
```
aneshlya, line 2630:
```suggestion
    std::string files[] = {"core.isph", "builtins.isph", "svml.isph", "stdlib.isph", "stdlib.ispc"};
```

aneshlya, line 2630: Hmm, I'd use more unique names for header files which would definitely not be used by user.

src/type.h
```diff
@@ -997,6 +997,10 @@ class FunctionType : public Type {
     /* Get string representation of calling convention */
     const std::string GetNameForCallConv() const;
 
+    bool IsInStdlib() const;
```
aneshlya, line 1000: I'd prefer `IsStdlibFunction` name.

src/binary_slim.cpp
```diff
@@ -21,10 +23,21 @@ void initializeBinaryType(const char *ISPCExecutableAbsPath) {
     llvm::sys::path::remove_filename(includeDir);
     llvm::sys::path::remove_filename(includeDir);
     llvm::SmallString<128> shareDir(includeDir);
-    // llvm::sys::path::append(includeDir, "include");
-    // lParseInclude(includeDir.c_str());
+    llvm::sys::path::append(includeDir, "include");
+    g->includePath.push_back(std::string(includeDir.c_str()));
     llvm::sys::path::append(shareDir, "share", "ispc");
     g->shareDirPath = std::string(shareDir.str());
+g->isSlimBinary = true;
```
aneshlya, line 30: formatting

builtins.isph
nurmukhametov: I have sorted them alphabetically inside groups of functions and reordered these groups.

src/builtins.cpp
```diff
@@ -321,957 +241,347 @@ static void lUpdateIntrinsicsAttributes(llvm::Module *module) {
...
+    const BitcodeLib *target =
+        g->target_registry->getISPCTargetLib(g->target->getISPCTarget(), g->target_os, g->target->getArch());
+    Assert(target);
+    llvm::Module *targetBCModule = target->getLLVMModule();
+
+    llvm::StringMap<int> targetBuiltins;
+    for (llvm::Function &F : targetBCModule->functions()) {
+        auto name = F.getName();
+        // if (!name.startswith("llvm") && !isPersistent(name.str())) {
+        if (!name.startswith("llvm")) {
```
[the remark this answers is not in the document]
nurmukhametov: Hm, it is strange. I don't see compilation fails in my setup (Linux).

builtins/target-sse4-i16x8.ll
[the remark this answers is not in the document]
nurmukhametov: Yes. I have removed them.

CMakeLists.txt
```diff
@@ -936,7 +952,7 @@ if (ISPC_INCLUDE_XE_EXAMPLES AND ISPC_INCLUDE_RT AND NOT ISPC_PREPARE_PACKAGE)
     else()
         ExternalProject_Add(ispc_xpu_examples
             PREFIX xpu_examples
-            DEPENDS ispc ispcrt
+            DEPENDS ispc stdlibs-bc ispcrt
```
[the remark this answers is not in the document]
nurmukhametov: Correct. I have added dependencies to tests and examples.

src/opt/ReplaceStdlibShiftPass.cpp
nurmukhametov: Done

tests/lit-tests/math-lib-value.ispc
nurmukhametov: Done

src/binary_slim.cpp
nurmukhametov: Done

src/ispc.h
```diff
@@ -734,6 +743,12 @@ struct Globals {
     /** Whether to dump IR to file. */
     bool dumpFile;
 
+    /** Wherther we are in special mode of generating IR for stdlib. */
```
aneshlya, line 746:
```suggestion
    /** Whether we are in special mode of generating IR for stdlib. */
```

src/builtins.cpp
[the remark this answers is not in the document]
aneshlya: There is two types of errors: 
```
no member named 'getInt8PtrTy' in 'llvm::Type'
```
```
no member named 'startswith' in 'llvm::StringRef'
```
I have clang version 19.1.0git (https://github.com/llvm/llvm-project.git 7af27be6633a18dd3d568594175feb6353320795)

src/builtins.cpp
```diff
@@ -321,957 +235,346 @@ static void lUpdateIntrinsicsAttributes(llvm::Module *module) {
...
-    // it removes unused declarations without definitions.
-    // This trick should be legal, as both modules use the same LLVMContext.
-    for (llvm::Function &f : *bcModule) {
-        if (f.isDeclaration()) {
-            // Declarations with uses will be moved by Linker.
-            if (f.getNumUses() > 0)
-                continue;
-            module->getOrInsertFunction(f.getName(), f.getFunctionType(), f.getAttributes());
+void lRemoveUnusedPersistentFuntions(llvm::Module *M) {
+    // TODO! make it clean. The idea here is to preserve all persistent
```
aneshlya, line 337: I'm agree with TODO here :). Looks like this function can be decomposed to several:

- extractUsedFunctions: Extracts the functions that are currently used based on llvm.compiler.used.
- findUsedPersistentGroups: Determines which persistent groups are in use.
- collectPreservedFunctions: Collects the functions that should be preserved based on the used persistent groups.
- updateLLVMUsedGlobal: Updates the llvm.compiler.used global variable with the new list of preserved functions.

src/builtins.cpp
```diff
@@ -321,957 +235,346 @@ static void lUpdateIntrinsicsAttributes(llvm::Module *module) {
...
+    // second run it preserves only subset of functions that are really
+    // used (or can be used) in module.
+    // M->print(llvm::errs(), nullptr);
+    llvm::GlobalVariable *llvmUsed = M->getNamedGlobal("llvm.compiler.used");
+    if (llvmUsed) {
+        std::unordered_map<llvm::Function *, int> usedFunctions;
+        llvm::ConstantArray *initList = llvm::cast<llvm::ConstantArray>(llvmUsed->getInitializer());
+        for (unsigned i = 0; i < initList->getNumOperands(); i++) {
+            auto *C = initList->getOperand(i);
+            // C->dump();
```
aneshlya, line 351: Please remove commented debug prints.

src/builtins.cpp
```diff
@@ -321,957 +235,346 @@ static void lUpdateIntrinsicsAttributes(llvm::Module *module) {
...
+            Assert(val);
+            // val->dump();
+            if (val->getNumUses() > 1) {
+                Assert(llvm::isa<llvm::Function>(val));
+                usedFunctions[llvm::cast<llvm::Function>(val)] = 1;
+            }
+        }
+
+        // Extend the list of preserved functions with the functions from corresponding persistent groups.
+        std::unordered_map<const builtin::PersistentGroup *, int> usedPersistentGroups;
```
aneshlya, line 365: I'm not quite understand why unordered_map is needed. Maybe unordered_set would be sufficient?

src/builtins.cpp
```diff
@@ -321,957 +235,346 @@ static void lUpdateIntrinsicsAttributes(llvm::Module *module) {
...
+    for (auto const &[group, functions] : persistentGroups) {
+        bool isGroupUsed = false;
+        for (auto const &name : functions) {
+            llvm::Function *F = M.getFunction(name);
+            if (F && F->getNumUses() > 0) {
+                // printf("used %s\n", F->getName().str().c_str());
+                // for (auto &use : F->uses()) {
+                //     llvm::User *U = use.getUser();
+                //     U->dump();
+                // }
```
aneshlya, line 474: Please remove commented debug prints

src/module.cpp
```diff
@@ -338,74 +341,101 @@ extern YY_BUFFER_STATE yy_create_buffer(FILE *, int);
...
-    // the Module constructor returns...)
-    {
-        llvm::TimeTraceScope TimeScope("DefineStdlib");
-        DefineStdlib(symbolTable, g->ctx, module, g->includeStdlib);
-    }
+    initCPPBuffer();
 
-    bool runPreprocessor = g->runCPP;
+    if (!g->isSlimBinary && !g->genStdlib) {
+        refs.push_back("#line 1 \"core.isph\"");
```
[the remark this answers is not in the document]
nurmukhametov: I used `getVirtualFileRef` of `FileManager` to create a virtual file from in-memory buffer.

src/ispc.h
nurmukhametov: Done

docs/design/TargetRedesign.md
```diff
@@ -0,0 +1,107 @@
...
+
+![Image 1](./TargetRedesign_Scheme.png)
+
+## User Code Compilation
+
+The compilation process of an ISPC program is structured into the following steps:
+
+1. Include header files: `core.isph` and `stdlib.isph` (if needed).
+2. Translate the ISPC source code into LLVM IR, which includes:
+1. Preprocess the ISPC source,
```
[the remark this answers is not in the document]
nurmukhametov: Done

src/module.cpp
```diff
@@ -338,74 +341,101 @@ extern YY_BUFFER_STATE yy_create_buffer(FILE *, int);
...
+    }
+
+    bool isFileInput = !IsStdin(filename);
+    const char *infilename = isFileInput ? filename : "-";
+    clang::FrontendInputFile inputFile(infilename, clang::InputKind());
+
+    llvm::ErrorOr<std::unique_ptr<llvm::MemoryBuffer>> buf =
+        isFileInput ? llvm::MemoryBuffer::getFile(inputFile.getFile()) : llvm::MemoryBuffer::getSTDIN();
+    if (!buf) {
+        // TODO!
```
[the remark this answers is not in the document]
nurmukhametov: It is not needed anymore after https://github.com/ispc/ispc/pull/2743#discussion_r1732579578

src/builtins.cpp
```diff
@@ -321,957 +235,346 @@ static void lUpdateIntrinsicsAttributes(llvm::Module *module) {
...
-    // New linker is kind of smart and think it knows better what to do, so
-    // it removes unused declarations without definitions.
-    // This trick should be legal, as both modules use the same LLVMContext.
-    for (llvm::Function &f : *bcModule) {
-        if (f.isDeclaration()) {
-            // Declarations with uses will be moved by Linker.
-            if (f.getNumUses() > 0)
-                continue;
-            module->getOrInsertFunction(f.getName(), f.getFunctionType(), f.getAttributes());
+void lRemoveUnusedPersistentFuntions(llvm::Module *M) {
```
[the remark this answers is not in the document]
nurmukhametov: Done

src/builtins.cpp
nurmukhametov: Done

src/builtins.cpp
nurmukhametov: No specific reason for that, unordered set will suit.

src/builtins.cpp
nurmukhametov: Done.

docs/design/TargetRedesign.md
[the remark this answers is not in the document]
aneshlya: What I see in the rendered file is:
image

src/module.cpp
```diff
@@ -338,74 +341,71 @@ extern YY_BUFFER_STATE yy_create_buffer(FILE *, int);
...
+
+int Module::CompileFile() {
+    llvm::TimeTraceScope CompileFileTimeScope(
+        "CompileFile", llvm::StringRef(filename + ("_" + std::string(g->target->GetISAString()))));
+    ParserInit();
 
-        const int numErrors = execPreprocessor(!IsStdin(filename) ? filename : "-", bufferCPP->os.get());
-        errorCount += (g->ignoreCPPErrors) ? 0 : numErrors;
+    if (g->forceAlignment != -1) {
+        llvm::GlobalVariable *alignment = module->getGlobalVariable("memory_alignment", true);
```
aneshlya, line 388: Would be good to check `alignment != nullptr`

aneshlya (approved):
There are couple of minor issues remain but in general the change looks good to me!

cmake/CommonStdlibBuiltins.cmake
```diff
@@ -0,0 +1,143 @@
...
+    endif()
+
+    set(${CPP_LIST} ${${CPP_LIST}} PARENT_SCOPE)
+    set(${BC_LIST} ${${BC_LIST}} PARENT_SCOPE)
+endfunction()
+
+# Common function to call stdlib_to_cpp or target_ll_to_cpp. The actual name
+# is passed in func argument.
+function (generate_stdlib_or_target_builtins func ispc_name CPP_LIST BC_LIST)
+  message(STATUS "func ${func}")
```
dbabokin, line 71: Looks like a debug output, which was forgotten to be removed.

src/opt/RemovePersistentFuncs.cpp
```diff
@@ -0,0 +1,25 @@
+/*
+  Copyright (c) 2022-2024, Intel Corporation
```
dbabokin, line 2: NIT: Looks like it needs to be just `2024`. It's a new file, right?

docs/design/TargetRedesign_Scheme.png
dbabokin, line 1: Could you remind me why do we need to have a separate copy of pre-compiled `stdlib.ispc` for every target now?

dbabokin, line 1: Where's the logic for function lookup sequence during linking of builtins?

dbabokin (changes requested):

stdlib/include/core.isph
```diff
@@ -0,0 +1,764 @@
...
+// always and unconditionally.
+
+// It is also included to compile stdlib.ispc but under
+// ISPC_INTERNAL_STDLIB_COMPILATION. This allows us to differentiate these
+// situations if needed.
+
+// All macros that used here without definitions are defined in module.cpp
+// in lSetPreprocessorOptions function.
+
+#ifndef ISPC_CORE_ISPH
```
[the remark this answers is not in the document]
nurmukhametov: Fixed.

stdlib/include/stdlib.isph
```diff
@@ -0,0 +1,1131 @@
...
+*/
+
+/** @file stdlib.isph
+
+    @brief Portion of the ISPC standard library declarations
+           that are implementated in ISPC.
+*/
+
+// This whole file in included for every user code compilation unless
+// --nostdlib provided.
```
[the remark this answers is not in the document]
nurmukhametov: Fixed.

stdlib/include/stdlib.isph
```diff
@@ -0,0 +1,1131 @@
+// -*- mode: c++ -*-
+/*
```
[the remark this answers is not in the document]
nurmukhametov: Done.

tests/lit-tests/memory-alignment.ispc
```diff
@@ -0,0 +1,20 @@
+// This test checks that the math library is correctly set based on the user
+// provided command line option --force-alignment
```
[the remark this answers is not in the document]
nurmukhametov: It is copy-paste error. Fixed.

tests/lit-tests/memory-alignment.ispc
```diff
@@ -0,0 +1,20 @@
...
+// __memory_alignment value is defined in the core.isph file without the initialization
+// when ISPC compiles stdlib bitcode giles during ISPC build. Later, this value
```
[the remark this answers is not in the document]
nurmukhametov: Done.

src/main.cpp
```diff
@@ -115,7 +116,7 @@ static void lPrintVersion() {
     printf("    [-MT <filename>]\t\t\tWhen used with `-M', changes the target of the rule emitted by dependency "
            "generation\n");
     printf("    [--no-omit-frame-pointer]\t\tDisable frame pointer omission. It may be useful for profiling\n");
-    printf("    [--nostdlib]\t\t\tDon't make the ispc standard library available\n");
+    printf("    [--nostdlib]\t\t\t Don't make the ispc standard library available\n");
```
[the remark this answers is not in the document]
nurmukhametov: Done.

src/main.cpp
```diff
@@ -200,19 +201,20 @@ static void lPrintVersion() {
 [[noreturn]] static void devUsage(int ret) {
     lPrintVersion();
     printf("\nusage (developer options): ispc\n");
-    printf("    [--ast-dump=user|all]\t\tDump AST for user code or all the code including stdlib. If no argument is "
-           "given, dump AST for user code only\n");
+    printf("    [--ast-dump]\t\tDump AST for user code or all the code including stdlib.\n");
```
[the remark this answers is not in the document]
nurmukhametov: I have shortened this to `Dump AST`.

src/main.cpp
```diff
@@ -697,7 +701,12 @@ int main(int Argc, char *Argv[]) {
             errorHandler.AddError(
                 "Option \"link\" can't be used in compilation mode. Use \"ispc link --help\" for details");
         } else if (!strcmp(argv[i], "--support-matrix")) {
-            g->target_registry->printSupportMatrix();
+            std::vector<std::string> missedFiles;
+            g->target_registry->printSupportMatrix(missedFiles);
```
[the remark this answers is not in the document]
nurmukhametov: It was more like ad-hoc debugging functionality for checking correctness. I outlined missedFiles checking logic to a separate function and introduced `--check-bitcode-libs` option for calling it.

cmake/CommonStdlibBuiltins.cmake
nurmukhametov: Fixed.

src/opt/RemovePersistentFuncs.cpp
nurmukhametov: Right, it is just a copy-paste leftover.

docs/design/TargetRedesign_Scheme.png
nurmukhametov: Because they are target specific, mostly it is a width and mask type specific code. Probably, we can reduce their amount to `widths*masks` where `widths=len([4,8,16,32,64]), masks=len([1,8,16,32,64])`.

nurmukhametov: It is to be done later. I want it to be separate from a big change in the current PR.

stdlib/include/stdlib.isph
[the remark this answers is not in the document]
dbabokin: It's just partially fixed. Need to fix the second line as well: `// --nostdlib is provided.`

tests/lit-tests/memory-alignment-win.ispc
```diff
@@ -0,0 +1,20 @@
+// This test checks that the math library is correctly set based on the user
+// provided command line option --force-alignment
+// __memory_alignment value is defined in the core.isph file without the initialization
+// when ISPC compiles stdlib bitcode giles during ISPC build. Later, this value
```
dbabokin, line 4: One last instance of "giles" left here.
```suggestion
// when ISPC compiles stdlib bitcode files during ISPC build. Later, this value
```

src/main.cpp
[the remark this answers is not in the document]
dbabokin: Sounds good.

src/main.cpp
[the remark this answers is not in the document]
dbabokin: This solution does make more sense. But why still doing that only on request. So, it basically verifies integrity of the installation, right?

I'm wondering if we should do that unconditionally on every run? It's an open question, I don't know if should do that or now.

Also, are you introducing any ENV variable or configuration mechanisms to locate these libs? It's also an open question. There was a statement in the design doc that this might be a way to distro owners to build ISPC, but the location might be different in every distro. Again, it's not a call to action, but would be good to understand this more clearly.

docs/design/TargetRedesign_Scheme.png
dbabokin: It used to be one copy to width. Which does make sense. Mask length is equal to target width. But **types** of the mask are different for different targets. Though it was not a problem for a previous implementation, so it's basically a question of something has changed during the redesign.

Let's clear that out, as I think it's quite fundamental thing. `stdlib.ispc` supposed to be more-or-less target agnostic - we do all the target specific magic though builtins. So would be good to understand if we are missing something here.

If we can "fix" that, then we'll avoid the only regression that we have - in the code size.

It's not a roadblock for merging this PR, but if we clear that out that it should be just one copy per length, then we should create an issue for that and follow up with fix for this.

dbabokin: 👍🏻 
Are you tracking it somewhere? It not, could you please create an issue with remaining changes? That would be really nice to have spelled out somewhere.

dbabokin (commented):
I think we are almost converged on the review of this PR! We just need to resolve all open discussions and create a follow up list for some of them.

I also would like to sped some time reviewing `module.cpp` later today and that's probably it.

src/builtins.cpp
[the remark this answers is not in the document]
aneshlya: Could you please double check the build with LLVM 18 and LLVM 19? I still see compilation errors.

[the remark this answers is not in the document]
nurmukhametov: With LLVM 18, I managed to reproduce them. They should be fixed now.

stdlib/include/stdlib.isph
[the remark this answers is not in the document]
nurmukhametov: Indeed, fixed.

tests/lit-tests/memory-alignment-win.ispc
nurmukhametov: Done

docs/design/TargetRedesign_Scheme.png
nurmukhametov: > It used to be one copy to width. Which does make sense. Mask length is equal to target width. But **types** of the mask are different for different targets. Though it was not a problem for a previous implementation, so it's basically a question of something has changed during the redesign.

It is used to be one copy per mask bits. The target width was provided during every compilation.

dbabokin (approved):
Looks good.

Issue #2713, opened 2023-11-30
Compiling for `neon-i32x8` results in ICE


Jasper-Bekkers:
```ispc
#include "image.ispc"

#define M_PI 3.14159265358979

static inline float clean(float t)
{
    const float EPSILON = .0000125f;
    if (abs(t) < EPSILON)
        return 0.0f;
    return (float)t;
}

static inline float sinc(float x)
{
    x = (x * M_PI);

    // if ((x < 0.01f) && (x > -0.01f))
    //     return 1.0f + x * x * (-1.0f / 6.0f + x * x * 1.0f / 120.0f);

    return sin(x) / x;
}

static inline float lanczos3_filter(float t)
{
    t = abs(t);

    if (t < 3.0)
        return clean(sinc(t) * sinc(t / 3.0));
    else
        return 0.0f;
}

static inline float frac(float f) {
    float absf = abs(f);
    return absf - floor(absf);
}

static inline float byte_to_float(uint b) {
    //return floatbits(0x3f800000 | (b << (23 - 8))) - 1.0;
    return (float)b;
}

static inline uint8<4> resample_internal(uniform Image src_image, float<2> uv, uniform uint8 num_channels) {
    float<4> col = 0.0;
    float weight = 0.0;
    // Truncate floating point coordinate to integer:
    const int<2> src_coord = uv * src_image.size;

    // Sample the lanczos3 filter from -2.5 to 2.5.
    // Pixel 0, 0 in the source will be at the bottom-right of the center of the target pixel,
    // hence reading source pixels in the range [-3, 2] will correctly read pixels to the left, top,
    // right and bottom of the target pixel.
    for (uniform int x = -3; x < 3; x++) {
        for (uniform int y = -3; y < 3; y++) {
            float wx = lanczos3_filter((uniform float)x + 0.5);
            float wy = lanczos3_filter((uniform float)y + 0.5);

            float w = wx * wy;
            int<2> texel_offset = {x, y};
            int<2> src_kernel_coord = src_coord + texel_offset;
            // TODO: Let the user specify a boundary mode!
            // https://github.com/Traverse-Research/ispc-downsampler/issues/25#issuecomment-1584915050
            src_kernel_coord.x = clamp(src_kernel_coord.x, 0, src_image.size.x - 1);
            src_kernel_coord.y = clamp(src_kernel_coord.y, 0, src_image.size.y - 1);

[...]

PR #2749, opened 2024-02-06
Disable loop alignment for Windows aarch64 targets

nurmukhametov:
Backport LLVM patch [#6ae36c01272](https://github.com/llvm/llvm-project/commit/6ae36c01272) to fix the issue #2713. It requires the rebuild of LLVM. The regression test will be committed later in different PR after LLVM 16 update in CI.


aneshlya (approved):

aneshlya:
Please connect the issue you're fixing to this PR.

PR #2751, opened 2024-02-07
tests: add reproducer for #2713

nurmukhametov:
This PR contains a regression test for issue #2713. We need to wait until LLVM rebuild happens (after merge PR[#2749](https://github.com/ispc/ispc/pull/2749)) to check that everything works as expected.


nurmukhametov:
I have updated LLVM 16 in CI. It contains now the patch from https://github.com/ispc/ispc/pull/2749. This PR can be merged.

aneshlya (approved):

Issue #2753, opened 2024-02-12
Incorrect returned bool value


nurmukhametov:
Consider the following code:
```ispc
export uniform bool foo() {
    return true;
}
```

Returned bool value depends on ISPC optimization level:
```asm
# with -O2
foo:                                    # @foo
        mov     al, 1
        ret

# with -O0
foo:                                    # @foo
        vpcmpeqd        xmm0, xmm0, xmm0
        vmovdqa xmmword ptr [rsp - 24], xmm0
        vxorps  xmm0, xmm0, xmm0
        vmovdqa xmmword ptr [rsp - 40], xmm0
        mov     qword ptr [rsp - 48], 0
        mov     byte ptr [rsp - 49], -1
        mov     al, byte ptr [rsp - 49]
        ret
```

nurmukhametov:
This was introduced in 1.13 release. ISPC 1.12 returns 1 in both cases.

Issue #2752, opened 2024-02-08
Incorrect bool values with MSVC


nurmukhametov:
Returning `bool` values from ISPC code to C++ code is broken when compiled with MSVC. It results in incomparable ISPC and C++ true values.

Consider the following example:

```ispc
export void set_ref(uniform bool &result) {
    result = true;
}

export void set_ptr(uniform bool * uniform result) {
	*result = true;
}

export uniform bool set_ret() {
	return true;
}
```

```c++
#include <iostream>
#include "bool_ispc.h"

using namespace ispc;
int main(int argc, char **argv) {
    bool b, v, r;

    set_ref(b);
    std::cout << "b=" << b << ", b==true is " <<  (b == true) << "\n";

	set_ptr(&v);
    std::cout << "v=" << v << ", v==true is " <<  (v == true) << "\n";

	r = set_ret();
    std::cout << "r=" << r << ", r==true is " <<  (r == true) << "\n";
    return 0;
}
```

Output is:
```bash
>.\bool.exe
b=255, b==true is 0
v=255, v==true is 0
r=255, r==true is 0
```

The temporal workaround is to use `!!b` value.

nurmukhametov:
Return value behavior depends on optimizaiton level, so issue #2753 was created for this.

nurmukhametov:
ISPC returns `true` as `-1`:

```asm
set_ref:                                # @set_ref
        mov     byte ptr [rdi], -1
        ret
set_ptr:                                # @set_ptr
        mov     byte ptr [rdi], -1
        ret
```

Godbolt [link](https://ispc.godbolt.org/z/1WM5raj3e).

It contradicts even to SystemV ABI. It explicitly says about bool that when the value is returned or passed in a register or on the stack, bit 0 contains the truth value and bits 1 to 7 shall be zero.

PR #2755, opened 2024-02-12
Zero extend bool values for storage representation

nurmukhametov:
This PR fixes ABI incompatibility of `bool` type covered by issues #2752, #2753 that were introduced in 1.13 release by commit #https://github.com/ispc/ispc/commit/c889cdee2b5c7ebecb2f159b3173ea22f4318474. That commit split `bool` type to two different representations (internal corresponding to mask and in-memory/storage corresponding to C/C++). Storage representation of `bool` has to comply with ABI not only about their size but about their values either. We need to zero extend values when converting boolean values for storage representation.

Copied from comments:
Internal representation of bool matches mask behavior that requires `true` value to be `-1`. It is an implication of some instruction requirements, e.g., `blendvps` (the most significant bit should be 1). Whereas, storage representation matches SystemV ABI (or windows one). It requires `bool` to be presented as one `byte` with the least significant bit equal `0` or `1` and others `7` bits equal to `0`, i.e., `true` is `1` when we read or write values to/from `C/C++`. Both `fromType` and `toType` can be vector or scalar and storage or internal `bool` representation. We need to support here conversion from internal vector type to storage vector type, it may be conversions: 1) `<4 x i32>` to `<4 x i8>` or 2) `<4 x i8>` to `<4 x i32>`, e.g., 1) `<-1, 0,...>` to `<1, 0,...>` 2) `<1, 0,...>` to `<-1, 0,...>`. To support all cases with less code do bool casting in two stages:
1) truncate to LLVM IR native bool types `i1` or `<N x i1>`
2) zero or sign extend from native bool types to storage or internal correspondingly.

### DIFF
--- a/src/ctx.cpp
+++ b/src/ctx.cpp
@@ -1,5 +1,5 @@
 /*
-  Copyright (c) 2010-2023, Intel Corporation
+  Copyright (c) 2010-2024, Intel Corporation
 
   SPDX-License-Identifier: BSD-3-Clause
 */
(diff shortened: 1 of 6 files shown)


aneshlya (approved):
Looks reasonable to me!

PR #2756, opened 2024-02-13
Allow select expr to be a lvalue

### LINKED ISSUE #2754: Compiler crash dealing with ternary operator and double pointers
Following code can not compile. this affects all ispc versions.

```
varying float test(uniform bool bCond, uniform float* uniform * uniform ptrA, uniform float* uniform* uniform ptrB) 
{
    return (bCond ? *ptrA : *ptrB)[programIndex];
}
```

This happened in part of my code and I was able to reproduce it with this function. the main trigger for the bug is use of ternary operator and dereferencing double pointers within. it works if I use if-else statement.

```
/usr/local/src/ispc/src/expr.cpp:4492: Assertion failed: "ptrType != nullptr".
***
***
/usr/local/src/ispc/src/main.cpp(404): FATAL ERROR: Unhandled signal sent to process; terminating.
Compiler returned: 139
```

nurmukhametov:
This PR allows to use select expr (ternary operator) as lvalue, e.g., dereference it. This fixes #2754

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -3648,6 +3648,15 @@ const Type *SelectExpr::GetType() const {
                                  becomesVarying, vectorSize);
 }
 
+const Type *SelectExpr::GetLValueType() const {
+    const Type *t = GetType();
+    if (CastType<PointerType>(t) != nullptr) {
+        return t;
+    } else {
+        return nullptr;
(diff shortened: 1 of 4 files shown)


aneshlya (approved):
LGTM!

PR #2757, opened 2024-02-14
Updates for Xe targets (25)

nurmukhametov:
This PR updates dependencies:
- Level zero loader dependency to [v1.15.1](https://github.com/oneapi-src/level-zero/releases/tag/v1.15.1),
- SPIRV-LLVM Translator,
- vc-intrinsics.

This change lets to use math spir-v builtins instead of llvm.genx intrinsics.

It also applied InstCombine patch to LLVM 17.0 on xe targets as well as some minor fixes to superbuild.


aneshlya (approved):
Looks good!

PR #2758, opened 2024-02-14
Support cross-compiling to Windows targets on a Linux host

ishland:
I'm currently working on a mod that optimizes mainly world generation of Minecraft privately and impressed by the high performance of ISPC generated code. 

But all developments take place on Linux and I do not want to set up a Windows VM. So this PR adds cross-compiling support to Windows targets on a Linux host to ease development and testing. 

~~Note that this also introduces [a test failure](https://github.com/ishland/ispc/actions/runs/7898564986/job/21556215643#step:6:282) that I can't seem to understand.~~ Fixed

Cross-compiled binary from internal sources works fine on Windows 10 according to an internal tester.


dbabokin:
> I'm currently working on a mod that optimizes mainly world generation of Minecraft privately and impressed by the high performance of ISPC generated code.

Thanks for kind words and I'm happy to see that ISPC is delivering performance in your use-case!

> But all developments take place on Linux and I do not want to set up a Windows VM. So this PR adds cross-compiling support to Windows targets on a Linux host to ease development and testing.

I totally understand that Windows VM on Linux is inconvenience and having local cross-platform builds is a definitely a good thing. But we generally trying to avoid the complexity of cross-build combinations that are not widely used, because we need to maintain these combinations (and hence test them and maintain CI for that combination, etc). And Linux->Windows and Linux->macOS are not the workflows that people typically use. Also when it comes to Windows platform we target just native Windows environment (vs GNU environment, which mingw is) for the same reason - it has very limited use, as for Windows users who need GNU environment, WSL is a way to go nowadays.

So this PR would add two things with very rare use cases:
- Linux->Windows compilation;
- "non-native" (GNU) environment on Windows.

So, before considering these changes for merging, I would like to understand your use case better and check if there are other ways to solve your problem:
- are you building your Windows **releases** locally on Linux? Would cloud CI solve this problem by building native Windows binary? (Github Actions, basically, or AppveyorCI).
- does local ISPC build work for you, or you want this to be part of the ISPC released binaries?

ishland:
> are you building your Windows releases locally on Linux?

No. But I build windows binaries for testing. 

> Would cloud CI solve this problem by building native Windows binary?

Technically yes, but that would add quite some complexity to the CI since all existing CI are only linux-based. 

> does local ISPC build work for you, or you want this to be part of the ISPC released binaries?

The local build does work well for me. 
I can see why such combination is not widely used, but I would appreciate it if this ever gets merged into upstream. If not, I'm still happy to maintain my own fork for the project.

dbabokin:
So, you use `mingw` for local Linux builds, but you've said it's for testing only. Does this mean that you build Windows releases differently? Do you use `mingw` for the releases? 

If ISPC were able to cross compile from Linux to native Window (non-`mingw` toolchain), would that solve your problem?

It's `mingw` part that adds most of maintenance complexity, not Linux->Windows cross-compilation path.

ishland:
> So, you use `mingw` for local Linux builds, but you've said it's for testing only. Does this mean that you build Windows releases differently? Do you use `mingw` for the releases?

This is still a WIP project so currently there isn't any releases yet. I would prefer building Windows releases on linux. 

> If ISPC were able to cross compile from Linux to native Window (non-`mingw` toolchain), would that solve your problem?

Yes.

nurmukhametov:
What part of ISPC standard library (or runtime?) requires headers from `-isystem${ISPC_MINGW_PATH}/${debian_triple}/include`.

dbabokin:
> What part of ISPC standard library (or runtime?) requires headers from `-isystem${ISPC_MINGW_PATH}/${debian_triple}/include`.

I wish we documented this part better, as this question comes up regularly. The only dependency on system headers that we have is the headers for building [builtins-c-cpu.cpp](https://github.com/ispc/ispc/blob/main/builtins/builtins-c-cpu.cpp), i.e. `print()` function. The rest of the stdlib is mostly OS agnostic and doesn't depends on target platform headers. One exception is memory allocation/deallocation, but we don't use headers for them, we just rely on stable signature of `malloc`/`free` (and their aligned versions).

The code that deals with system headers includes for `builtins-c-cpu.cpp` is [here](https://github.com/ispc/ispc/blob/aa6610460368619567f0dcda037230dfae369af8/cmake/GenerateBuiltins.cmake#L213) - watch for `includePath` var.

One other thing that is different for `mingw` is that it assumes different debug info format, as the tool chain is gcc-based.

heyer2:
I stumbled upon the lack of Linux -> Windows support a while ago as well.
We compile a project to `x86-64 (Linux)`, `ARM64 (Linux)` and `x86-64 (Windows)` using CMake on a single machine.
Since the overhead associated with a VM or Windows machine is too much for us, we use LLVM-MinGW.
Currently we disable ISPC on Windows builds.

I can't speak for whether the maintenance burden is worth it or not for the ISPC team,
but MinGW support would allow us to use ISPC more extensively.

.github/workflows/scripts/install-build-deps.sh
```diff
@@ -38,3 +38,12 @@ fi
 [ -n "$LLVM_REPO" ] && wget --retry-connrefused --waitretry=5 --read-timeout=20 --timeout=15 -t 5 $LLVM_REPO/releases/download/llvm-$LLVM_VERSION-ispc-dev/$LLVM_TAR
 tar xf $LLVM_TAR
 echo "${GITHUB_WORKSPACE}/bin-$LLVM_VERSION/bin" >> $GITHUB_PATH
+
+# Fetch llvm-mingw: version of llvm doesn't matter as building only needs the headers in it
```
nurmukhametov, line 42: I don't think we need to include mingw headers at all. I looked into `builtins-c-cpu.cpp` and it has only one platform specific function (`num_cores`) with implementation both for windows and posix(?). If we target windows we need just use the windows specific one.

Moreover, I noted one strange discrepancy in `builtins-c-cpu.cpp`:
`num_cores` platform specific implementation and its dependency are guarded by different `ifdef`s., here https://github.com/ispc/ispc/blob/9c2e40ad9282d20ee7a725caf8271634c9887967/builtins/builtins-c-cpu.cpp#L38 and here https://github.com/ispc/ispc/blob/9c2e40ad9282d20ee7a725caf8271634c9887967/builtins/builtins-c-cpu.cpp#L160

I think both should be guarded in same way, i.e., `#include <unistd.h` under `#if !defined(_MSC_VER) && !defined(__MINGW32__)`

@ishland , was this the only place that required mingw headers?

nurmukhametov:
> > What part of ISPC standard library (or runtime?) requires headers from `-isystem${ISPC_MINGW_PATH}/${debian_triple}/include`.
> 
> I wish we documented this part better, as this question comes up regularly. The only dependency on system headers that we have is the headers for building [builtins-c-cpu.cpp](https://github.com/ispc/ispc/blob/main/builtins/builtins-c-cpu.cpp).

Yes, this is indeed the only file that requires target system headers (`stdio.h`, `stdlib.h`, `string.h`) and symbols defined there (`fputs`, `fflush`, `stdout`, `snprintf`).

I have encountered a very interesting blog post about cross compiling from Linux to Windows [link](https://jake-shadle.github.io/xwin/). They cross-compile from `x86_64-unknown-linux-gnu` to `x86_64-pc-windows-msvc` using native Windows 10 SDK. I checked that this approach works for ISPC as well.

boberfly:
Hi all,

I started to look into this for my own purposes before discovering this PR. I am approaching it slightly differently and using `clang-cl` so that it is compatible with the `msvc` toolchain, and yes using `xwin` that @nurmukhametov mentioned in the last comment.

I'll clean-up where I was with this. I am using something along the lines of `set(includePath /vctoolsdir ${ISPC_WINDOWS_VCTOOLS_PATH} /winsdkdir ${ISPC_WINDOWS_SDK_PATH})` with `xwin`'s generated `crt` and `sdk` directories and the `clang-cl` executable, I haven't tested yet if this works or not but I am glad by looking at this PR that I am along the right track...

aneshlya:
Closing this in favor of https://github.com/ispc/ispc/pull/2841

PR #2763, opened 2024-02-15
lit-tests: adjust tests for arm

nurmukhametov:
Do not try to match x86 asm when arm asm generated. I think we need to create CI jobs that will run tests on `aarch64` worker that we have in our CI to avoid this in future.

When `host` can generate both `x86` and `asm` it is not clear how to distinguish when host is arm target when it is not (especially minding that universal binary contains both), so `X86_ENABLED && !ARM_ENABLED` is used. We need to figure out that properly later.

`bool-store-vector.ispc` checks that ISPC sign extends `i1` for internal `bool` representation. I am not sure whether there is a requirement of internal `true` to be `-1` for `arm`, but it looks like ISPC matched their behavior before. We need to figure out that later.


aneshlya (commented):
Looks good!

PR #2764, opened 2024-02-15
Update docs for v1.23.0

nurmukhametov:


docs/ispc.rst
```diff
@@ -568,6 +569,29 @@ aligned, please refer to `Function Templates`_ section for more details.
...
+* Variables with const qualifiers can be initialized using the values of
+  previously initialized const variables including arithmetic operations above
+  them. It now works also with varying types.
+* Enumeration type values can be used as constants.
+
+The result of selection operator can now be used as lvalue if it has suitable
+type.
+
+Now, command-line switch `--dump-file=<dir>` forces dumping of the whole IR
+modules after each pass.
```
aneshlya, line 593: I don't think this option is needed to be described here since `--dump-file` is an option for ISPC developers (under `--help-dev`).

docs/ispc.rst
```diff
@@ -568,6 +569,29 @@ aligned, please refer to `Function Templates`_ section for more details.
...
 
+Updating ISPC Programs For Changes In ISPC 1.23.0
+-------------------------------------------------
+
+This release contains the following changes that may affect compatibility with
+older versions:
+
+* `true` `bool` values in storage were changed from `-1` to `1` to match ABI.
+  It may have resulted to incomparable `true` values with C/C++ `true` values
+  since 1.13.0 release.
```
aneshlya, line 580: I think it should be clarified a bit:
```suggestion
* `true` `bool` values in storage were changed from `-1` to `1` to match C/C++ ABI.
Previously, ISPC treated `bool` values similarly to C/C++ in terms of size, but incorrectly interpreted their actual values. This meant that `true` in ISPC might not have translated correctly to true in C/C++. This issue was introduced in version 1.13.0. Starting now, ISPC correctly stores and interprets `true` values in a way that aligns with C/C++ expectations.

```

aneshlya (approved):

PR #2767, opened 2024-02-15
Update google-benchmarks to v1.8.3 and google-tests to v1.14

aneshlya:
Use latest releases of submodules to avoid pip vulnerabilities.


nurmukhametov (approved):
LGTM!

PR #2772, opened 2024-02-20
Remove unused code

nurmukhametov:
This PR deletes an unused code from `ImproveMemoryOps.cpp` file. 

It depends on #2771


aneshlya:
I can confirm that Xe-related functions can be safely removed. They had to be removed when we switched from svm loads/stores to generic ones.

dbabokin (approved):
Not sure why this depends on #2771, as it just removes the code, but it looks good anyway.

PR #2773, opened 2024-02-20
Fix function lookup

nurmukhametov:
This PR fixes the pattern of searching for function call replacement rules in `ImproveMemoryOps.cpp` and `ReplacePseudoMemoryOps.cpp`. They used an array that created and searched for to match called function upon every call visit.

This fix introduces an unordered_map instead. It is created once and used to lookup called function more efficiently. As a result, ISPC works a few percent faster.

This PR depends on #2771 and #2772

### DIFF
--- a/src/opt/ImproveMemoryOps.cpp
+++ b/src/opt/ImproveMemoryOps.cpp
@@ -1084,6 +1084,44 @@ static llvm::Constant *lGetOffsetScaleVec(llvm::Value *offsetScale, llvm::Type *
     return llvm::ConstantVector::get(scales);
 }
 
+struct GatherImpInfo {
+    GatherImpInfo(const char *lmName, const char *bmName, llvm::Type **st, int a)
+        : align(a), load(lmName), blend(bmName), sType(st) {}
+    bool isFactored() const { return !g->target->useGather(); }
+    llvm::Function *loadMaskedFunc() const { return m->module->getFunction(load); }
+    llvm::Function *blendMaskedFunc() const { return m->module->getFunction(blend); }
+    llvm::Type *scalarType() const { return *sType; }
+    llvm::Type *baseType() const {
+        // Pseudo gather base pointer element type (the 1st argument of the intrinsic) is int8
+        // e.g. @__pseudo_gather_base_offsets32_i8(i8 *, i32, <WIDTH x i32>, <WIDTH x MASK>)
+        return LLVMTypes::Int8Type;
+    }
+    const int align;
+
+  private:
+    const char *load;
+    const char *blend;
+    llvm::Type **sType;
+};
+
+struct ScatterImpInfo {
+    ScatterImpInfo(const char *msName, llvm::Type **vpt, int a) : align(a), store(msName), vpType(vpt) {}
+    bool isFactored() const { return !g->target->useScatter(); }
+    llvm::Function *maskedStoreFunc() const { return m->module->getFunction(store); }
+    llvm::Type *vecPtrType() const { return *vpType; }
+    llvm::Type *baseType() const {
+        // Pseudo scatter base pointer element type (the 1st argument of the intrinsic) is int8
+        // e.g. @__pseudo_scatter_base_offsets32_i8(i8 * nocapture, i32, <WIDTH x i32>, <WIDTH x i8>, <WIDTH x
+        // MASK>)
+        return LLVMTypes::Int8Type;
+    }
+    const int align;
+
+  private:
+    const char *store;
+    llvm::Type **vpType;
+};
+
 /** After earlier optimization passes have run, we are sometimes able to
     determine that gathers/scatters are actually accessing memory in a more
     regular fashion and then change the operation to something simpler and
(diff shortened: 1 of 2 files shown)


dbabokin:
I'll have a look at it when #2771 and #2772 are merged and this PR is rebased.

src/opt/ImproveMemoryOps.cpp
```diff
@@ -653,226 +653,155 @@ static bool lOffsets32BitSafe(llvm::Value **offsetPtr, llvm::Instruction *insert
...
+    llvm::Function *baseOffsetsFunc() const {
+        const char *name = use() ? pgbo : pgfbo;
+        return m->module->getFunction(name);
+    }
+    llvm::Function *baseOffsets32Func() const {
+        const char *name = use() ? pgbo32 : pgfbo32;
+        return m->module->getFunction(name);
+    }
+    const char *pgbo;
+    const char *pgfbo;
```
aneshlya, line 703: I would add "64" suffix to these struct members similar to `pgbo32`/`pgfbo32`.
Same for `baseOffsetsFunc`.

src/opt/ImproveMemoryOps.cpp
```diff
@@ -1489,60 +1343,36 @@ static llvm::Instruction *lGSToLoadStore(llvm::CallInst *callInst) {
 ///////////////////////////////////////////////////////////////////////////
 // MaskedStoreOptPass
 
+static std::unordered_map<std::string, int> maskedStoreAlign = {
+    {__pseudo_masked_store_i8, 1},     {__pseudo_masked_store_i16, 2},   {__pseudo_masked_store_half, 2},
```
aneshlya, line 1347: It's not directly related to this PR but would be good to fix: the alignment values are used at least 4 times in this file. I suggest saving them into separate enum and using it instead of explicit integer values.

src/opt/ImproveMemoryOps.cpp
```diff
@@ -653,226 +653,155 @@ static bool lOffsets32BitSafe(llvm::Value **offsetPtr, llvm::Instruction *insert
...
-        GSInfo(__pseudo_prefetch_write_varying_3,
-               g->target->hasVecPrefetch() ? __pseudo_prefetch_write_varying_3_native : __prefetch_write_varying_3,
-               g->target->hasVecPrefetch() ? __pseudo_prefetch_write_varying_3_native : __prefetch_write_varying_3,
-               false, true),
+// Contains a function call substitution info for gather/scatter/prefetch calls.
+// We construct this class oblects in program constructors phase, i.e., before initialization of m->module, g->target
+// and any of LLVMTypes values. Moreover, g->target values are not same during multi-target compilation. Because of
+// that, we postpone evaluation of types, choosing the proper function before actual call replacement here and in
+// similar types across this file.
+struct GSInfo {
```
dbabokin, line 661: Any specific reason why it's a struct, but not a class? Some fields and functions are private, so probably it makes sense to make them private.

src/opt/ImproveMemoryOps.cpp
```diff
@@ -653,226 +653,155 @@ static bool lOffsets32BitSafe(llvm::Value **offsetPtr, llvm::Instruction *insert
...
+    static GSInfo Gather(const char *pgbo, const char *pgfbo) { return GSInfo(pgbo, pgfbo, Type::Gather); }
+    static GSInfo Scatter(const char *pgbo, const char *pgfbo, const char *pgbo32, const char *pgfbo32) {
+        return GSInfo(pgbo, pgfbo, pgbo32, pgfbo32, Type::Scatter);
+    }
+    static GSInfo Scatter(const char *pgbo, const char *pgfbo) { return GSInfo(pgbo, pgfbo, Type::Scatter); }
+    static GSInfo Prefetch(const char *pgbo, const char *pgfbo) { return GSInfo(pgbo, pgfbo, Type::Prefetch); }
+    bool isGather() const { return type == GSInfo::Type::Gather; }
+    bool isScatter() const { return type == GSInfo::Type::Scatter; }
+    bool isPrefetch() const { return type == GSInfo::Type::Prefetch; }
+    bool use() const {
```
dbabokin, line 683: Function name is not really descriptive. Either need a comment or a better name.

src/opt/ImproveMemoryOps.cpp
```diff
@@ -653,226 +653,155 @@ static bool lOffsets32BitSafe(llvm::Value **offsetPtr, llvm::Instruction *insert
...
+    }
+    const char *pgbo;
+    const char *pgfbo;
+    const char *pgbo32;
+    const char *pgfbo32;
+    GSInfo::Type type;
+};
+
+// A map for call replacement used in lGSToGSBaseOffsets.
+static std::unordered_map<std::string, GSInfo> replGSToGSBaseOffsets = {
```
dbabokin, line 710: I have slight preference for this map to be a local static over the global static. If you declare it local static (inside the function), it will be initialized on the first entrance to the function. This way you will not pay for it's initialization if the code doesn't reach this point (i.e. `ispc --help` invocation, for example).

src/opt/ImproveMemoryOps.cpp
```diff
@@ -918,137 +918,97 @@ static llvm::CallInst *lGSToGSBaseOffsets(llvm::CallInst *callInst) {
     return newCall;
 }
 
+static std::unordered_map<std::string, int> checksDecompositionOfArgs = {
```
dbabokin, line 921: Again, my slight preference is to have this local static.

dbabokin, line 921: `std::unordered_map<std::string, int>` - I'm scratching my head why is it `std::string`? Why not a pointer, because that's what the identifier is and that's more efficient.

But does this really mean that we need a typedef for a builtin type?

src/opt/ImproveMemoryOps.cpp
```diff
@@ -1082,6 +1082,106 @@ static llvm::Constant *lGetOffsetScaleVec(llvm::Value *offsetScale, llvm::Type *
     return llvm::ConstantVector::get(scales);
 }
 
+struct GatherImpInfo {
```
dbabokin, line 1085: Same questions for this commit:
- why not local static?
- why not class?

dbabokin (commented):
Nice changes, but need some more polishing.

src/opt/ImproveMemoryOps.cpp
nurmukhametov: Fixed

src/opt/ImproveMemoryOps.cpp
nurmukhametov: `Alignment` enum is added

src/opt/ImproveMemoryOps.cpp
nurmukhametov: No reason, it just was so. Fields are private now.

src/opt/ImproveMemoryOps.cpp
nurmukhametov: I've renamed it to `chooseBaseOrFactoredFunc`

src/opt/ImproveMemoryOps.cpp
nurmukhametov: Moved it (and others) to function scope and renamed some of them to get rid of bulky names.

src/opt/ImproveMemoryOps.cpp
nurmukhametov: The problem is in equality/hashing. With `string`, equality checks the content of the string. With pointers, it checks pointer value (i.e., the address of the string). Even if two `char*` pointers point to identical strings, they are considered different keys if they point to different addresses.

src/opt/ImproveMemoryOps.cpp
nurmukhametov: I've moved them to function scope. It was a struct.

aneshlya (approved):

src/opt/ImproveMemoryOps.cpp
dbabokin: But the c-string identifying the builtin is unique. At least with the redesign the builtin is assumed to be declared once and used only by the handle. So I don't see an issue here.

I'm fine with the string, but it's kinda unexpected pivot after the builtins redesign. I would say it's up to you if you'd like keep it this way or introduce the type and use it as a key for the map.

dbabokin (approved):
I approve the changes if you prefer to keep the string as a key for the map.

But think if this looks logical to you ;-)

src/opt/ImproveMemoryOps.cpp
nurmukhametov: > But the c-string identifying the builtin is unique. 

They are but loaded LLVM module has no connection with them, so `call->GetParent()->GetName()` does not refer/contain a c-string from `builtin-decl.cpp`.

PR #2779, opened 2024-02-28
Refactor TemplateArgs to support non-type parameters

aneshlya:
Refactor TemplateArgs to support non-type parameters:
1. Wrap `std::pair<const Type *, SourcePos>` to `TemplateArg`
2. Wrap `std::vector<std::pair<const Type *, SourcePos>>` to `TemplateArgs`

`TemplateArg` will be extended with NonType arguments in a separate PR.

### DIFF
--- a/src/func.h
+++ b/src/func.h
@@ -89,41 +89,12 @@ class TemplateArg {
     std::string GetString() const;
     // Returns `true` if this argument is a Type.
     bool IsType() const;
-    // Compares two `TemplateArg` instances for equality.
-    bool IsEqual(const TemplateArg &other) const;
     // Mangles the stored type to a string representation.
     std::string Mangle() const;
     // Transforms the stored type to its varying equivalent.
     void SetAsVaryingType();
-};
-
-// Represents a list of template arguments. This is used
-// to store all arguments provided to a template, preserving their order and types.
-class TemplateArgs {
-  public:
-    TemplateArgs() = default;
-    TemplateArgs(const std::vector<TemplateArg> &args);
-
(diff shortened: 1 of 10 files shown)


dbabokin (approved):
❤️  🚀  👍🏻

src/func.h
```diff
@@ -65,12 +65,65 @@ class TemplateParms : public Traceable {
...
+    bool IsEqual(const TemplateArg &other) const;
+    // Mangles the stored type to a string representation.
+    std::string Mangle() const;
+    // Transforms the stored type to its varying equivalent.
+    void SetAsVaryingType();
+};
+
+// Represents a list of template arguments. This is used
+// to store all arguments provided to a template, preserving their order and types.
 class TemplateArgs {
```
nurmukhametov, line 102: Do we really need a class here? Why `typedef std::vector<TemplateArg> TemplateArgs` is not enough?

aneshlya: I created a separate class to maintain consistency with the `TemplateParams` class. However, I agree that it doesn't introduce any functionality beyond that of a basic `std::vector`. Additionally, in the forthcoming PR, I plan to introduce a `TemplateParam` class that will encapsulate the logic currently implemented in the `TemplateParams::IsEqual` method, so the `TemplateParams` class will become redundant as well. I am considering transitioning both to `std::vector`.

nurmukhametov: Ok

nurmukhametov (approved):
LGTM!

src/func.h
aneshlya: Replaced `TemplateArgs` class with typedef.

Issue #2777, opened 2024-02-28
Extra vmovd, vpinsrd instruction generated


nurmukhametov:
This was initially reported in https://github.com/ispc/ispc/issues/2611, but it needs to be tracked separately.

https://ispc.godbolt.org/z/3KnEqnfsq

```ispc
struct FVector4f
{
	float V[4];
};

// Extra vmovd, vpinsrd after 2x vmaxps 
inline uniform FVector4f VectorMax(const uniform FVector4f& V1, const uniform FVector4f& V2)
{
	varying float S0, S1, Result;
	*((uniform FVector4f *uniform)&S0) = *((uniform FVector4f *uniform)&V1);
	*((uniform FVector4f *uniform)&S1) = *((uniform FVector4f *uniform)&V2);

	Result = max(S0, S1);

	return *((uniform FVector4f *uniform)&Result);
}

export void foo(uniform float A[], uniform float B[])
{

	uniform FVector4f *uniform pA = (uniform FVector4f *uniform)A;
	uniform FVector4f *uniform pB = (uniform FVector4f *uniform)B;

        *pA = VectorMax(*pA, *pB);
}
```

The ideal codegen:
```
        vmovups xmm0, xmmword ptr [rsi + 4*rcx]
        vmaxps  xmm0, xmm0, xmmword ptr [rdx + 4*rcx]
        vmaxps  xmm0, xmm0, xmmword ptr [rdi + 4*rcx]
        vmovdqu xmmword ptr [rdi + 4*rcx], xmm0
```

The actual codegen:
```
        vmovups xmm0, xmmword ptr [rdi]
        vmaxps  xmm0, xmm0, xmmword ptr [rsi]
        vmovd   eax, xmm0
        vpinsrd xmm0, xmm0, eax, 0
        vmovdqu xmmword ptr [rdi], xmm0
```

nurmukhametov:
ISPC generates bitcode that looks fine:
```
declare <8 x float> @llvm.x86.avx.max.ps.256(<8 x float>, <8 x float>) #0

define void @foo(float* noalias nocapture %A, float* noalias nocapture readonly %B) local_unnamed_addr {
allocas:
  %V1_load_bitcast_load.unpack.unpack.i = load float, float* %A, align 4
  %V1_load_bitcast_load.unpack.elt17.i = getelementptr inbounds float, float* %A, i64 1
[...]

PR #2781, opened 2024-02-29
llvm_patches: add x86-isel fix

nurmukhametov:
This backported patch fixes generation of redundant vmovd, vpinsrd instructions.

Test case added as XFAIL until LLVM rebuild in CI.

It resolves issue #2777

### DIFF
--- a/llvm_patches/16_0_17_0-X86-getFauxShuffleMask-handle-fp-int-bitcast.patch
+++ b/llvm_patches/16_0_17_0-X86-getFauxShuffleMask-handle-fp-int-bitcast.patch
@@ -1,0 +1,40 @@
+# This patch is required to fix the issue #2777
+From 7aa6ddfea210b6b114ac649c4d3219138e9cc52a Mon Sep 17 00:00:00 2001
+From: Simon Pilgrim <llvm-dev@redking.me.uk>
+Date: Thu, 29 Feb 2024 10:32:37 +0000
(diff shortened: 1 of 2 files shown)


llvm_patches/16_0_17_0_-X86-getFauxShuffleMask-handle-fp-int-bitcast.patch
```diff
@@ -0,0 +1,39 @@
+From 7aa6ddfea210b6b114ac649c4d3219138e9cc52a Mon Sep 17 00:00:00 2001
```
dbabokin, line 1: Why underscore after `17_0`? Do we need that for pattern matching?

dbabokin, line 1: Need to mention ISPC issue that describes the problem.

dbabokin (approved):
LGTM

llvm_patches/16_0_17_0_-X86-getFauxShuffleMask-handle-fp-int-bitcast.patch
nurmukhametov: No, it is just an artifact of copy-pasting.

nurmukhametov: done

Issue #2784, opened 2024-03-05
Compile failure with Apple Clang 15 and LLVM 17.0.6: no type 'std::va_list'


aumuell:
Compilation of current ispc against LLVM 17.0.6 fails on macOS 14.4 Sonoma with Apple Clang 15. The problem is also present in ispc 1.22.0 and 1.23.0, but not 1.21.1. Both LLVM and ispc have been built by spack.

ispc version: `9c2e40ad9282d20ee7a725caf8271634c9887967`

Error message:
```
/var/folders/zv/qr55pmd9065glf0mcltpx5bm000102/T/ma/spack-stage/spack-stage-ispc-main-jbt4bz2t3xpwmfp34wozgyh7qvo3xncs/spack-src/src/util.cpp:50:5: error: no type named 'va_list' in namespace 'std'; did you mean simply 'va_list'?
    std::va_list list;
    ^~~~~~~~~~~~
    va_list
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/sys/_types/_va_list.h:32:26: note: 'va_list' declared here
typedef __darwin_va_list va_list;
                         ^
```

Compiler version:
```
Apple clang version 15.0.0 (clang-1500.1.0.2.5)
Target: arm64-apple-darwin23.4.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

Compile command:
```
[...]

PR #2785, opened 2024-03-05
macos/llvm 17 compile fix: std::va_list undefined

aumuell:
Use va_list from global namespace, as <stdarg.h> does not seem to put va_list also into std namespace.
This fixes compilation on macOS 14 against LLVM 17.0.6 as reported in #2784.

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -47,7 +47,7 @@
 // See https://libcxx.llvm.org/UsingLibcxx.html#overriding-the-default-termination-handler
 // It is not quite clear why and where this symbol is used.
 void std::__libcpp_verbose_abort(char const *format, ...) {
-    std::va_list list;
+    va_list list;
     va_start(list, format);
     std::vfprintf(stderr, format, list);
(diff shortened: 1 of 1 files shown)


dbabokin:
In C++ it's defined in `<cstdarg>`, but it's not directly included. I think the right fix is to include it.

Could you verify that it would fix the problem on your end?

aumuell:
I can confirm that replacing <stdarg.h> with <cstdarg> together with std::va_list fixes the problem, too. This just seems to be a little inconsistent, as all the other headers in src/util.cpp are the C headers.

dbabokin:
It's a fair point, even though it's a C++ file.

I'm not quite sure what's going on in these lines, but I think the proposed fix is fine. I would let @nurmukhametov decide on that, as he brought these line.

aumuell:
Thank you for your feedback: the meaning of the failed check_copyright was not clear to me.

nurmukhametov (approved):
LGTM!

PR #2788, opened 2024-03-11
Fix ISPC build with LLVM 18.1 and trunk

aneshlya:
Fix ISPC build with recently released LLVM 18.1 and trunk.
There are several failing lit tests which require further investigation:
ispc :: aossoa_alignment.ispc
ispc :: assume_uniform.ispc
ispc :: interleaved.ispc


src/module.cpp
```diff
@@ -495,7 +503,7 @@ Expr *lCreateConstExpr(ExprList *exprList, const AtomicType::BasicType basicType
     } else {
         // T equals int8_t* and etc.
         using PointToType = typename std::remove_pointer<T>::type;
-        vals = (T)alloca(N * sizeof(PointToType));
+        vals = (T)ALLOCA(N * sizeof(PointToType));
```
dbabokin, line 506: @nurmukhametov 
I think `alloca()` is not needed here. Why not T to be a set of `{bool, int8_t, ..., llvm::APVal}` instead of `{bool*, int8_t*, ..., std::vector<llvm::APVal>}` and inside of the function define `std::vector<T> vals`? That would simplify the logic and will allow getting rid of `alloca()`.

This should not block this PR though.

dbabokin (approved):
`alloca()` issue is weird, I think we should get rid of it. But it should not block this PR if you prefer to go forward and merge it. It's up to you.

src/module.cpp
nurmukhametov: > Why not T to be a set of {bool, int8_t, ..., llvm::APVal} instead of {bool*, int8_t*, ..., std::vector<llvm::APVal>} and inside of the function define std::vector<T> vals?

See comment here https://github.com/ispc/ispc/pull/2739#discussion_r1471150928

nurmukhametov (changes requested):
Do we need to re-apply some patches for LLVM 18 in this PR?

src/module.cpp
dbabokin: You don't need vector here. `T vals[N];` should do the job.

src/ispc.cpp
```diff
@@ -51,6 +51,12 @@
 #include <llvm/Target/TargetMachine.h>
 #include <llvm/Target/TargetOptions.h>
 
+#if ISPC_LLVM_VERSION > ISPC_LLVM_17_0
+#define CODEGEN_OPT_LEVEL llvm::CodeGenOptLevel
```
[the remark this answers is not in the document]
dbabokin: I like this proposal!

tests/lit-tests/ftz_daz_flags_x86.ispc
```diff
@@ -4,19 +4,19 @@
...
-// CHECK: define void @export_void_test_ftz_daz___
+// CHECK-LABEL: define {{.*}} @export_void_test_ftz_daz___
 // CHECK-NOT: stmxcsr
 // CHECK-NOT: ldmxcsr
-// CHECK: define float @externC_test_ftz_daz()
+// CHECK-LABEL: define {{.*}} @externC_test_ftz_daz()
 // CHECK: stmxcsr
 // CHECK-COUNT-2: ldmxcsr
-// CHECK: define float @export_test_ftz_daz(
+// CHECK-LABEL: define {{.*}} @export_test_ftz_daz(
```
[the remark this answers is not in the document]
aneshlya: The return type didn't change. New `noundef` keyword is appeared. I removed return type by accident but anyway we don't care about it anyway, so I'll keep only @NAME here.

aneshlya:
> Do we need to re-apply some patches for LLVM 18 in this PR?

Added patches.

src/module.cpp
aneshlya: This will be fixed in separate PR as discussed.

nurmukhametov: VLA won't work due to MSVC. With static big enough array `T vals[64]` there is another problem. `llvm::APFloat` doesn't have the default constructor, so I don't understand how to declare `llvm::APFloat vals[64]` without compilation error.

nurmukhametov (approved):
LGTM!

src/module.cpp
dbabokin: use `std::array`?

nurmukhametov: > use `std::array`?

Is there any difference?

```c++
#include <array>
#include "llvm/ADT/APFloat.h"

int main() {
    llvm::APFloat X[64];
    std::array<llvm::APFloat, 64> Y;
    return 0;
}
```

Both declarations are compilation errors due to missing constructor.

dbabokin: My bad, I assumed that `std::array` may be initialized like `std::vector` from a single element. Looks like there's no elegant way to construct it on the stack.

PR #2789, opened 2024-03-12
Prerequisite changes for templates with non-type parameters support

aneshlya:
This PR includes 3 changes:
1. Adding `Mangle()` method to `Expr` which is needed to produce mangled names of template functions with non-type parameters.
2. Encapsulation of template parameters logic in `TemplateParam` class
3. Support of non-type template arguments.

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -32,6 +32,7 @@
 #include <algorithm>
 #include <list>
 #include <set>
+#include <sstream>
 #include <stdio.h>
 
 #include <llvm/ExecutionEngine/GenericValue.h>
@@ -5852,6 +5853,64 @@ llvm::Value *ConstExpr::GetValue(FunctionEmitContext *ctx) const {
     }
 }
 
+bool ConstExpr::IsEqual(const ConstExpr *ce) const {
+    if (ce == nullptr)
+        return false;
+
+    if (!Type::EqualIgnoringConst(type, ce->type))
+        return false;
+
+    for (int i = 0; i < Count(); ++i) {
+        switch (getBasicType()) {
+        case AtomicType::TYPE_BOOL:
+            if (boolVal[i] != ce->boolVal[i])
+                return false;
+            break;
+        case AtomicType::TYPE_INT8:
+            if (int8Val[i] != ce->int8Val[i])
+                return false;
+            break;
+        case AtomicType::TYPE_UINT8:
+            if (uint8Val[i] != ce->uint8Val[i])
+                return false;
+            break;
+        case AtomicType::TYPE_INT16:
+            if (int16Val[i] != ce->int16Val[i])
+                return false;
+            break;
+        case AtomicType::TYPE_UINT16:
+            if (uint16Val[i] != ce->uint16Val[i])
+                return false;
+            break;
+        case AtomicType::TYPE_INT32:
+            if (int32Val[i] != ce->int32Val[i])
+                return false;
+            break;
+        case AtomicType::TYPE_UINT32:
+            if (uint32Val[i] != ce->uint32Val[i])
+                return false;
+            break;
+        case AtomicType::TYPE_INT64:
+            if (int64Val[i] != ce->int64Val[i])
+                return false;
+            break;
+        case AtomicType::TYPE_UINT64:
+            if (uint64Val[i] != ce->uint64Val[i])
+                return false;
+            break;
+        case AtomicType::TYPE_FLOAT16:
+        case AtomicType::TYPE_FLOAT:
+        case AtomicType::TYPE_DOUBLE:
+            if (fpVal[i] != ce->fpVal[i])
+                return false;
+            break;
+        default:
+            FATAL("unimplemented const type");
+            return false;
+        }
+    }
+    return true;
+}
 /* Type conversion templates: take advantage of C++ function overloading
    rules to get the one we want to match. */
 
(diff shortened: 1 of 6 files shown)


src/expr.cpp
```diff
@@ -6177,61 +6185,68 @@ int ConstExpr::EstimateCost() const { return 0; }
 
 ConstExpr *ConstExpr::Instantiate(TemplateInstantiation &templInst) const { return new ConstExpr(this, pos); }
 
-void ConstExpr::Print(Indent &indent) const {
-    indent.Print("ConstExpr", pos);
-
-    printf("[%s] (", GetType()->GetString().c_str());
+std::string ConstExpr::GetValuesAsStr(const std::string &separator) const {
```
nurmukhametov, line 6188: What do you think about changing the method name to `GetString` and making a default value for argument? It makes sense for debugging (`Print` interface is annoying to use) and matches method name of `Type` classes hierarchy.

dbabokin: > `Print` interface is annoying to use

It's here for you:
https://github.com/ispc/ispc/blob/c72af29f0bd82678918b941654aff892fbd7dc8a/src/ast.h#L175-L176

nurmukhametov: And here how it doesn't work for interactive debugging:
```bash
$ cat max.ispc
int foo(int a, int b) { return max(a, b); }
$ gdb --args ./bin/ispc --target=host max.ispc
(gdb) b ReturnStmt::EmitCode
Breakpoint 1 at 0x3064424: file /localdisk2/anurmukh/applications.compilers.ispc.core/src/stmt.cpp, line 2957.
(gdb) r                                                                                                                                                                                                                                                     Starting program: /localdisk2/anurmukh/applications.compilers.ispc.core/build-debug/bin/ispc --target=host max.ispc
Breakpoint 1, ispc::ReturnStmt::EmitCode (this=0x55555ae9ed50, ctx=0x7fffffffd440) at /localdisk2/anurmukh/applications.compilers.ispc.core/src/stmt.cpp:2957
2957        if (!ctx->GetCurrentBasicBlock())
(gdb) p expr
$1 = (ispc::Expr *) 0x55555ae9ecf0
(gdb) call expr->Print()
Cannot evaluate function -- may be inlined
```
ISPC is configured with `cmake ../ -DCMAKE_BUILD_TYPE=Debug`.

Moreover, there is an interesting observation related to `Print` functionality here https://github.com/ispc/ispc/pull/2716#issuecomment-1850921990 that it can interfere with compiler logic. Of course, this comment is completely unrelated to this PR.

dbabokin: Interesting, I implemented this function to use in the debugger and it did work for me. Worth investigating why it's not working.

dbabokin: It works for me on macOS with `lldb`:
```
DMITRYs-MacBook-Pro-2:~/ispc/dbabokin_ispc/build (main)> lldb -- bin/ispc --target=host max.ispc -o max.o
(lldb) target create "bin/ispc"
Current executable set to '/Users/dbabokin/ispc/dbabokin_ispc/build/bin/ispc' (arm64).
(lldb) settings set -- target.run-args  "--target=host" "max.ispc" "-o" "max.o"
(lldb) b ReturnStmt::EmitCode
Breakpoint 1: where = ispc`ispc::ReturnStmt::EmitCode(ispc::FunctionEmitContext*) const + 56 at stmt.cpp:2957:10, address = 0x000000010024e354
(lldb) r
Process 71055 launched: '/Users/dbabokin/ispc/dbabokin_ispc/build/bin/ispc' (arm64)
Process 71055 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
    frame #0: 0x000000010024e354 ispc`ispc::ReturnStmt::EmitCode(this=0x0000600002af7cf0, ctx=0x000000016fdfda88) const at stmt.cpp:2957:10
   2954	ReturnStmt::ReturnStmt(Expr *e, SourcePos p) : Stmt(p, ReturnStmtID), expr(e) {}
   2955
   2956	void ReturnStmt::EmitCode(FunctionEmitContext *ctx) const {
-> 2957	    if (!ctx->GetCurrentBasicBlock())
   2958	        return;
   2959
   2960	    if (ctx->InForeachLoop()) {
Target 0: (ispc) stopped.
(lldb) p expr->Print()
`-FunctionCallExpr <stdlib.ispc:81.86 - 81.113> [varying float16]
  |-(func) FunctionSymbolExpr <stdlib.ispc:81.86 - 81.110> [ varying float16(varying unsigned int16 )] function name: __halfbits_varying_int16
  `-(args) ExprList <stdlib.ispc:81.111 - 81.112>
    `-SymbolExpr <stdlib.ispc:81.111 - 81.112> [varying unsigned int16] symbol name: a
```

dbabokin: @nurmukhametov does adding `__attribute__((used))` to `Print()` declaration help in your environment?

aneshlya: For me it doesn't help.
```
Breakpoint 1, ispc::BinaryExpr::GetValue (this=0x55555b968d20, ctx=0x7fffffffd238)
    at /export/users/aneshlya/ispc/applications.compilers.ispc.core/src/expr.cpp:2101
2101        if (!arg0 || !arg1) {
(gdb) p Print()
Too few arguments in function call.
```

dbabokin: Actually, no need for this attribute. But we don't need security hardening linker options in the debug build. Specifically this one is doing the harm:
https://github.com/ispc/ispc/blob/c72af29f0bd82678918b941654aff892fbd7dc8a/CMakeLists.txt#L615
I verified it on Linux.

src/expr.cpp
```diff
@@ -8233,6 +8248,8 @@ void SymbolExpr::Print(Indent &indent) const {
     indent.Done();
 }
 
+std::string SymbolExpr::Mangle() const { return symbol->type->Mangle() + symbol->name; }
```
dbabokin, line 8251: Do we need constant init value to be part of symbol mangling?

src/expr.h
```diff
@@ -83,6 +83,11 @@ class Expr : public ASTNode {
     virtual Expr *Instantiate(TemplateInstantiation &templInst) const = 0;
 
     virtual bool HasAmbiguousVariability(std::vector<const Expr *> &warn) const;
+
+    /** Returns a string that represents the mangled type and value (for use in
+    mangling of function symbol names). This method is needed when expression is part of a function template
+    instantiation (like ConstExpr or SymbolExpr. */
```
dbabokin, line 89: Function template parameters is the only case when expressions need to be mangled. Would be good to specifically mention that.

src/expr.cpp
```diff
@@ -6177,61 +6185,68 @@ int ConstExpr::EstimateCost() const { return 0; }
...
+            result << std::fixed << std::setprecision(6) << llvm::APFloat(fpVal[i]).convertToFloat();
             break;
-        }
         case AtomicType::TYPE_FLOAT:
-            printf("%f", fpVal[i].convertToFloat());
+            result << std::fixed << std::setprecision(6) << fpVal[i].convertToFloat();
             break;
         case AtomicType::TYPE_DOUBLE:
-            printf("%f", fpVal[i].convertToDouble());
+            result << std::fixed << std::setprecision(6) << fpVal[i].convertToDouble();
```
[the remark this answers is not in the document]
aneshlya: I checked AST dump on ast_expr.ispc test and it is identical to what was printed before.

src/expr.cpp
aneshlya: Good point! Fixed.

src/expr.cpp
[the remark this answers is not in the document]
dbabokin: Everything below `0.000001` will be reported as `0.000000`. Basically you are limiting the number of decimal digits after the period.

As long as you are not using it for mangling (floating points are not allowed as template parameters), you should print in the most informative way - i.e. ideally the full number needs to be printed, without truncation. So `std::fixed` doesn't sound like a good idea as well as limiting the precision.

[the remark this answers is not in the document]
aneshlya: I changed it to std::`to_string` which matches the current behavior using `printf `https://en.cppreference.com/w/cpp/string/basic_string/to_string. But it doesn't solve problem with truncated value and this problem exists with current implementation as well. Default precision for `printf` and `snprint` is 6.
If you think that it makes sense to fix it and print the value till the first significant digits, I can do it in separate PR.

src/expr.cpp
aneshlya: Fixed in https://github.com/ispc/ispc/pull/2797

src/expr.cpp
```diff
@@ -6177,56 +6236,61 @@ int ConstExpr::EstimateCost() const { return 0; }
...
-        case AtomicType::TYPE_FLOAT16: {
-            llvm::APFloat V(fpVal[i]);
-            printf("%f", V.convertToFloat());
+        case AtomicType::TYPE_FLOAT16:
+            result << std::to_string(llvm::APFloat(fpVal[i]).convertToFloat());
             break;
-        }
         case AtomicType::TYPE_FLOAT:
-            printf("%f", fpVal[i].convertToFloat());
+            result << std::to_string(fpVal[i].convertToFloat());
```
dbabokin, line 6277: Why `float16` and `float` cases are different? I looks like it's not intentional.

src/expr.cpp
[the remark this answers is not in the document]
dbabokin: Would be nice to have it fixed if you have time.

dbabokin (approved):
Looks good, but have a look at `float` vs `float16` cases difference.

src/expr.cpp
aneshlya: Yes, it was not intentional. Fixed.

PR #2793, opened 2024-03-13
builtins: align all implementations of __round_varying

### LINKED ISSUE #2780: Round varying values with avx512
Consider the following code:
```ispc
float RoundToInt(float x)
{
    return round(x);
}
```

when compiled as folow:
```bash
ispc --target=avx512skx-x8 round.ispc -o -
```

it generates suspicious rounding instruction:

```asm
RoundToInt___vyf:                       # @RoundToInt___vyf
        vroundps        ymm0, ymm0, 12
        ret
```

Imm8 operand value 12 (`0b1100`) means that it selects the source of rounding mode control from MXCSR.RC (not from immediate value itself where it is `0b00`, i.e., rounding), so it may under some circumstances produce wrong result.

It also doesn't match codegen with uniform value (imm8 = 8), and codegen for `avx2` and `sse4` (imm8 = 8).

nurmukhametov:
Make all implementations of __round_varying across different targets and types to actually do roundeven.

This PR fixes #2780

### DIFF
--- a/builtins/target-avx2-i16x16.ll
+++ b/builtins/target-avx2-i16x16.ll
@@ -1,4 +1,4 @@
-;;  Copyright (c) 2020-2023, Intel Corporation
+;;  Copyright (c) 2020-2024, Intel Corporation
 ;;
 ;;  SPDX-License-Identifier: BSD-3-Clause
 
@@ -199,14 +199,14 @@ define double @__ceil_uniform_double(double) nounwind readonly alwaysinline {
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; round/floor/ceil varying float/doubles
 
-declare <8 x float> @llvm.nearbyint.v8f32(<8 x float> %p)
+declare <8 x float> @llvm.roundeven.v8f32(<8 x float> %p)
 declare <8 x float> @llvm.floor.v8f32(<8 x float> %p)
 declare <8 x float> @llvm.ceil.v8f32(<8 x float> %p)
 
 define <16 x float> @__round_varying_float(<16 x float> %v) nounwind readonly alwaysinline {
   v16tov8(float, %v, %v0, %v1)
-  %r0 = call <8 x float> @llvm.nearbyint.v8f32(<8 x float> %v0)
-  %r1 = call <8 x float> @llvm.nearbyint.v8f32(<8 x float> %v1)
+  %r0 = call <8 x float> @llvm.roundeven.v8f32(<8 x float> %v0)
+  %r1 = call <8 x float> @llvm.roundeven.v8f32(<8 x float> %v1)
   v8tov16(float, %r0, %r1, %r)
   ret <16 x float> %r
 }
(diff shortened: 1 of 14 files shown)


aneshlya (approved):
Looks good!

dbabokin (changes requested):
Why `@llvm.roundeven`, but not `@llvm.round`?

nurmukhametov:
> Why `@llvm.roundeven`, but not `@llvm.round`?

I suggest that this is the actual semantic of ISPC `round` stdlib function. The reason I think so because of its implementation for other targets https://github.com/ispc/ispc/blob/71f6866107da93bb963e9e1a5d2313df493a7ce5/builtins/target-sse4-common.ll#L22 https://github.com/ispc/ispc/blob/71f6866107da93bb963e9e1a5d2313df493a7ce5/builtins/target-sse4-i32x4.ll#L100 https://github.com/ispc/ispc/blob/71f6866107da93bb963e9e1a5d2313df493a7ce5/builtins/target-avx-common-8.ll#L92

dbabokin:
My understanding of the rounding functions (and other math lib functions, actually) design in ISPC math library is to be close to C versions of the same routines. There's no point to give them any other definitions, at least I don't see such reasons.

Otherwise it's just confuses people. Especially as we don't give strict definition of these functions.

dbabokin:
Ok, looks like I start to understand what's going on.

x86 has just 4 rounding modes ([link](https://xem.github.io/minix86/manual/intel-x86-and-64-manual-vol2/o_b5573232dd8f1481-1245.html)). These 4 modes correspond to `ceil`, `floor`, `trunk`, and `roundeven`.

`round` and `roundeven` are not much different. `round` rounds `x.5` to `x+1` for positive numbers, while `roundeven` rounding depends on `x` being odd or even. So numerically is not much different and the difference will not matter for many cases.

Original implementation is indeed uses `roundeven` mode and is a single x86 instruction - but this is mostly for efficiency, as it directly maps to single hardware instruction. Though it difficult to say for sure, I was not involved in original implementation and I can only guess.

`nearbyint` appeared originally in AVX512 implementation. This implementation was done by an intern and I probably didn't catch that.

`nearbyint` uses dynamic rounding and should be avoided for `round` implementation for sure.

Some historical background on LLVM support for rounding intrinsics
LLVM 3.3 (June 2013): `ceil`/`floor/`trunk`/`rint`/`nearbyint` were added.
LLVM 3.4 (January 2014): `round` was added.
LLVM 11.0 (October 2020): `roundeven` was added.

> I couldn't manage to find any round implementation that matches C math round function.

It's [@llvm.round.*](https://llvm.org/docs/LangRef.html#llvm-round-intrinsic), it specifically defined as "This function returns the same values as the libm round functions would, and handles error conditions in the same way."

> It is confusing right now.

Yes, that's the problem and that's what we are trying to fix :)

Given all this background, I still don't know what's the best way to proceed.
1. use `@llvm.round.*` uniformly. Pro: it's fully compliant with libm. Cons: implementation is 4 instructions and it might break some code.
2. use `@llvm.roundeven.*` uniformly. Pro: don't break existing code (except when switching from `nearbyint`), more efficient on x86. Cons: will not fully comply with libm definition.

I'm slightly leaning towards 2, which is current patch is. Though documenting that behavior is needed, I think.

nurmukhametov:
> > I couldn't manage to find any round implementation that matches C math round function.
> 
> It's [@llvm.round.*](https://llvm.org/docs/LangRef.html#llvm-round-intrinsic), it specifically defined as "This function returns the same values as the libm round functions would, and handles error conditions in the same way."

I meant ISPC round implementation that matches C math round.

> Given all this background, I still don't know what's the best way to proceed.
> 
> 1. use `@llvm.round.*` uniformly. Pro: it's fully compliant with libm. Cons: implementation is 4 instructions and it might break some code.
> 2. use `@llvm.roundeven.*` uniformly. Pro: don't break existing code (except when switching from `nearbyint`), more efficient on x86. Cons: will not fully comply with libm definition.
> 
> I'm slightly leaning towards 2, which is current patch is. Though documenting that behavior is needed, I think.

I'm sure that the second option is the only reasonable one. The ISPC documentation doesn't mention any math compliancy to `libm` C functions. It actually makes a quite opposite statement:

> Standard rounding functions are provided for float16, float and double types. (On machines that support Intel®SSE or Intel® AVX, these functions all map to variants of the roundss and roundps instructions, respectively.)

dbabokin:
Another option is to use `@llvm.round.*` for `round()` implementation and introduce `roundeven()` and document the semantics and add recommendations to use them with respect to performance expectations.

The idea is that we don't only need to comply with current documentation, which is sometimes vague and incomplete. But rather to define our vision for the solution that we'd like to have and move towards that. Over the years we definitely saw the demand for more strictly defined and portable math lib and complying with libm definitions is probably the best way to provide intuitive semantics.

nurmukhametov:
> Another option is to use `@llvm.round.*` for `round()` implementation and introduce `roundeven()` and document the semantics and add recommendations to use them with respect to performance expectations.
> 

But it breaks correctness (the expected result) of the legacy code.

dbabokin:
> But it breaks correctness (the expected result) of the legacy code.

It could break potentially. We should document that in release notes. We should check with big customers that we know to see the impact if we decide to go this path.

But if the impact on existing code is low, then benefits of having cleaner definition outweighs it.

aneshlya:
I think we should proceed with `roundeven` and explicitly state in the documentation that ISPC's `round` function aligns with `roundeven` + describe rationale behind it (direct mapping to a single x86 instruction, offering enhanced performance — a core objective of ISPC).
Current docs have some mentions about it but very implicit: `On machines that support Intel®SSE or Intel® AVX, these functions all map to variants of the ``roundss`` and ``roundps`` instructions, respectively.`

I believe that there is no need to introduce one more rounding function unless one of "big customers" requests it.

dbabokin (approved):
I'm ok with Arina's suggestion. Note, that docs needs clarification.

nurmukhametov:
I've rebased the branch and updated docs.

aneshlya (approved):

aneshlya:
One lit test requires update on Windows.

PR #2795, opened 2024-03-14
Fix path to clang_rt.profile library

nurmukhametov:
This PR fixes path to `clang_rt.profile.lib` in superbuild.

It also updates name of LLVM artifacts with `-lto` for reusable build triggered with `LTO=ON`. The minor thing is to provide defaults for reusable builds.

### DIFF
--- a/.github/workflows/rebuild-llvm17.yml
+++ b/.github/workflows/rebuild-llvm17.yml
@@ -20,11 +20,15 @@ on:
 
 jobs:
   llvm17:
+    strategy:
+      fail-fast: false
+      matrix:
+        lto: ['ON', 'OFF']
     uses: ./.github/workflows/reusable.rebuild.yml
     with:
       version: '17.0'
       full_version: '17.0.6'
-      lto: 'OFF'
+      lto: ${{ matrix.lto }}
       ubuntu: '18.04'
       vs_generator: 'Visual Studio 16 2019'
       vs_version_str: 'vs2019'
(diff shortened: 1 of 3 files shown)


.github/workflows/rebuild-llvm17.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     with:
       version: '17.0'
       full_version: '17.0.6'
-      lto: 'OFF'
+      lto: 'ON'
```
aneshlya, line 27: I think it makes sense to make `lto` as an input parameter since we need both `lto` and not-`lto` versions for testing.

nurmukhametov: Yes, it is but this is a mostly push triggered job.

aneshlya: Let's say we added some new patches and we need to rebuild LLVM. The job will be triggered automatically with LTO=ON. But some tests jobs require LLVM with LTO=OFF. How to trigger rebuild of LLVM with LTO=OFF?

nurmukhametov: I see no solution better than splitting the every LLVM job into two: LTO=ON, LTO=OFF.

aneshlya: Looks like it would be the best. I'd use matrix for that
```
    strategy:
      matrix:
        lto: ['ON', 'OFF']
...
lto: ${{ matrix.lto }}
```

aneshlya (approved):
Feel free to merge when it's ready.

PR #2796, opened 2024-03-15
CI LLVM 17 and LLVM 18 updates

aneshlya:
Depends on https://github.com/ispc/ispc/pull/2795.

1. Use LLVM 17.0 by default in CI and Dockerfiles
2. Use LLVM 17.0 and corresponding Xe deps in osPresets
3. Added workflow to build LLVM 18 in CI


nurmukhametov:
I have rebased and updated this PR here #2828

PR #2797, opened 2024-03-15
Fix security options for Debug build

aneshlya:
Remove `-fdata-sections`, `-ffunction-sections` and `-Wl,--gc-sections` from builds with debug info to avoid removal of unused methods which might be useful for debugging purposes like `Print()`.
Related to discussion https://github.com/ispc/ispc/pull/2789#discussion_r1523138858


dbabokin (approved):
Looks good to me.

nurmukhametov (approved):

PR #2798, opened 2024-03-15
Replace alloca with heap allocation

nurmukhametov:
This PR removes `alloca` usage as requested here https://github.com/ispc/ispc/pull/2788#discussion_r1521970362

It also removes unused its declaration in `src/opt.cpp`. There is one more left in `src/util.cpp` - I'm not sure if it is worth changing it.

### DIFF
--- a/src/module.cpp
+++ b/src/module.cpp
@@ -115,14 +115,6 @@
 #define strcasecmp stricmp
 #endif
 
-// Clang defines alloca which interferes with standard library function.
-// It happenned after newly included clang/Basic/Builtins.h in PR71709
-#if ISPC_LLVM_VERSION > ISPC_LLVM_17_0
-#define ALLOCA __builtin_alloca
-#else
-#define ALLOCA alloca
-#endif
-
 using namespace ispc;
 
 // The magic constants are derived from https://github.com/intel/compute-runtime repo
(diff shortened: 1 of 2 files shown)


src/module.cpp
```diff
@@ -502,8 +499,10 @@ Expr *lCreateConstExpr(ExprList *exprList, const AtomicType::BasicType basicType
         }
     } else {
         // T equals int8_t* and etc.
+        // We allocate PointToType[N] on the heap. It is managed by unique_ptr.
         using PointToType = typename std::remove_pointer<T>::type;
-        vals = (T)ALLOCA(N * sizeof(PointToType));
+        managedMemory = std::make_unique<PointToType[]>(N);
```
aneshlya, line 504: I would add some comments here from the discussions https://github.com/ispc/ispc/pull/2788 and https://github.com/ispc/ispc/pull/2739#discussion_r1471150928 to explain why it is implemented this way and what are the problems if you try to implement it "easier".

aneshlya (approved):

dbabokin (approved):
Looks good to me.

I also agree with Arina that more comments would be welcome for future generations who will deal with this code.

src/module.cpp
nurmukhametov: I have added the comment.

PR #2800, opened 2024-03-18
SwitchBoolSize: distinguish bool type by argument

### LINKED ISSUE #2799: sse4-i8x16 target is broken
Nightly test report that it's broken with ~315 test failing and the whole run for -O2 takes about 2 hours, which is suspiciously too long. Here's an example of a failed [run](https://github.com/ispc/ispc/actions/runs/8313685464/job/22750101032).

I can reproduce this locally.

According to the nightly runs it work [here](https://github.com/ispc/ispc/actions/runs/7897628896) and doesn't work [here](https://github.com/ispc/ispc/actions/runs/7912249250), which mean it's a regression between this [working commit](https://github.com/ispc/ispc/commit/36577b39251479950048bec33395884fd8a1d8d2) and this [failing commit](https://github.com/ispc/ispc/commit/555325b7e807f0fb018bc0ae932053f3c9a74078).

Problematic commit is one of [these](https://github.com/ispc/ispc/commits/main/?since=2024-02-12&until=2024-02-14). And looks like the latest release is broken.

nurmukhametov:
Storage and internal bool types can have same type width, so use extra flag to choose the proper sign/zero extension mode.

This PR fixes #2799

### DIFF
--- a/src/ctx.h
+++ b/src/ctx.h
@@ -1,5 +1,5 @@
 /*
-  Copyright (c) 2010-2023, Intel Corporation
+  Copyright (c) 2010-2024, Intel Corporation
 
   SPDX-License-Identifier: BSD-3-Clause
 */
@@ -478,10 +478,12 @@ class FunctionEmitContext {
     llvm::Value *AddElementOffset(AddressInfo *basePtrInfo, int elementNum, const llvm::Twine &name = "",
                                   const PointerType **resultPtrType = nullptr);
 
-    /** Bool is stored as i8 and <WIDTH x i8> but represented in IR as i1 and
-     * <WIDTH x MASK>. This is a helper function to match bool size at storage
-     * interface. */
-    llvm::Value *SwitchBoolSize(llvm::Value *value, llvm::Type *toType, const llvm::Twine &name = "");
+    /** Bool is stored as i8 and <WIDTH x i8> (storage type) but represented in
+     * IR as i1 and <WIDTH x MASK> (mask type). These are two helper functions
+     * to match bool sizes.  */
+    llvm::Value *SwitchBoolToMaskType(llvm::Value *value, llvm::Type *toType, const llvm::Twine &name = "");
+    llvm::Value *SwitchBoolToStorageType(llvm::Value *value, llvm::Type *toType, const llvm::Twine &name = "");
+
     /** Load from the memory location(s) given by lvalue, using the given
         mask.  The lvalue may be varying, in which case this corresponds to
         a gather from the multiple memory locations given by the array of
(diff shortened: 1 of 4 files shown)


tests/lit-tests/2799.ispc
```diff
@@ -0,0 +1,12 @@
+// RUN: %{ispc} --pic --target=avx2-i8x32 %s -o %t.o
+// RUN: %{cc} %t.o -o %t.bin
+// RUN: %t.bin | FileCheck %s
+
+// REQUIRES: !MACOS_HOST && X86_ENABLED
```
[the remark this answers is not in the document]
dbabokin: Ahhh, I see the problem and it's not about macOS. It's about cross-targets. You need not just `X86_ENABLED`, you need your target to be x86.

This should fail on Arm Linux running X86-enabled build.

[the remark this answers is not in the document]
dbabokin: We either need to add a macro which reflects host architecture or rely `run_tests.py` for runtime tests.

LIT is not really designed for runtime tests.

tests/lit-tests/2799.ispc
```diff
@@ -0,0 +1,8 @@
+// RUN: %{ispc} --pic --target=avx2-i8x32 %s -O2 --emit-llvm-text -o - | FileCheck %s
+
+// CHECK: store i64 496, {{.*}} %print_arg
```
dbabokin, line 3: Still need `X86_ENABLED`.

nurmukhametov:
Full tests run with all green i8xN targets is [here](github.com/ispc/ispc/actions/runs/8379368406).

tests/lit-tests/2799.ispc
nurmukhametov: Fixed that.

src/ctx.h
```diff
@@ -479,9 +479,12 @@ class FunctionEmitContext {
                                   const PointerType **resultPtrType = nullptr);
 
     /** Bool is stored as i8 and <WIDTH x i8> but represented in IR as i1 and
-     * <WIDTH x MASK>. This is a helper function to match bool size at storage
-     * interface. */
-    llvm::Value *SwitchBoolSize(llvm::Value *value, llvm::Type *toType, const llvm::Twine &name = "");
+     * <WIDTH x MASK>. This is a helper function to match bool size. The
+     * argument toStorageType is true when conversion to storage type is
+     * needed.  */
+    llvm::Value *SwitchBoolSize(llvm::Value *value, llvm::Type *toType, bool toStorageType,
```
aneshlya, line 485: Personally I don't like bool function parameters because they reduce readability of the code quite a lot - you need go to function declaration to understand what this parameter does. I would do 2 separate functions `SwitchBoolSizeToI1` and `SwitchBoolSizeToStorageType`. 
But it's up to you. We already have so many functions with similar bool parameters so this change doesn't make any difference in this regard :)

aneshlya (approved):

src/ctx.h
nurmukhametov: I have introduced two wrappers `SwitchBoolToStorageType` and `SwitchBoolToMaskType` that are to be used to convert types.

aneshlya (approved):

PR #2801, opened 2024-03-18
Fix build with LLVM trunk

aneshlya:
Fix ISPC build after https://github.com/llvm/llvm-project/commit/9997e0397156ff7e01aecbd17bdeb7bfe5fb15b0

### DIFF
--- a/src/ctx.cpp
+++ b/src/ctx.cpp
@@ -1599,8 +1599,14 @@ void FunctionEmitContext::EmitVariableDebugInfo(Symbol *sym) {
 
     llvm::DebugLoc diLoc =
         llvm::DILocation::get(scope->getContext(), sym->pos.first_line, sym->pos.first_column, scope, nullptr, false);
-    llvm::Instruction *declareInst = m->diBuilder->insertDeclare(sym->storageInfo->getPointer(), var,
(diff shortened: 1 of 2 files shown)


src/ctx.cpp
```diff
@@ -1599,8 +1599,14 @@ void FunctionEmitContext::EmitVariableDebugInfo(Symbol *sym) {
 
     llvm::DebugLoc diLoc =
         llvm::DILocation::get(scope->getContext(), sym->pos.first_line, sym->pos.first_column, scope, nullptr, false);
-    llvm::Instruction *declareInst = m->diBuilder->insertDeclare(sym->storageInfo->getPointer(), var,
-                                                                 m->diBuilder->createExpression(), diLoc, bblock);
+    llvm::Instruction *declareInst =
+#if ISPC_LLVM_VERSION > ISPC_LLVM_18_1
```
dbabokin, line 1603: Probably `ISPC_LLVM_VERSION >= ISPC_LLVM_19_0`?

What if they ship 18.2?

aneshlya: Updated

dbabokin (approved):
LGTM

PR #2806, opened 2024-03-20
Fix memorySSA enablement under debugPM

### LINKED ISSUE #2804: LLVM ERROR with --dump-file
The following example:
```ispc
typedef bool<1> bool1; 
task void f_f(uniform float RET[], uniform float a[]) { 
   RET[programIndex] = 0; 
   bool1 vec1 = {true}; 
   bool1 vec = vec1 == vec1; 
    if (!vec.x) 
       RET[0] = -1; 
}
```
when dumping IR to files with the following command:
```bash
$ ispc --target=avx2-i8x32 --debug-phase=first:last -O2 --dump-file=dir test.ispc
```

produces the LLVM ERROR: LICM requires MemorySSA (loop-mssa).

nurmukhametov:
This PR fixes #2804

### DIFF
--- a/src/opt.cpp
+++ b/src/opt.cpp
@@ -185,7 +185,10 @@ class DebugModulePassManager {
     // Start a new group of loop passes
     void initLoopPassManager();
     // Add loop passes to the FunctionPassManager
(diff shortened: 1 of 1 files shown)


aneshlya (approved):
Looks good.

PR #2807, opened 2024-03-20
Rename Print() method for debuggers

### LINKED ISSUE #2805: Too few arguments for Print() in gdb
Following-up the discussion in comments here https://github.com/ispc/ispc/pull/2789#discussion_r1521935655 and this fix https://github.com/ispc/ispc/pull/2797, I need to report a one more case when `Print()` function of `ASTNode`/`Expr` class hierarchy doesn't work in gdb.

For the following test:
```ispc
typedef bool<1> bool1; 
task void f_f(uniform float RET[], uniform float a[]) { 
   RET[programIndex] = 0; 
   bool1 vec1 = {true}; 
   bool1 vec = vec1 == vec1; 
    if (!vec.x) 
       RET[0] = -1; 
}
```

the following workflow:
```bash
$ gdb --quiet --args ./bin/ispc --nostdlib --target=avx2-i8x32 -o test.o test.ispc
(gdb) b FunctionEmitContext::SwitchBoolSize
(gdb) r
(gdb) f 2
(gdb) p this
$1 = (const ispc::SymbolExpr *) 0x55555b67e710
(gdb) p this->Print()
Too few arguments in function call.
(gdb) p ((ispc::Expr*)this)->Print()
`-SymbolExpr <../tests/short-vec-bool3.ispc:13.17 - 13.21> [varying bool<1>] symbol name: vec1
$2 = void
```

Renaming `Print` to something else, e.g., `Dump` allows the debugger to avoid this issue.

nurmukhametov:
It is needed to avoid name conflict that happens with overloaded virtual method Print(Indent). Debuggers struggle (at least gdb) to resolve it.

This fixes #2805

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -1,5 +1,5 @@
 /*
-  Copyright (c) 2011-2023, Intel Corporation
+  Copyright (c) 2011-2024, Intel Corporation
 
   SPDX-License-Identifier: BSD-3-Clause
 */
@@ -101,7 +101,7 @@ void Indent::Done() {
 
 ASTNode::~ASTNode() {}
 
-void ASTNode::Print() const {
+void ASTNode::Dump() const {
     Indent indent;
     indent.pushSingle();
     Print(indent);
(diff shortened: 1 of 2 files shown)


dbabokin (approved):
ok, if it works better this way, let it be `Dump()`.

PR #2809, opened 2024-03-22
WIP: opt: add ShrinkVector pass

nurmukhametov:
It traverse bitcode for masked stores that have the turned-off second half. We can safely replace them with narrow masked stores. This can help the back-end to generate better code (no extra spills, assigning narrow registers).

This fixes that last part of issue #2611

### DIFF
--- a/src/opt/ShrinkVectors.cpp
+++ b/src/opt/ShrinkVectors.cpp
@@ -1,0 +1,270 @@
+/*
+  Copyright (c) 2024, Intel Corporation
+
+  SPDX-License-Identifier: BSD-3-Clause
+*/
+
+#include "ShrinkVectors.h"
+
+#include <llvm/IR/IRBuilder.h>
+
+namespace ispc {
+
+bool lIsSecondHalfAllFalse(llvm::Value *mask) {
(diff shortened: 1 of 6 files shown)


src/opt/ShrinkVectors.cpp
```diff
@@ -0,0 +1,270 @@
...
+    }
+    while (!worklist.empty()) {
+        llvm::Value *val = worklist.front();
+        worklist.pop_front();
+        if (llvm::CallInst *call = llvm::dyn_cast<llvm::CallInst>(val)) {
+            llvm::Function *func = call->getCalledFunction();
+            assert(func);
+
+            // TODO! We may probably want to consider more
+            // instruction/intrinsics to propagate shrinking fact.
```
aneshlya, line 249: Feels like the list of instructions will be huge :)

aneshlya:
Currently, the pass appears to be narrowly focused on a specific scenario and lacks generality. 
Do I understand correctly that the core concept here is to optimize memory operations by downsizing vector size and adjusting associated operations accordingly **when it's clear that the full vector isn't required**? Is it clear on the front-end stage?

PR #2810, opened 2024-03-25
Remove no spr dispatch

### LINKED ISSUE #2774: Remove dispatch-no_spr
As comment [here](https://github.com/ispc/ispc/blob/aa6610460368619567f0dcda037230dfae369af8/builtins/dispatch.c#L24) says `dispatch-no_spr` should be removed (we already dropped LLVM 13 support).

It also not clear for me why we need to keep `dispatch-*.ll` files at all. We may generate bitcode straight from `dispatch.c` during build process.

nurmukhametov:
This PR fixes #2774


aneshlya (approved):
LGTM

cmake/GenerateBuiltins.cmake
```diff
@@ -72,13 +72,19 @@ function(target_ll_to_cpp llFileName bit os_name resultFileName)
...
+    if (${os_name} STREQUAL "macos")
+        set(DISP_TYPE -DMACOS)
+        set(DISP_NAME "dispatch-macos")
+    endif()
+    set(output ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_CFG_INTDIR}/builtins-${DISP_NAME}.cpp)
     add_custom_command(
         OUTPUT ${output}
-        COMMAND ${M4_EXECUTABLE} -DLLVM_VERSION=${LLVM_VERSION} ${inputFilePath}
-            | \"${Python3_EXECUTABLE}\" bitcode2cpp.py ${inputFilePath} --type=dispatch --os=${os_name} --llvm_as ${LLVM_AS_EXECUTABLE} --opaque_flags="${LLVM_TOOLS_OPAQUE_FLAGS}"
+        COMMAND ${CLANGPP_EXECUTABLE} -x c ${ISPC_OPAQUE_FLAGS} ${DISP_TYPE} -march=x86-64 -mtune=generic -O2 -emit-llvm -S -c ${inputFilePath} -o -
```
dbabokin, line 86: I have this error on ARM macOS without cross-build enabled:
```
[  5%] Generating builtins-dispatch-macos.cpp
clang++: error: unsupported argument 'x86-64' to option '-march='
[  7%] Generating builtins-dispatch.cpp
clang++: error: unsupported argument 'x86-64' to option '-march='
```
which is not surprising.

Also, looks like we are missing ARM builds in CI. We do have hosted ARM Linux runner, we should enabled a ARM-only and full ARM build there. We used to have Travis to test ARM, but it's not running anymore, I'm not sure what happened to it.

dbabokin (changes requested):
These are scary changes (to me), that's why I did some unification of this code over the years, but never went all the way to making dispatch compiled during the build. So I applaud your courage here :)

My only concern (beyond that it's not fully working for non-x86 builds for now) is that it's completely missing test coverage. When I was updating dispatch, I was manually verifying the changes - looking at the resulting code and running under SDE to check that it correctly behaves under all different conditions.

Current testing infra is not designed to support such testing. I suggest adding new class of tests. Something like "integrations tests" and try multi-target under SDE with all conditions that are interesting for us. Probably it should be just hardcoded via the script. 

Also probably some other tests will need to be added to this class - for example ABI (C <->ISPC) testing and some of the runtime tests that you were trying to add to lit tests.

nurmukhametov:
OK, then reduce the scope of this PR to just removing `dispatch-no-spr.ll`. 

I created issues about everything else #2815, #2816, #2817 .

dbabokin (approved):
LGTM

PR #2814, opened 2024-03-26
CI: upload alloy_results logs on failure

nurmukhametov:
This PR also splits x86 and x86-64 linux jobs to different one matrix jobs.

As a result, we will be able to see the logs with the real issues that sometimes are unclear to reproduce locally.

### DIFF
--- a/.github/workflows/ispc-ci.yml
+++ b/.github/workflows/ispc-ci.yml
@@ -348,6 +348,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


aneshlya (approved):
You didn't add matrix to `linux-test-xe-llvm16-release` and `linux-test-llvm16-release` jobs. Is it intentional? I understand that the change is motivated by desire to investigate avx512 failures (and this jobs doesn't run avx512 targets) but it just look strange in CI run.

aneshlya (approved):

Issue #2479, opened 2023-04-06
Builtins compilation error


nurmukhametov:
On rare occasion, e.g., user stops building process with `ctrl+c`, `bitcode2cpp.py` may produce incorrect `builtins-target-*.cpp` files. 

Like this:
```c++
#include "bitcode_lib.h"

using namespace ispc;

extern const unsigned char builtins_bitcode_unix_avx1_i32x16_64bit[] = {
```

With the following compilation error:
```bash
D:\users\anurmukh\ispc\build\builtins-target-avx1-i32x16-64bit-unix.cpp(5): fatal error C1075: '{': no matching token found
```

PR #2823, opened 2024-04-01
WIP: Two stage build

nurmukhametov:


CMakeLists.txt
```diff
@@ -650,45 +783,99 @@ foreach(clangLib ${CLANG_LIBRARY_LIST})
...
+# This function configures executable targets ispc-slim and ispc-whole.
+function(configure_ispc TARGET TARGET_SPECIFIC_SOURCES)
+    add_executable(${TARGET} ${TARGET_SPECIFIC_SOURCES})
+    configure_ispc_obj(${TARGET})
+    target_link_options(${TARGET} PRIVATE ${LINK_OPTIONS_PRIVATE})
+    target_link_options(${TARGET} PUBLIC ${LINK_OPTIONS_PUBLIC})
+    target_link_libraries(${TARGET} ${LINK_LIBRARIES} common)
+endfunction()
+
+# Target common contains all sources that are shared between slim and whole binaries.
```
turinevgeny, line 850: An idea to refer to the old binary as "composite".
"slim" could also be named as "lean". Just a suggestion for consideration.

src/bitcode_lib.cpp
```diff
@@ -9,25 +9,56 @@
            Builtiins-c, or ISPCTarget).
 */
 
+#include <stdlib.h>
```
turinevgeny, line 12: Do the ISPC sources adhere to the LLVM coding style of includes?
https://llvm.org/docs/CodingStandards.html#include-style

src/builtins.cpp
```diff
@@ -987,91 +987,60 @@ static void lSetInternalFunctions(llvm::Module *module) {
...
-            // to update attributes because Xe intrinsics are
-            // separated from the others and it is not done by default
-            lUpdateIntrinsicsAttributes(bcModule);
-        }
+    llvm::Module *bcModule = lib->getLLVMModule();
+
+    // FIXME: this feels like a bad idea, but the issue is that when we
+    // set the llvm::Module's target triple in the ispc Module::Module
+    // constructor, we start by calling llvm::sys::getHostTriple() (and
+    // then change the arch if needed).  Somehow that ends up giving us
```
turinevgeny, line 995:
```suggestion
    // then change the arch if needed). Somehow that ends up giving us
```

src/builtins.cpp
```diff
@@ -987,91 +987,60 @@ static void lSetInternalFunctions(llvm::Module *module) {
...
+    // strings like 'x86_64-apple-darwin11.0.0', while the stuff we
+    // compile to bitcode with clang has module triples like
+    // 'i386-apple-macosx10.7.0'.  And then LLVM issues a warning about
```
turinevgeny, line 998:
```suggestion
    // 'i386-apple-macosx10.7.0'. And then LLVM issues a warning about
```

src/slim.cpp
```diff
@@ -0,0 +1,278 @@
...
+    llvm::sys::path::remove_filename(includeDir);
+    llvm::SmallString<128> shareDir(includeDir);
+    // llvm::sys::path::append(includeDir, "include");
+    // lParseInclude(includeDir.c_str());
+    llvm::sys::path::append(shareDir, "share", "ispc");
+    g->shareDirPath = std::string(shareDir.str());
+}
+
+// Dispatch bitcode_libs
+// TODO! strangly enough this TargetOS has no sense and this dispatch lib is used on windows
```
turinevgeny, line 31:
```suggestion
// TODO! strangely enough this TargetOS has no sense and this dispatch lib is used on windows
```

src/slim.cpp
```diff
@@ -0,0 +1,278 @@
...
+#endif // defined(ISPC_WINDOWS_TARGET_ON) && defined(ISPC_XE_ENABLED)
+
+#ifdef ISPC_WASM_ENABLED
+static BitcodeLib cpp_32_web_wasm32("builtins-cpp-32-web-wasm32.cpp", TargetOS::web, Arch::wasm32);
+static BitcodeLib cpp_64_web_wasm64("builtins-cpp-64-web-wasm64.cpp", TargetOS::web, Arch::wasm64);
+#endif // ISPC_WASM_ENABLED
+
+// Target-specific built-in bitcode_libs
+// clang-format off
+#if defined(ISPC_UNIX_TARGET_ON) && defined(ISPC_X86_ENABLED)
```
turinevgeny, line 107: Is it possible to simplify the patterns below with, maybe, macro magic?

CMakeLists.txt
```diff
@@ -650,45 +783,99 @@ foreach(clangLib ${CLANG_LIBRARY_LIST})
...
+endif()
+
+
+# This function configures compile definitions, options and include dirs.
+function(configure_ispc_obj TARGET)
+    target_compile_definitions(${TARGET} PRIVATE ${COMPILE_DEFINITIONS})
+    target_include_directories(${TARGET} PRIVATE ${INCLUDE_DIRECTORIES})
+    target_compile_options(${TARGET} PRIVATE ${COMPILE_OPTIONS})
+
+    # Set hidden visibility for inline functions.  This is needed to be in sync
```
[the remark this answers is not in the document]
nurmukhametov: Fixed

CMakeLists.txt
```diff
@@ -488,151 +559,213 @@ endif()
...
 else()
-    target_compile_definitions(${PROJECT_NAME} PRIVATE NOMINMAX)
+    list(APPEND COMPILE_DEFINITIONS NOMINMAX)
     if (NOT CMAKE_BUILD_TYPE STREQUAL "DEBUG" )
-        target_compile_definitions(${PROJECT_NAME} PRIVATE NDEBUG)
+        list(APPEND COMPILE_DEFINITIONS NDEBUG)
     endif()
 endif()
 
+# It may be confusing that some CMake variables start with ISPC but some not.
```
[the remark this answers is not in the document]
nurmukhametov: Yes, sure. I've added it.

src/bitcode_lib.cpp
nurmukhametov: I am not sure. It looks like almost except system headers place.

src/builtins.cpp
nurmukhametov: Done

src/builtins.cpp
nurmukhametov: Done

src/slim.cpp
nurmukhametov: Done

src/slim.cpp
nurmukhametov: Under assumption that target name in file contains `_` instead of `-`, it is possible to _simplify_ with something like this (not really tested though):
```c
#define SSE2_TARGETS(M) \
  M(sse2_i32x4); \
  M(sse2_i32x8);

#define SSE4_TARGETS(M) \
  M(sse4_i8x16) \
  M(sse4_i16x8) \
  M(sse4-i32x4) \
  M(sse4-i32x8)

#define AVX1_TARGETS(M) \
  M(avx1-i32x8) \
  M(avx1-i32x16) \
  M(avx1-i64x4)

#define AVX2_TARGETS(M) \
  M(avx2-i8x32) \
  M(avx2-i16x16) \
  M(avx2-i32x4) \
  M(avx2-i32x8) \
  M(avx2-i32x16) \
  M(avx2-i64x4)

#define AVX512KNL_TARGETS(M) \
  M(avx512knl-x16)

#define AVX512SKX_TARGETS(M) \
  M(avx512skx-x4) \
  M(avx512skx-x8) \
  M(avx512skx-x16) \
  M(avx512skx-x32) \
  M(avx512skx-x64)

#define AVX512SPR_TARGETS(M) \
  M(avx512spr-x4) \
  M(avx512spr-x8) \
  M(avx512spr-x16) \
  M(avx512spr-x32) \
  M(avx512spr-x64)

#define X86_TARGETS(M) \
  SSE2_TARGETS(M) \
  SSE4_TARGETS(M) \
  AVX1_TARGETS(M) \
  AVX2_TARGETS(M) \
  AVX512KNL_TARGETS(M) \
  AVX512SKX_TARGETS(M) \
  AVX512SPR_TARGETS(M)

#define TARGET_BITCODE_LIB(TARGET, OS_NAME, OS_TARGET, BIT, ARCH) static BitcodeLib target_ ## TARGET ## _ ## BIT ## bit_ ## OS_NAME("builtins-" #TARGET "-" #BIT "bit-" #OS_NAME ".bc", ISPCTarget::TARGET, TargetOS::OS_TARGET, Arch::ARCH)

#define TARGET_BITCODE_LIB_UNIX_X86_64(TARGET) TARGET_BITCODE_LIB(TARGET, unix, linux, 64, x86_64)
#define TARGET_BITCODE_LIB_UNIX_X86(TARGET) TARGET_BITCODE_LIB(TARGET, unix, linux, 32, x86)

#define TARGET_BITCODE_LIB_WINDOWS_X86_64(TARGET) TARGET_BITCODE_LIB(TARGET, windows, windows, 64, x86_64)
#define TARGET_BITCODE_LIB_WINDOWS_X86(TARGET) TARGET_BITCODE_LIB(TARGET, windows, windows, 32, x86)

X86_TARGETS(TARGET_BITCODE_LIB_UNIX_X86_64)
X86_TARGETS(TARGET_BITCODE_LIB_UNIX_X86)

X86_TARGETS(TARGET_BITCODE_LIB_WINDOWS_X86_64)
X86_TARGETS(TARGET_BITCODE_LIB_WINDOWS_X86)
```

turinevgeny: It might be worth considering if the future extensions or support of this code would be easier.

CMakeLists.txt
```diff
@@ -650,45 +784,99 @@ foreach(clangLib ${CLANG_LIBRARY_LIST})
...
+
+if (ISPC_SLIM_BINARY)
+    configure_ispc(ispc src/slim.cpp)
+    add_dependencies(ispc builtins)
+    message(STATUS "ispc is slim binary")
+else()
+    configure_ispc(ispc-slim src/slim.cpp)
+    add_dependencies(ispc-slim builtins)
+
+    configure_ispc(ispc src/whole.cpp)
```
aneshlya, line 863: I would expect this target to be named `ispc-whole` to be aligned with comments to `configure_ispc` function.

CMakeLists.txt
```diff
@@ -650,45 +784,99 @@ foreach(clangLib ${CLANG_LIBRARY_LIST})
...
+endfunction()
+
+# Target common contains all sources that are shared between slim and whole binaries.
+add_library(common OBJECT ${COMMON_SOURCES})
+configure_ispc_obj(common)
+
+if (ISPC_SLIM_BINARY)
+    configure_ispc(ispc src/slim.cpp)
+    add_dependencies(ispc builtins)
+    message(STATUS "ispc is slim binary")
```
aneshlya, line 858: Let's use "ISPC will be built as a slim binary" message instead. Same for "whole".

CMakeLists.txt
aneshlya: I'm OK with "slim" and I like the idea to have "composite" instead of "whole".

src/dist.h
```diff
@@ -0,0 +1,15 @@
+/*
+  Copyright (c) 2024, Intel Corporation
+
+  SPDX-License-Identifier: BSD-3-Clause
+*/
+
+/** @file dist.h
+    @brief Defines function with distribution specific implementations.
```
aneshlya, line 8: It doesn't look for me as "distribution", it's about binary type.
I would rename this file to "binary_type.h".
Also I prefer separating build-related sources from compiler-core sources so I'd move dist.h, slim.h and whole.h to separate src/binary_type folder and rename them to binary_type.h, slim.h and composite.h. If you don't like idea of creating a subfolder it can be binary_type.h, binary_slim.h and binary_composite.h in src folder.

aneshlya, line 8: If you want to define a common interface here why not to create a base class and inherit slim and composite from it?

src/dist.h
```diff
@@ -0,0 +1,15 @@
...
+*/
+
+/** Print the binary type: slim or whole. */
+void printBinaryType();
+
+/** Initializes distribution specific paths based on main executable abspath. */
+void initializeShareDirPath(const char *MainExecutableAbsPath);
```
aneshlya, line 15: Since composite binary doesn't use share dir path I suggest to rename this method to just `initialize`.

src/slim.cpp
aneshlya: Maybe we can use CMake configuration template here?

aneshlya (commented):
I'd like to do more thorough review but here are some initial comments.

src/slim.cpp
nurmukhametov: > Maybe we can use CMake configuration template here?

What for?

CMakeLists.txt
nurmukhametov: Done.

src/dist.h
nurmukhametov: Done.

nurmukhametov: What issue/problem does it solve?

aneshlya: I didn't notice that only one of them can be used in compile time. It's OK then.

src/slim.cpp
aneshlya: > What for?

To avoid manually maintaining the list of builtins, which is prone to errors. I think it can be more efficient to utilize CMake. CMake is already aware of all target names, architectures, and compilation definitions. It would also help to concentrate all logic regarding build in one place.

nurmukhametov: > To avoid manually maintaining the list of builtins, which is prone to errors.
I doubt that CMake code would be free of it. 

> I think it can be more efficient to utilize CMake. CMake is already aware of all target names, architectures, and compilation definitions. It would also help to concentrate all logic regarding build in one place.

I disagree that this is build logic.

aneshlya: I still can hardly imagine how to support the list of `BitcodeLib` in binary_slim.cpp up-to-date and in sync with logic in GenerateBuiltins.cmake `generate_common_builtins` function.
I'd do something like this:
```
# GenerateBuiltins.cmake
file(WRITE ${CMAKE_BINARY_DIR}/bitcode_libs_generated.cpp "

macro(write_bitcode_lib bitcode_file)
    file(APPEND ${CMAKE_BINARY_DIR}/bitcode_libs_generated.cpp
        "static BitcodeLib ${bitcode_file};\n")
endmacro()

function(builtin_to_cpp bit os generic_arch)
...
write_bitcode_lib(${bc})
```
Same for targets.

CMakeLists.txt
```diff
@@ -413,10 +431,42 @@ message(STATUS "ISPC opaque pointers mode is " ${ISPC_OPAQUE_PTR_MODE})
...
 
+add_custom_target(dispatch-builtins-bc DEPENDS ${DISPATCH_BUILTIN_BC_FILES})
+add_custom_target(common-builtins-bc DEPENDS ${COMMON_BUILTIN_BC_FILES})
+add_custom_target(target-builtins-bc DEPENDS ${TARGET_BUILTIN_BC_FILES})
+add_custom_target(builtins-bc DEPENDS dispatch-builtins-bc target-builtins-bc common-builtins-bc)
+
+add_custom_target(dispatch-builtins-cpp DEPENDS dispatch-builtins-bc)
+add_custom_target(common-builtins-cpp DEPENDS common-builtins-bc)
+add_custom_target(target-builtins-cpp DEPENDS target-builtins-bc)
+add_custom_target(builtins-cpp DEPENDS dispatch-builtins-cpp target-builtins-cpp common-builtins-cpp)
```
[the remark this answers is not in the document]
nurmukhametov: Fixed

CMakeLists.txt
```diff
@@ -413,10 +431,42 @@ message(STATUS "ISPC opaque pointers mode is " ${ISPC_OPAQUE_PTR_MODE})
...
+
+if (MSVC)
+    # Group generated files inside Visual Studio
+    source_group("Generated Dispatch Builtins" FILES ${DISPATCH_BUILTIN_CPP_FILES})
+    source_group("Generated Common Builtins" FILES ${COMMON_BUILTIN_CPP_FILES})
+    source_group("Generated Target Builtins" FILES ${TARGET_BUILTIN_CPP_FILES})
+endif()
+set_source_files_properties(${DISPATCH_BUILTIN_CPP_FILES} PROPERTIES GENERATED true)
+set_source_files_properties(${COMMON_BUILTIN_CPP_FILES} PROPERTIES GENERATED true)
+set_source_files_properties(${TARGET_BUILTIN_CPP_FILES} PROPERTIES GENERATED true)
```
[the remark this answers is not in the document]
nurmukhametov: I am not really sure that I follow all but I have moved all builtin logic inside `GenerateBuiltins.cmake`.

bitcode2cpp.py
```diff
@@ -152,6 +138,3 @@
 else:
     sys.stderr.write("Unknown argument for --type: " + args[0].type + "\n")
     sys.exit(1)
-
-as_out.wait()
-sys.exit(as_out.returncode)
```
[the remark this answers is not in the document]
nurmukhametov: I changed this script. It now takes input from the file and write output to the file. Previously, it operated with stdin/stdout. It is cleaner to use in cmake (e.g., not escape like this needed `| \"${Python3_EXECUTABLE}\"`) and it may allows us to resolve https://github.com/ispc/ispc/issues/2479 (write to temp file and then atomically rename it).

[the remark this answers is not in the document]
nurmukhametov: I have added the fix to mentioned ticket as well.

aneshlya (approved):
It looks great now!

Issue #2719, opened 2023-12-15
Masked stores used when setting shorter width registers in some cases


rkaric:
Given the following code: https://godbolt.org/z/Kraa491fn

ISPC will operate on XMM registers up until the foreach() is encountered. Since the result is available in XMM0, it can safely perform an aligned/unaligned store using that register.

ISPC seems to broadcast to YMM0, load YMM1, mask store YMM0, and then clear the upper bits on YMM0.

nurmukhametov:
This looks like a duplicate of #2611.

pbrubaker:
Closing as duplicate of #2611.  This issue is mentioned in the other issue's thread.

PR #2836, opened 2024-04-16
opt: add ReplaceMaskedMemOps pass

nurmukhametov:
It traverses bitcode for masked stores that have the turned-off second part and the turned-on first part. We can safely replace them with narrow unmasked stores and loads with the following shuffle with the passthrough value. This can help the back-end to generate better code (no extra spills, assigning narrow registers).

This fixes that last part of issue https://github.com/ispc/ispc/issues/2611

### DIFF
--- a/src/opt.cpp
+++ b/src/opt.cpp
@@ -606,6 +606,7 @@ void ispc::Optimize(llvm::Module *module, int optLevel) {
 #else
         optPM.addFunctionPass(llvm::GVN(), 301);
 #endif
+        optPM.addFunctionPass(ReplaceMaskedMemOpsPass());
         optPM.addFunctionPass(IsCompileTimeConstantPass(true));
         optPM.addFunctionPass(IntrinsicsOpt());
         optPM.addFunctionPass(InstructionSimplifyPass());
--- a/src/opt/ReplaceMaskedMemOps.cpp
+++ b/src/opt/ReplaceMaskedMemOps.cpp
@@ -1,0 +1,253 @@
+/*
+  Copyright (c) 2024, Intel Corporation
+
+  SPDX-License-Identifier: BSD-3-Clause
+*/
+
+#include "ReplaceMaskedMemOps.h"
+
+#include <llvm/IR/IRBuilder.h>
+#include <llvm/Transforms/Utils/Local.h>
+
+#include <numeric>
+
+namespace ispc {
+
+bool lIsPowerOf2(unsigned n) { return (n > 0) && !(n & (n - 1)); }
+
+// Check if the mask is a constant vector with the first part being all ones
+// and the second part being all zeros. The length of the first part is
+// returned in TrueSubmaskLength argument. We are looking for the mask with the
+// true prefix of length 2^k only. This is a limitation of the current
+// implementation of lMaskedMergeVectors.
+bool lCheckMask(llvm::Value *mask, unsigned &TrueSubmaskLength) {
+    if (auto *CV = llvm::dyn_cast<llvm::ConstantVector>(mask)) {
+        auto N = CV->getType()->getNumElements();
+        if (!lIsPowerOf2(N)) {
+            return false;
+        }
+
+        unsigned TruePrefixLength = 0;
+        for (auto i = 0; i < N; i++) {
+            llvm::Constant *E = CV->getAggregateElement(i);
+            if (auto *C = llvm::dyn_cast<llvm::ConstantInt>(E)) {
+                if (C->isOne()) {
+                    TruePrefixLength++;
+                    continue;
+                } else {
+                    break;
+                }
+            }
+        }
+
+        for (auto i = TruePrefixLength + 1; i < N; i++) {
+            llvm::Constant *E = CV->getAggregateElement(i);
+            if (auto *C = llvm::dyn_cast<llvm::ConstantInt>(E)) {
+                if (C->isZero()) {
+                    continue;
+                } else {
+                    return false;
+                }
+            }
+        }
+
+        if (!lIsPowerOf2(TruePrefixLength)) {
+            return false;
+        }
+
+        TrueSubmaskLength = TruePrefixLength;
+        return true;
+    }
+    return false;
+}
+
+// Extract the first part of the vector with the shufflevector instruction. The
+// length of the part is passed in SubVectorLength.
+llvm::Value *lShrinkVector(llvm::IRBuilder<> &B, llvm::Value *originalVector, unsigned SubVectorLength) {
+    // Create a vector of indices with [0, 1, 2, ..., SubVectorLength).
+    std::vector<unsigned> indices(SubVectorLength);
+    std::iota(indices.begin(), indices.end(), 0);
+    llvm::Constant *mask = llvm::ConstantDataVector::get(B.getContext(), indices);
+
+    // Use shufflevector instruction to create the new vector
+    llvm::Twine name = originalVector->getName() + ".part";
+    return B.CreateShuffleVector(originalVector, llvm::UndefValue::get(originalVector->getType()), mask, name);
+}
+
+// Create a new vector containing the second part of the const vector.
+// The length of the first part is passed in SubVectorLength. The new vector
+// has the length of N - SubVectorLength, where N is the length of the original
+// vectors we are working with.
+llvm::Constant *lExtractSecondPartOfConstVec(llvm::LLVMContext &context, llvm::Constant *originalMask,
+                                             unsigned SubVectorLength) {
+    auto *vecType = llvm::dyn_cast<llvm::VectorType>(originalMask->getType());
+    Assert(vecType);
+    unsigned N = vecType->getElementCount().getKnownMinValue();
+
+    // Iterate over the second part of the elements and add them to the new vector.
+    std::vector<llvm::Constant *> newValues;
+    for (unsigned i = SubVectorLength; i < N; ++i) {
+        auto *element = originalMask->getAggregateElement(i);
+        if (!element) {
+            return nullptr;
+        }
+        newValues.push_back(element);
+    }
+
+    llvm::ArrayRef<llvm::Constant *> newValuesRef(newValues);
+    return llvm::ConstantVector::get(newValuesRef);
+}
+
+// Merge two vectors into a single vector with the sequence of shufflevector
+// instructions. The first vector length is the power of 2 (M). The overall
+// length is the power of 2 either (N). This means that we can do the merge in
+// the log2(N) - log2(M) steps. The second vector has the length of N. In
+// resulting vector, we want to have the first part and the rest elements from
+// the second vector. In other words, change in the second vector the first M
+// elements with the elemenst from the first vector.
+// Ideally, we would like to merge vector of different lengths. This would
+// simplify the current implemenation, but shuffle instruction requires the
+// vector to have same length.
+llvm::Value *lMaskedMergeVectors(llvm::IRBuilder<> &B, llvm::Value *firstVector, llvm::Constant *secondVector,
+                                 const llvm::Twine &name) {
+    auto *firstVecType = llvm::dyn_cast<llvm::VectorType>(firstVector->getType());
+    auto *secondVecType = llvm::dyn_cast<llvm::VectorType>(secondVector->getType());
+    Assert(firstVecType && secondVecType);
+    unsigned M = firstVecType->getElementCount().getKnownMinValue();
+    unsigned N = secondVecType->getElementCount().getKnownMinValue();
+
+    for (unsigned length = M; length < N; length *= 2) {
+        // Create a vector of indices [0, 1, 2, ..., length*2).
+        std::vector<unsigned> indices(length * 2);
+        std::iota(indices.begin(), indices.end(), 0);
+        llvm::Constant *mask = llvm::ConstantDataVector::get(B.getContext(), indices);
+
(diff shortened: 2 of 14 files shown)


nurmukhametov:
The test from #2719 was added.

turinevgeny (commented):
Would it makes sense to add unit LIT-tests? E.g. for given input IR check the new pass output.

nurmukhametov:
> Would it makes sense to add unit LIT-tests? E.g. for given input IR check the new pass output.

It would but at the moment I don't see such tests for other passes from `src/opt` as well as ability to run specific pass only on input IR.

src/opt/ReplaceHalfMaskedMemOps.h
```diff
@@ -0,0 +1,29 @@
...
+
+namespace ispc {
+
+// This pass transforms some masked loads and stores into their half-width
+// vector unmasked counterparts. This is done to improve the performance of
+// the generated code in cases when only half of the registers are used.
+// The masked.load and masked.store intrinsics are directly mapped to machine
+// instructions with the specified full width of vector values being loaded or
+// stored. This transformation allows the backend to generate shorter vector
+// loads and stores avoidind extra spills.
```
dbabokin, line 19: Is it only about shorter memory ops or it's also about short math operations? I assumed that it's both.

Also, what do you mean by "spill" here? It's reads/writes of the user visible memory, while spills refer to storing/restoring from temporary memory location when register allocator runs out of registers.

dbabokin (commented):
Does this PR supersedes #2809?

src/opt.cpp
```diff
@@ -731,6 +731,18 @@ void ispc::Optimize(llvm::Module *module, int optLevel) {
 
         optPM.addFunctionPass(PeepholePass());
         optPM.addFunctionPass(llvm::ADCEPass());
+        optPM.addFunctionPass(ReplaceHalfMaskedMemOpsPass());
```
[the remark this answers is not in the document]
turinevgeny: I agree, it should be an earlier pass, to open more optimization doors.

src/opt/ReplaceHalfMaskedMemOps.cpp
```diff
@@ -0,0 +1,216 @@
...
+#include <llvm/IR/IRBuilder.h>
+
+namespace ispc {
+
+bool lIsSecondHalfAllFalse(llvm::Value *mask) {
+    if (auto *CV = llvm::dyn_cast<llvm::ConstantVector>(mask)) {
+        auto N = CV->getType()->getNumElements();
+        for (auto i = N / 2; i < N; i++) {
+            llvm::Constant *E = CV->getAggregateElement(i);
+            if (!E || !llvm::isa<llvm::ConstantInt>(E) || !llvm::cast<llvm::ConstantInt>(E)->isZero()) {
```
turinevgeny, line 18: Is it possible to use `llvm::all_of` in this function?

src/opt/ReplaceHalfMaskedMemOps.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+    // Use shufflevector instruction to create the new vector
+    llvm::Twine name = originalVector->getName() + ".half";
+    return B.CreateShuffleVector(originalVector, llvm::UndefValue::get(vecType), mask, name);
+}
+
+llvm::Value *lBitcastPointerType(llvm::IRBuilder<> &B, llvm::Value *ptr, llvm::Value *value) {
+    auto *vecType = llvm::cast<llvm::VectorType>(value->getType());
+    auto *newPtrType = llvm::PointerType::get(vecType, 0 /* TODO! */);
+    // TODO! opaque pointer is no-op here, any special handling?
```
turinevgeny, line 75: Does it make sense to check if the pointer is opaque and skip the bitcast in this case?

src/opt/ReplaceHalfMaskedMemOps.cpp
```diff
@@ -0,0 +1,216 @@
...
+    return B.CreateBitCast(ptr, newPtrType);
+}
+
+llvm::Constant *lShrinkConstVec(llvm::LLVMContext &context, llvm::Value *originalValue) {
```
turinevgeny, line 79: It would clearer if the method accepted `llvm::Constant*` argument.

src/opt/ReplaceHalfMaskedMemOps.cpp
```diff
@@ -0,0 +1,216 @@
...
+        }
+    }
+
+    for (auto CI : storesToReplace) {
+        lReplaceMaskedStore(builder, CI);
+    }
+
+    for (auto CI : loadsToReplace) {
+        lReplaceMaskedLoad(builder, CI);
+    }
```
turinevgeny, line 209: Should `llvm::PreservedAnalyses::all()` be returned in case both `storesToReplace` and `loadsToReplace` are empty?

src/opt/ReplaceHalfMaskedMemOps.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+    // Create a new constant vector with the first half of the elements
+    llvm::ArrayRef<llvm::Constant *> newValuesRef(newValues);
+    return llvm::ConstantVector::get(newValuesRef);
+}
+
+// This function replaces, e.g.,
+//
+// %ptr = bitcast %v8_uniform_FVector4f* %Result.i to <8 x float>*
```
turinevgeny, line 103: What would be code generation if the original `N = 4` and we cut it to `load <2 x float>`? We need to ensure it wouldn't be worse than just `masked.load/store.v4f32`.

src/opt/ReplaceHalfMaskedMemOps.cpp
```diff
@@ -0,0 +1,216 @@
...
+    return false;
+}
+
+llvm::Value *lMergeVectors(llvm::IRBuilder<> &B, llvm::Value *firstVector, llvm::Value *secondVector,
+                           llvm::Twine &name) {
+    auto *firstVecType = llvm::cast<llvm::VectorType>(firstVector->getType());
+    auto *secondVecType = llvm::cast<llvm::VectorType>(secondVector->getType());
+    unsigned N = firstVecType->getElementCount().getKnownMinValue();
+
+    assert(firstVecType == secondVecType);
```
aneshlya, line 41: I suggest using ispc's `Assert` defined in util.h

src/opt/ReplaceHalfMaskedMemOps.cpp
```diff
@@ -0,0 +1,216 @@
...
+            }
+
+            llvm::Function *CF = CI->getCalledFunction();
+            if (!(CF && CF->isIntrinsic())) {
+                continue;
+            }
+
+            if (CF->getIntrinsicID() == llvm::Intrinsic::masked_store) {
+                llvm::Value *mask = CI->getOperand(3);
+                if (lIsSecondHalfAllFalse(mask)) {
```
aneshlya, line 185: Here and in other similar places, please check that `mask != nullptr`.
I suggest running static analysis on this PR before merging.

src/opt/ReplaceHalfMaskedMemOps.cpp
```diff
@@ -0,0 +1,216 @@
...
+  SPDX-License-Identifier: BSD-3-Clause
+*/
+
+#include "ReplaceHalfMaskedMemOps.h"
+
+#include <llvm/IR/IRBuilder.h>
+
+namespace ispc {
+
+bool lIsSecondHalfAllFalse(llvm::Value *mask) {
```
aneshlya, line 13: The function's name suggests it only checks if the second half of the vector is zero, but the code also checks if the first half contains all ones. Either adjust the name and documentation to reflect this dual purpose or separate the functionality into two distinct functions, depending on your use case.

src/opt/ReplaceHalfMaskedMemOps.cpp
```diff
@@ -0,0 +1,216 @@
...
+            }
+        }
+        return true;
+    }
+    return false;
+}
+
+llvm::Value *lMergeVectors(llvm::IRBuilder<> &B, llvm::Value *firstVector, llvm::Value *secondVector,
+                           llvm::Twine &name) {
+    auto *firstVecType = llvm::cast<llvm::VectorType>(firstVector->getType());
```
aneshlya, line 37: Why do you use `llvm::cast` and not `llvm::dyn_cast`?

```
auto *firstVecType = llvm::dyn_cast<llvm::VectorType>(firstVector->getType());
auto *secondVecType = llvm::dyn_cast<llvm::VectorType>(secondVector->getType());
```

src/opt/ReplaceHalfMaskedMemOps.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+    assert(firstVecType == secondVecType);
+
+    // Create a vector of indices to merge two vectors
+    std::vector<llvm::Constant *> indices;
+    for (unsigned i = 0; i < N * 2; ++i) {
+        indices.push_back(llvm::ConstantInt::get(B.getInt32Ty(), i));
+    }
+    llvm::ArrayRef<llvm::Constant *> indicesRef(indices);
+    llvm::Constant *mask = llvm::ConstantVector::get(indicesRef);
```
aneshlya, line 49: Maybe it would be more efficient to create mask as
```
llvm::Constant *mask = llvm::ConstantDataVector::get(B.getContext(), indices);
```
where indices is `std::vector<unsigned> indices(N * 2);`

src/opt/ReplaceHalfMaskedMemOps.h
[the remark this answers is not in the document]
nurmukhametov: Done

nurmukhametov: Changed to "This transformation allows the backend to generate shorter vector memory operations and corresponding math operations avoiding extra spills of temporal values to memory"

src/opt/ReplaceHalfMaskedMemOps.cpp
nurmukhametov: Fixed

src/opt/ReplaceHalfMaskedMemOps.cpp
nurmukhametov: Done

src/opt/ReplaceHalfMaskedMemOps.cpp
nurmukhametov: I have changed that in the last revision completely. I have realized that we need to support more cases, e.g., when only 4 first elements are true. It is needed to generate code using narrow registers for targets with `x16`.

src/opt/ReplaceHalfMaskedMemOps.cpp
nurmukhametov: I am not quite sure why it is needed. Doesn't sane LLVM IR guarantee that mask is non-null? Moreover, nullptr is safe here because inside the next function call we just return `false` for it.

src/opt/ReplaceHalfMaskedMemOps.cpp
nurmukhametov: Done

src/opt/ReplaceHalfMaskedMemOps.cpp
nurmukhametov: I agree, fixed

src/opt/ReplaceHalfMaskedMemOps.cpp
nurmukhametov: As I understand, no bitcast is generated in such case. See `tests/lit-tests/2611.ll`

src/opt/ReplaceHalfMaskedMemOps.cpp
nurmukhametov: I am not quite sure how to use it for aggregates.

src/opt.cpp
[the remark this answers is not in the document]
nurmukhametov: I moved it earlier.

src/opt/ReplaceHalfMaskedMemOps.cpp
aneshlya: I still see mix of `llvm::dyn_cast` and `llvm::cast` in the code. It's not a problem if you're certain of the type you are casting to, and the program logic guarantees that the type is correct. If it's not, I suggest using `llvm::dyn_cast`.

aneshlya:
Please format lit-tests with `clang-format`.

nurmukhametov:
> Please format lit-tests with `clang-format`.

Done

src/opt/ReplaceHalfMaskedMemOps.cpp
nurmukhametov: I have changed all `llvm::cast` to `llvm::dyn_cast`

src/opt/ReplaceMaskedMemOps.cpp
```diff
@@ -0,0 +1,253 @@
...
+#include "ReplaceMaskedMemOps.h"
+
+#include <llvm/IR/IRBuilder.h>
+#include <llvm/Transforms/Utils/Local.h>
+
+#include <numeric>
+
+namespace ispc {
+
+bool lIsPowerOf2(unsigned n) { return (n > 0) && !(n & (n - 1)); }
```
turinevgeny, line 16: There are such functions in LLVM, but it would need additional headers and libs to link.

turinevgeny (approved):
LGTM!

Issue #2813, opened 2024-03-26
Several lit tests fail when ISPC is built with LLLVM 18.1


aneshlya:
ispc :: aossoa_alignment.ispc
ispc :: assume_uniform.ispc
ispc :: interleaved.ispc

PR #2837, opened 2024-04-19
ImproveMemOps: consider zext in lOffsets32BitSafe

nurmukhametov:
After LLVM change [d77067d08](https://github.com/llvm/llvm-project/commit/d77067d08), InstCombine substitutes 
```llvm-ir
%v2 = sext <4 x i32> %v1 to <4 x i64>
```
with
```llvm-ir
%v2 = zext nneg <4 x i32> %v1 to <4 x i64>
```

`lOffsets32BitSafe` should consider `zext` too to be able to replace `__pseudo_scatter64_i32` with `__pseudo_scatter_factored_base_offsets32_i32`. Otherwise, it replaces it with `__pseudo_scatter_factored_base_offsets32_i32` that somehow results in generation of `vunpckhps`.

This fixes https://github.com/ispc/ispc/blob/main/tests/lit-tests/interleaved.ispc (issue #2813) with LLVM 18.1.

### DIFF
--- a/src/opt/ImproveMemoryOps.cpp
+++ b/src/opt/ImproveMemoryOps.cpp
@@ -569,9 +569,13 @@ static bool lOffsets32BitSafe(llvm::Value **variableOffsetPtr, llvm::Value **con
 
     if (variableOffset->getType() != LLVMTypes::Int32VectorType) {
         llvm::SExtInst *sext = llvm::dyn_cast<llvm::SExtInst>(variableOffset);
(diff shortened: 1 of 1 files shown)


turinevgeny (approved):

dbabokin (approved):
Looks good.

I'm a bit concerned that `nneg` is not checked for `zext`, as strictly speaking `zext` on 32 bit value is not equivalent to just 32 bit value - i.e. arithmetics will apply differently to negative values. But it seems for our purposes it's more or less safe.

aneshlya (approved):

PR #2838, opened 2024-04-22
 Add InferAlignment pass invocation

nurmukhametov:
This PR fixes failed tests with LLVM 18.1 from #2813 (`aossoa_alignment.ispc` and `assume_uniform.ispc`).

It also fixes `ISPC_LLVM_18_1` macro.


dbabokin (commented):
Alignment is a subtle matter, which sometimes difficult to verify. I suggest recomnputing it at least on the first and the last invocations of InstCombine. So it's available for the optimizer, not just for codegen.

nurmukhametov:
> Alignment is a subtle matter, which sometimes difficult to verify. I suggest recomnputing it at least on the first and the last invocations of InstCombine. So it's available for the optimizer, not just for codegen.

Done.

dbabokin (approved):
LGTM

aneshlya (approved):

Issue #2847, opened 2024-05-10
Report the absence of ncurses library on CMake stage


nurmukhametov:
Consider the scenario when ISPC is built with `ISPC_STATIC_LINK=ON`. It happens, e.g., with `ISPC_PREPARE_PACKAGE=ON`. If `tinfo.a` and `curses.a` are missing then it is reported only on linking stage. See here https://github.com/ispc/ispc/blob/3f22e4fde21b288c1984e59143ce1a2f55610212/CMakeLists.txt#L749

It is better to report that immediately on configuration stage.

nurmukhametov:
This dependency is to be removed soon as discussed here #2855

PR #2848, opened 2024-05-10
Miscellaneous fixes

nurmukhametov:
This fixes build-related issue https://github.com/ispc/ispc/issues/2847

It fixes `ispc-opt` to check that provided pass name is correct.


CMakeLists.txt
```diff
@@ -746,20 +746,15 @@ if (WIN32)
...
-            # this contains the tinfo library, if found
-            list(APPEND LINK_LIBRARIES ${CURSES_EXTRA_LIBRARY})
-        endif()
-        # Recommended way to get Curses library is to use CURSES_LIBRARIES variable
-        # https://cmake.org/cmake/help/v3.27/module/FindCurses.html
-        # but it contains both libform.so and libcurses.so.
-        # We need libcurses.so only, so exclude libform.so from the list.
-        list(FILTER CURSES_LIBRARIES EXCLUDE REGEX ".*libform.*")
-        list(APPEND LINK_LIBRARIES ${CURSES_LIBRARIES})
+        find_library(CURSES_LIB REQUIRED NAMES curses)
```
aneshlya, line 755: This change is not equivalent to what was before. What about tinfo library in case of dynamic linking? 
Even if it works in our environment I recommend double check it. There was an issue related to that.

nurmukhametov: Right. I have changed this to follow the LLVM logic [here](https://github.com/llvm/llvm-project/blob/main/llvm/cmake/modules/FindTerminfo.cmake)

aneshlya: Have the set of link libraries changed for release package or stayed the same as it was?

nurmukhametov: The release package used and uses the path under `ISPC_STATIC_LINK`. So, the `find_package(Curses REQUIRED)` and the following logic has never used in the release build. The set of linked libraries hasn't changed.

dbabokin (commented):
What libs do we need to be able to print output in terminal in different colors and detect terminal width (`--no-wrap` disables that)?

nurmukhametov:
> What libs do we need to be able to print output in terminal in different colors and detect terminal width (`--no-wrap` disables that)?

ISPC uses none libraries to do that. See here https://github.com/ispc/ispc/blob/0f2b9facbd1ee42ef1f75e3cf33e9246bcc36737/src/util.cpp#L68 and here https://github.com/ispc/ispc/blob/0f2b9facbd1ee42ef1f75e3cf33e9246bcc36737/src/util.cpp#L87

Given that it looks like the ncurses dependency is obsolete (https://github.com/ispc/ispc/issues/2855).

dbabokin:
> Given that it looks like the ncurses dependency is obsolete (#2855).

Thanks for explaining this. If we can drop this dependency, we should probably do that. But we need to make sure that output works as it used to be.

dbabokin (commented):
Why removing tinfo only for LLVM 18.0+? It makes the logic more complex - i.e it's ISPC version * LLVM version, rather than just ISPC version.

nurmukhametov:
> Why removing tinfo only for LLVM 18.0+? It makes the logic more complex - i.e it's ISPC version * LLVM version, rather than just ISPC version.

Because I don't want to rebuild the previous versions.

dbabokin:
> Because I don't want to rebuild the previous versions.

Up to you, but I vote for removing it once and for all versions - so we get better testing, less problems explaining different requirements when others are building ISPC and simpler code.

nurmukhametov:
At the moment, I have dropped curses/terminfo related changes from ISPC CMake here. 

I will proceed further in the following order:
1. Merge it with TERMINFO disabled for LLVM builds.
2. Rebuild and update LLVM.
3. Create a separate PR to completely remove the curses/terminfo search from the ISPC CMake.

PR #2850, opened 2024-05-13
Add user-visible definitions of numeric limits

### LINKED ISSUE #2794: Add support for basic type min/max values
In C/C++ it's possible to use values such as INT_MAX or FLT_MAX by including a standard library header such as <limits.h> or <float.h>. It would be nice if there was some way for ISPC to offer those as well. Perhaps through predefined preprocessor symbols or a standard library header?

https://en.cppreference.com/w/c/types/limits

nurmukhametov:
This PR fixes #2794

### DIFF
--- a/src/module.cpp
+++ b/src/module.cpp
@@ -2891,6 +2891,26 @@ static void lSetPreprocessorOptions(const std::shared_ptr<clang::PreprocessorOpt
     opts->addMacroDef("ISPC");
     opts->addMacroDef("PI=3.1415926535");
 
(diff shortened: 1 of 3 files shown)


aneshlya (approved):
LGTM!

dbabokin (commented):
I'm late to the party, but should we mention availability of these macros in the documentation?

nurmukhametov:
> I'm late to the party, but should we mention availability of these macros in the documentation?

I mentioned them in docs in #2858

PR #2857, opened 2024-05-15
Fix path processing

### LINKED ISSUE #2856: FATAL_ERROR when directory name used as input filename
ISPC fails with FATAL_ERROR for the following command:

```bash
$ ispc -g ../
util.cpp:535: Assertion failed: "basenameStart[0] != '\0'".
***
***
/localdisk2/anurmukh/ispc/src/main.cpp(406): FATAL ERROR: Unhandled signal sent to process; terminating.
Aborted
```

nurmukhametov:
This PR fixes #2856

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1100,6 +1100,20 @@ int main(int Argc, char *Argv[]) {
         exit(1);
     }
 
+    if (strncmp(file, "-", 2)) {
+        // If the input is not stdin then check that the file exists and it is
+        // not a directory.
+        if (!llvm::sys::fs::exists(file)) {
+            Error(SourcePos(), "File \"%s\" does not exist.", file);
+            exit(1);
+        }
+
+        if (llvm::sys::fs::is_directory(file)) {
+            Error(SourcePos(), "File \"%s\" is a directory.", file);
+            exit(1);
+        }
+    }
+
     // If [no]discard-value-names is explicitly specified, then use this value.
     // Otherwise enable it only if the output is some form of bitcode.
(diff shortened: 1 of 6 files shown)


src/main.cpp
```diff
@@ -1100,6 +1100,20 @@ int main(int Argc, char *Argv[]) {
         exit(1);
     }
 
+    if (strcmp(file, "-")) {
```
dbabokin, line 1103: `strncmp`, otherwise it will not pass security review.

src/util.cpp
```diff
@@ -503,39 +505,23 @@ std::vector<std::string> ispc::MatchStrings(const std::string &str, const std::v
...
+    if (llvm::sys::path::is_absolute(relativeName)) {
+        // We may actually get not relative path but absolute, then just use it.
         fullPath = relativeName;
-    else {
-        fullPath = g->currentDirectory;
-        if (fullPath[fullPath.size() - 1] != '/')
-            fullPath.push_back('/');
-        fullPath += relativeName;
+    } else {
+        fullPath = currentDirectory;
```
dbabokin, line 516: Interesting, it used `g->currentDirectory` before. Is it something different?

nurmukhametov: Both use cases have `currentDirector == g->currentDirectory` so nothing changed except the removed inconsistency.

src/main.cpp
nurmukhametov: fixed

dbabokin (approved):
LGTM

PR #2872, opened 2024-05-30
Add lit-test for the dispatcher

nurmukhametov:
This PR fixes fix `check_dispatch` test and add a new one `dispath.ispc`.

This also installs SDE for build jobs to run these tests.

### DIFF
--- a/.github/workflows/scripts/install-build-deps.sh
+++ b/.github/workflows/scripts/install-build-deps.sh
@@ -22,6 +22,10 @@ do
   fi
 done
 
+echo -U "$USER_AGENT" --retry-connrefused --waitretry=5 --read-timeout=20 --timeout=15 -t 5 https://downloadmirror.intel.com/"$SDE_MIRROR_ID"/"$SDE_TAR_NAME"-lin.tar.xz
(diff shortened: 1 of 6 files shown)


dbabokin (approved):
LGTM

.github/workflows/scripts/install-build-deps.sh
```diff
@@ -22,6 +22,10 @@ do
   fi
 done
 
+echo -U "$USER_AGENT" --retry-connrefused --waitretry=5 --read-timeout=20 --timeout=15 -t 5 https://downloadmirror.intel.com/"$SDE_MIRROR_ID"/"$SDE_TAR_NAME"-lin.tar.xz
```
dbabokin, line 25: You need to defined `USER_AGENT` in the build job to make it work. See here:
https://github.com/ispc/ispc/blob/8ec6a69016fc10f62827fe0b275005a07f1d1ed0/.github/workflows/ispc-ci.yml#L27

nurmukhametov: Right!

PR #2878, opened 2024-06-03
Fix devcontainer build deps

nurmukhametov:
It also updates the base image up to `ubuntu22.04` to be able to run clang from ispc.dependencies.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -2,9 +2,9 @@
   "__comment1__": "Copyright (c) 2023-2024, Intel Corporation",
   "__comment2__": "SPDX-License-Identifier: BSD-3-Clause",
   "name": "ISPC developers environment (CPU only)",
-  "image": "mcr.microsoft.com/devcontainers/universal:2",
+  "image": "mcr.microsoft.com/devcontainers/base:ubuntu22.04",
   "features": {
-    "ghcr.io/devcontainers-contrib/features/apt-get-packages:1": { "packages" : "libc6-dev,libc6-dev-i386,flex,wget,libtbb-dev" },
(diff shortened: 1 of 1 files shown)


.devcontainer/devcontainer.json
```diff
@@ -2,9 +2,9 @@
   "__comment1__": "Copyright (c) 2023-2024, Intel Corporation",
   "__comment2__": "SPDX-License-Identifier: BSD-3-Clause",
   "name": "ISPC developers environment (CPU only)",
-  "image": "mcr.microsoft.com/devcontainers/universal:2",
+  "image": "mcr.microsoft.com/devcontainers/base:ubuntu22.04",
```
dbabokin, line 5: Why not `24.04`? It should be available already.

dbabokin (approved):

.devcontainer/devcontainer.json
nurmukhametov: Hm, it doesn't seem working for me. Although, there are ubuntu24.04 tags [here](https://mcr.microsoft.com/v2/devcontainers/base/tags/list), but not everywhere.

dbabokin: Here I have `dev-ubuntu24.04` kinda working: https://github.com/dbabokin/ispc/tree/devcontainer

But we have a problem on Ubuntu 24.04. We build clang artifact on Ubuntu 18.04, which has `libtinfo5`. Version 6 is available starting with Ubuntu 20.04. But on Ubuntu 24.04 only `libtinfo6` is available, version 5 is not available. So clang binary fails.

You can reproduce this by starting Codespace in my branch and running clang in the terminal.

So, probably staying on Ubuntu 22.04 for devcontainer is a good thing for now.

dbabokin (approved):
LGTM

.devcontainer/devcontainer.json
nurmukhametov: Ah, I was trying without `dev` prefix, i.e., `ubuntu24.04` and it doesn't work. 

`libtinfo` version is the one more reason to get rid of `TERMINFO` dependency.

PR #2883, opened 2024-06-10
Do not use vrsqrt14 in `rsqrt` stdlib function

### LINKED ISSUE #2882: rsqrt discrepancy between avx2 and avx512
`rsqrt` standard library function returns values that are not same for different `x86` ISAs.

For example, we have the following results for `avx2-i32x8` and `avx512skx-x16`.

```bash
avx2-i32x8: rsqrt(1309.0) = 0.027639
avx512skx-x16: rsqrt(1309.0) = 0.027640
```

It is the part of #2782.

nurmukhametov:
This fixes #2882


dbabokin (changes requested):
I don't think that it's the right thing to do and I elaborated that in this comment: https://github.com/ispc/ispc/issues/2882#issuecomment-2159321313

Issue #2867, opened 2024-05-28
avx512knl-* targets EOL


dbabokin:
LLVM is removing support for Xeon Phi in `19.0` (current trunk), so we have to remove/disable `avx512knl-*` targets for builds with LLVM 19.0. No need to remove it with older LLVMs.

For the reference, Intel Xeon Phi is a family of accelerators based on x86 cores and it was the first AVX512 hardware available. AVX512 flavor in KNL is slightly different from Xeons, but for majority of instructions it's the same (AVX512F). 

Here's the PR that remove the support in LLVM: https://github.com/llvm/llvm-project/pull/75580

dbabokin:
KNL targets need to be removed from 19.0 build - it started failing, as KNL intrinsics are not available. See example of failed run here: https://github.com/ispc/ispc/actions/runs/9675033610/job/26700429544

Failed tests:
```
Failed Tests (19):
  ispc :: 1323_2.ispc
  ispc :: 1408.ispc
  ispc :: 1469.ispc
  ispc :: 1470.ispc
  ispc :: 1608.ispc
  ispc :: 1609.ispc
  ispc :: 1963.ispc
  ispc :: 2134.ispc
  ispc :: check_dispatch.ispc
  ispc :: deps.ispc
  ispc :: disable_gather_scatter.ispc
  ispc :: dispatch.ispc
  ispc :: fp16_code_gen.ispc
  ispc :: gatherScaleConst.ispc
  ispc :: gather_scatter.ispc
  ispc :: interleaved.ispc
  ispc :: knl-deprecation.ispc
  ispc :: perf-warnings.ispc
  ispc :: targets_x86.ispc
```

dbabokin:
@aneshlya @nurmukhametov Let's remove KNL target for LLVM 19, otherwise we have too many lit code failures and LLVM 19 is not tested.

pbrubaker:
Bye, bye, Larrabee. ;)

nurmukhametov:
Resolved by #3187

PR #2885, opened 2024-06-12
Deprecate avx512knl-x16 target

nurmukhametov:
This is to address #2867. It just adds a warning when avx512knl-x16 target is used.


dbabokin (commented):
I assume we should remove it for the build with LLVM 19.0+, right?

dbabokin (approved):
LGTM

PR #2887, opened 2024-06-14
Remove unused CI jobs

nurmukhametov:
This PR removes unused CI jobs and related configuration files:

1. calcite configuration
2. Travis job
3. ~~`ispc-svml` it doesn't look any different from the jobs that we run in `ispc-ci`.~~


dbabokin (changes requested):
I don't really like removing svml job. It's useful. Iwould suggest updateing it to the working state and make it run weekly.

Otherwise SVML support will get broken again.

nurmukhametov:
I am ready to update it but explain how it is different from the jobs in `ispc-ci.yml` that use `svml`.

dbabokin (approved):
LGTM

aneshlya (approved):

PR #2889, opened 2024-06-14
Fix nightly builds

nurmukhametov:
Fix and re-enable nightly builds.

The change to them is needed after https://github.com/ispc/ispc/pull/2848


dbabokin (approved):
Looks good.

With new allowances for shared workers (4 CPUs) it's easier to build LLVM is one stage and that's good!

Issue #2880, opened 2024-06-05
extend CI with FreeBSD alike job


nurmukhametov:
To catch errors like these #2874 early, we can add FreeBSD alike jobs into CI:

1. FreeBSD
1. Linux build with https://libcxx.llvm.org/

See also https://github.com/ispc/ispc/pull/2875#pullrequestreview-2094463893

Issue #2877, opened 2024-06-03
Build clang GLIBCXX dep version


nurmukhametov:
Clang, we build in `ubuntu:18.04` docker, requires `libstdc++` with at least `GLIBCXX_3.4.30`. So, it is not working inside both ubuntu 18.04 (GLIBCXX*.21) and 20.04 (GLIBCXX*.25).

It is caused by `libstdc++-9-dev` package installed from `ppa` repo here: https://github.com/ispc/ispc/blob/26a85fcd4bc6aba1339dd771c06ec620b24688b6/docker/ubuntu/18.04/cpu_ispc_build/Dockerfile#L31

It also resulted in the broken devcontainer after updating to LLVM 17.

PR #2892, opened 2024-06-20
Use libc++ to build LLVM/ISPC

nurmukhametov:
CI typically builds ISPC by statically linking Clang and LLVM libraries into ISPC binaries. Usually, Clang and LLVM libraries are built in a different environment than the subsequent ISPC builds. For example, currently, CI builds libraries in Ubuntu 18.04 but builds ISPC in Ubuntu 22.04. This means that the libraries are built with a different version of the C++ standard library than the ISPC code. We need to provide the same version alongside the compiler or use the same OS when building Clang and ISPC.

There are the following options to provide the same C++ standard library version:
1. Copy it from the Clang and LLVM build environment.
2. Build it from source and install it alongside the compiler, then use it to build Clang/LLVM libraries and ISPC.

The first option is tricky and probably impractical. The second option can be implemented using two different implementations of the C++ standard library: GCC's libstdc++ and LLVM's libc++.

This PR introduces the following options. Superbuild CMake has `USE_LLVM_LIBCXX` to build Clang and LLVM libraries with LLVM's libc++. ISPC CMake has the corresponding option `ISPC_USE_LLVM_LIBCXX` to link with that C++ library instead of the system one.

It also enables them both for Clang/LLVM and ISPC builds.

It relates to #2877 and #2880


superbuild/README.md
```diff
@@ -105,3 +105,26 @@ real world source codes complemented with scripts to build them. It allows us
 to collect profile meaningful from the customer point of view. On the last
 third stage, stage1 compiler uses that profile to optimize LLVM and ISPC code
 with PGO.
+
+# Usage of Built LLVM Libraries
```
dbabokin, line 109:
```suggestion
# Usage of Pre-built LLVM Libraries
```

superbuild/README.md
```diff
@@ -105,3 +105,26 @@ real world source codes complemented with scripts to build them. It allows us
...
+use the same OS when building Clang and ISPC.
+
+There are the following options to provide the same C++ standard library version:
+1. Copy it from the Clang and LLVM build environment.
+2. Build it from source and install it alongside the compiler, then use it to
+   build Clang/LLVM libraries and ISPC.
+
+The first option is tricky and probably impractical. The second option can be
+implemented using two different implementations of the C++ standard library:
+GCC's libstdc++ and LLVM's libc++.
```
dbabokin, line 126:
```suggestion
GCC's `libstdc++` and LLVM's `libc++`.
```

.github/workflows/compiler-warnings.yml
```diff
@@ -57,7 +59,11 @@ jobs:
 
     - name: Install dependencies
       run: |
-        .github/workflows/scripts/install-build-deps.sh
+        sudo apt update && \
+        sudo apt install -y \
```
dbabokin, line 63: So, you are switching to system packages instead of our own LLVM. Sound ok for checking compiler compiler warnings. But probably I'm missing the real reason for doing that.

.github/workflows/compiler-warnings.yml
```diff
@@ -57,7 +59,11 @@ jobs:
 
     - name: Install dependencies
       run: |
-        .github/workflows/scripts/install-build-deps.sh
+        sudo apt update && \
```
dbabokin, line 62: By not using the script, you are missing fail-retry mechanism for packages installation. It's there for a reason - sometimes Github share workers connectivity degrades and jobs start failing really frequently without fail-retry. I recommend keeping that.

.github/workflows/compiler-warnings.yml
```diff
@@ -57,7 +59,11 @@ jobs:
 
     - name: Install dependencies
       run: |
-        .github/workflows/scripts/install-build-deps.sh
```
dbabokin, line 60: Looks like with `gcc` package and `ubuntu-latest`, it's not gcc-11, but gcc-13. Not a big deal, but the job name might be missleading.

src/util.cpp
```diff
@@ -42,7 +42,7 @@
 #include <llvm/Support/FileSystem.h>
 #include <llvm/Support/Path.h>
 
-#ifdef ISPC_HOST_IS_APPLE
+#if defined(ISPC_HOST_IS_APPLE) || defined(ISPC_HOST_IS_LINUX)
```
dbabokin, line 45: Should it be the check for `libc++`? I.e. `#ifdef ( _LIBCPP_VERSION)`.

CMakeLists.txt
```diff
@@ -226,6 +226,10 @@ if (UNIX)
     else()
         option(ISPC_STATIC_LINK "Link statically (all except glibc)" OFF)
     endif()
+    if (ISPC_STATIC_LINK)
+        set(ISPC_STATIC_STDCXX_LINK ON)
```
dbabokin, line 230: I'm confused with the naming. `ISPC_STATIC_STDCXX_LINK` refers to `libstdc++`, while `ISPC_USE_LLVM_LIBCXX` refers specifically to `libc++`. As I understand they can be used together. That's confusing.

dbabokin (approved):
Overall looks good. But I'm a bit concerned how complicated the build process is. I'm not sure that we have way to avoid it, though.

nurmukhametov:
> Overall looks good. But I'm a bit concerned how complicated the build process is. I'm not sure that we have way to avoid it, though.

Yes indeed, it became more complicated than I anticipated. One possible approach is to build a proper toolchain starting from GCC like here [branch](https://github.com/ispc/ispc/tree/build-gcc) but free runners are not able to build it [link](https://github.com/ispc/ispc/actions/runs/9601521664).

I think, the easiest solution is to build Clang/LLVM and ISPC in same docker/runner environment, e.g., ubuntu-22.04 like I do on this [branch](https://github.com/ispc/ispc/tree/llvm-ispc-same-image).

So, I will reduce this PR to change related to CMake files only, just to have option to build LLVM/ISPC with `libc++` in future.

dbabokin:
> So, I will reduce this PR to change related to CMake files only, just to have option to build LLVM/ISPC with `libc++` in future.

I'm ok with current set of changes.

The idea of building in Ubuntu 18.04 is to enable users to download these binaries and use for ISPC dev build. But I don't think it's used much. All we use these binaries for is to run in CI. So we should be fine using any environment that works fine for CI.

Another story is release build. But we build full package from scratch anyway, so it's also not affected.

superbuild/README.md
nurmukhametov: Fixed

.github/workflows/compiler-warnings.yml
nurmukhametov: Not applicable anymore

.github/workflows/compiler-warnings.yml
nurmukhametov: Not applicable anymore

src/util.cpp
nurmukhametov: It is a good point. Fixed.

CMakeLists.txt
nurmukhametov: I understand that in the other way. I treated `ISPC_STATIC_STDCXX_LINK` as link statically any of `libstdc++`, `libc++`. Whereas, `ISPC_USE_LLVM_LIBCXX` is just link with `libc++` dynamically or statically.

But, the other problem that the difference between `ISPC_STATIC_LINK` and `ISPC_STATIC_STDCXX_LINK` will be very subtle after we merge #2895.

nurmukhametov: I have rebased on the current main and removed `ISPC_STATIC_STDCXX_LINK`.

dbabokin (changes requested):
I suggest closing the PR, but committing `src/util.cpp` changes. The reason for not committing the PR is that (1) it will not be used right now, and (2) it needs to be enabled for macOS.

PR #2893, opened 2024-06-21
Fixes for LLVM trunk

dbabokin:
- Fix lit test `simplifyCFG.ispc` lit test to tolerate new `range <>` return value attribute.
- Avoid using deprecated `StringRef::equal()` method.
- Fix a problem with debug info - it's being redesigned in LLVM right now, so stick to old version for now. This needs more investigation, as we'll need to use new representation eventually.
- Fix `alloy.py` LLVM build on macOS. We should probably switch to superbuild called from `alloy.py`, but it's not the scope of this changes.
- LLVM 18.1 to build `18.1.8`.

Debug info changes are long sanding project, which finally hit trunk. For example, see discussion [here](https://discourse.llvm.org/t/rfc-instruction-api-changes-needed-to-eliminate-debug-intrinsics-from-ir/68939).

With these changes ISPC build with LLVM `trunk` should be successful and all lit tests pass.


nurmukhametov (approved):
LGTM.

But I suggest to merge #2894 first to repair CI.

PR #2894, opened 2024-06-24
Use same image to build Clang/LLVM and ISPC

nurmukhametov:
Build Clang/LLVM in same ubuntu as ISPC to avoid issues with linking to different `libstdc++` versions. See #2892 and #2877 for context.


dbabokin (approved):
Looks good.

PR #2895, opened 2024-06-25
Finalize removing of ncurses/terminfo dependency

### LINKED ISSUE #2855: Consider removing ncurses dependency
It is not quite clear whether we actually use that via LLVM. We may build LLVM with `LLVM_ENABLE_TERMINFO=OFF`. After that, we don't need to provide ncurses, e.g., here: https://github.com/ispc/ispc/blob/2eb90ed85bdb3bf71ec8a39a8054265be140d343/docker/centos/7/cpu_ispc_build/Dockerfile#L74, as well as track it as third-party open-source dependency and etc..

nurmukhametov:
This fixes #2855.

This PR:
1. Remove installation of libncurses and libtinfo in docker/CI/devcontainers.
2. Remove CMake logic to find them.
3. Remove the license entry from `third-party-programs.txt`

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -4,7 +4,7 @@
   "name": "ISPC developers environment (CPU only)",
   "image": "mcr.microsoft.com/devcontainers/base:ubuntu22.04",
   "features": {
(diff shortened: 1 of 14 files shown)


aneshlya (approved):
Looks good!

dbabokin (approved):

PR #2897, opened 2024-06-26
Report error when var declarations have incompatible type specifiers

### LINKED ISSUE #2835: variables can be declared with unmasked keyword
This code compiles and does not throw an error, even though it should.

```c++
void Test(const unmasked uniform int32 a)
{
    unmasked float x;
    unmasked float y = 1;
}
```

nurmukhametov:
This PR fixes #2835.

It fixes the reporting logic for cases when unsuited qualifiers provided for variable declarations.


dbabokin (approved):
Looks good.

Issue #2905, opened 2024-07-09
Add support for float/double types in atomic functions


aneshlya:
LLVM has support for fadd, fsub, fmax and fmin operations for atomics.
For more context see [comment](https://github.com/ispc/ispc/issues/2904#issuecomment-2218995955).

PR #2907, opened 2024-07-12
Added support for float/double atomics (add, sub, min, max)

aneshlya:
Fixes: https://github.com/ispc/ispc/issues/2905
On Xe targets these atomic functions are not supported, so the error is generated.

The tests for the new functionality have been added in the same manner as the existing atomic tests. While the test coverage for atomics is not exhaustive, it seems acceptable because the implementation of atomics is consistent across preprocessor definitions. Similarly, the newly added tests do not cover 100% of the functionality.


docs/ispc.rst
```diff
@@ -5471,23 +5471,16 @@ There are also ``int64`` equivalents as well as variants that take
...
-  float atomic_max_local(uniform float * uniform ptr, float value)
-  float atomic_swap_local(uniform float * uniform ptr, float value)
-
-For global atomics, only atomic swap is available for these types:
-
-::
-
-  float atomic_swap_global(uniform float * uniform ptr, float value)
-  double atomic_swap_global(uniform double * uniform ptr, double value)
+  float atomic_add_{local,global}(uniform float * uniform ptr, float value)
```
nurmukhametov, line 5479: I slightly prefer having both explicitly stated. Otherwise, one might miss them when trying to find or grep for them.

nurmukhametov (approved):
LGTM!

docs/ispc.rst
aneshlya: I did it in the same manner as other atomics described https://github.com/ispc/ispc/pull/2907/files#diff-f014868f1ccd5256b5fa04823a766153c82b0eb98c0b3d9a2b4d154e2ee7275aR5464-R5471

dbabokin (approved):
LGTM.

PR #2908, opened 2024-07-16
KNL targets partial removal

aneshlya:
Remove KNL targets from lit-tests, full testing and trunk testing.
Addresses https://github.com/ispc/ispc/issues/2867#issuecomment-2228893891

I don't want to remove the target completely right now, it should go through deprecation. It's the right approach to communicate about changed targets to users. However (since it's deprecated) we can stop testing it.

### DIFF
--- a/.github/workflows/ispc-ci.yml
+++ b/.github/workflows/ispc-ci.yml
@@ -33,7 +33,6 @@ env:
                    "avx1-i32x4", "avx1-i32x8", "avx1-i32x16", "avx1-i64x4",
                    "avx2-i8x32", "avx2-i16x16", "avx2-i32x4", "avx2-i32x8", "avx2-i32x16", "avx2-i64x4",
                    "avx2vnni-i32x4", "avx2vnni-i32x8", "avx2vnni-i32x16",
-                   "avx512knl-x16",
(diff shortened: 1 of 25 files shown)


dbabokin (changes requested):
I would suggest testing it conditionally for LLVM < 19.0. We probably don't need all of the tests that we test right now, but some basic functionality would be good to test - i.e. that it's working in general. We will be shipping ISPC with LLVM < 19.0, so it's good to avoid shipping completely broken target.

Also, I assume making KNL conditional for LLVM < 19.0 builds should come in different PR. But it a trivial one-line change, so probably it's worth doing with one shot.

aneshlya:
> I would suggest testing it conditionally for LLVM < 19.0. We probably don't need all of the tests that we test right now, but some basic functionality would be good to test - i.e. that it's working in general. We will be shipping ISPC with LLVM < 19.0, so it's good to avoid shipping completely broken target.

There is a `tests/lit-tests/knl-deprecation.ispc` which is basically a sanity check that knl target is enabled and deprecated. The test is enabled for LLVM < 19.0

> Also, I assume making KNL conditional for LLVM < 19.0 builds should come in different PR. But it a trivial one-line change, so probably it's worth doing with one shot.

Done

dbabokin (approved):
Looks good!

PR #2912, opened 2024-07-18
ci: add coverity scans

aneshlya:
Our Coverity open source project is set up here: https://scan.coverity.com/projects/intel-ispc?tab=overview.
Open source Coverity has limitations on number of runs: https://scan.coverity.com/faq#frequency so I'm setting it to run by schedule or on workflow_dispatch.
Here is example run: https://github.com/aneshlya/ispc/actions/runs/9998651864

Feel free to contribute to this PR if you think something is missing.

### DIFF
--- a/.github/workflows/scripts/run-coverity.sh
+++ b/.github/workflows/scripts/run-coverity.sh
@@ -1,0 +1,97 @@
+#!/bin/bash -e
+
+# Environment check
+required_vars=(
+  "COVERITY_SCAN_PROJECT_NAME"
+  "COVERITY_SCAN_NOTIFICATION_EMAIL"
+  "COVERITY_SCAN_BUILD_COMMAND"
+  "COVERITY_SCAN_TOKEN"
+  "COVERITY_SCAN_BRANCH_PATTERN"
+)
+
+for var in "${required_vars[@]}"; do
+  if [ -z "${!var}" ]; then
+    echo "ERROR: $var must be set"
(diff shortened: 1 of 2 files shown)


.github/workflows/scripts/run-coverity.sh
```diff
@@ -0,0 +1,96 @@
...
+
+echo "Adding Coverity to PATH..."
+TOOL_DIR=$(find "$TOOL_BASE" -type d -name 'cov-analysis*')
+export PATH="$TOOL_DIR/bin:$PATH"
+
+# Clean up the build directory
+cd "$GITHUB_WORKSPACE"
+cmake --build build --target clean
+
+echo "Running Coverity analysis..."
```
nurmukhametov, line 70: In my opinion, this comment is slightly misleading. The analysis occurs on the remote Coverity server, which is why there are strict limits on the number of runs. Here, the ISPC build is captured along with the source code information.

nurmukhametov (approved):
LGTM!

dbabokin:
Should we add the condition to the workflow that it's the main repo, but not the fork?

.github/workflows/scripts/run-coverity.sh
aneshlya: Fixed

aneshlya:
> Should we add the condition to the workflow that it's the main repo, but not the fork?

Added check

PR #2921, opened 2024-07-26
Fix the return type of lanemask in the doc

zephyr111:
Hello,

I found that the documentation is not up to date about the doc about the `lanemask` intrinsic. Here is a (minor) quick fix.


aneshlya (approved):
Thanks for this fix!

Issue #2922, opened 2024-07-27
Need to add support for LLVM 20.0 (trunk) and LLVM 19.0 as a release version.


dbabokin:
LLVM has branched version 19 and current trunk identifies itself as 20.0. So keep building with LLVM trunk ISPC need to recognize this version and build with it.

Also new LLVM rebuild workflow needs to be added for LLVM 19 with a tag `llvmorg-19.1.0-rc1`, and 19.1 definition needs to be added to `alloy.py` and `superbuild/CMakeLists.txt`.

PS. This issue is created to test "Copilot Workspace".

PR #2923, opened 2024-07-29
Add support for LLVM 20.0 (trunk) and LLVM 19.1 tag.

dbabokin:
Related to #2922

Add support for LLVM 20.0 and LLVM 19.1.

- **alloy.py**:
  - Add recognition for LLVM 19.1 in `checkout_LLVM` function with `llvmorg-19.1.0-rc1` tag.
  - Update the list of supported LLVM versions in the `only` option.

- **superbuild/CMakeLists.txt**:
  - Add definition for LLVM version 19.1 with the tag `llvmorg-19.1.0-rc1`.

- **.github/workflows/rebuild-llvm19.yml**:
  - Add new workflow for rebuilding LLVM 19.1.

- **src/ispc_version.h**:
  - Add support for LLVM 20.0.
  - Update the range of supported LLVM versions to include 19.1 and 20.0.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/ispc/ispc/issues/2922?shareId=2919ca6b-6a3e-4031-8d69-93b02b8c1cdb).


src/ispc_version.h
```diff
@@ -20,9 +20,10 @@
 #define ISPC_LLVM_17_0 170000
 #define ISPC_LLVM_18_1 180100
 #define ISPC_LLVM_19_0 190000
```
aneshlya, line 22: Should the version be `ISPC_LLVM_19_1 190100`?

dbabokin: I was going with it forth and back. And I think no.

For `18_1` we have the only usage case `>= 18_1`, which is ok, since `18_0` doesn't exist anymore. But it's really `>= 18_0`.

For `19_0` we have the same usages: `>= 19_0`, which looks right. So I decided not to change it and not to introduce `19_1`.

aneshlya (approved):

Issue #2832, opened 2024-04-09
why this shuffle is inefficient?


MkazemAkhgary:
following code

```c++
unmasked int32 Test(uniform int32* uniform data, int32 i, int32 j)
{
    int32 a;
    for(int32 _i = 0; _i <= 8; _i++)
    {
        if(i == _i) a = shuffle(data[programIndex + programCount * _i], j);
    }
    return a;
}
```
compiles to bunch of `vpinsrd` and `vpextrd` instead of using one `vpermd`.

```asm
Test___UM_un_3C_uni_3E_vyivyi:          # @Test___UM_un_3C_uni_3E_vyivyi
        push    rbp
        mov     rbp, rsp
        push    r14
        push    rbx
        and     rsp, -32
        sub     rsp, 32
        vpextrd eax, xmm1, 1
        vpextrd ecx, xmm1, 2
        vmovd   edx, xmm1
        vpextrd esi, xmm1, 3
        vextracti128    xmm1, ymm1, 1
        vpextrd r8d, xmm1, 1
        vpextrd r9d, xmm1, 2
        vmovd   r10d, xmm1
        vpextrd r11d, xmm1, 3
        vpxor   xmm4, xmm4, xmm4
        vpcmpeqd        ymm2, ymm2, ymm2
        xor     ebx, ebx
        vpbroadcastd    ymm3, dword ptr [rip + .LCPI1_0] # ymm3 = [7,7,7,7,7,7,7,7]
        and     edx, 7
        and     eax, 7
        and     ecx, 7
        and     esi, 7
        and     r10d, 7
        and     r8d, 7
        and     r9d, 7
        and     r11d, 7
        vpcmpeqd        ymm5, ymm5, ymm5
        jmp     .LBB1_1
.LBB1_2:                                #   in Loop: Header=BB1_1 Depth=1
        vpminud ymm6, ymm4, ymm3
        vpcmpeqd        ymm6, ymm4, ymm6
        vpsubd  ymm4, ymm4, ymm2
        vpand   ymm5, ymm6, ymm5
        vmovmskps       r14d, ymm5
        add     ebx, 32
        test    r14d, r14d
        je      .LBB1_3
.LBB1_1:                                # =>This Inner Loop Header: Depth=1
        vpcmpeqd        ymm6, ymm4, ymm0
        vpand   ymm6, ymm6, ymm5
        vmovmskps       r14d, ymm6
        test    r14d, r14d
        je      .LBB1_2
[...]

PR #2924, opened 2024-07-29
Target-specific implementations of shuffle operations

### LINKED ISSUE #2770: improve shuffle(T, T, int) implementation.
The second variant of the **shuffle** function, which permutes over two values, generates un-optimized code when the permutation is not static.

Here is an example:

```c++
unmasked varying int32 Shuffle(varying int32 a, varying int32 b, varying int32 perm)
{
    return shuffle(a, b, perm);
}
```

```asm
Shuffle___UM_vyivyivyi:                 # @Shuffle___UM_vyivyivyi
        push    rbp
        mov     rbp, rsp
        and     rsp, -64
        sub     rsp, 128
        vmovaps ymmword ptr [rsp + 32], ymm1
        vmovaps ymmword ptr [rsp], ymm0
        vmovd   eax, xmm2
        cdqe
        vpextrd ecx, xmm2, 1
        vpextrd edx, xmm2, 2
        movsxd  rcx, ecx
        movsxd  rdx, edx
        vpextrd esi, xmm2, 3
        vextracti128    xmm0, ymm2, 1
        vmovd   edi, xmm0
        vpextrd r8d, xmm0, 1
        movsxd  rdi, edi
        movsxd  r8, r8d
        vpextrd r9d, xmm0, 2
        movsxd  r9, r9d
        vpextrd r10d, xmm0, 3
        vmovd   xmm0, dword ptr [rsp + 4*rdi]   # xmm0 = mem[0],zero,zero,zero
        vpinsrd xmm0, xmm0, dword ptr [rsp + 4*r8], 1
        vpinsrd xmm0, xmm0, dword ptr [rsp + 4*r9], 2
        movsxd  rdi, r10d
        vpinsrd xmm0, xmm0, dword ptr [rsp + 4*rdi], 3
        vmovd   xmm1, dword ptr [rsp + 4*rax]   # xmm1 = mem[0],zero,zero,zero
        vpinsrd xmm1, xmm1, dword ptr [rsp + 4*rcx], 1
        vpinsrd xmm1, xmm1, dword ptr [rsp + 4*rdx], 2
        movsxd  rax, esi
        vpinsrd xmm1, xmm1, dword ptr [rsp + 4*rax], 3
        vinserti128     ymm0, ymm1, xmm0, 1
        mov     rsp, rbp
        pop     rbp
        ret
```

_The generated code is even worse for AVX512 targets._

---

Proposed implementation

```c++
unmasked varying int32 Shuffle(varying int32 a, varying int32 b, varying int32 perm)
{
    varying int32 permA = shuffle(a, perm);
    varying int32 permB = shuffle(b, perm - programCount);
    return select(perm < programCount, permA, permB);
}
```

```asm
.LCPI1_0:
        .long   4294967288                      # 0xfffffff8
.LCPI1_1:
        .long   8                               # 0x8
Shuffle___UM_vyivyivyi:                 # @Shuffle___UM_vyivyivyi
        vpermd  ymm0, ymm2, ymm0
        vpbroadcastd    ymm3, dword ptr [rip + .LCPI1_0] # ymm3 = [4294967288,4294967288,4294967288,4294967288,4294967288,4294967288,4294967288,4294967288]
        vpaddd  ymm3, ymm2, ymm3
        vpermd  ymm1, ymm3, ymm1
        vpbroadcastd    ymm3, dword ptr [rip + .LCPI1_1] # ymm3 = [8,8,8,8,8,8,8,8]
        vpcmpgtd        ymm2, ymm3, ymm2
        vblendvps       ymm0, ymm1, ymm0, ymm2
        ret
```

---

Remarks:
- If the permutation is static (such as `perm = {1, 3, 8, 10, 6, 14, 7, 15}`), the current implementation already generates optimized code.
- proposed implementation is only good for int32/float types on AVX2/AVX512 targets. (otherwise generated code is worse)
- If the permutation is out of range (`perm < 0 || perm >= 2*programCount`), the behavior for the current implementation is undefined, while the behavior for the proposed implementation is equivalent to the following function (although not very meaningful):
  
```c++
unmasked varying int32 Shuffle(varying int32 a, varying int32 b, varying int32 perm)
{
    varying int32 permUnderflow = (programCount + (perm % programCount)) % programCount; // rotates around a
    varying int32 permOverflow = ((perm - programCount) % programCount) + programCount; // rotates around b
    return shuffle(a, b, perm < 0 ? permUnderflow : perm >= programCount * 2 ? permOverflow : perm);
}
```

Since the behavior is undefined for the current implementation with invalid permutations, I assume that the behavior change in this case should not matter.

aneshlya:
Fixes #2770, #2832

First of all. This implementation would be much cleaner after redesign targets and it's one additional example why we need it.

This PR introduces a more refined and target-specific implementation for shuffle operations. 

Previously, all shuffle-like operations (broadcast, rotate, shift, and shuffle) were generically implemented in util.m4 under a single `define_shuffles` macro.
This PR separates these operations into distinct macros, specifically targeting non-constant index shuffles.

- shuffle1(T): Corresponds to the standard library's shuffle(T, int) function.
- shuffle2_const(T): A helper function designed for shuffle2 operations with constant indices.
- shuffle2(T, T): Corresponds to the standard library's shuffle(T, T, int), utilizing shuffle2_const for non-constant index operations.

The code generated for constant indexes is already optimal (produces shufflevector after InstCombine) so I only add target-specific implementations for non constant part of `shuffle2` and `shuffle1`

**Target-specific changes**
- AVX-AVX2 provides `vpermilps` instruction (which operates within 128-bit lanes) and `vperm(ps|d)` (only on AVX2 256-bit).
- AVX-512 provides `vpermi2var` instructions that map almost directly to our shuffle2 functions signature. The only difference is that permutation vector should be the same size as data vectors (e.g vec of i8 for i8 type, i64 for double type etc). But since the behavior is undefined in case of permutation index > 2*programCount in ISPC, the casts from i32 to corresponding type should be fine. AVX512 also provides `vperm(ps|d)` as AVX2.

**Perf changes**
The results collected on `09_shuffles` micro benchmark and reported on 64 << 10 benchmark size.

|                 | avx512skx-x4 | avx512skx-x8 | avx512skx-x16 | avx2-i32x4,avx1-i64x4 | avx2-i32x8 |
|-----------------|--------------|--------------|---------------|------------|------------|
| shuffle1_int8   | N/A          | N/A          | -4.25%        | N/A        | N/A        |
| shuffle1_int16  | N/A          | -81.99%      | -87.15%       | N/A        | N/A        |
| shuffle1_int32  | -65.90%      | -83.66%      | -80.20%       | -67.11%    | -61.96%    |
| shuffle1_float  | -66.01%      | -84.05%      | -80.01%       | -65.86%    | -61.42%    |
| shuffle1_double | -69.01%      | N/A          | N/A           | N/A        | N/A        |
| shuffle1_int64  | -70.87%      | N/A          | N/A           | N/A        | N/A        |
|                 |              |              |               |            |            |
| shuffle2_int8   | N/A          | N/A          | N/A        | N/A        | N/A        |
| shuffle2_int16  | N/A          | -71.75%      | -82.96%       | N/A        | N/A        |
| shuffle2_int32  | -53.72%      | -69.84%      | -72.79%       | -49.00%    | -58.15%    |
| shuffle2_float  | -53.44%      | -70.15%      | -72.73%       | -50.09%    | -57.81%    |
| shuffle2_double | -22.61%      | N/A          | N/A           | N/A        | N/A        |
| shuffle2_int64  | -27.31%      | N/A          | N/A           | N/A        | N/A        |

No changes on other targets.

The results are obtained on 12th Gen Intel(R) Core(TM) i9-12900 (Alder Lake) for AVX-AVX2 targets and Intel(R) Xeon(R) Platinum 8360Y (Ice Lake) for AVX512.


builtins/util.m4
```diff
@@ -2014,7 +2014,94 @@ divert`'dnl
...
+  >
+forloop(i, 0, eval(WIDTH-1), `
+  %index_`'i = extractelement <WIDTH x i32> %2, i32 i')
+
+  %isc = call i1 @__is_compile_time_constant_varying_int32(<WIDTH x i32> %2)
+  br i1 %isc, label %is_const, label %not_const
+
+is_const:
+  ; extract from the requested lanes and insert into the result; LLVM turns
+  ; this into good code in the end
```
dbabokin, line 2083: What does "good code" mean here? Is it `shufflevector` LLVM instruction? If so, what pass does this? Is it InstCombine? Do we need a test that this happens in our optimization pipeline?

I see that you are not the author of this code, but still would be good to answer these questions.

dbabokin (commented):
I suggest adding benchmarks for shuffles, measuring on the platforms that we care about and posting it here in the PR.

Please also make sure that you run `alloy.py` / `run_test.py` testing for all targets to make sure that correctness is not compromised.

builtins/util.m4
aneshlya: I updated comment there and added tests. LLVM coverts it to `shufflevector` during InstCombine pass across all our targets except avx512*-x64 targets (I'm going to create an issue for that in LLVM).

aneshlya:
> I suggest adding benchmarks for shuffles, measuring on the platforms that we care about and posting it here in the PR.
> 
> Please also make sure that you run `alloy.py` / `run_test.py` testing for all targets to make sure that correctness is not compromised.

I tested on all targets: https://github.com/aneshlya/ispc/actions/runs/10171241038 and internal tests. Working on benchmarks.

builtins/util.m4
dbabokin: 🔥

benchmarks/01_trivial/09_shuffle.cpp
```diff
@@ -0,0 +1,132 @@
+#include "../common.h"
```
dbabokin, line 1: Copyright.

benchmarks/01_trivial/09_shuffle.ispc
```diff
@@ -0,0 +1,39 @@
+// For easier checking on the C++ side:
```
dbabokin, line 1: Copyright.

dbabokin (commented):
Benchmarks looks good. But please mention the specific HW that you tested on, as benchmarks are micro-arch sensitive. And you run it for `Range(64, 64 << 10)`, please specify for what input size you are giving the numbers.

benchmarks/01_trivial/09_shuffle.cpp
aneshlya: Added

benchmarks/01_trivial/09_shuffle.ispc
aneshlya: Added

nurmukhametov (approved):
It looks good to me. The only thing is bothered me why `shuffle2_int8` for `avx512skx-x16` is slower.

benchmarks/01_trivial/01_aossoa.cpp
```diff
@@ -1,3 +1,9 @@
+/*
```
dbabokin, line 1: NIT: Do we really so much vertical spacing? I would prefer to have it as compact as possible, i.e.:
```cpp
// Copyright (c) 2021-2024, Intel Corporation
// SPDX-License-Identifier: BSD-3-Clause
```

tests/lit-tests/const_shuffle-1.ispc
```diff
@@ -0,0 +1,101 @@
+// This test fails on avx512-x64 targets.
+
+// RUN: %{ispc} --target=avx512skx-x64 --nowrap -O2 --emit-llvm-text %s -o - | FileCheck %s
+
+// XFAIL: *
+
```
dbabokin, line 6: +`REQUIRES: X86_ENABLED`

tests/lit-tests/const_shuffle-1.ispc
```diff
@@ -0,0 +1,101 @@
+// This test fails on avx512-x64 targets.
```
dbabokin, line 1: Why does this fail? is it not-yet implemented?

aneshlya: An InstCombine bug (or not yet implemented for 64 vectors)

dbabokin: Let's mention this explicitly. If there's an issue, please put it here. So this XFAIL will be more actionable.

benchmarks/01_trivial/01_aossoa.cpp
aneshlya: I don't mind it. Changed to compact version.

tests/lit-tests/const_shuffle-1.ispc
aneshlya: Done

aneshlya:
> It looks good to me. The only thing is bothered me why `shuffle2_int8` for `avx512skx-x16` is slower.

It was slower because LLVM couldn't optimize GEPs after extracting the part related to non-constant indexes into a separate function, due to the presence of an alloca. I reworked it by extracting the logic related to the constant part instead. The new version is much cleaner and works identical to how it was before.

tests/lit-tests/const_shuffle-1.ispc
dbabokin: Do we have an issue filed on InstCombine?

aneshlya: Not yet

dbabokin (approved):

PR #2927, opened 2024-07-31
Support __attribute__ syntax.

nurmukhametov:
Supported syntaxes:
- `__attribute__((attr))`
- `__attribute__((attr(2)))`
- `__attribute__((attr("str")))`

They can be placed on variable declarations (local, global, parameters) and function declaration/definitions.

Placing them on struct declarations is not supported yet.

### DIFF
--- a/docs/ispc.rst
+++ b/docs/ispc.rst
@@ -133,6 +133,11 @@ Contents:
     * `Structure of Array Types`_
 
   + `Declarations and Initializers`_
+  + `Attributes`_
+
+    * `noescape`_
+    * `address_space`_
+
   + `Expressions`_
 
     * `Dynamic Memory Allocation`_
@@ -632,6 +637,12 @@ Updating ISPC Programs For Changes In ISPC 1.25.0
 
 ``avx512knl-x16`` target is deprecated and will be removed in future releases.
 
+ISPC language was extended to support ``__attribute__(())`` syntax for variable
+and function declarations. Two attributes are supported now: ``noescape`` and
+``address_space(N)``. The macro ``ISPC_ATTRIBUTE_SUPPORTED`` is defined if the
+ISPC compiler supports attribute syntax. Please, reger to `Attributes`_ for
+more details.
+
 Getting Started with ISPC
 =========================
 
@@ -1053,6 +1064,9 @@ preprocessor runs:
   * - ISPC_UINT_IS_DEFINED
     - 1
     - The macro is defined if uint8/uint16/uint32/uint64 types are defined in the ``ispc`` (it's defined in 1.13.0 and later)
+  * - ISPC_ATTRIBUTE_SUPPORTED
+    - 1
+    - The macro is defined if the ``ispc`` compiler supports ``__attribute__(())`` syntax.
   * - ISPC_FP16_SUPPORTED
     - 1
     - The macro is defined if float16 type is supported by the ``ispc`` target.
@@ -1764,7 +1778,7 @@ The following reserved words from C89 are also reserved in ``ispc``:
 ``foreach_unique``, ``in``, ``inline``, ``noinline``, ``__regcall``,
 ``__vectorcall``, ``int8``, ``int16``, ``int32``, ``int64``, ``launch``,
 ``new``, ``print``, ``uint8``, ``uint16``, ``uint32``, ``uint64``, ``soa``,
-``sync``, ``task``, ``true``, ``uniform``, and ``varying``.
+``__attribute__``, ``sync``, ``task``, ``true``, ``uniform``, and ``varying``.
 
 
 Lexical Structure
@@ -1929,7 +1943,7 @@ The following identifiers are reserved as language keywords: ``bool``,
 ``switch``, ``sync``, ``task``, ``template``, ``true``, ``typedef``,
 ``typename``, ``uint``, ``uint8``, ``uint16``, ``uint32``, ``uint64``,
 ``uniform``, ``union``, ``unsigned``, ``varying``, ``__regcall``,
-``__vectorcall``, ``void``, ``volatile``, ``while``.
+``__vectorcall``, ``__attribute__``, ``void``, ``volatile``, ``while``.
 
 ``ispc`` defines the following operators and punctuation:
 
@@ -2802,6 +2816,59 @@ them:
     ...
     const varying int x = { 1, 2, 3, 2 + 2 };
     const varying int y = x * 2;
+
+
+Attributes
+----------
+
+ISPC provides GNU style attributes syntax using ``__attribute__`` keyword.
+This section contains the list of currently supported attributes.
+
+noescape
+--------
+
+``__attribute__((noescape))`` can be placed on a function parameter of a
+uniform pointer type. It informs the compiler that the pointer cannot escape,
+i.e., no reference to the object the pointer points to, derived from the
+parameter value, will survive after the function returns. Applying this
+attribute to a varying pointer type is not supported.
+
+::
+
+    uniform int *uniform global_ptr;
+
+    void nonescaping(__attribute__((noescape)) uniform int * uniform ptr) {
+        // OK, because ptr doesn't escape the function
+        *ptr = 1;
+    }
+
+    void escaping(__attribute__((noescape)) uniform int * uniform ptr) {
+        // Not OK, because ptr escapes the function
+        uniform int *uniform global_ptr = ptr;
+    }
+
+
+address_space
+-------------
+
+``__attribute__((address_space(N)))`` is Xe specific attribute that can be
+applied to a pointer type or a reference type.  The value of this type points
+to or refers to the value allocated in the provided address space.  The address
+space is a non-negative integer value, and the default address space is 0.
+ISPC doesn't support allocation of data in address spaces other than the
+default one.  Placing this attribute to a varying pointer or reference type is
+not supported.
+
+::
+
+    uniform int example(__attribute__((address_space(1))) uniform int *uniform ptr) {
+        // ptr points to value in address space 1
+        return *ptr;
+    }
+
+    // allocation of data in non-default address space is not supported
+    __attribute__((address_space(1))) uniform int x;
+
 
 Expressions
 -----------
--- a/src/decl.cpp
+++ b/src/decl.cpp
@@ -23,6 +23,34 @@
 
 using namespace ispc;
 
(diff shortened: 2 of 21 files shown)


src/decl.cpp
```diff
@@ -141,6 +141,114 @@ static const Type *lApplyTypeQualifiers(int typeQualifiers, const Type *type, So
...
+void Attribute::Print() const {
+    printf("%s", name.c_str());
+    arg.Print();
+}
+
+AttributeList::AttributeList() {}
+
+AttributeList::AttributeList(const AttributeList &attrList) {
+    auto attrs = attrList.attributes;
+    for (unsigned int i = 0; i < attrs.size(); ++i) {
```
dbabokin, line 192: Why `unsigned int`?
Use range-based for instead: `for (auto i : attrs.size())`.

If you want to be an old school, use `std::size_t`, which is the type returned by `vector::size()`.

src/decl.cpp
```diff
@@ -141,6 +141,114 @@ static const Type *lApplyTypeQualifiers(int typeQualifiers, const Type *type, So
...
+    for (unsigned int i = 0; i < attrList.attributes.size(); ++i) {
+        Attribute *attr = attrList.attributes[i];
+        if (!HasAttribute(attr->name)) {
+            AddAttribute(*attr);
+        }
+    }
+}
+
+void AttributeList::CheckForUnknownAttributes(SourcePos pos) const {
+    for (size_t i = 0; i < attributes.size(); ++i) {
```
dbabokin, line 237: Hmm, here you use `size_t`. I suggest being consistent through the patch. My slight preference is range-based for, but `size_t` is also ok.

docs/ispc.rst
```diff
@@ -632,6 +632,20 @@ Updating ISPC Programs For Changes In ISPC 1.25.0
 
 ``avx512knl-x16`` target is deprecated and will be removed in future releases.
 
+ISPC language was extended to support ``__attribute__(())`` syntax for variable
+and function declarations. Two attributes are supported now: ``noescape`` and
+``address_space(N)``. The macro ``ISPC_ATTRIBUTE_SUPPORTED`` is defined if the
+ISPC compiler supports these attributes.
+
+``__attribute__((noescape))`` can be placed on a function parameter of a
```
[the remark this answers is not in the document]
aneshlya: Agree. I would also add a couple of code examples because they illustrate the feature better:
```
 void example1(__attribute__((noescape)) uniform int *ptr) {
     // ptr cannot escape the function
 }

 uniform int * example2(__attribute__((address_space(1))) uniform int *x) {
     // x is in address space 1
     return x;
 }
```

docs/ispc.rst
```diff
@@ -632,6 +632,20 @@ Updating ISPC Programs For Changes In ISPC 1.25.0
...
+uniform pointer type. It informs the compiler that the pointer cannot escape,
+i.e., no reference to the object the pointer points to, derived from the
+parameter value, will survive after the function returns.
+
+``__attribute__((address_space(N)))`` can be applied to a variable or parameter
```
aneshlya, line 645: Can these attributes be applied to function pointers?

I tried this and it doesn't compile:
```
#define ADDR_SPACE(N) __attribute__((address_space(N)))

void foo(int a) {
    print("foo: %d\n", a);
}

void bar(int a) {
    print("bar: %d\n", a);
}

typedef void (ADDR_SPACE(1) *func_ptr_t)(int);

void call_function(func_ptr_t func, int value) {
    func(value);
}

void test() {
    func_ptr_t fp1 = (func_ptr_t)foo;
    func_ptr_t fp2 = (func_ptr_t)bar;

    call_function(fp1, 42);
    call_function(fp2, 84);

    }
```

aneshlya (commented):
Should the attribute be always before type? For example should the test below compile?
```
int test(uniform int * uniform NOESCAPE a, int value)
```

docs/ispc.rst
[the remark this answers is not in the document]
nurmukhametov: I have introduced the separate `Attributes` section with description and examples.

src/decl.cpp
nurmukhametov: No specific reasons. Replaced with range-based syntax all loops.

docs/ispc.rst
nurmukhametov: This code doesn't seem to work in clang either https://godbolt.org/z/bYvE5vvPb

aneshlya: Here is fixed example: https://godbolt.org/z/5TqEjn9q3

nurmukhametov: > Should the attribute be always before type? 

No, it should not. Moreover, the meaning can be different as it is for `address_space`. Consider the following example: https://godbolt.org/z/YYqeqo3eP.

src/decl.h
```diff
@@ -55,6 +55,85 @@ class Declarator;
...
+class Attribute {
+  public:
+    Attribute(const std::string &n);
+    Attribute(const std::string &n, AttrArgument a);
+    Attribute(const Attribute &a);
+
+    /** Returns true if the attribute is known/supported false otherwise. */
+    bool IsKnownAttribute() const;
+
+    void Print() const;
```
dbabokin, line 93: No tests with `--ast-dump` were added / modified. Are the attributes printed / printed correctly with `--ast-dump`?

Do we need `void Print(&Indent)` version?

docs/ispc.rst
aneshlya: So I see that the following code is valid in C++ but ISPC produces an error
```
void foo(int a) {
    print("foo: %d\n", a);
}
typedef void (*func_ptr_t)(int);
ADDR_SPACE(1) func_ptr_t f1 = foo;

 Error: Can't convert between pointer types with different address spaces " void(varying int32 a) * uniform" and " void(varying int32 __anon_parameter_0) addrspace(1) * varying" for initializer.
ADDR_SPACE(1) func_ptr_t f1 = foo;
```
I think the error reporting for attributes with function pointers may be improved. Or we can prohibit the usage of it with function pointers and explicitly say about it in docs.

dbabokin (changes requested):
Generally looks good, but I still have a few question, see specific comments below.

I haven't played with examples, but I rely on Arina's review for that.

docs/ispc.rst
```diff
@@ -2803,6 +2817,51 @@ them:
...
+----------
+
+ISPC provides GNU style attributes syntax using ``__attribute__`` keyword.
+This section contains the list of currently supported attributes.
+
+noescape
+--------
+
+``__attribute__((noescape))`` can be placed on a function parameter of a
+uniform pointer type. It informs the compiler that the pointer cannot escape,
```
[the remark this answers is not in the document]
nurmukhametov: The problem here is that varying pointer is not a pointer in LLVM IR. It is just a vector of `int`, while `nocapture` is applicable only to pointer types.

docs/ispc.rst
```diff
@@ -2803,6 +2817,51 @@ them:
...
+    void escaping(__attribute__((noescape)) uniform int * uniform ptr) {
+        // Not OK, because ptr escapes the function
+        uniform int *uniform global_ptr = ptr;
+    }
+
+
+address_space
+-------------
+
+``__attribute__((address_space(N)))`` can be applied to a variable or parameter
```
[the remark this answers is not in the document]
nurmukhametov: I would like to know the answers to these questions as well. From my understanding of the current implementation, the type system supports `address_space` for pointer and reference types only with the semantics "points to something allocated in the provided address space".  Therefore, the user cannot define/declare a variable that is to be allocated in non-default address space.

The specific reason for me to support `address_space` attribute is to declare the following LLVM IR function
```llvm-ir
define void @__do_assert_uniform(i8 addrspace(2) *%str, i1 %test, <WIDTH x MASK> %mask)
```
in ISPC as:
```ispc
extern "C" void __do_assert_uniform(__attribute__((address_space(2))) uniform int8 * uniform str, MASK_TYPE mask);
```
The proper and thorough support of `address_space` in the language is beyond the scope of this PR.

docs/ispc.rst
nurmukhametov: But the error is valid because foo is not placed in `address_space = 1`. Consider that this is absolutely valid:
```ispc
ADDR_SPACE(0) func_ptr_t f1 = foo;
```

src/decl.h
nurmukhametov: I have added test `ast-dump-attrs.ispc` but I don't see any traces of `DeclSpecs` and `Declarator` in AST dumps.

docs/ispc.rst
aneshlya: Ah, got it. OK!

docs/ispc.rst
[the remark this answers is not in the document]
aneshlya: > Is it specifically for Xe backend?

It is. It was implicitly added to the type system to produce LLVM IR matching IGC backend contract (and only for that). There is no support of address_space in the language.

docs/ispc.rst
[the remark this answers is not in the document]
dbabokin: So we should probably say something for varying pointer - that it's not supported or will have no effect.

Most likely it will cause questions from the users, as they don't understand implementation details.

Though I'm now sure how to best document it.

docs/ispc.rst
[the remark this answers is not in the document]
dbabokin: Should we add a clause in the documentation that we don't support address spaces when allocating the data, but the pointers (and variables?) might have the address space and it's user's responsibility to handle it correctly?

By the way, is it only uniform pointers again or both uniform and varying?

src/decl.h
dbabokin: Looking at the test, I assume these attributes are printed as part of AST dump. If so, I'm good with it.

aneshlya (approved):
LGTM

docs/ispc.rst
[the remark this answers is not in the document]
nurmukhametov: I have added that it is not supported for varying pointers.

docs/ispc.rst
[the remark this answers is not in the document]
nurmukhametov: I addressed it in the `address_space` section of docs.

dbabokin (approved):
Changes look good.

But the branch is based on `cve-bin-tool` branch instead of `main`, so it contains unrelated changes. Please take care of it before merging.

PR #2929, opened 2024-08-01
Declare libcpp_verbose_abort on Linux also

nurmukhametov:
It is needed when we build with libc++.

Commit this patch as separate PR as requested here: https://github.com/ispc/ispc/pull/2892#pullrequestreview-2208334079


dbabokin (approved):
LGTM

Issue #2913, opened 2024-07-18
Automate security scans


aneshlya:
- [x] Coverity scans
- [x] Binary check: https://github.com/intel/cve-bin-tool
- [x] Trivy for containers
- [ ] Bandit scan for Python
- [x] Malware scanning (we can use ClamAV as part of ubuntu CI runner)

Later we can add 

- [x] Fuzzing for loops (Yarpgen)
- [ ] Fuzzing for different types of inputs: https://github.com/google/oss-fuzz)

llindsaya:
Would it be easier to use https://github.com/intel/cve-bin-tool-action?

PR #2935, opened 2024-08-06
CI: add cve-bin-tool jobs

nurmukhametov:
It adds a job to scan trunk and release archives with cve-bin-tool.

This partially addresses https://github.com/ispc/ispc/issues/2913.


aneshlya:
I'm not sure if it is ready for review but I see that you use `intel/cve-bin-tool-action` in one workflow and install cve-bin and perform a scan manually in another. 
Any approach works for me, but it should be unified.
Other than that looks fine.

nurmukhametov:
> Any approach works for me, but it should be unified. 

I have fixed it.

aneshlya (approved):
LGTM!

PR #2936, opened 2024-08-06
lexer: remove self-fuzzing functionality

nurmukhametov:
This PR removes `--fuzz-target` and `--fuzz-seed=<int>` developer functionality. It is not used anymore. It will eventually be replaced by fuzzing with external tools.


aneshlya (approved):
Looks good to me!

dbabokin (approved):

Issue #2236, opened 2022-01-20
Quantify math library maximum error


dbabokin:
ISPC supports 4 math libraries: `default`, `fast`, `svml`, `system`. To make a conscious decision between them user needs information about their precision on specific targets. The standard way to quantify precision is to have maximum error in [ulp](https://en.wikipedia.org/wiki/Unit_in_the_last_place). Also, if some of the implementations is returning NaN or Inf when it's supposed to return a number, this should be specifically noted.

I'm attaching the draft implementation of the utility that does error estimation for `rsqrt()`: [ulp_eval.tar.gz](https://github.com/ispc/ispc/files/7908918/ulp_eval.tar.gz)

RoyiAvital:
What about having a choice for the different levels of SVML?

dbabokin:
If you are talking about the `svml` option - i.e. when we use externally linked SVML implementations provided by Intel C/C++ compiler, then yes we can do that. But I would say that to give this feature a priority, I'd like to see that someone would like to use it, so I would justify spending my time implementing that. Do you have such use case? What precision are you interested in? AFAIK there 3 or 4 precision modes for SVML.

RoyiAvital:
Another question would be, why not use SVML always and that's it?
Isn't SVML open sourced now? Then use it and use its internal precision instead of rewriting them.

dbabokin:
SVML is not open sourced. Some parts of SVML were open sourced as contributions to glibc and LLVM project.

SVML is part of Intel C/C++ compiler, which is not shipped separately.

There are two drawbacks of using SVML library in ISPC - first, the project needs to be linked by Intel C/C++ compiler. Second, as implementation is available only on the link step, SVML functions are not inlined. These limitations might or might not be a problem for specific use cases, but they need to be taken into account.

As for precision / speed trade off, it's very reasonable to provide a set possible options. Intel C/C++ Compiler provides it as a choice of maximum relative error (in ULPs) required from library functions. ISPC now is much less informative about library implementation and this issue is exactly about providing more specific details about our library implementation.

RoyiAvital:
Well, Regarding SVML, since it Intel, and the compiler is free now, maybe you can do something within the organization to have SVML natively be supported in this project?

Another high quality alternative would be [Sleef](https://github.com/shibatch/sleef) by @shibatch.

dbabokin:
[...]

Issue #2920, opened 2024-07-26
Sub-optimal bool/masked operations


zephyr111:
Hello,

I found out that the code generated by ISPC is clearly sub-optimal when a (`varying`) condition scan variables that are of type `bool` as opposed to `uint8`. It seems the compiler want to check (only) the first bit of each boolean resulting in more instructions and a less efficient code (especially in AVX-512). 
AFAIK a `bool` value must always contains 0 or 1 values, and even if this would not be true (certainly UB -- undefined behaviour), then I expect the C behaviour to apply: zeros values are booleans set to `false` and non-zeros values are booleans set to `true` (which does not seems to applies here). If this is UB, then there is no need to perform checks.

Here is a simple example of ISPC masked mean targetting AVX-512:

```ispc
unmasked uniform float compute(uniform float v[], uniform bool m[], uniform int size)
{
    uniform const float NaN = floatbits(0x7FC00000); // Side note: is there no direct NaN in ISPC?
    float vsum = 0.0f;
    int vcount = 0;

    // For sake of clearity: generate a more readable code
    assume(size % programCount == 0);

    #pragma unroll(1) // For sake of clearity too
    foreach (i = 0 ... size)
    {
        const float tmp = v[i]; // must be here for sake of performance

        if (m[i]) // Fast for int8 but not for bool
        {
            vsum += tmp;
            vcount += 1;
        }
    }

    const uniform float sum = reduce_add(vsum);
    const uniform int count = reduce_add(vcount);
    return select(count > 0, sum / (uniform float)count, NaN);
}
```

The code of the generated function is the following (see on [GodBolt](https://godbolt.org/z/v7nTn9cjd)):

```lang-asm
.LBB1_3:
        vmovdqu xmm3, xmmword ptr [rsi + rcx]
        vpsllw  xmm4, xmm3, 7
        vpmovb2m        k1, xmm4
        vaddps  zmm0 {k1}, zmm0, zmmword ptr [rdi + 4*rcx]
        vpand   xmm3, xmm3, xmm2
        vpmovzxbd       zmm3, xmm3
        vpaddd  zmm1, zmm1, zmm3
        add     rcx, 16
        cmp     rcx, rax
        jb      .LBB1_3
```

And here it is if with replace the `bool` type by `int8` :

```lang-asm
.LBB1_3:                                # =>This Inner Loop Header: Depth=1
        vmovdqu xmm2, xmmword ptr [rsi + rcx]
[...]

PR #2937, opened 2024-08-07
Improve 32-bit log standard math functions

zephyr111:
Hello,

This PR fix few issues in the 32-bit implementation of `exp` and `log` (for the default ISPC math mode). Most of the issues have been mentioned [here](https://github.com/ispc/ispc/issues/2236#issuecomment-2267633771). More specifically, this PR:
 - Fix an overflow in the `exp` implementation previously resulting in bad results for `-Inf`, `NaN` and big numbers.
 - Fix a missing support for `NaN` and `Inf` in `log`.
 - Fix a numerical instability for input values close to 1 in `log` (more specifically values in the range 1-1.25). The results was initially very inaccurate and they are now about 1000 times more precise in this case. This was especially a serious issue when `pow` was called with values like `pow(1.001, 2000)` or `pow(1.01, 287.5)`.

Regarding the `exp`, I did not notice a significant performance impact.

Regarding the `log`, supporting NaN had a minor impact on performance. Surprisingly, checking infinity made the code nearly 50% slower for no apparent reason (just a `|| x_full == inf`) on the `avx2-i32x16` target, but no significant impact on the `avx2-i32x8` one. I guess this is certainly because of register spilling (due to the many constants), especially since the double-pumping target is now slower. The fix of the numerical instability also results in a significant impact on performance. In the end, all the modifications resulted in about 45% performance loss with the target `avx2-i32x8` and about 75% with the target `avx2-i32x16` on my (Zen2) laptop. This should only impact the default ISPC math version and not the fast one.  Thus, overall, the `log` function is significantly more reliable/accurate but this improvement is not free.

The precision can be improved further but I think this does not worth the additional performance overhead : 5-10% so to get about 30 ULPs precision on input numbers in the range 1.0-1.25 instead of about 120 ULPs in the PR and even 10_000-100_000 ULPs currently. Let me know if you want to increase the precision further.

### DIFF
--- a/stdlib.ispc
+++ b/stdlib.ispc
@@ -3464,49 +3464,53 @@ __declspec(safe) static inline float log(float x_full) {
         z = x_full + y;
         return z + 0.693359375 * fe;
     } else if (__math_lib == __math_lib_ispc) {
+
+        // See the uniform version for more information
+
         float reduced;
         int exponent;
 
-        const int NaN_bits = 0x7fc00000;
-        const int Neg_Inf_bits = 0xFF800000;
+        const int NaN_bits = 0x7FFFFFFF;
+        const int Inf_bits = 0x7F800000;
         const float NaN = floatbits(NaN_bits);
-        const float neg_inf = floatbits(Neg_Inf_bits);
-        bool use_nan = x_full < 0.;
-        bool use_inf = x_full == 0.;
-        bool exceptional = use_nan || use_inf;
+        const float inf = floatbits(Inf_bits);
+        const bool use_nan = !(x_full >= 0.);
+        const bool use_inf = x_full == 0. || x_full == inf;
+        const bool exceptional = use_nan || use_inf;
         const float one = 1.0;
 
-        float patched = exceptional ? one : x_full;
+        const float patched = select(exceptional, one, x_full);
         __range_reduce_log(patched, &reduced, &exponent);
 
-        const float ln2 = 0.693147182464599609375;
+        const float ln2 = 6.931471825e-01;
+        const float c1 = +4.999968200e-01;
+        const float c2 = +3.333395239e-01;
+        const float c3 = +2.506880353e-01;
+        const float c4 = +1.964300405e-01;
+        const float c5 = +1.429149029e-01;
+        const float c6 = +3.786718051e-01;
+        const float c7 = -5.571787647e-01;
+        const float c8 = +8.494745916e-01;
 
-        float x1 = one - reduced;
-        const float c1 = 0.50000095367431640625;
-        const float c2 = 0.33326041698455810546875;
-        const float c3 = 0.2519190013408660888671875;
-        const float c4 = 0.17541764676570892333984375;
-        const float c5 = 0.3424419462680816650390625;
-        const float c6 = -0.599632322788238525390625;
-        const float c7 = +1.98442304134368896484375;
-        const float c8 = -2.4899270534515380859375;
-        const float c9 = +1.7491014003753662109375;
+        const bool close_to_one = abs(x_full - one) < 0.12;
+        reduced = select(close_to_one, x_full, reduced);
+        const float log_from_exponent = select(close_to_one, 0.0, (float)exponent * ln2);
 
-        float result = x1 * c9 + c8;
-        result = x1 * result + c7;
-        result = x1 * result + c6;
-        result = x1 * result + c5;
-        result = x1 * result + c4;
-        result = x1 * result + c3;
-        result = x1 * result + c2;
-        result = x1 * result + c1;
-        result = x1 * result + one;
+        const float x = one - reduced;
+        float result = c8;
+        result = x * result + c7;
+        result = x * result + c6;
+        result = x * result + c5;
+        result = x * result + c4;
+        result = x * result + c3;
+        result = x * result + c2;
+        result = x * result + c1;
+        result = x * result + one;
 
-        // Equation was for -(ln(red)/(1-red))
-        result *= -x1;
-        result += (float)(exponent)*ln2;
+        result = log_from_exponent - x * result;
 
-        return exceptional ? (use_nan ? NaN : neg_inf) : result;
+        const float exceptional_result = select(use_nan, NaN, select(x_full == inf, x_full, -inf));
+        return select(exceptional, exceptional_result, result);
     }
 }
 
@@ -3546,49 +3550,70 @@ __declspec(safe) static inline uniform float log(uniform float x_full) {
         z = x_full + y;
         return z + 0.693359375 * fe;
     } else if (__math_lib == __math_lib_ispc) {
+
+        // Precision: <15 ULP (for normal numbers). The error can be much higher for subnormal numbers.
+        // Support special input values like NaN, -Inf, +Inf and subnormal numbers correctly.
+
         uniform float reduced;
         uniform int exponent;
 
-        const uniform int NaN_bits = 0x7fc00000;
-        const uniform int Neg_Inf_bits = 0xFF800000;
+        const uniform int NaN_bits = 0x7FFFFFFF;
+        const uniform int Inf_bits = 0x7F800000;
         const uniform float NaN = floatbits(NaN_bits);
-        const uniform float neg_inf = floatbits(Neg_Inf_bits);
-        uniform bool use_nan = x_full < 0.;
-        uniform bool use_inf = x_full == 0.;
-        uniform bool exceptional = use_nan || use_inf;
+        const uniform float inf = floatbits(Inf_bits);
+        const uniform bool use_nan = !(x_full >= 0.); // Check if x_full is either NaN or less than 0
+        const uniform bool use_inf = x_full == 0. || x_full == inf;
+        const uniform bool exceptional = use_nan || use_inf;
         const uniform float one = 1.0;
 
-        uniform float patched = exceptional ? one : x_full;
+        const uniform float patched = select(exceptional, one, x_full);
         __range_reduce_log(patched, &reduced, &exponent);
 
-        const uniform float ln2 = 0.693147182464599609375;
+        // Use a 8-degree polynomial to approximate the expression `-(ln(reduced)/(1-reduced))`.
+        // Then, we multiply the polynomial by `reduced-1` so the resulting value is guaranteed
+        // to approximate `ln(reduced)` pretty accurately when `reduced ~= 1`.
+        // This is critical for the relative error to be small since `ln(1) = 0`.
+        // Besides, note the Taylor expansion of `ln(reduced)` is `reduced - 1 + O(1)`.
+        // Coefficients can be found with `curve_fit` from `scipy.optimize` in Python.
+        // There is a catch though: `reduced` lies in the [0.5;1.0] interval and
+        // there is a wrap-around when `x_full` is just above 1.0.
+        // In this specific case, `reduced` is set to ~0.5 and the approximation is
+        // not guaranteed to be very accurate anymore.
+        // Moreover, a catastrophic cancellation happens because we need to add `ln(2**exponent)`.
+        // To avoid this problem, we extend the right bound of the wrap-around slightly above 1.
+        // As a result, the domain of definition for the polynomial is [0.50;1.12]
+        // instead of simply [0.5;1.0].
 
-        uniform float x1 = one - reduced;
-        const uniform float c1 = 0.50000095367431640625;
-        const uniform float c2 = 0.33326041698455810546875;
-        const uniform float c3 = 0.2519190013408660888671875;
-        const uniform float c4 = 0.17541764676570892333984375;
-        const uniform float c5 = 0.3424419462680816650390625;
-        const uniform float c6 = -0.599632322788238525390625;
-        const uniform float c7 = +1.98442304134368896484375;
-        const uniform float c8 = -2.4899270534515380859375;
-        const uniform float c9 = +1.7491014003753662109375;
+        const uniform float ln2 = 6.931471825e-01;
+        const uniform float c1 = +4.999968200e-01;
+        const uniform float c2 = +3.333395239e-01;
+        const uniform float c3 = +2.506880353e-01;
+        const uniform float c4 = +1.964300405e-01;
+        const uniform float c5 = +1.429149029e-01;
+        const uniform float c6 = +3.786718051e-01;
+        const uniform float c7 = -5.571787647e-01;
+        const uniform float c8 = +8.494745916e-01;
 
-        uniform float result = x1 * c9 + c8;
-        result = x1 * result + c7;
-        result = x1 * result + c6;
-        result = x1 * result + c5;
-        result = x1 * result + c4;
-        result = x1 * result + c3;
-        result = x1 * result + c2;
-        result = x1 * result + c1;
-        result = x1 * result + one;
+        const uniform bool close_to_one = abs(x_full - one) < 0.12;
+        reduced = select(close_to_one, x_full, reduced);
+        const uniform float log_from_exponent = select(close_to_one, 0.0, (uniform float)exponent * ln2);
 
-        // Equation was for -(ln(red)/(1-red))
-        result *= -x1;
-        result += (uniform float)(exponent)*ln2;
+        const uniform float x = one - reduced;
+        uniform float result = c8;
+        result = x * result + c7;
+        result = x * result + c6;
+        result = x * result + c5;
+        result = x * result + c4;
+        result = x * result + c3;
+        result = x * result + c2;
+        result = x * result + c1;
+        result = x * result + one;
 
-        return exceptional ? (use_nan ? NaN : neg_inf) : result;
+        // The equation was for -(ln(reduced)/(1-reduced))
+        result = log_from_exponent - x * result;
+
+        const uniform float exceptional_result = select(use_nan, NaN, select(x_full == inf, x_full, -inf));
+        return select(exceptional, exceptional_result, result);
     }
 }
 


zephyr111:
Please note I used the following code to test the `exp` and `log` functions:

```
void test_ext(uniform float x, uniform bool verbose)
{
    const uniform float res = log(x);
    const uniform float dres = (uniform float)log((uniform double)x);
    const uniform float inf = floatbits(0x7F800000);

    if(verbose)
        print("%: %\n", x, res);

    if(isnan(res) != isnan(dres))
        print("ERROR: NaN mismatch: x=%\n", x);
    else if(res != dres && (abs(res) == inf || abs(dres) == inf))
        print("ERROR: Inf mismatch: x=% res=% dres=%\n", x, res, dres);
    else if(res != dres && abs(res-dres)/dres >= 130e-7) // Usual error check: >= 4~5 ULP
        print("ERROR: value mismatch: x=% res=% dres=% error_x1000=%\n", x, res, dres, abs(res-dres)*1000/dres);

    if(verbose)
        print("\n");
}

void test(uniform float x)
{
    test_ext(x, true);
}

extern "C" uniform int main()
{
    uniform bool full_check = true;
    uniform bool measure_perf = true;

    uniform float pinf  = floatbits(0x7F800000);
    uniform float ninf  = floatbits(0xFF800000);
    uniform float pnan  = floatbits(0x7FC00000);
    uniform float nnan  = floatbits(0xFFC00000);
    uniform float pnan2 = floatbits(0x7FC00123);

    test(0.0f);
    test(-0.0f);
    test(1.0f);
    test(1.0000005f); // very imprecise for initial log
    test(1.0000010f);
    test(1.0000011f); // very imprecise for initial log
    test(1.0000012f);
    test(-1.0f);
    test(-10.0f);
    test(10.0f);
    test(1.15e+5);
    test(1.15e+36);
    test(1.15e+42);
    test(1.15e-36);
    test(pinf);
    test(ninf);
    test(pnan);
    test(nnan);
    test(pnan2);

    if(full_check)
    {
        uniform int64 maxi = ((uniform int64)1) << 32;
        for(uniform int64 i=0; i<maxi; ++i)
        {
            test_ext(floatbits((uniform int32)i), false);

            if(i % 100000000 == 0)
                print("%/100\n", i*100/maxi);
        }
    }

    if(measure_perf)
    {
        uniform int32 maxi = (((uniform int64)1) << 31) - 1; // skip negative numbers and few NaN to avoid overflows
        float checksum = 0.0;
        const uniform int64 startTime = clock();

        foreach(i=0...maxi)
            checksum += log(floatbits(i));

        const uniform int64 endTime = clock();
        print("checksum: %, clock: %e9\n", reduce_add(checksum), (uniform float)(endTime-startTime)*1e-9);
    }
}
```

It may be useful for another use-case (e.g. other math functions) or possibly for tests. It is quite convenient to check all possible float values (since tests based on random numbers are not really reliable).

nurmukhametov:
Thank you for you contribution!

I suggest splitting this PR into two because it addresses two different problems.

To make it look neat, I suggest formatting the code (see the output of [this](https://github.com/ispc/ispc/actions/runs/10278978811/job/28456359829?pr=2937) job) and formatting the git message as follows: the first line is a short description, e.g., `stdlib: fix overflow in exp(float)` then an empty line, followed by a detailed description of any length, formatted to a width of 80 characters.

aneshlya:
Could you wrap your test code into a trivial microbenchmark, similar to what we have here: [ISPC Benchmarks](https://github.com/ispc/ispc/tree/main/benchmarks/01_trivial)? This will allow us to retain and integrate your testing code effectively.

aneshlya:
Does your change anyhow utilizes Sleef algorithms?

dbabokin:
> Could you wrap your test code into a trivial microbenchmark, similar to what we have here: [ISPC Benchmarks](https://github.com/ispc/ispc/tree/main/benchmarks/01_trivial)? This will allow us to retain and integrate your testing code effectively.

We already have it for `pow`, `log`, and `exp`: https://github.com/ispc/ispc/blob/ff58297a64be598c5e1c3d3da36d4ae76044eed5/benchmarks/01_trivial/06_math.ispc#L88

stdlib.ispc
```diff
@@ -3256,7 +3256,7 @@ __declspec(safe) static inline float exp(float x_full) {
 
         float scaled = x_full * one_over_ln2;
         float k_real = floor(scaled);
-        int k = (int)k_real;
+        int k = (int)k_real; // Be careful: float to int convertions are UB if the float value is too huge
```
dbabokin, line 3259: I understand the comment in general, but I don't understand what useful information it brings. It has no call for action - i.e. if the overflow happens, then it happens. It seems there's nothing we can fix here, right? And I don't think we are going to tune this algorithm.

Please correct me if I get this wrong.

stdlib.ispc
```diff
@@ -3468,45 +3472,51 @@ __declspec(safe) static inline float log(float x_full) {
         int exponent;
 
         const int NaN_bits = 0x7fc00000;
-        const int Neg_Inf_bits = 0xFF800000;
```
dbabokin, line 3471: Where the log algorithm is coming from? Possibly we need to put a reference to the origin of the algorithm to make sure we are not breaking the license.

I know that algorithms are not copyrightable, only specific implementation is. But we'd like to understand the origin and give the credit.

dbabokin (commented):
Please let us know if you need help fixing formatting issues - we use `clang-format` version 12 for formatting. I can push the fix to your branch, if you'd like.

For "copyright check" failure - it's not related to your changes, just rebase to main to fix that.

dbabokin:
Also I measured the impact of the changes on Apple M3 chip with `neon-i32x4` target. `exp` is not affected. `log` is slower approximately 50%.

stdlib.ispc
zephyr111: This is my own modification from the initial ISPC algorithm. I fixed the special values from the initial algorithm and then generated a new lower-order polynomial (with my a Python script optimization algorithm I written myself) since a it is a bit faster and the precision is enough thanks to the correction close to log(1). Finally, I added a correction based on the Taylor series at log(1). I just checked the functions/errors with the great Geogebra tool. 
Thus, there is not breaking any license :) .

stdlib.ispc
zephyr111: I added this comment because it was the source of the initial bug : AFAIR we used the `k` variable in a case where it was UB because of the overflow. So I fixed that by using directly `k_real` instead and found useful to mention this UB so to warn future people about this problem (avoiding them to repeat the same mistake again). Maybe it is not needed. I can remove the comment if you think so.

zephyr111:
> Does your change anyhow utilizes Sleef algorithms?

No. This is my own modification.

> Could you wrap your test code into a trivial microbenchmark, similar to what we have here: [ISPC Benchmarks](https://github.com/ispc/ispc/tree/main/benchmarks/01_trivial)? This will allow us to retain and integrate your testing code effectively.

> We already have it for pow, log, and exp

Ok, so there is finally nothing to add in the microbenchmark, isn't it?

aneshlya:
> Ok, so there is finally nothing to add in the microbenchmark, isn't it?

Yep, no changes.

zephyr111:
I improved the speed of the `log` function. It is now "only" 25% slower than the initial code for the target `avx2-i32x8` and 55% for the target `avx2-i32x16`. This mitigate the strange issue related to the `Inf` (the assembly code seems minor but the performance impact is surprisingly huge). I am going to investigate if the `avx2-i32x16` target can be optimized further. I am not sure there is much more to do now regarding performance of the target `avx2-i32x8` (or Neon). I think 25% is the price to pay for the fixes.

stdlib.ispc
dbabokin: I don't think that we are going to hack this algorithm in the future. So this comment belongs more to the commit description, rather than to code itself. Let's remove it.

stdlib.ispc
dbabokin: That's impressive to say the least!

Two things then:
1. Let's put implementation note before the algorithm to explain how you come up with it and what it precision characteristics are there. This information definitely belongs to the code, as we might find bugs in it, consider improvement, etc. And understanding the background is very important in this case.
2. Have you "shopped around" for the alternatives? I.e. have you compared it's precision and performance with SVML and Sleef alternatives? It would be good to understand where it stands compared other proved implementations. If needed we can borrow algorithms from both libraries, it just needs to be done properly (license, copyright, etc).

zephyr111:
I tried to understand what was more precisely the numerical error so to use a more efficient approach for the correction close to x=1 and it worked well. The new approach is not only faster for the `avx2-i32x8` target but also significantly more accurate. I described the approach more precisely in comments of the submitted ISPC file.

In the end, the new code is only 8% slower than the initial one for the `avx2-i32x8` target and still 50% slower with the `avx2-i32x16` (due to register spilling and apparently LLVM optimization issues too). The new code is slightly less precise (only few ULP) for most values, but much more precise for values close  to x=1 : about 10-15 times better! Overall the new code has a relative precision of <10 ULP in practice (so 1_000-10_000 time more than the initial code).

I also fixed the formatting issues. 
I am going to certainly remove the `exp` fix from this PR and open a new one as proposed by @nurmukhametov .

stdlib.ispc
zephyr111: Thank you.

1. There was already a comment in the code about the precision (in ULPs). I added information about how the method works.
2. I never used Sleef yet, but it seems interesting/promising. I'm going to give it a try, but I think this can be done later, independently of this PR.

zephyr111:
It should be fine now and ready to be merged :) !

stdlib.ispc
zephyr111: I finally performed a quick check/benchmark with Sleef (the one available on Ubuntu 24.04 LTS system packages) since is relatively simple in practice (as long as we use a standard C interface and not LLVM IR one which is apparently not documented).
The C interface might be slightly slower since there is a function call. However, the Sleef functions are not masked so adding a masking introduces slight overhead. In the end, the two might cancel out so the performance may be representative of what we can get with direct a IR interface (if any).
Sleef provide 2 groups of functions: ones with 1 ULP error (typically due to rounding), and ones with a 3.5 ULP error (let say 4). The former can be significantly slower (i.e. up to 10 time slower) than the later and I think 4 ULP is a good reference for a project like ISPC where people focus on performance (and also because the current ISPC math log/exp function have a 4 to 10 ULP error with the new PRs). Not to mention the performance of 4 ULP functions are more stable. See the benchmark for more information: https://sleef.org/nontrigsp.png and https://sleef.org/trigsp.png . 
I wrote a  "corrected" implementation (not pushed) so to get a <5 ULP error (instead of <10 ULP) in order not to mix apples and oranges when comparing this with Sleef. It simply use a 9-degreee polynomial instead of a (current) 8-degree one.
Regarding performance, on my machine (Zen2), I get (in clocks, for the entire 32-bit float range):
```
This PR implementation (default math):                   2.11e9
Corrected PR implementation:                   2.26e9
Sleef avx2-i32x8 with C interface and no masking:   2.73e9
```
Thus, in this case, Sleef is about 30% slower than the less-precise PR implementation and about 20% slower than the similarly-precise corrected PR implementation :) !
A part of the overhead might come from the C call but probably not all of it.
It shows that this PR implementation is pretty good since it is competitive with Sleef (which is itself competitive with the SVML based on the provided Sleef's benchmarks).

Overall, aside from this specific `log` implementation, IMHO, Sleef is pretty interesting for ISPC since it has the benefit to provide: 
- (pretty good) bounded-precision and support for special numbers (eg. NaN, Inf, subnormals, etc.)
- challenging performance
- certainly more tested functions than ISPC ones
- (fast) double-precision functions
- a flexible license certainly compatible with the ISPC one
- an alternative implementation (useful when an implementation has issues)

nurmukhametov:
> In the end, the new code is only 8% slower than the initial one for the `avx2-i32x8` target and still 50% slower with the `avx2-i32x16` (due to register spilling and apparently LLVM optimization issues too).

It looks like if we move `exceptional_result` definition up like this:
```ispc
@@ -3477,6 +3477,7 @@ __declspec(safe) static inline float log(float x_full) {
         const bool use_nan = !(x_full >= 0.);
         const bool use_inf = x_full == 0. || x_full == inf;
         const bool exceptional = use_nan || use_inf;
+        const float exceptional_result = select(use_nan, NaN, select(x_full == inf, x_full, -inf));
         const float one = 1.0;
 
         const float patched = select(exceptional, one, x_full);
@@ -3509,7 +3510,6 @@ __declspec(safe) static inline float log(float x_full) {
 
         result = log_from_exponent - x * result;
 
-        const float exceptional_result = select(use_nan, NaN, select(x_full == inf, x_full, -inf));
         return select(exceptional, exceptional_result, result);
     }
 }
```

then no spills are generated. Could you try this?

zephyr111:
In practice, there are 6 spills in the current code and 7 spill if I move the line. The x16 target code is also 1% slower (not really significant). The x8 version seems not affected (as expected). Thus, it unfortunately does not help on my machine. 

By the way, note that I saw different results between Godbolt and my local ISPC (with no other change than this PR) and I do not know why yet. Did you use that to check spilling?

I expected LLVM to track the dependency and perform a good register allocation so to reduce spilling as much as possible. In practice, such problem are not rare in ISPC with double-pumping. That being said, here, it is difficult for LLVM to produce a good code (except on AVX-512 thanks to cheap broadcasting embedded in instructions) based on the high number of registers to load and the number of variables. Still, there is a room for improvement since the x16 target is significantly slower than the x8 target and I think LLVM could theoretically just reorder most instructions so to generate operation on the first x8 items and then the second group. This is mostly the case but not totally. However, in practice, the boolean operation results in fused instructions preventing partially such optimization. This is a bit complicated since I think merging can be better when people do a lot of boolean operations on SSE/AVX-1/AVX-2/Neon targets while it is not a good idea otherwise except maybe in pathological cases (maybe it can reduce register pressure). I am not sure ISPC can make good decision while LLVM can certainly (thanks to a global optimization step). I am clearly not an expert of this so I may completely miss-understand how such thing work or how things should be improved. I think this problem is related to [this issue](https://github.com/ispc/ispc/issues/2920). Still, I am not sure this is the only issue here.

For example, generated instructions are sub-optimal when spilling happens:
```
vblendvps	ymm2, ymm2, ymm6, ymm5
vblendvps	ymm2, ymm14, ymm2, ymm4
vmovups	ymm3, ymmword ptr [rsp - 112]   # 32-byte Reload
vaddps	ymm3, ymm3, ymm1
vmovups	ymmword ptr [rsp - 112], ymm3   # 32-byte Spill
vmovups	ymm1, ymmword ptr [rsp - 112]   # 32-byte Reload    <--------------------
vaddps	ymm0, ymm0, ymm2
add	eax, 16
cmp	eax, 2147483632
jne	.LBB109_1
```

Here we can see a useless reload which can be replaced by a basic `vmovps ymm1, ymm3`. 

I think spilling can can responsible for up to a 30% slowdown on my machine so it may not fully explain the performance drop of the double-pumping version...

nurmukhametov:
> In practice, there are 6 spills in the current code and 7 spill if I move the line. The x16 target code is also 1% slower (not really significant). The x8 version seems not affected (as expected). Thus, it unfortunately does not help on my machine.
> 
> By the way, note that I saw different results between Godbolt and my local ISPC (with no other change than this PR) and I do not know why yet. Did you use that to check spilling?

I used my local build but the generated code looks similar to compiler-explorer's code [link](https://godbolt.org/z/r988YqdMn). I suggest to check which LLVM version you use. Release ISPC and trunk ISPC in compiler-explorer are built with the patched LLVM (see llvm_patches dir in repo). But I don't see difference on my machine between patched LLVM 17 and not-patched LLVM 18 for this example.

zephyr111:
**TL;DR:** the overhead of th target `avx2-i32x16` (on Zen2) appears to be mainly **the price to pay for the much better precision and correct support of special value**. I do not think we can reduce it much more (certainly not more than 10%) unless a better approach is found (and I do not think there is a much better approach reaching a <5 ULP precision).

I profiled more intensively the double pumping implementation to better know what is happening on Zen2 with this PR. The code is bound by the number of instruction/uops which is about 50% higher with the PR. Since the IPC is good both before/after the modification (~2 FP uops/cycle), this makes sense to get a 50% slowdown.

Here are few difference between the two versions for the target `avx2-i32x16`:

- the number of `vblendvps` instruction increases from 6 to 12 : this is normal because of the patch close to 0 so to increase the precision. Indeed, 3 new `select` are done (2 for the patch close to zero and 1 for the support of Inf).
- 2 FMA instructions seems to be split in sub+mul instruction (I do not know why and if they can theoretically be fused).
- the number of comparison instructions increases from 4 to 8 (new `vcmpnleps`/`vcmpnltps`) : this is normal due to the better support of Inf (first additional check) and the check of values close to 0 (second additional check).
- the number of `vxorps`+`vpor` increases from 4 to 8 overall : this is certainly due to the check too though the `vxorps` could certainly be optimized out.
- the number of mov-like instructions increases from 14 to 19 : this is due to register spilling (3 spill vs 6 spill) which is due to a higher register pressure itself certainly due to more constant to load and the additional operations. There is maybe a way to reduce it further but certainly not to reach the same amount as before. 
- 2 additional `vandps` instructions are generated : this is due to the check close to 0 (i.e. the call to `abs`).
- few additional new instructions (x7) like `vpshufd`, `vpackssdw` and `vpacksswb` and `vextracti128` are generated and I think this is missed optimization of LLVM/ISPC related to boolean operations. Indeed, we pack boolean of YMM registers so to perform the logical-or and then unpack the XMM registers so to get YMM ones back while we could just perform YMM operations when the number of booleans operation is not big (hard to do in practice without a non-trivial global optimization pass though).

Note that the loop calling `log` is unrolled twice (hence you should divide the number of instruction by two in the above list to get the number of instructions per call to `log`).

The target `avx2-i32x8` is less impacted by the overheads than the `x16` one simply because the `avx2-i32x8` code is more latency bound. Thus, **the additional instructions can be executed in parallel without adding a significant overhead**. Meanwhile, the `avx2-i32x16` code already hide latency thanks to double-pumping so additional instruction are not free anymore. 

While the generated assembly code can certainly be slightly improved, **most of the overheads directly come from the approach/fix** so I think there is not much to do (besides using possibly another better approach, if any). 

> I used my local build but the generated code looks similar to compiler-explorer's code [link](https://godbolt.org/z/r988YqdMn). I suggest to check which LLVM version you use. Release ISPC and trunk ISPC in compiler-explorer are built with the patched LLVM (see llvm_patches dir in repo). But I don't see difference on my machine between patched LLVM 17 and not-patched LLVM 18 for this example.

Ok. I use the commit 9e749459630c8fc6 @ 20240806 and LLVM 18.1.3. I think I use a non-patched LLVM (from the Ubuntu system packages unless Ubuntu developers patched it).

stdlib.ispc
```diff
@@ -3546,49 +3550,70 @@ __declspec(safe) static inline uniform float log(uniform float x_full) {
         z = x_full + y;
         return z + 0.693359375 * fe;
     } else if (__math_lib == __math_lib_ispc) {
+
+        // Precision: <10 ULP (for all input values).
```
nurmukhametov, line 3554: What about ULP for x=1.120036 (0x3f8f5d57)?

nurmukhametov (commented):
I measured performance both on GameDev tests and microbenchmarks. For both cases, `avx2-i32x8` performance is 20~25 % worse (for tests using `log`). So, it looks we sacrifice performance for precision for `avx2-32x8` also.

stdlib.ispc
zephyr111: The relative error formula gives 7.88694e-7 which looks like <10 ULP at first glance but as we discussed by mail, the formula is not very precise and the number of ULP can be a bit bigger than that. In practice the result is 0x3de829c5 instead of 0x3de829b9 for the truncated value. This means an error of 12 ULP so the error bound is not correct (because of the imprecise formula to check results).

zephyr111:
> I measured performance both on GameDev tests and microbenchmarks. For both cases, `avx2-i32x8` performance is 20~25 % worse (for tests using `log`). So, it looks we sacrifice performance for precision for `avx2-32x8` also.

Yes. Note that the results are dependent of the target architecture. Out of curiosity, on which architecture/CPU did you run this benchmark?

stdlib.ispc
zephyr111: If I modify the checking code so to compute the error directly in ULP with the formula `abs((uniform int64)intbits(res) - (uniform int64)intbits(dres))`, then I get a maximum error of 12 ULP. This maximum error is reached for values closed to the one you mention. Note that this error is maximal for such values because of the problem mentioned above in this PR and present in the current ISPC code too (though the error is much bigger).
This means the actual error bound should not be "<10 ULP" but 12 ULP. Let say "<15 ULP" so to be safe.

zephyr111: If I modify the checking code so to compute the error directly in ULP with the formula `abs((uniform int64)intbits(res) - (uniform int64)intbits(dres))`, then I get a maximum error of 12 ULP for *normal numbers*. This maximum error is reached for values closed to the one you mention. Note that this error is maximal for such values because of the problem mentioned above in this PR and present in the current ISPC code too (though the error is much bigger).

This means the actual error bound should not be "<10 ULP" but 12 ULP. Let say "<15 ULP" so to be safe.

For subnormal numbers (so values close to zeros and output close to -inf), the error is significantly bigger. I expected the mantissa to be >0.5 but I think this is not true with subnormal numbers. I need to check that.

zephyr111: For the subnormal numbers, it turns out that the problem was already present in the current ISPC implementation so this is not a problem introduced by this PR. It can be fixed/mitigated, but again, at the expense of a slower implementation. Note the error can be pretty significant for small subnormal numbers. For example, for 1.20206185e-39 (0xd16dc), the results should be -89.616783 while it is -87.932335 (and -87.932327 for the current ISPC implementation which is not really better).
Tell me if you want to improve the precision for subnormal numbers (certainly at the expense of an even slower implementation).

nurmukhametov: > 
> This means the actual error bound should not be "<10 ULP" but 12 ULP. Let say "<15 ULP" so to be safe.

This is aligned with my estimation.

nurmukhametov:
> > I measured performance both on GameDev tests and microbenchmarks. For both cases, `avx2-i32x8` performance is 20~25 % worse (for tests using `log`). So, it looks we sacrifice performance for precision for `avx2-32x8` also.
> 
> Yes. Note that the results are dependent of the target architecture. Out of curiosity, on which architecture/CPU did you run this benchmark?

I tested on Intel i9-12900 (GameDev tests) and AMD 7840HS (microbenchmarks).

zephyr111:
I am closing this PR since it is superseded by the recent new one.

PR #2940, opened 2024-08-08
Fix Benchmarks for ARM

dbabokin:
- Add `-mavx2` switch for benchmarks only on x86 - so it can work on ARM. I'm not sure why it was there in the first place, but currently I don't have x86 machine handy to check what will happen if it's removed.
- Update Google Benchmark `v1.8.3` -> `v1.8.5`.


aneshlya (approved):
LGTM!

Issue #2347, opened 2022-07-17
Add `fmod` for `double` and `float` to the standard library


pbrubaker:
We currently have `trunc()` so it's possible for users to implement it themselves, but it would be nice to have it in the stdlib.

```
float fmodf(float x, float y)
{
  return x - trunc(x / y) * y;
}
```

PR #2942, opened 2024-08-08
Add fmodf for float16, float, and double

aneshlya:
Fixes https://github.com/ispc/ispc/issues/2347.


nurmukhametov (approved):

docs/ispc.rst
```diff
@@ -4252,6 +4252,18 @@ use Newton-Raphson.
     float rcp_fast(float v)
     uniform float rcp_fast(uniform float v)
 
+The ``fmodf()`` functions compute the floating-point remainder of the division
+operation x/y (its implementation is based on the ``rcp`` and ``trunc`` functions):
```
dbabokin, line 4256: Do we want implementation details note in the docs?

I would rather point out that it's semantics is equivalent to C/C++ lib functions.

docs/ispc.rst
```diff
@@ -4252,6 +4252,18 @@ use Newton-Raphson.
     float rcp_fast(float v)
     uniform float rcp_fast(uniform float v)
 
+The ``fmodf()`` functions compute the floating-point remainder of the division
```
dbabokin, line 4255: `f` suffix is C language mangling of `float` argument. I.e. see: https://man7.org/linux/man-pages/man3/fmod.3.html
 
So C has `fmod(double, double)`, `fmodf(float, float)`, and `fmodl(long double, long double)`.

C++ has overloaded `fmod()` function: https://en.cppreference.com/w/cpp/numeric/math/fmod

As long as we have overloading in the language, we should name it `fmod()`, not `fmodf()`.

dbabokin (changes requested):

docs/ispc.rst
aneshlya: Fixed

docs/ispc.rst
aneshlya: Fixed

dbabokin (approved):
LGTM

PR #2944, opened 2024-08-09
stdlib: fix overflow in exp(float) and exp(uniform float)

zephyr111:
Hello.

This PR fix an overflow in the exp implementation previously resulting in bad results for -Inf, NaN and big numbers (especially due to an integer overflow). The implementation is now 100% valid with a relative precision of <5 ULP for normal numbers (possibly much more for subnormal numbers).

It was initially a part of the PR #2937 .

### DIFF
--- a/stdlib.ispc
+++ b/stdlib.ispc
@@ -3250,6 +3250,9 @@ __declspec(safe) static inline float exp(float x_full) {
         x_full = ldexp(z, n);
         return x_full;
     } else if (__math_lib == __math_lib_ispc) {
+
+        // See the uniform version for more information
+
         const float ln2_part1 = 0.6931457519;
         const float ln2_part2 = 1.4286067653e-6;
         const float one_over_ln2 = 1.44269502162933349609375;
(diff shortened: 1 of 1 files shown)


nurmukhametov (approved):
LGTM!

How did you estimate ULP? Is it same for both variants (`uniform` and `varying`)?

zephyr111:
> How did you estimate ULP?

With the kind of code provided in the other (initial) PR (about `log`). I compute the formula of the relative error for each possible 32-bit float value and then report if the error is bigger than a threshold. With a threshold of 5e-7, I assume that the error is less than 5 ULP (since 1.0 + 1 ULP is about 1+1.2e7). Just to be safe, I checked this logic worked on few values by checking the binary representation of the numbers. 

Note that I did not really consider subnormals during the evaluation. After a deeper analysis, It looks like they are truncated to zero.  but I do not expect people using ISPC to complain much about them. I am not sure it is worth doing anything about it. I do expect a performance loss for a benefit that only very very few ISPC user will certainly ever need. This is why I mention that the error in ULP for subnormal number inputs can be much bigger than normal numbers (but after a deeper analysis, I think we can say that subnormal numbers are not supported or can be truncated). What do you think about it?

Note the analysis only focus on the default math functions (the one modified in this PR), not fast/system ones. 

> Is it same for both variants (uniform and varying)?

There is no reason for them to be different, but to be on the safe side, I just check `extract(exp(x), 1)` values was as good as `exp(x)` with `x` being a `uniform float`. Thus, yes.

aneshlya (approved):

PR #2948, opened 2024-08-10
stdlib: massively improve the precision of asin (especially close to 0) and make it faster

### LINKED ISSUE #3807: Improve asin precision

zephyr111:
Hello,

This PR strongly improves the relative precision of `asin` close to 0. As mentioned in #2236, the error was previously huge:

> `asin` appears not to support subnormal numbers nor even just small numbers like `x=2.980232e-08` considered as a 0 (truncation).

With this PR, the relative precision is now <20 ULP (about 15 ULP in practice using AVX2 & FMAs). This include subnormal numbers which are perfectly supported. I wish to reduce it to 10 ULP but the performance impact was huge when I tried so far. I think it is possible by rewriting (partially) the implementation, but 20 ULP is already pretty good compared to the previous implementation.
The performance impact is about 5% for the target `avx2-i32x8` and I did not see any performance impact for the target `avx2-i32x16` (default math mode).

### DIFF
--- a/stdlib/stdlib.ispc
+++ b/stdlib/stdlib.ispc
@@ -3510,10 +3510,14 @@ __declspec(safe) static inline float sinh(float x) {
 }
 
 __declspec(safe) static inline float asin(float x0) {
-    bool isneg = x0 < 0;
-    float x = abs(x0);
-    bool isnan = (x > 1);
-    float v;
+
+    // See the uniform implementation for more information
+
+    const bool isneg = x0 < 0;
+    const float x = abs(x0);
+    const float y = x0 * x0 + 1e-30;
+    const bool isnan = x > 1;
+    float v, w;
 
     if (__have_native_trigonometry && !__is_xe_target) {
         return __asin_varying_float(x0);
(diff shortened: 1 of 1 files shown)


zephyr111:
I finally rewrote the implementation. It is now significantly **more precise** (<5 ULP) and **also significantly faster**! The fast version might be just slightly less precise for values close to 1 (absolute error) but the relative error is far better (<20 ULP). I think this implementation is (nearly) optimal.

Here is performance results on the full float range (on a Zen2 CPU):

```
Default math-lib before/after:
    - avx2-i32x8:   2.0e9  =>  1.5e9
    - avx2-i32x16:  2.0e9  =>  1.6e9

Fast math-lib before/after:
    - avx2-i32x8:   1.2e9 =>  1.3e9
    - avx2-i32x16:  1.2e9 =>  1.2e9

Sleef: 3.1e9
```

We can see that the default version is faster (while being also more precise). It is actually significantly better than Sleef achieving a similar precision (3.5 ULP). The fast version is a bit slower (about 5-10%) but it clearly worth it since the relative precision is several order of magnitude better for values close to 0 (not to mention it now support subnormal-numbers) and very similar to the previous implementation for values close to 1.

zephyr111:
With a better way to measure the error in ULP (i.e. `abs((uniform int64)intbits(res) - (uniform int64)intbits(dres))`), the error is 6 ULP (instead of 4) for the default math version. Thus, I think we can specify <10 ULP (just to be safe). For the fast math version, it is 25, so we can specify <30 ULP (again to be safe). Note this include subnormal numbers.

nurmukhametov:
The default `asin` for `float` looks good to me.

However, I'm slightly concerned about the fast implementation because it's slower. Is it possible to make it, let's say, `<50 ULP` while maintaining the same performance as before?

zephyr111:
> The default `asin` for `float` looks good to me.

Ok!

> However, I'm slightly concerned about the fast implementation because it's slower. Is it possible to make it, let's say, `<50 ULP` while maintaining the same performance as before?

Unfortunately, no, this is not possible to both make it faster and get <50 ULP (at least using the current approach). There is a balance between precision and performance. Here, to get better performance, the main way is to reduce the order of one of the two polynomial. The thing is their precision is very dependent of their range of definition. Thus, if we reduce the precision of one polynomial, then the other must be defined on a significantly wider range which sharply reducing its precision.

My best try with a very-precise hand-tuning of the polynomial coefficients is 60 ULP. Let's say 65 ULP so to be safe. Note that at this scale, the order of the polynomials is so low and the target precision so small that algorithms tends to fail to find the best coefficient. Thus, things become very tricky. The resulting performance gain is 5-6% on my machine. Thus, this less precise implementation is roughly as fast as the initial implementation (i.e.  the one in the main branch). I am not sure the ~5% performance gain is worth the loss of precision (about twice less precise).

An alternative solution might be not to support NaN outside the range so to make the function a bit faster with no loss of precision. Most fast math functions tends not to support special cases/values. That being said, I think we should discuss about that for all fast math functions (so either all of them support special cases/values or none of them).

What do you think about that?

zephyr111:
I modified the default version so to get an **even higher precision**. Indeed, I found out that the first polynomial is *not numerically stable* not only for value very close to 0 but also even for values like 0.3~0.5. Thus, I increased the precision of the other polynomial so to better balance the precision between both. I also manually tuned (with sub-ULP precision) the constants so the polynomials are certainly optimal for the target order, precision, and range. This results in a ~3 ULP error (as opposed to 6 previously) using still the new (accurate) way of measuring ULPs. Thus, I reported <5 ULP so to be safe. Interestingly, the new version is also **slightly faster** on my machine. It is now very close to the Sleef implementation in term of precision while being more than twice faster :) .

Here are updated results:

```
Default math-lib before/after (ie. main branch vs last version of the PR):
    - avx2-i32x8:   2.0e9  =>  1.4e9
    - avx2-i32x16:  2.0e9  =>  1.4e9
Sleef: 3.1e9
```

I did not change the fast version so far (see the previous message).

zephyr111:
I rebased the branch on main and check everything works well. I confirm the <5 ULP precision.
Tests are already written for this function so this seems fine.

Here are benchmark results on my machine (i5-9600KF) for the two ISPC math-lib implementation on the full float range:

```
	- without subnormals:
		- Current (default):    2.679913e9
		- This PR (default):    2.141665e9
		- Current (fast):       1.546451e9
		- This PR (fast):       1.853974e9
		- Sleef:                3.403672e9
		- Current (default):    3.826855e9
		- This PR (default):    2.851407e9
		- Current (fast):       2.130620e9
		- This PR (fast):       2.566657e9
		- Sleef:               15.252493e9
```

Is everything OK so to merge this?

**EDIT:** I also ran another benchmark with random inputs in the uniform range `[-1,1]`. It is certainly a more realistic benchmark than the full range since asin is only defined in this range. The timing are in clocks. The RNG take a non-negligible part of the time, but I think the timings could affected by the speed of the `sqrt` which is AFAIK variable on my CPU (possibly faster for out-of-range values).

```
	- Current (default):    7.152750e9
	- This PR (default):    5.848970e9
	- Current (fast):       5.114508e9
	- This PR (fast):       5.566213e9
	- Sleef:                6.557322e9
	- RNG loop without computing asin (RNG overhead): 2.116213e9
```

aneshlya (approved):
Looks good to me!
@nurmukhametov, do you have any concerns or comments?

elena901:
> An alternative solution might be not to support NaN outside the range so to make the function a bit faster with no loss of precision.

@zephyr111 When this PR was introduced i'd say "Let's remove the NaN support!" but now since the time has passed maybe there is a better solution? 

P.S. At least it looks like the default-math version could be safely merged anyway.

nurmukhametov (approved):

nurmukhametov:
> Looks good to me! @nurmukhametov, do you have any concerns or comments?

LGTM!

zephyr111:
@elena901

I agree for the NaNs. That being said, I tried to removed the select for the NaN support and ran benchmarks again and found no significant difference in the performance results. I think it is why I actually supported them: it does not impact the performance (at least on my machine) although I do see a diffrence in the generated assembly code (see the [test function with a select](https://pastebin.com/QZyzivcF) and [without it](https://pastebin.com/Fij3mq2p)). This is certainly because the blend does not impact the saturated ports (the ones doing the FMAs and `sqrt`). It might be slightly slower (5%?) on AMD zen3/zen4/zen5 but one need to test to be sure. Do you see any performance impact on your machine? And if so, which CPU do you have?

```
	- without subnormals:
		- This PR (fast no-nan-select):       1.851723e9
		- This PR (fast no-nan-select):       2.567942e9
```

elena901:
> Do you see any performance impact on your machine?

@zephyr111 I will try later but Aleksei already said that all is ok!



> maybe there is a better option?

I mean to make it faster on a specific range for example.

zephyr111:
> I mean to make it faster on a specific range for example.

@elena901 This is possible with the fast-math version using a conditional branch but it would be significantly slower in other cases (e.g. on the boundary of the branch), so I think it does not worth it here.

zephyr111:
@nurmukhametov @aneshlya Should an issue be created for this PR like you did with #3801 for the log in order to update the planning accordingly? Btw, can't PR be directly linked in the planning?

nurmukhametov:
> @nurmukhametov @aneshlya Should an issue be created for this PR like you did with #3801 for the log in order to update the planning accordingly? Btw, can't PR be directly linked in the planning?

We have been doing so, but I don't remember why we chose not to link PRs directly. I have created issue #3807

PR #2950, opened 2024-08-11
docs: fix English typo/mistakes

zephyr111:
Hello,

As he name of the PR suggest, it fixes minor typo / English mistakes.

By the way, please note that there is a surprising "XXXX todo" in the performance guide (even [visible online](https://ispc.github.io/perfguide.html#understanding-memory-read-coalescing)).


dbabokin (approved):
Thanks for proofreading our docs! Changes look good!

As for perf guide, we haven't touch it for a while. Would be good to refresh and update it. If you see where you can contribute - feel free to do so, either minor or major changes.

PR #2953, opened 2024-08-13
builtins: remove internal from __shuffle_*

aneshlya:
Extracted from https://github.com/ispc/ispc/pull/2743.
It's not related to targets redesign. It happened after https://github.com/ispc/ispc/pull/2924 because I changed the interface in the last moment and forgot to remove `internal` in several places.

-------------------------------------------------------------------

1. It doesn't match with other __shuffle_* definitions in other files.

2. The more important thing is the following one. After the target redesign, builtins module is linked into the stdlib^. It means that no builtin function used in stdlib.ispc should be defined with internal attribute (C static keyword). Otherwise, the linker gets rid of __shuffle_ from builtins instead of resolving __shuffle_ symbol used in stdlib. It would result into unresolved symbol. Only functions that are internal to builtins, i.e., without usage in stdlib, can be marked as internal although I don't really expect it matters from the optimization point of view.

^ actually user code with stdlib code

### DIFF
--- a/builtins/target-avx1-i64x4base.ll
+++ b/builtins/target-avx1-i64x4base.ll
@@ -27,14 +27,14 @@ shuffle1(double)
 shuffle1(i64)
 
 declare <4 x float> @llvm.x86.avx.vpermilvar.ps(<4 x float>, <4 x i32>)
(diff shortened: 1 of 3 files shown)


dbabokin (approved):
Looks good.

PR #2957, opened 2024-08-16
ci: fix yarpgen workflow

aneshlya:
The second checkout action is overriding the directory created by first one (somehow not always).
This PR fixes this problem.


dbabokin (approved):
I'm not surprised, checkout action is not doing just plain checkout, it's doing something way more complex.

Looks good to me, if it works well.

Issue #2079, opened 2021-05-08
cbrt(x) support?


Remotion:
Currently ISPC does not support cube root: 
https://godbolt.org/z/338aeEr89 .

Would it be possible to add cbrt support to ISPC ?

dbabokin:
I generally support adding `cbrt()` as it's part of C math library.

Though, the problem is to have an efficient implementation for it. ISPC has 4 type of math library that it can use: `default` (ispc's own implementation), `fast` (ispc's own implementation, which less accurate), `svml` (call SVML library, fast vector implementation, but SVML needs to be provided at link step), and `system` (calls scalar version in glibc).

For `default` and `fast` we need to find an implementation. Would be `pow(x, 1/3)` be good enough as a first cut?

Remotion:
`cbrt()` is there since C99.
https://en.cppreference.com/w/c/numeric/math/cbrt

I mean `cbrt()` like this one.
https://github.com/Remotion/fast_cbrt/blob/d79c9e44c09a2219c588c41fc144a81b25d7aa0e/ispc/cbrt.ispc#L87

JeffGreen:
There is a more generic and optimal solution here than what ISPC is doing at present.  The current state of affairs has a few issues -

* default doesn't have implementations for all functions and types (double implementations are missing, for example)
* default's tradeoff between performance and precision could be optimized
* SVML isn't bundled and isn't yet open sourced so it's more of a niche use case

I propose that we integrate Sleef (https://sleef.org/) which is an open source SVML-like library that has float32 and float64 implementations of all math functions supported by ISPC.  It supports many architectures as is listed on its homepage.  Furthermore, its build process allows generating LLVM IR for each implementation which makes tight integration with ISPC possible and removes function call overhead (-DSLEEF_ENABLE_LLVM_BITCODE in cmake).  Finally, many of the transcendental functions it provides offer two precision / speed options (_u35 and _u10 suffixes for 3.5 and 1.0 ULPs).  

Benchmarks as compared with SVML are here:
https://sleef.org/benchmark.xhtml

zephyr111:
Using `pow(x, 1.d/3.d)` results in several issues. First, it does not support negative numbers (because the exponent is not an integer). Moreover, it seems not to support well +inf/-inf nor NaN. Finally, it is rather slow compared to an optimized `cbrt` implementation.

[...]

PR #2958, opened 2024-08-18
stdlib: add a cbrt function

zephyr111:
Hello,

This PR add a fast and precise `cbrt` function for floats (there is also a double-precision implementation calling the standard library like for other functions).
This fix the issue #2079.

Its precision is 4 ULP on my Zen2 machine (using the new more reliable formula to compare values : `abs((uniform int64)intbits(res) - (uniform int64)intbits(dres))`) so I mentioned <5 ULP. I tested the target `sse2-i32x8`, `avx2-i32x8` and `avx2-i32x16` so to see the impact of the FMA on the precision (i.e. no impact).

Here are performance results on the full 32-bit float range (using the target `avx2-i32x8`):
```
 - This PR: 2.69e9 cycles
 - Sleef:   4.65e9 cycles
```

We can see that this implementation is significantly faster than the one of Sleef (while having a similar precision : 3.5 ULP vs 4 ULP). Note that enabling or disabling subnormal numbers (with `--opt=reset-ftz-daz`) does not seems to impact much result of both implementations (<2% difference).

Note the fast version is a bit faster. It runs in 2.26e9 cycles.

I am just concerned about one thing: functions like `_cbrt_newton_iter` can certainly be called from user code. I just prefixed them with a `_` so to warn users to not use them. Is there a (better) way to hide them from the users?


nurmukhametov (changes requested):
Could you rebase this PR on the current main? CI should work fine after the fix of #2959 issue.

There are also two things that would be good to do:
1. Add documentation in `docs/ispc.rst`.
2. Add at least some run-time tests, e.g., like here #2942 (files under `tests`, now `tests/func-tests`).

zephyr111:
I am closing this PR since it has been merged in #3369 instead.

PR #2960, opened 2024-08-20
Fixes for LLVM 19 build and nightly testing

aneshlya:
1. Fixed lit tests for LLVM 19+ build unlocking the testing in CI
2. Skip avx512 targets on x86 target in nightly and trunk testing.
3. Remove LLVM rebuild in nightly LLVM 17.0 testing. BTW I'm not fully understand why do we need this workflow at all, this testing is covered by ispc-ci.yml
4. Fixed shuffles for avx512-knl targets (deprecated, not tested anywhere except nightly 17.0 testing)

### DIFF
--- a/tests/lit-tests/loop_unroll_varying.ispc
+++ b/tests/lit-tests/loop_unroll_varying.ispc
@@ -43,23 +43,7 @@ void foo_for() {
 // CHECKO0:         br label %{{[a-zA-Z_][a-zA-Z0-9_.]*}}
 // CHECKO0:         }
 
-// CHECKO2:         for_loop{{.*}}:
-// CHECKO2:         call void @goo_for_unroll8___
-// CHECKO2:         for_loop{{.*}}:
-// CHECKO2:         call void @goo_for_unroll8___
-// CHECKO2:         for_loop{{.*}}:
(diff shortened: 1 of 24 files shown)


tests/lit-tests/loop_unroll_varying.ispc
```diff
@@ -43,21 +43,13 @@ void foo_for() {
 // CHECKO0:         br label %{{[a-zA-Z_][a-zA-Z0-9_.]*}}
 // CHECKO0:         }
 
-// CHECKO2:         for_loop{{.*}}:
 // CHECKO2:         call void @goo_for_unroll8___
```
nurmukhametov, line 46: If this test count the number of calls to `goo_for_unroll8___`, so it probably needs to check that with `CHECK-COUNT-N`.

nurmukhametov (approved):

dbabokin (approved):
Looks good to me.

Just fix copyright and I agree with Alexey's comment.

tests/lit-tests/loop_unroll_varying.ispc
aneshlya: Fixed.

PR #2961, opened 2024-08-21
Enable NewGVN for CPU also

### LINKED ISSUE #2954: Extra mov's generated when returning a struct from a function
Given the following code: https://godbolt.org/z/b58eWxGeY, ISPC can eliminate the extra mov instructions return the results more efficiently

nurmukhametov:
This fixes #2954 extra moves issue.


dbabokin (approved):
LGTM

aneshlya (approved):

PR #2963, opened 2024-08-22
Support more attributes for target redesign

### LINKED ISSUE #2956: Support readnone, readonly attributes
Support them to assist the target redesign change.

nurmukhametov:
This PR supports attributes: `unmangled`, `memory("read")`, `memory("none")`. 

This fixes #2956.

### DIFF
--- a/src/decl.cpp
+++ b/src/decl.cpp
@@ -199,6 +199,7 @@ bool Attribute::IsKnownAttribute() const {
     static std::unordered_map<std::string, int> lKnownParamAttrs = {
         {"noescape", 1},
         {"address_space", 1},
+        {"unmangled", 1},
     };
 
     if (lKnownParamAttrs.find(name) != lKnownParamAttrs.end()) {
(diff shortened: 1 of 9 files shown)


dbabokin (commented):
Need to add description in the docs.

src/decl.cpp
```diff
@@ -798,14 +801,25 @@ void Declarator::InitFromType(const Type *baseType, DeclSpecs *ds) {
             Warning(pos, "\"unmasked\" qualifier is redundant for exported "
                          "functions.");
 
+        if (isUnmangled) {
```
aneshlya, line 804: A test covering different combinations of attributes/function qualifiers would be good.

nurmukhametov: I have added a couple of tests.

nurmukhametov:
> Need to add description in the docs.

I have added it for `unmangled`. About `memory`, I don't think we need to expose them to users.

aneshlya (approved):
Looks good!

PR #2964, opened 2024-08-22
Updates for Xe targets (27)

aneshlya:
- Added `xe2hpg` and `xe2lpg` targets for Xe2 Battlemage and Lunar Lake HW.
- Updater Meteor Lake device names.
- Updated dependencies versions.


aneshlya:
Looks like Agama driver doesn't have necessary support yet.

PR #2965, opened 2024-08-23
Restrict security scans to ispc/ispc repo

dbabokin:
Main problem is that regular scans run in forks and create traffic to nightly builds on appveyor hosting, which is pretty limited on traffic.

Would be good to move nightly builds to github releases.


aneshlya (approved):
LGTM!

PR #2967, opened 2024-08-26
Bandit scan + minor CI fixes

aneshlya:
1. Added bandit scan
2. Excluded `avx512knl-x16` from x86 testing
3. Updated result check in fmod float tests (was failing on x64 targets with -O0)
4. Added description of ispc optimization options to docs

### DIFF
--- a/.github/workflows/bandit.config
+++ b/.github/workflows/bandit.config
@@ -1,0 +1,405 @@
+# Copyright 2024, Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause
+
+### Bandit config file generated from:
+# './bandit/bandit/cli/config_generator.py --out bandit.config'
+
+### This config may optionally select a subset of tests to run or skip by
+### filling out the 'tests' and 'skips' lists given below. If no tests are
+### specified for inclusion then it is assumed all tests are desired. The skips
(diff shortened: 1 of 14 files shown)


dbabokin (approved):
Looks good, but it's surprising that library math calls with -O0 return different results, while they use exactly the same algorithms. Do you have any explanation for that?

aneshlya:
> Looks good, but it's surprising that library math calls with -O0 return different results, while they use exactly the same algorithms. Do you have any explanation for that?

I don't have explanation for that. I was experimenting with it and I noticed that having any branch inside the `fmod` function affects the precision of the result. LLVM IR for `x - trunc(x * rcp(y)) * y` expression inside fmod function and in test looks exactly the same.

dbabokin (changes requested):
Thinking about this problem, I'd like to understand the problem better before merging it. Possibly the fix should be different.

Could try running -O2 with `--opt=disable-fma`? I think this should be enough to disable all transformation that affect precision.

aneshlya:
> Could try running -O2 with `--opt=disable-fma`? I think this should be enough to disable all transformation that affect precision.

With  `--opt=disable-fma` results are the same.

dbabokin (approved):
LGTM

PR #2970, opened 2024-08-29
dispatch.c: make system symbols weak

### LINKED ISSUE #2969: Avoid duplicates of dispatcher
At the moment, when user application contains several ISPC modules, every module contains its own copy of `__get_system_isa`, `__set_system_isa` and `__system_best_isa`. They are not eliminated in the final executable and every module calls its own `__get_system_isa` to determine the best ISA even the other module has already done that.

nurmukhametov:
This reduces the number of copies when a user application consists of several multi-target modules. Before this change, each module had its own copy, leading to multiple instances of system ISA detection.

This PR fixes #2969 leading to the only instance of dispatcher.

### DIFF
--- a/builtins/dispatch.c
+++ b/builtins/dispatch.c
@@ -24,7 +24,10 @@
 #include <stdint.h>
 #include <stdlib.h>
 
-static int __system_best_isa = -1;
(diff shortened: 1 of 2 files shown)


aneshlya (approved):
LGTM!

PR #2972, opened 2024-09-02
Do not emit error for typedef struct foo {} foo

### LINKED ISSUE #2351: typedef struct with same name doesn't work
Give the following code https://godbolt.org/z/3ha18KEaG
```
typedef struct Foo {
    int a;
} Foo;   
```
Doesn't work in ISPC, but is fully legal in C/C++. The way C/C++ handles this is that struct names and typedefs are in different namespaces thus this works, while ISPC reports `Error: Multiple types provided for declaration`

nurmukhametov:
This PR fixes #2351


dbabokin (approved):
LGTM

aneshlya (approved):
Looks good!

PR #2973, opened 2024-09-03
Use typedef name for anon struct

### LINKED ISSUE #2350: $anon0 struct name is used instead of typedef
Given the following code
```
typedef struct {
    float t;
} TestStruct;

export uniform TestStruct test() {
    uniform TestStruct s;
    s.t = 1.0;
    return s;
}
```
This gets generated in the header:
```
// (Header automatically generated by the ispc compiler.)
// DO NOT EDIT THIS FILE.
//

#pragma once
#include <stdint.h>



#ifdef __cplusplus
namespace ispc { /* namespace */
#endif // __cplusplus

#ifndef __ISPC_ALIGN__
#if defined(__clang__) || !defined(_MSC_VER)
// Clang, GCC, ICC
#define __ISPC_ALIGN__(s) __attribute__((aligned(s)))
#define __ISPC_ALIGNED_STRUCT__(s) struct __ISPC_ALIGN__(s)
#else
// Visual Studio
#define __ISPC_ALIGN__(s) __declspec(align(s))
#define __ISPC_ALIGNED_STRUCT__(s) __ISPC_ALIGN__(s) struct
#endif
#endif

#ifndef __ISPC_STRUCT_$anon0__
#define __ISPC_STRUCT_$anon0__
struct $anon0 {
    float t;
};
#endif


// Functions exported from ispc code
#if defined(__cplusplus) && (! defined(__ISPC_NO_EXTERN_C) || !__ISPC_NO_EXTERN_C )
extern "C" {
#endif // __cplusplus
    extern struct $anon0 test();
#if defined(__cplusplus) && (! defined(__ISPC_NO_EXTERN_C) || !__ISPC_NO_EXTERN_C )
} /* end extern C */
#endif // __cplusplus


#ifdef __cplusplus
} /* namespace */
#endif // __cplusplus
```
It seems like the typedef name isn't used at all here and only the `$anon0` struct name which is a bit confusing.

nurmukhametov:
If anonymous struct has typedef then we can use typedef name instead of `$anonN`.
This fixes #2350.

### DIFF
--- a/src/parse.yy
+++ b/src/parse.yy
@@ -3008,9 +3008,15 @@ lAddDeclaration(DeclSpecs *ds, Declarator *decl) {
         return;
 
     decl->InitFromDeclSpecs(ds);
-    if (ds->storageClass == SC_TYPEDEF)
-        m->AddTypeDef(decl->name, decl->type, decl->pos);
-    else {
+    if (ds->storageClass == SC_TYPEDEF) {
+        const StructType *st = CastType<StructType>(decl->type);
+        if (st && st->IsAnonymousType()) {
+            st = st->GetAsNamed(decl->name);
+            m->AddTypeDef(decl->name, st, decl->pos);
+        } else {
+            m->AddTypeDef(decl->name, decl->type, decl->pos);
+        }
+    } else {
         if (decl->type == nullptr) {
             Assert(m->errorCount > 0);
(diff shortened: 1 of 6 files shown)


dbabokin (approved):
Looks reasonable.

PR #2975, opened 2024-09-03
Fix linkage of template instantiations

### LINKED ISSUE #2959: Crash of ISPC with templated functions in the standard library
Hello,

There is apparently a bug happening in the tests of the PR #2958 causing ISPC to suddenly crash with the following message:

> /home/richard/Bureau/projects/ispc/src/module.cpp(1478): FATAL ERROR: Resulting module verification failed!

I can reproduce it on my machine. Strangely, it works well on my testing code. It apparently happens when there is a templated function in the `stdlib.ispc` and when this templated function is also called. Here is an example:

```
template <typename T> __declspec(safe) static inline void testing_func(T x) {}

__declspec(safe) static inline void caller_func(uniform float x) {
    return testing_func(x);
    // return testing_func<uniform float>(x);   // also fail
}
```

The command of the test causing the issue is the following:

> echo "" | ./bin/ispc -o - --target=avx2-i32x8 -E --emit-asm -

It crashes with all x86 targets the same way. 

-----

Note that with ARM ones (e.g. `neon-i32x4`), the error is different:

> stdlib.ispc:1496:48: Error: Undeclared symbol "__num_cores". Did you mean "num_cores"?

However, this problem happen even without changing any file.

Both issues happen with the latest commit so far.

aneshlya:
This PR fixes #2959.
It also has a minor CMakeLists fix regarding `_FORTIFY_SOURCE=2` compile definition that can't be used in Debug mode (triggers warning with new gcc)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -549,11 +549,13 @@ list(APPEND COMPILE_DEFINITIONS
     BUILD_VERSION=\"${GIT_COMMIT_HASH}\"
 )
 if (UNIX)
-    # Compile-time protection against static sized buffer overflows.
-    list(APPEND COMPILE_DEFINITIONS _FORTIFY_SOURCE=2)
+    if (NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
+        # Compile-time protection against static sized buffer overflows.
+        list(APPEND COMPILE_DEFINITIONS _FORTIFY_SOURCE=2)
+    endif()
 else()
     list(APPEND COMPILE_DEFINITIONS NOMINMAX)
-    if (NOT CMAKE_BUILD_TYPE STREQUAL "DEBUG" )
+    if (NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
         list(APPEND COMPILE_DEFINITIONS NDEBUG)
     endif()
 endif()
--- a/src/func.cpp
+++ b/src/func.cpp
@@ -689,12 +689,10 @@ void Function::GenerateIR() {
         // ISPC external functions.
         if (type->isExported || type->isExternC || type->isExternSYCL || type->IsISPCExternal() ||
             type->IsISPCKernel()) {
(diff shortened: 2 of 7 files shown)


CMakeLists.txt
```diff
@@ -549,11 +549,13 @@ list(APPEND COMPILE_DEFINITIONS
     BUILD_VERSION=\"${GIT_COMMIT_HASH}\"
 )
 if (UNIX)
-    # Compile-time protection against static sized buffer overflows.
-    list(APPEND COMPILE_DEFINITIONS _FORTIFY_SOURCE=2)
+    if (CMAKE_BUILD_TYPE STREQUAL "Release")
```
dbabokin, line 552: What about `RelWithDebInfo`?

src/func.cpp
```diff
@@ -993,6 +991,29 @@ void FunctionTemplate::Print() const {
...
+llvm::GlobalValue::LinkageTypes lGetTemplateInstantiationLinkage(TemplateInstantiationKind kind) {
+    llvm::GlobalValue::LinkageTypes linkage = llvm::GlobalValue::LinkageTypes::InternalLinkage;
+    switch (kind) {
+    // Function can be defined multiple times across different translation units without causing conflicts.
+    // The linker will choose a definition for the function based on its default behavior.
+    case TemplateInstantiationKind::Explicit:
+        linkage = llvm::GlobalValue::WeakODRLinkage;
+        break;
+    // The function is only allowed to be defined once across all translation units, but it can be discarded if unused.
+    // If multiple definitions of the function are present across different translation units, the             linker
```
dbabokin, line 1013: to many whitespaces in the end of the line :)

dbabokin (approved):
LGTM

CMakeLists.txt
aneshlya: Fixed

src/func.cpp
aneshlya: Fixed

nurmukhametov (approved):
LGTM

PR #2977, opened 2024-09-04
Fix anon struct renaming

nurmukhametov:
This PR is fix for #2973. `GetAsNamed` always return named struct. This prevents renaming struct when new aliases created with `typedef`.


aneshlya (approved):
LGTM!

dbabokin (approved):
LGTM

PR #2985, opened 2024-09-05
Dependabot version update

aneshlya:
This PR improves configuration of dependabot.yml. In theory (I can't test it until it's merged) it should create one PR for all dependencies update.
It also reverts 7431a910b5be098cc03ef556661d5f91bd91c8f7. Pinning the version didn't help to eliminate warnings like this one: https://github.com/ispc/ispc/security/code-scanning/88. Seems like it should be pinned by hash which is IMO overkill for the tools that are used for testing.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,4 +9,9 @@ updates:
     schedule:
       interval: "weekly"
     commit-message:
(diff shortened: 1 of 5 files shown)


dbabokin (approved):
Looks ok, let's try it.

PR #2987, opened 2024-09-05
ci: fix groups in dependabot.yml

aneshlya:
Fixed groups in dependabot.yml.
Now it should create one PR for all the GA dependencies update like this one: https://github.com/aneshlya/ispc/pull/19.


dbabokin (approved):
LTGM

PR #2992, opened 2024-09-09
Support short vectors based on template parameters

### LINKED ISSUE #2522: Enable SOA and short vector types based on template type parameters
Given the following code: https://ispc.godbolt.org/z/4YvcjPE8h

```cpp
template<typename T>
T Calculate(T x, T y, T z) {
    T<TARGET_WIDTH> soaValue;

    return x + y + z;
}
```

ISPC will give the following error:

><source>:3:10: Error: Only atomic types (int, float, ...) are legal 
>        for vector types. 
>    T<TARGET_WIDTH> soaValue;
>         ^^^^^^^^
>
>Compiler returned: 1

aneshlya:
Partially fixes #2522. Support of soa<N> and array types will be added in a similar way.
It also adds a symbol kind to the Symbol class, which is used now only to check if the symbol in the vector size declaration is a TemplateNonTypeParameter. However, I believe it could be useful in other scenarios as well. 
The PR opens up the possibility of using constants in vector/array declarations (as seen in func_template_short_vec_9.ispc), but resolving these constants is non trivial, so I prefer not to introduce it at this stage.

### DIFF
--- a/src/builtins.cpp
+++ b/src/builtins.cpp
@@ -177,7 +177,7 @@ Symbol *ispc::CreateISPCSymbolForLLVMIntrinsic(llvm::Function *func, SymbolTable
     }
     FunctionType *funcType = new FunctionType(returnType, argTypes, noPos);
     Debug(noPos, "Created Intrinsic symbol \"%s\" [%s]\n", name.c_str(), funcType->GetString().c_str());
-    Symbol *sym = new Symbol(name, noPos, funcType);
+    Symbol *sym = new Symbol(name, noPos, Symbol::SymbolKind::Function, funcType);
     sym->function = func;
     symbolTable->AddIntrinsics(sym);
     return sym;
--- a/src/decl.cpp
+++ b/src/decl.cpp
@@ -281,7 +281,7 @@ DeclSpecs::DeclSpecs(const Type *t, StorageClass sc, int tq) {
     storageClass = sc;
     typeQualifiers = tq;
     soaWidth = 0;
-    vectorSize = 0;
+    vectorSize = std::monostate{};
     attributeList = nullptr;
     if (t != nullptr) {
         if (m->symbolTable->ContainsType(t)) {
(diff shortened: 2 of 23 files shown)


dbabokin (approved):
Looks good to me!

nurmukhametov:
By some reason, the following example doesn't work:

```ispc
varying int32 func(varying int32 a, varying int32 b);
uniform int32 func(uniform int32 a, uniform int32 b);

template <typename T, uint N>
T<N> func(T<N> a, T<N> b)
{
    T<N> r;
    foreach (i = 0 ... N) {
        r[i] = func(a[i], b[i]);
    }
    return r;
}

uniform int<4> foo2(uniform int<4> a, uniform int<4> b) {
    return func<uniform int, 4>(a, b);
}
```
with the following error:
```bash
: Error: Unable to resolve the value of the symbol "N" used to specify the size of a vector type. 
template <typename T, uint N>
                                     ^^^^^^

: Error: Unable to find any matching overload for call to function "func".  Passed types: (varying int32, varying int32) 
         
         r[i] = func(a[i], b[i]);
                ^^^^
```

aneshlya:
Looks like the logic is broken when template function and regular function have the same name. The case below is hanging on main branch.

```
varying int32 func(varying int32 a, varying int32 b);
uniform int32 func(uniform int32 a, uniform int32 b);

template <typename T>
T func(T a, T b)
{
    T r = func(a, b);
    return r;
}
uniform int foo2(uniform int a, uniform int b) {
    return func<uniform int>(a, b);
}
```

aneshlya:
The issue with `Unable to find any matching overload for call to function "func"` error is not related to this PR and happens on main branch on the following test case:
```
varying int32 func(varying int32 a, varying int32 b);
uniform int32 func(uniform int32 a, uniform int32 b);
template <typename T>
uniform T func(uniform T a, uniform T b, uniform int c)
{
    uniform T r = func(a, b);
    return r;
}
uniform int foo2(uniform int a, uniform int b, uniform int c) {
    return func<uniform int>(a, b, c);
}
```
Looks like templates and regular functions can't have the same name.
https://github.com/ispc/ispc/blob/main/src/expr.cpp#L8327-L8339

aneshlya:
Anyway the error: `Error: Unable to resolve the value of the symbol "N" used to specify the size of a vector type. ` should not happen in this case and it's fixed now. I realized that error reporting should not happen in type.cpp. If anything wrong, it must be detected earlier in pipeline.
Created #3016, #3015 for the issues above.

nurmukhametov (approved):

PR #2993, opened 2024-09-10
Run zx in all threads in CI during LLVM build

dbabokin:
`xz` by default runs in one thread. With `tar cJvf` command running more than 20 minutes during LLVM build, it makes sense to speed this up by running in all threads.

To run in all threads we can use `xz`'s arg `-T`:
```
  -T, --threads=NUM   use at most NUM threads; the default is 0 which uses
                      as many threads as there are processor cores
```

On Linux we can pass it through `XZ_DEFAULTS` env variable. But on macOS tar is `bsdtar` and for some reason it doesn't act like regular `tar` with `XZ_DEFAULTS` has no effect. But we can run it explicitly.

### DIFF
--- a/.github/workflows/reusable.rebuild.yml
+++ b/.github/workflows/reusable.rebuild.yml
@@ -101,6 +101,7 @@ jobs:
         rm -rf cache.local
         tar xvf result.tar usr/local/src/llvm
         mv usr/local/src/llvm/bin-${{ inputs.version }} .
+        export XZ_DEFAULTS="-T0" # run zx in all threads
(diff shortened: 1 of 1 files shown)


dbabokin:
I've triggered LLVM 17.0 rebuild to verify the changes: https://github.com/ispc/ispc/actions/runs/10784567271

dbabokin:
I've removed macOS part, as `zx` was not available on our runner.

Not it's ready for review.

dbabokin:
Packing time has reduced 4x from 22min to ~5:30min

nurmukhametov (approved):
LGTM!

aneshlya (approved):
Looks good!

PR #2994, opened 2024-09-11
CI: add hardening checks

nurmukhametov:
This PR adds a CI job to check hardening flags of Linux and Windows release binaries. Example of run: [link](https://github.com/nurmukhametov/ispc/actions/runs/10810929910)

### DIFF
--- a/.github/workflows/hardening-check.yml
+++ b/.github/workflows/hardening-check.yml
@@ -1,0 +1,76 @@
+## Copyright 2024 Intel Corporation
+## SPDX-License-Identifier: BSD-3-Clause
+
+name: Hardening Checks (trunk)
+
+permissions: read-all
+
+on:
+  workflow_dispatch:
+  schedule:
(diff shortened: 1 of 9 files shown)


.github/workflows/hardening-check.yml
```diff
@@ -0,0 +1,77 @@
...
+## SPDX-License-Identifier: BSD-3-Clause
+
+name: Run Hardening Checks
+
+permissions: read-all
+
+on:
+  workflow_dispatch:
+    inputs:
+      release_tag:
```
aneshlya, line 11: I think we should run it similar to ClamAV - on release and trunk builds. Otherwise if necessary compiler flag is removed, we'll know about it too late.

nurmukhametov: Done.

aneshlya (approved):
Looks good!

PR #2996, opened 2024-09-11
CI: add job to build ISPCRT

nurmukhametov:
This PR adds a CI job to build ISPCRT using different tasking libraries. Example: [link](https://github.com/nurmukhametov/ispc/actions/runs/10812832056)


aneshlya (approved):
Looks good

Issue #2902, opened 2024-06-30
Dead-code elimination of alternative implementation of exported functions


zephyr111:
Hello,

When a function is exported, ISPC usually generates 2 implementations. One which can be called from outside the translation unit (eg. from a C/C++ code) and a local one. When the local one is not useful (ie. never called nor even referenced somewhere), the dead code is not eliminated. This is common with exported functions since they are not called/reference internally most of the time.

[Here](https://godbolt.org/z/sfc1KGTW8) is an example on GodBolt where we can see that the function `compute` is compiled to 2 versions; `compute___un_3C_und_3E_unu` and `compute`. I think the former is not useful and should be automatically removed from the generated assembly code (and binaries). 

This issue results in bigger binaries and less readable assembly code (harder to analyse due to a twice bigger code).

Can this be done automatically? 
If no, is there any way to tell the compiler not to generate the `compute___un_3C_und_3E_unu` function (or to do a more aggressive dead-code optimisation)?

Version: ISPC trunk (>=1.24.0)

nurmukhametov:
It is not the compiler's job to optimize them out. It is up to the linker to do so.

Usually, in the case of `C/C++`, this is done by providing `-ffunction-sections` to the compiler and `--gc-sections` to the linker. The compiler places every function in a separate section, and the linker removes unused sections when the final executable is linked.

Unfortunately, `ISPC` does not currently have an option to place functions in separate sections. However, this can be addressed to resolve the issue.

aneshlya:
If you don't need declaration of exported function in the header file, you can try replacing `export` specifier with `extern "C"`. It will generate only one version of function. You can check the difference between `export` and `extern "C"` functions in the table [here](https://ispc.github.io/ispc.html#interoperability-overview).

zephyr111:
Thank you @aneshlya, It perfectly fits my needs : it divides by ~2 the size of the ISPC generated assembly code (which is great for debugging/profiling), reduces the size of the generated binaries and also speed up compilation!
Thank you @nurmukhametov for this information. I did not know that.

Should I close this issue or left it open (regarding the implementation of linker sections)?

aneshlya:
Let's keep it open.

pbrubaker:
If you are reviewing disassembly interactively on a tool like Compiler Explorer, you don't need to add `extern "C"` you can just remove the export keyword.

PR #2997, opened 2024-09-11
Add -ffunction-sections flag

nurmukhametov:
It forces the compiler to place each function in its own section.

This PR addresses the issue #2902. One can eliminate unused copies of exported functions providing this flag and enabling a linker flag that removes unused sections (`-gc-sections` for `ld` and compatible linkers).

### DIFF
--- a/docs/ispc.rst
+++ b/docs/ispc.rst
@@ -642,8 +642,13 @@ Updating ISPC Programs For Changes In ISPC 1.25.0
 ISPC language was extended to support ``__attribute__(())`` syntax for variable
 and function declarations. Two attributes are supported now: ``noescape`` and
 ``address_space(N)``. The macro ``ISPC_ATTRIBUTE_SUPPORTED`` is defined if the
-ISPC compiler supports attribute syntax. Please, reger to `Attributes`_ for
-more details.
+ISPC compiler supports attribute syntax. Please, refer to `Attributes`_ for
+more details and the full list of supported attributes.
+
+This release supports ``-ffunction-sections`` command line flag to generate
+each function in a separate section. This flag is useful for reducing the size
+of the final executable by removing unused functions. Please, refer to
+`Basic Command-line Options`_ for more details.
 
 Getting Started with ISPC
 =========================
@@ -851,6 +856,14 @@ for use in a shared library. The ``--PIC`` flag can be used to generate
 position-independent code suitable for dynamic linking avoiding any limit on
 the size of the global offset table. When no ``--pic`` or ``--PIC`` flag is
 provided, the compiler enforces target-specific default behavior.
+
+The ``-ffunction-sections`` flag can be used to generate each function in a
+separate section. This flag is useful for reducing the size of the final
+executable by removing unused functions when it is combined with linker flag
+that removes unused sections: ``--gc-sections`` for ``GNU ld`` and ``/OPT:REF``
+for ``MSVC link.exe``. On macOS, this flag does not have any effect (as in
+clang) because dead stripping ``-dead_strip`` for ``ld64`` works differently.
+The ``-fno-function-sections`` disables this behavior.
 
 Selecting The Compilation Target
 --------------------------------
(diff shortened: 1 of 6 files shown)


src/main.cpp
```diff
@@ -104,6 +104,7 @@ static void lPrintVersion() {
     printf("        fast\t\t\t\tUse high-performance but lower-accuracy math functions\n");
     printf("        svml\t\t\t\tUse the Intel(r) SVML math libraries\n");
     printf("        system\t\t\t\tUse the system's math library (*may be quite slow*)\n");
+    printf("    [-ffunction-sections]\t\t\tPlace each function in its own section\n");
```
aneshlya, line 107: Maybe 
```suggestion
    printf("    [-f[no]function-sections]\t\t\tPlace each function in its own section\n");
```

aneshlya (approved):
Looks good!

src/main.cpp
nurmukhametov: Done

src/main.cpp
```diff
@@ -104,6 +104,7 @@ static void lPrintVersion() {
     printf("        fast\t\t\t\tUse high-performance but lower-accuracy math functions\n");
     printf("        svml\t\t\t\tUse the Intel(r) SVML math libraries\n");
     printf("        system\t\t\t\tUse the system's math library (*may be quite slow*)\n");
+    printf("    [-f[no-]function-sections]\t\t\tPlace each function in its own section\n");
```
dbabokin, line 107:
```suggestion
    printf("    [-f[no-]function-sections]\t\tPlace each function in its own section\n");
```

dbabokin (changes requested):
The changes don't work on macOS - the function is still in TEXT section.

But I tried clang and had the same behavior, i.e. no effect on emitted assembler. Do you know why? It's probably ok to have the same behavior as clang (i.e. accept the option which has no effect), but would be good to understand what's going on.

nurmukhametov:
> The changes don't work on macOS - the function is still in TEXT section.
> 
> But I tried clang and had the same behavior, i.e. no effect on emitted assembler. Do you know why? It's probably ok to have the same behavior as clang (i.e. accept the option which has no effect), but would be good to understand what's going on.

Yes. It looks like a specific mechanism is slightly different with macOS toolchain. Although, I didn't manage to find quickly what it is. Anyway, the usage of `-dead_strip` linker option ([docs](https://github.com/llvm/llvm-project/blob/9cd93774098c861c260090a690f428b7ae031c65/clang/include/clang/Basic/AttrDocs.td#L83)) eliminates unused copies regardless of `-ffunction-sections` flag.

src/main.cpp
nurmukhametov: Done

dbabokin (approved):
Could you put this knowledge about macOS (that it doesn't work in the same way on macOS, that symbols are stripped anyway, and that we are following clang, which does have this option on macOS, but it does nothing) to comments in the test?

After that it's good to merge.

nurmukhametov:
> Could you put this knowledge about macOS (that it doesn't work in the same way on macOS, that symbols are stripped anyway, and that we are following clang, which does have this option on macOS, but it does nothing) to comments in the test?
> 
> After that it's good to merge.

I have added this comment and updated docs.

docs/ispc.rst
```diff
@@ -852,6 +857,14 @@ position-independent code suitable for dynamic linking avoiding any limit on
 the size of the global offset table. When no ``--pic`` or ``--PIC`` flag is
 provided, the compiler enforces target-specific default behavior.
 
+The ``-ffunction-sections`` flag can be used to generate each function in a
+separate section. This flag is useful for reducing the size of the final
+executable by removing unused functions when it is combined with linker flag
+that removes unused sections: ``--gc-sections`` for ``GNU ld`` and ``/OPT:REF``
+for ``MSVC link.exe``. On macOS, this flag does not have any effect (as clang)
```
dbabokin, line 864:
```suggestion
for ``MSVC link.exe``. On macOS, this flag does not have any effect (as in clang)
```

Issue #2995, opened 2024-09-11
Add compiler option to only emit C++ callable export functions


rkaric:
Given the following code: https://godbolt.org/z/E3449vvdG

ISPC generates two callable functions, one for C++ and another for C. It would be very helpful to have a compiler option to only generate the C++ version.

If the C version is unused it will be removed as dead code in optimized builds. This would still be useful for Debug builds and generally to improve compile times.

nurmukhametov:
It looks same as #2902.

Indeed, there are two versions but one is for C/C++ (unmangled one `ArrayMax`) and the other is for ISPC code (`ArrayMax___un_3C_Cunf_3E_un_3C_Cunf_3E_un_3C_unf_3E_Cuni` mangled one and contains an extra mask argument).

If some function is for external only calls (from C/C++) then you may specify `extern "C"` instead of `export`, i.e., `extern "C"` is the option you are looking for.

rkaric:
While that is an option, there are situations where you want to exclusively call the C++ version and not the C version. The generate functions prototypes are placed into their own namespace and declared for you as part of the compilation process. This is not the case when using `extern "C"`.

One other thing to note, since it's up to the linker to remove dead code there may variation between toolchains on how well that works. Epic Games has reported that the Apple toolchain does not remove the dead C code while I've confirmed that MSVC does.

nurmukhametov:
I don't understand what it is C++ version you are talking about. In the generated code there are only two functions, the one is for ISPC code, the other for C/C++.

nurmukhametov:
> This is not the case when using `extern "C"`.

Ah, OK. The problem of `extern "C"` is that ISPC doesn't generate declarations for them in the `-h` header file (it is unexpected for me).

Probably, there are twoapproaches (except the `-ffunction-sections`/`--gc-section` mentiond #2902):
1. Add declarations for `extern "C"` functions to the generated header file. Users then need to change `export` to `extern "C"` in their code to avoid extra copies.
2. Add an extra qualifier for `export` keyword, e.g., `export "C"`. In that case, ISPC generates only the externally called function and add its declaration as it happens now for `export` keyword. Users need to change `export` to `export "C"` to avoid an extra function copy. We may also have `export "ISPC"`.

rkaric:
> I don't understand what it is C++ version you are talking about. In the generated code there are only two functions, the one is for ISPC code, the other for C/C++.

[...]

PR #3002, opened 2024-09-13
Introduce external_only attribute

nurmukhametov:
If ~~export "C" ``__attribute__((remove_ispc_copy))``~~  ``__attribute__((external_only))`` is provided then generate only C/C++ version of exported function.

It addresses #2995

### DIFF
--- a/docs/ispc.rst
+++ b/docs/ispc.rst
@@ -650,6 +650,13 @@ each function in a separate section. This flag is useful for reducing the size
 of the final executable by removing unused functions. Please, refer to
 `Basic Command-line Options`_ for more details.
 
+In some cases like shared libraries, the mentioned ``-ffunction-sections`` flag
+is not enough to remove unused ISPC copies of exported functions. To address
+this, the ``external_only`` attribute was introduced. It can be applied only
+to exported functions and it instructs the compiler to remove the ISPC version
+of the function. Please, refer to `Attributes`_ and
+`Functions and Function Calls`_ for more details.
+
 Getting Started with ISPC
 =========================
 
@@ -2958,6 +2965,17 @@ to the additional features they introduce.
     __attribute__((unmangled)) void foo(int a, int b);
 
 
+external_only
+----------------
+
+``__attribute__((external_only))`` can be applied to a function with
+``export`` qualifier. It informs the compiler that it should not generate a
+ISPC version of the function. This is useful for functions that are only called
+from C/C++ in case when the user wants to reduce the size of the generated
+code. Same effect can be achieved by using ``-ffunction-sections`` compiler
+option but not in all cases (e.g., shared libraries with ISPC code), so this
+attribute is provided as a more fine-grained control.
+
 Expressions
 -----------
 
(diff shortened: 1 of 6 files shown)


tests/lit-tests/export-c.ispc
```diff
@@ -0,0 +1,7 @@
+// RUN: %{ispc} --target=host --emit-asm %s -o - 2>&1 | FileCheck %s
+
+// CHECK-NOT: foo___:
+// CHECK-NOT: Error: syntax error, unexpected "C".
+// CHECK: foo:
+
+export "C" void foo() { return; }
```
aneshlya, line 7: I would also check that `export "C"` function is present in generated header file.

aneshlya:
The implementation looks fine. The only thing remain is to describe the new qualifier in docs.

However, I’m still hesitant about introducing a new function keyword (as I mentioned in #2995). The main concern is that it can be difficult to explain to people when and how to use it correctly. We already have a table in the documentation outlining all supported function qualifiers, but the reality is that most people don’t read the docs.

How can we effectively communicate the purpose of this new keyword?

One idea is to generate a warning that suggests using `export "C"` instead of `export` when the internal ISPC version isn’t called inside the module (which I believe will cover about 95% of cases).

An alternative approach would be to introduce a compiler option (e.g., --strip-unused or any other appropriate name), which would treat all ISPC functions (except for `export` external function and `extern "C"`) as internal by default, allowing unused ones to be removed through DCE. This would avoid the need to modify existing ISPC code, though it would require changes to the build scripts.

Let me know your thoughts and why you prefer introducing a new function keyword over adding a compiler option.

aneshlya:
As we discussed, let's implement a new `attribute` instead of `export "C"` to mark the "external only" functions. This approach is more explicit and avoids adding the additional complexity of introducing a new function keyword.

nurmukhametov:
> As we discussed, let's implement a new `attribute` instead of `export "C"` to mark the "external only" functions. This approach is more explicit and avoids adding the additional complexity of introducing a new function keyword.

Done

tests/lit-tests/export-c.ispc
nurmukhametov: Done

docs/ispc.rst
```diff
@@ -2958,6 +2964,17 @@ to the additional features they introduce.
     __attribute__((unmangled)) void foo(int a, int b);
 
 
+remove_ispc_copy
+----------------
+
+``__attribute__((remove_ispc_copy))`` can be applied to a function with
+``export`` qualifier. It informs the compiler that it should not generate a
```
aneshlya, line 2971: I tried to find a place in the docs where we explain that `export` function produces 2 variants and I can't find any mention of it. It would be helpful to include this information, and the 'Functions and Function Calls' section seems like a good place for it. It would also be useful to add a link to this section here

nurmukhametov: I have added explanation about two function versions to the mentioned section.

aneshlya (approved):
LGTM!

nurmukhametov:
> I don't really like the name of the attribute. `remove_ispc_copy` sounds like too many words.
> 
> What about, for example, `external_only`? What do you think?

I have no preference regarding naming, so I have changed to `external_only`.

dbabokin (approved):

aneshlya (approved):

Issue #2870, opened 2024-05-29
Loop counter codegen


nurmukhametov:
Consider the following code:

```ispc
unmasked void foo(uniform float Data[], const uniform int N) {
        foreach(i = 0 ... N) {
                Data[i] = Data[i] + 1;
        }
}
```

The inner loop asm with ISPC v1.20.0 is:
```asm
.LBB1_3:
        movsxd  rdx, edx
        vaddps  ymm1, ymm0, ymmword ptr [rdi + rdx]
        vmovups ymmword ptr [rdi + rdx], ymm1
        add     eax, 8
        add     edx, 32
        cmp     eax, ecx
        jl      .LBB1_3
```

It has redundant instruction `movsxd` that was already reported on similar test case #2611 and was supposed to be fixed by #2494. 
 However, since v1.21.0 ISPC generates:

```asm
.LBB1_3:
        mov     r8d, edx
        and     r8d, -32
        vaddps  ymm1, ymm0, ymmword ptr [rdi + r8]
        vmovups ymmword ptr [rdi + r8], ymm1
        add     rax, 8
        add     rdx, 32
        cmp     rax, rcx
        jb      .LBB1_3
```

that has `mov` and `and` instead of `movsxd`.

The reverted change #2864 transformed this code to

```asm
.LBB0_3:
        mov     r8d, edx
        vaddps  ymm1, ymm0, ymmword ptr [rdi + r8]
        movsxd  rdx, edx
        vmovups ymmword ptr [rdi + rdx], ymm1
        add     rax, 8
        add     edx, 32
        cmp     rax, rcx                                                                                                                                                                                                                                            
        jb      .LBB0_3
```

All these versions look not optimal, e.g., GCC generates for similar loops only the following:

```asm
.L4:
        vaddps  ymm0, ymm1, YMMWORD PTR [rax]
        add     rax, 32
        vmovups YMMWORD PTR [rax-32], ymm0
        cmp     rax, rdx
        jne     .L4
```

Compiler explorer link with examples: https://ispc.godbolt.org/z/hbYPr6soK

nurmukhametov:
With `--addressing=64`, ISPC generates code similar to GCC:
```asm
.LBB1_3:
        vaddps  ymm1, ymm0, ymmword ptr [rdi + 4*rax]
        vmovups ymmword ptr [rdi + 4*rax], ymm1
        add     rax, 8
        cmp     rax, rcx
        jb      .LBB1_3
```

[...]

PR #3009, opened 2024-09-16
Propagate IRFlags for binary operators in lExtractFirstVectorElement

nurmukhametov:
This PR fixes issue #2870.

It propagates IR flags from vector operations to extracted scalar ones. This influences the quality of loop counter code. `nsw`/`nuw` flags were assigned to vector operations in https://github.com/ispc/ispc/commit/a3975be118ffdb8f52151ab25ce84cfceb0a1478 to exploit UB to extend loop counters (to 64-bit) avoiding extra redundant sext/zext instructions.

After this, we can reapply https://github.com/ispc/ispc/commit/99a78930624aed76b401e52a4ebae5a77a31b10d

### DIFF
--- a/ispc-opt.cpp
+++ b/ispc-opt.cpp
@@ -30,6 +30,8 @@
 
 using namespace llvm;
 
+static cl::opt<int> Addressing("addressing", cl::desc("Select 32- or 64-bit addressing."), cl::init(32),
+                               cl::value_desc("model"));
 static cl::opt<std::string> Passes("passes", cl::desc("Passes to run"));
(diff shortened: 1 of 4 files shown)


aneshlya (approved):
LGTM!
Looks like the test should be updated for MacOS.

dbabokin (commented):
This is potentially risky change, so would be good to have an option to disable it. `[--[no-]wrap-signed-int]` switch seems to be relevant for that.

nurmukhametov:
> This is potentially risky change, so would be good to have an option to disable it. `[--[no-]wrap-signed-int]` switch seems to be relevant for that.

Done

PR #3012, opened 2024-09-17
Add clang-tidy script and CI job

### LINKED ISSUE #3005: Add clang-tidy github workflow
Add clang-tidy check with the following configuration:
```
Checks: '-*, bugprone-*, cert-*, clang-analyzer-*, cppcoreguidelines-*, hicpp-*, misc-*, modernize-*, readability-*,'

CheckOptions:
  - key: readability-function-cognitive-complexity.Threshold
    value: 15
  - key: readability-function-cognitive-complexity.DescribeBasicIncrements
    value: false
  - key: readability-function-size.LineThreshold
    value: 100
  - key: readability-function-size.StatementThreshold
    value: 50

FormatStyle: 'none'
```

nurmukhametov:
This PR fixes #3005.

At the moment, it is not enabled by default because there are a lot of warnings. Example run [here](https://github.com/nurmukhametov/ispc/actions/runs/10904199532).


clang_tidy.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
aneshlya, line 1: Lets' move this file to the scripts folder.

clang_tidy.sh
```diff
@@ -0,0 +1,29 @@
...
+#  Copyright (c) 2024, Intel Corporation
+#  SPDX-License-Identifier: BSD-3-Clause
+
+set -e
+
+BUILD_FODLER=${1:-"build"}
```
aneshlya, line 7:
```suggestion
BUILD_FOLDER=${1:-"build"}
```

clang_tidy.sh
```diff
@@ -0,0 +1,29 @@
...
+FILES=$(ls                                  \
+    src/*.{cpp,h}                           \
+    src/opt/*.{cpp,h}                       \
+    *.cpp                                   \
+    builtins/*{cpp,hpp,c}                   \
+    common/*.h                              \
+)
+
+# Run clang-tidy with the compilation database from the given build folder in parallel
+NUM_CORES=$(nproc)
```
aneshlya, line 28: You can use `run-clang-tidy` for parallel run:
https://github.com/llvm/llvm-project/blob/main/clang-tools-extra/clang-tidy/tool/run-clang-tidy.py

.github/workflows/clang-tidy.yml
```diff
@@ -0,0 +1,52 @@
...
+      with:
+        submodules: true
+
+    - name: Install dependencies
+      run: |
+        .github/workflows/scripts/install-build-deps.sh
+
+    - name: Install clang-tidy
+      run: |
+        sudo apt install -y clang-tidy
```
aneshlya, line 38: Let's pin clang-tidy version

.clang-tidy
```diff
@@ -0,0 +1,13 @@
+Checks: '-*,bugprone-*,cert-*,clang-analyzer-*,cppcoreguidelines-*,hicpp-*,misc-*,modernize-*,readability-*'
```
aneshlya, line 1: Let's reduce the number of checks. Could you review the logs to identify the most useful diagnostics and retain only those checks for now (perhaps just `cppcoreguidelines` and `bugprone`)? This will allow us to address the issues and then run clang-tidy with a minimal configuration on each PR.

aneshlya (changes requested):

dbabokin (commented):
I think it should run for every PR, isn't it?

clang_tidy.sh
nurmukhametov: Done

clang_tidy.sh
nurmukhametov: Fixed

.clang-tidy
nurmukhametov: I disabled some checks that were noisy or not really useful. For others, fixes are provided. From what I saw, I think these two are useful `misc-include-cleaner`, `misc-unused-parameters`, but before enabling we need to fix their warnings.

nurmukhametov:
> I think it should run for every PR, isn't it?

Yes, sure.

dbabokin (approved):

aneshlya (approved):
Looks good!

PR #3014, opened 2024-09-17
Add release LLVM 19.1 support

dbabokin:
- Added `19.1.0` tag support for both `alloy.py` and superbuild.
- Since `19.1` LLVM's config flag `LLVM_ENABLE_TERMINFO` was removed and the default is to build without terminfo, more details are [here](https://github.com/llvm/llvm-project/blob/a4bf6cd7cfb1a1421ba92bca9d017b49936c55e4/llvm/docs/ReleaseNotes.rst?plain=1#L108). So I made it conditional depending on the version. Also I added this flag to the build through `alloy.py`.
- Fixed python warnings in `alloy.py` - strings which look like escape sequence (specifically `\.` and `\;`) and need to be keep this way need to be raw-strings.


aneshlya (approved):
Look good!

PR #3027, opened 2024-09-25
CMake: link with tinfo and pthread if LLVM requires them

### LINKED ISSUE #2989: Use `llvm-config` to figure out list of system libs to link
After [recent change](https://github.com/ispc/ispc/issues/2855) to remove the dependency on `ncurses`, we build LLVM without `ncurses` and do not link it.

So when using LLVM that was built with `ncurses` support, ISPC fails to build, as `ncurses` dependency is not satisfied. While it's the right thing to do for release builds, I think we should support using any LLVM. We'll get a bunch of complains from packages maintainers for sure after the next release.

`llvm-config` has `--system-libs` switch, which reports the list of system libraries that LLVM libs need to be linked against. I think this needs to be used to figure out if `ncurses` should be linked or not.

nurmukhametov:
This PR fixes #2989

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -816,10 +816,19 @@ if (WIN32)
         list(APPEND LINK_LIBRARIES ws2_32.lib)
     endif()
 else()
-    # ISPC doesn't use pthreads directly, but LLVM does. Although, it looks
-    # like we do not use LLVM code that depend on pthreads, don't we?
-    find_library(PTHREAD_LIB REQUIRED NAMES pthread)
-    list(APPEND LINK_LIBRARIES ${PTHREAD_LIB})
+    run_llvm_config(LLVM_SYSTEM_LIBS "--system-libs")
+    if (LLVM_SYSTEM_LIBS MATCHES "pthread")
+        # ISPC doesn't use pthreads directly, but some LLVM does. Although, it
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -815,10 +815,19 @@ if (WIN32)
...
-    list(APPEND LINK_LIBRARIES ${PTHREAD_LIB})
+    run_llvm_config(LLVM_SYSTEM_LIBS "--system-libs")
+    if (LLVM_SYSTEM_LIBS MATCHES "pthread")
+        # ISPC doesn't use pthreads directly, but some LLVM does. Although, it
+        # looks like we do not use LLVM code that depend on pthreads, don't we?
+        find_library(PTHREAD_LIB REQUIRED NAMES pthread)
+        message(STATUS "Link with pthread library ${PTHREAD_LIB}")
+        list(APPEND LINK_LIBRARIES ${PTHREAD_LIB})
+    endif()
+    if (LLVM_SYSTEM_LIBS MATCHES "tinfo")
```
dbabokin, line 826: Here's what `llvm-config --system-libs` returns on my system:
```
-lm -lcurses -lxml2
```

So this fix doesn't work for my environemnt.

dbabokin (approved):
Works for me, looks good.

Issue #2737, opened 2024-01-24
Add support for non-type template arguments


atafra:
Having support for type template arguments is very useful but it's often insufficient. In many cases, non-type arguments, e.g. int or enum, are also very important. A typical use case for these are high-performance kernels (e.g. convolution, machine learning) which require several different variants, each having different parameters (most commonly integer ones, e.g. kernel size). Sometimes the lack of template arguments could be worked around with macros, which is very cumbersome and also has a major limitation: it's not possible to add pragmas inside a macro (e.g. `#pragma unroll`). This can also be worked around but it's even more awkward, and doesn't scale well with the number of required parameter combinations.

nurmukhametov:
Support of numerical template parameters can unlock a reasonable way to implement stdlib functions for short vectors as requested, e.g., in #2707.

Something like this for `max` (although it needs to be defined in the header of ISPC stdlib that is not exist yet):
```ispc
template <typename T, uint N>
uniform T<N> max(uniform const T<N> a, uniform const T<N> b)
{
    uniform T<N> r;
    foreach (i = 0 ... N) {
        r[i] = max(a[i], b[i]);
    }
    return r;
}

uniform double<4> foo(uniform const double<4> a, uniform const double<4> b) {
    return max<double, 4>(a, b);
}
```

samayala22:
Another use case would be to simplify code generation for various indexing behaviors. 

Example: 

```
template<typename T>
inline void axpy(uniform int n, uniform T alpha, uniform T x[], uniform int incx, uniform T y[], uniform int incy) {
    // This can be replaced by a template once it is supported
    if (incx == 1 && incy == 1) {
        #pragma unroll(4)
        foreach(i = 0 ... n) {
            y[i] += alpha * x[i]; // 2 loads + 1 store
        }
    } else if (incx == 1 && incy != 1) {
        #pragma unroll(4)
        foreach(i = 0 ... n) {
            y[i * incy] += alpha * x[i]; // 1 gather + 1 load + 1 scatter
        }
    } else if (incx != 1 && incy == 1) {
        #pragma unroll(4)
        foreach(i = 0 ... n) {
            y[i] += alpha * x[i * incx]; // 1 load + 1 gather + 1 store
        }
    } else {
        #pragma unroll(4)
        foreach(i = 0 ... n) {
            y[i * incy] += alpha * x[i * incx]; // 2 gathers + 1 scatter
        }
    }
}
```

could be replaced by:

```
[...]

PR #3028, opened 2024-09-25
Support non-type template parameters in arrays

aneshlya:
This is continuation of non-type template parameters support. This PR adds support of template parameters in arrays.
Related to #2522 and https://github.com/ispc/ispc/issues/2737

### DIFF
--- a/src/type.h
+++ b/src/type.h
@@ -696,7 +696,12 @@ class ArrayType : public SequentialType {
 
     int GetElementCount() const;
     const Type *GetElementType() const;
-
+    /* Checks if the array is unsized.
+       An array is considered unsized if its size is not explicitly set
+       as compile-time constant i.e. `fixedCount` is 0 and `symbolCount`
+       is nullptr. Unsized arrays are typically used as functions parameters.
+    */
+    bool IsUnsized() const;
     /** Returns a new array of the same child type, but with the given
         length. */
     virtual ArrayType *GetSizedArray(int length) const;
--- a/src/type.h
+++ b/src/type.h
@@ -127,7 +127,23 @@ class Type : public Traceable {
     bool IsNumericType() const { return IsFloatType() || IsIntType(); }
 
     /** Returns true if the type is any kind of dependent type. */
-    bool IsDependentType() const;
+    bool IsDependent() const;
+
+    /** This function checks whether the size of the collection (e.g., an array or vector)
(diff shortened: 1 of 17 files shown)


nurmukhametov (approved):
LGTM!

src/type.h
```diff
@@ -626,6 +626,8 @@ class ArrayType : public SequentialType {
                             example).
      */
     ArrayType(const Type *elementType, int numElements);
+    ArrayType(const Type *elementType, Symbol *num);
```
dbabokin, line 635: The different type of element count argument probably requires some comments about their intended usage scenarios and meaning.

src/type.h
```diff
@@ -127,7 +127,13 @@ class Type : public Traceable {
     bool IsNumericType() const { return IsFloatType() || IsIntType(); }
 
     /** Returns true if the type is any kind of dependent type. */
-    bool IsDependentType() const;
+    bool IsDependent() const;
+
+    /** Returns true if the type is count dependent (based on template non-type parameter). */
+    bool IsCountDependent() const;
```
dbabokin, line 133: The term "count dependent" is not a standard thing. Hence it requires a definition. I would assume that's only a vector and array types with a count parameter dependent on the non-type template parameter (so things like non-type template parameters being used in other contexts are not included). But dichotomy "type dependent" vs "count dependent" suggests that expressions that use non-type template parameters are also "count dependent".

I.e.
```cpp
template <typename T, int N> int foo(T in) {
    int my_var = N - 1; // is "N-1" count dependent?
    T array[N];
    array[my_var] = 123;
    return array[my_var];
}
```

src/type.h
```diff
@@ -672,7 +672,8 @@ class ArrayType : public SequentialType {
 
     int GetElementCount() const;
     const Type *GetElementType() const;
-
+    /* Returns true if array is unsized*/
```
dbabokin, line 675: Again, would be give some definition of "unsized".

src/type.h
aneshlya: Done

src/type.h
aneshlya: Done

src/type.h
aneshlya: Done

dbabokin (approved):
Looks good to me

PR #3029, opened 2024-09-26
CMake: use explicit x86 target for dispatch

### LINKED ISSUE #2990: ISPC build on ARM targeting x86 is broken
```
[  3%] Generating share/ispc/builtins_dispatch.bc
cd /Users/dbabokin/ispc/dbabokin_ispc && /Users/dbabokin/llvm/bin-no-ncureses/bin/clang++ -x c -DREGULAR -fPIC -O2 -emit-llvm -c builtins/dispatch.c -o /Users/dbabokin/ispc/dbabokin_ispc/build/bin/../share/ispc/builtins_dispatch.bc
builtins/dispatch.c:33:36: error: invalid output constraint '=a' in asm
   33 |     __asm__ __volatile__("cpuid" : "=a"(info[0]), "=b"(info[1]), "=c"(info[2]), "=d"(info[3]) : "0"(infoType));
      |                                    ^
builtins/dispatch.c:37:36: error: invalid output constraint '=a' in asm
   37 |     __asm__ __volatile__("cpuid" : "=a"(info[0]), "=b"(info[1]), "=c"(info[2]), "=d"(info[3]) : "0"(level), "2"(count));
      |                                    ^
builtins/dispatch.c:45:53: error: invalid output constraint '=a' in asm
   45 |     __asm__ __volatile__(".byte 0x0f, 0x01, 0xd0" : "=a"(rEAX), "=d"(rEDX) : "c"(0));
      |                                                     ^
builtins/dispatch.c:57:53: error: invalid output constraint '=a' in asm
   57 |     __asm__ __volatile__(".byte 0x0f, 0x01, 0xd0" : "=a"(rEAX), "=d"(rEDX) : "c"(0));
      |                                                     ^
4 errors generated.
```

I'm macOS, but this should not be different on Linux or Windows. The problem is that `clang++` is run in "native" mode, i.e. the target is not specified. So x86 inline asm doesn't work. Any way, the dispatch is assumed to be x86 anyway.

Adding `--target=x86_64-apple-macosx` in my case fixes the problem. Though I'm not sure what generic x86 target we should use here to make it work on all platforms.

nurmukhametov:
This PR fixes #2990

### DIFF
--- a/builtins/dispatch.c
+++ b/builtins/dispatch.c
@@ -22,7 +22,8 @@
 #endif
 
 #include <stdint.h>
-#include <stdlib.h>
+
+void abort();
 
 // __system_best_isa __get_system_isa and __set_system_isa are weak symbols
 // because we want to have the only version of them across user application
--- a/cmake/GenerateBuiltins.cmake
+++ b/cmake/GenerateBuiltins.cmake
@@ -116,7 +116,7 @@ function(generate_dispatcher os)
 
     add_custom_command(
         OUTPUT ${bc}
-        COMMAND ${CLANGPP_EXECUTABLE} -x c ${ISPC_OPAQUE_FLAGS} ${DISP_TYPE} ${EXTRA_OPTS} -O2 -emit-llvm -c ${input} -o ${bc}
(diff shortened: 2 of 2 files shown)


dbabokin (changes requested):
It doesn't work for me, as I mentioned in #2990, I'll be able to look at it a bit later.

dbabokin:
See my comment here: https://github.com/ispc/ispc/issues/2990#issuecomment-2379825954

aneshlya (approved):
Looks OK to me!

dbabokin (approved):
This builds nicely with x86 targets enabled on my ARM MacBook.

Issue #2661, opened 2023-09-22
Template parameterized type loses `uniform` qualifier when calling another template function (ISPC 1.21.0)


mikemccarty-vertex:
Sample code (https://gcc.godbolt.org/z/To8ofGWWG):
```
template<typename UVFloat, typename UVInt>
inline UVFloat level2(UVInt v) {
    return v * 1.f;
}

template<typename UVFloat, typename UVInt>
inline UVFloat level1(UVInt v) {
    return level2<UVFloat,UVInt>(v);
}

uniform float testI(uniform int i) {
#if 1
    return level1<uniform float,uniform int>(i);
#else
    return level2<uniform float,uniform int>(i);
#endif
}
```

This code fails to compile with error:
```
<source>:9:12: Error: Can't convert from type "varying float" to type 
        "uniform float" for return statement. 
    return level2<UVFloat,UVInt>(v);
           ^^^^^^^^^^^^^^^^^^^^^^^^
```

Toggling the `#ifdef` to call `level2` directly allows it to compile.

PR #3031, opened 2024-09-27
Fix incorrect template argument variability (#2661)

aneshlya:
Fixes https://github.com/ispc/ispc/issues/2661.


nurmukhametov (approved):
LGTM!

PR #3032, opened 2024-09-27
CI: use LLVM 18.1 by default

aneshlya:
Switch CI jobs to use LLVM 18.1 by default.

Do not merge until all tests are passed: https://github.com/aneshlya/ispc/actions/runs/11078718054

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -5,7 +5,7 @@
 configuration: Release
 
 environment:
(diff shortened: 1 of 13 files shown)


dbabokin (approved):
Looks all good.

Just one remark - we should probably consider removing Appveyor and adding "latest" build through Girhub Releases and Actions.

Issue #2998, opened 2024-09-12
Add RK superbuild to ISPC actions


aneshlya:
Create workflow to build RK superbuild: https://github.com/RenderKit/superbuild
Depending on how much time it takes run it on each PR or daily.

PR #3035, opened 2024-09-30
ci: add RK superbuild

aneshlya:
Fixes https://github.com/ispc/ispc/issues/2998.
(already uses LLVM 18. for trunk packages).


nurmukhametov (approved):
LGTM!

PR #3039, opened 2024-10-01
Add Dockerfile for v1.25.0 release

nurmukhametov:
This PR adds Dockerfile for v1.25.0 release. It is based on `ubuntu18.04` because of `centos7` EOL. It means that release ISPC binary will require at least GLIBC version 2.27.

### DIFF
--- a/docker/v1.25.0/ubuntu18.04/Dockerfile
+++ b/docker/v1.25.0/ubuntu18.04/Dockerfile
@@ -13,6 +13,7 @@ ARG SHA=v1.25.0
 ARG LTO=ON
 ARG PGO=OFF
 ARG TBB=default
+ARG XE_DEPS=ON
 
 # !!! Make sure that your docker config provides enough memory to the container,
 # otherwise LLVM build may fail, as it will use all the cores available to container.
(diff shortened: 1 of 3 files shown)


docker/v1.25.0/ubuntu18.04/Dockerfile
```diff
@@ -0,0 +1,101 @@
...
+#  SPDX-License-Identifier: BSD-3-Clause
+
+
+FROM ubuntu:18.04
+LABEL maintainer="Arina Neshlyaeva <arina.neshlyaeva@intel.com>"
+SHELL ["/bin/bash", "-c"]
+
+ARG REPO=ispc/ispc
+ARG SHA=main
+ARG LTO=OFF
```
aneshlya, line 13: For release docker I'd update default arguments values.
```suggestion
ARG SHA=v1.25.0
ARG LTO=ON
```

docker/v1.25.0/ubuntu18.04/Dockerfile
```diff
@@ -0,0 +1,101 @@
...
+    ./configure CFLAGS=-fPIC CXXFLAGS=-fPIC && make -j"$(nproc)" && make install && \
+    popd && rm -rf /usr/local/src/*
+
+# If you are behind a proxy, you need to configure git.
+RUN if [ -v http_proxy ]; then git config --global --add http.proxy "$http_proxy"; fi
+
+ENV LLVM_HOME=/usr/local/src/llvm
+WORKDIR /usr/local/src
+
+ENV PATH=$LLVM_HOME/bin-$LLVM_VERSION/bin:$PATH
```
aneshlya, line 66: i can't find where LLVM_VERSION is set.

aneshlya:
For previous release there was Ubuntu 18.04 Dockerfile: https://github.com/ispc/ispc/blob/main/docker/v1.24.0/ubuntu18.04/Dockerfile. Was it used for ARM binaries?
Does the v1.25.0 Dockerfile have everything what is needed to build ARM binaries?

aneshlya:
Let's pin version of container image and the versions of packages that are used to link with ISPC.

docker/v1.25.0/ubuntu18.04/Dockerfile
nurmukhametov: Done

docker/v1.25.0/ubuntu18.04/Dockerfile
nurmukhametov: Right, this is actually redundant.

nurmukhametov:
> For previous release there was Ubuntu 18.04 Dockerfile: https://github.com/ispc/ispc/blob/main/docker/v1.24.0/ubuntu18.04/Dockerfile. Was it used for ARM binaries? Does the v1.25.0 Dockerfile have everything what is needed to build ARM binaries?

Yes, we used cpu_image of ubuntu18.04 to build arm Linux binaries. I have added `XE_DEPS` arg for ubuntu18.04 docker that allows us to build CPU release either for ARM.

nurmukhametov:
> Let's pin version of container image and the versions of packages that are used to link with ISPC.

Done

aneshlya (approved):
Looks good!
Feel free to merge if you've tested the docker.

PR #3043, opened 2024-10-07
Bunch of CI and compiler build fixes

aneshlya:
1. Use clang-format 18 from pre-built LLVM binaries + simplify .clang-format
2. Use concurrency group
3. Nightly-17 workflow duplicates ispc-ci.yml for LLVM 18. Lets' use it to test with LLVM 19. Currently we can't add LLVM 19 to ispc-ci workflow because matrix is too big.
4. Fix CI runs in forks
5. Fix trunk build: use llvm::Intrinsic::lookupIntrinsicID
6. Minor fixes


.github/workflows/format-check.yml
```diff
@@ -7,9 +7,22 @@ permissions: read-all
 
 on: [push, pull_request]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  SDE_MIRROR_ID: 813591
```
dbabokin, line 15: Install SDE to run clang-format? No way.

Pulling our llvm distro is ok, but isn't it easier just to do `apt install clang-format-18`?

dbabokin: > apt installs clang-format 18.1.3 and it has different formatting in comparison to latest 18.1.8.

😨 😨 😨

dbabokin (approved):
Looks good!

nurmukhametov (approved):

PR #3049, opened 2024-10-10
Copy shl flags in ImproveMemoryOps

nurmukhametov:
This PR is the follow-up fix to https://github.com/ispc/ispc/issues/2870

The main reason to propagate nsw/nuw flags from operations above arguments of pseudo_scatter to corresponding operations of pseudo_factored_base_scatter. This influences the quality of loop counter code.  nsw/nuw flags are assigned to vector operations in a3975be118ffdb8f52151ab25 to exploit UB to extend loop counters and some induction variables to wider type to avoid sext/zext instructions.


aneshlya (approved):
It looks good but we need to understand (not as part of this PR) if we need to copy nsw flags from some other instructions.
Feel free to merge when tests are passing.

aneshlya (approved):

Issue #3048, opened 2024-10-10
NaN handling on ARM NEON for min/max


johguenther:
For min (analog for max) ISPC issues `fmin` or `vmin` instructions, which are not IEEE 754 compliant: in case any operand is NaN then NaN is the result. I suggest to better choose the alternative instructions `fminnm` / `vminnm`: those follow IEEE 754 and select the numerical value instead if NaN is one operand and are thus consistent to x86 behavior. OSPRay uses min/max to filter out NaNs.

aneshlya:
Will be fixed in v1.25.0.
Be aware that `vminnm/vmaxnm` will not be generated by default for `--arch=arm`. In addition you need to specify `--cpu=cortex-a35` or newer.
`fminnm/fmaxnm` will be generated by default when `--arch=aarch64`  is selected.

PR #3050, opened 2024-10-10
Generate fminnm/vminnm instructions on arm

aneshlya:
Fixes https://github.com/ispc/ispc/issues/3048.
`fminnm`/`fmaxnm` are supported for aarch64 targets.
`vminnm`/`vmaxnm` are supported on arm targets with fp-armv8 (starting cortext-a35).


dbabokin (approved):
looks good

PR #3051, opened 2024-10-11
Add CPU definitions for ARL, LNL and GNR

aneshlya:
ARL and LNL was added into LLVM 18.
GNR is in LLVM since LLVM 16.
Link to [X86TargetParser.cpp](https://github.com/llvm/llvm-project/blob/llvmorg-18.1.8/llvm/lib/TargetParser/X86TargetParser.cpp) for easier review


dbabokin (approved):
Looks good!

PR #3053, opened 2024-10-11
Release notes v1.25.0

aneshlya:


docs/ReleaseNotes.txt
```diff
@@ -1,3 +1,113 @@
...
+
+Deprecated targets:
+
+ - `avx512knl-x16`, `gen9-x8` and `gen9-x16` targets are deprecated and will be
+   removed in future releases.
+ - Adjusted device names for `xelpg-x8` and `xelpg-x16` to `mtl-u` and `mtl-h`.
+
+Standard library:
+
+ - Fixed integer overflow in `exp` for -Inf, NaN and big numbers.
```
dbabokin, line 51: The wording is a bit confusing, as it's floating point `exp()`, while overflow is integer.

docs/ReleaseNotes.txt
```diff
@@ -1,3 +1,113 @@
...
+ - Added a function for the floating-point remainder of division (`fmod`) for
+   double and float types.
```
dbabokin, line 53: What about "floating-point types" or just removing "double and float types"? It's double, float and float16.

docs/ReleaseNotes.txt
```diff
@@ -1,3 +1,113 @@
+=== v1.25.0 === (TBD October 2024)
```
dbabokin, line 1: Should we mention the changes in standard library implementation and new `stdlib.isph`?

nurmukhametov, line 1: Need to fix `TBD` to the actual date.

docs/ReleaseNotes.txt
```diff
@@ -1,3 +1,113 @@
...
+ - Extended template support for short vectors and array declarations.  You can
+   now use type and non-type parameters to specify the type and dimensions of
+   these types.
+ - Added error messages for variable declarations with incompatible type
+   specifiers.
+ - Supported typedef with same name as struct tag.
+
+Code generation:
+
+ - Switched to the NewGVN pass for global value numbering for CPU targets. This
```
nurmukhametov, line 23: We need to remove `NewGVN` entry because of #3056.

nurmukhametov (changes requested):

docs/ReleaseNotes.txt
aneshlya: Fixed.

docs/ReleaseNotes.txt
aneshlya: Updated

docs/ReleaseNotes.txt
aneshlya: Updated

dbabokin (approved):
Looks good to me.

PR #3056, opened 2024-10-14
opt: disable NewGVN for CPU

nurmukhametov:
This reverts commit c6543aef74c3690a60d168c9b121dfead39aa8b3 (PR #2961).

NewGVN fails to optimize out loads that drastically decreases performance for some CPUs for speficic GameDev tests.


aneshlya (approved):
LGTM!

PR #3059, opened 2024-10-15
Set correct target triple for linked bc libraries

nurmukhametov:
Fix target triple and data layout of linked bc library (common, builtin and stdlib) just like it was done before the target redesign.


aneshlya (approved):
Looks good!

PR #3060, opened 2024-10-16
docs: fix short vector templates and noescape example

johguenther:


aneshlya (approved):
Thanks for your contribution!

Issue #3062, opened 2024-10-16
Use of -MF flag in v1.25.0 emits compile warnings/errors


jsiwek:
I use a CMake integration that seems to add ISPC flags like `-M -MT foo.obj -MF foo.obj.d`.  When moving from v1.24.0 to v1.25.0, I see warnings/errors (depending on use of `-werror`) like:  `Emitting dependencies file, but filename "foo.obj.d" has suffix "d"?`

With all the confidence of one who briefly looked at the code for the first time to see if there's an easy hack I can do myself if I have to: getting back to equivalent behavior of v1.24.0 may be a simple change in `OutputTypeInfo::isSuffixValid()` to return true if the `validSuffixes` member vector is empty, like it is for dependencies.

https://github.com/ispc/ispc/blob/b0d935e52aa8d9b03cedd0113d6c44f5be233f57/src/module.cpp#L1550

https://github.com/ispc/ispc/blob/b0d935e52aa8d9b03cedd0113d6c44f5be233f57/src/module.h#L138-L150

https://github.com/ispc/ispc/blob/b0d935e52aa8d9b03cedd0113d6c44f5be233f57/src/module.cpp#L1574-L1578

aneshlya:
Thanks for quick report of this issue. We'll create a new minor release to address this.

PR #3063, opened 2024-10-16
Fix deps file generation

nurmukhametov:
This PR fixes warning wrongly emitted for dependency file suffixes (#3062).

It also fixes the content of this file, is should not contain in-memory ISPC headers (`/core.isph`, `/stdlib.isph`) as dependencies for make.


aneshlya (approved):
Looks good

PR #3069, opened 2024-10-18
Fix #3055 - changed llvm::Intrinsic::getDeclaration() function

### LINKED ISSUE #3055: Build failure with LLVM trunk (llvm::Intrinsic::getDeclaration)
Here's the problem https://github.com/ispc/ispc/actions/runs/11312281122/job/31462531355#step:6:382
```
/home/runner/work/ispc/ispc/src/module.cpp:505:37: error: no member named 'getDeclaration' in namespace 'llvm::Intrinsic'
  505 |         funcDecl = llvm::Intrinsic::getDeclaration(module, ID, argArr);
      |                    ~~~~~~~~~~~~~~~~~^
```

Here's LLVM commit that broke it: https://github.com/llvm/llvm-project/commit/fa789dffb1e12c2aece0187aeacc48dfb1768340

Interesting, that they are going to bring back `llvm::Intrinsic::getDeclaration`:

```
Rename the function to reflect its correct behavior and to be consistent
with `Module::getOrInsertFunction`. This is also in preparation of
adding a new `Intrinsic::getDeclaration` that will have behavior similar
to `Module::getFunction` (i.e, just lookup, no creation).
```

And that's actually how it's used in ISPC right now.

dbabokin:
Fixes #3055 

This pull request includes a change to the `Symbol *Module::AddLLVMIntrinsicDecl` function in the `src/module.cpp` file to ensure compatibility with different versions of LLVM. The most important change is the conditional compilation logic to handle the difference in the `getDeclaration` method between LLVM versions.

Compatibility with LLVM versions:

* [`src/module.cpp`](diffhunk://#diff-91eede7716565a700586788e3f9d285985bf5c449a366cdd3b28fc37c073a616R505-R509): Added conditional compilation to use `llvm::Intrinsic::getDeclarationIfExists` for LLVM version 20.0 and above, while retaining `llvm::Intrinsic::getDeclaration` for earlier versions.

### DIFF
--- a/src/module.cpp
+++ b/src/module.cpp
@@ -502,7 +502,11 @@ Symbol *Module::AddLLVMIntrinsicDecl(const std::string &name, ExprList *args, So
             }
         }
         llvm::ArrayRef<llvm::Type *> argArr(exprType);
+#if ISPC_LLVM_VERSION >= ISPC_LLVM_20_0
+        funcDecl = llvm::Intrinsic::getDeclarationIfExists(module, ID, argArr);
+#else
         funcDecl = llvm::Intrinsic::getDeclaration(module, ID, argArr);
+#endif
         llvm::StringRef funcName = funcDecl->getName();
 
         if (g->target->checkIntrinsticSupport(funcName, pos) == false) {


nurmukhametov (approved):
LGTM!

PR #3077, opened 2024-10-18
Fix deps file generation 2

### LINKED ISSUE #3066: core.isph and stdlib.isph may still be registered as dependencies in v1.25.1 on Windows?
Follow up to #3062, particularly bad9ac20aae0adbfdf5a285a10b31ca024d0cc04: in a CMake+Ninja build system using Windows OS and ISPC v1.25.1, I see ISPC object files always being treated as dirty and rebuilt.  They also report dependencies on non-existing files `c:/core.isph` and `c:/stdlib.isph` -- I don't know exactly whether having those as missing file dependencies entirely explains and translates to the undesired always-dirty+rebuild behavior I see, but it's a suspect.  Maybe this platform-dependent handling:

https://github.com/ispc/ispc/blob/bad9ac20aae0adbfdf5a285a10b31ca024d0cc04/src/module.cpp#L2994-L3000

simply needs to be factored out and re-used in the following logic rather than the hardcoded `/` file prefix or else use some other way to identify those "files" in a platform-independent way ?

https://github.com/ispc/ispc/blob/bad9ac20aae0adbfdf5a285a10b31ca024d0cc04/src/module.cpp#L163-L164

nurmukhametov:
This should fix deps file generation for ISPC windows binary.

This PR fixes #3066

### DIFF
--- a/tests/lit-tests/3066.ispc
+++ b/tests/lit-tests/3066.ispc
@@ -1,0 +1,8 @@
+// RUN: %{ispc} --target=host %s -M -MT target_name 2>&1 | FileCheck %s
+
+// REQUIRES: WINDOWS_ENABLED
+
+// CHECK-NOT: ^ c:/core.isph{{.*}}
+// CHECK-NOT: ^ c:/stdlib.isph{{.*}}
+
+void foo() {}
--- a/tests/lit-tests/3066.ispc
+++ b/tests/lit-tests/3066.ispc
@@ -1,4 +1,4 @@
-// RUN: %{ispc} --target=host %s -M -MT target_name 2>&1 | FileCheck %s
+// RUN: %{ispc} --target=host --target-os=windows %s -M -MT target_name 2>&1 | FileCheck %s
 
 // REQUIRES: WINDOWS_ENABLED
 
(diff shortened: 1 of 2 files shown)


tests/lit-tests/3066.ispc
```diff
@@ -0,0 +1,8 @@
+// RUN: %{ispc} --target=host %s -M -MT target_name 2>&1 | FileCheck %s
```
dbabokin, line 1: Did you meant to add `--target-os=windows`?

Alternatively you can have `WINDOWS_HOST` instead of `WINDOWS_ENABLED`

dbabokin (approved):
Looks good.

Though assigning filenames (that may exist on the disk) to virtual files is probably not the best idea and we should handle it differently in the longterm.

tests/lit-tests/3066.ispc
dbabokin:
```suggestion
// RUN: %{ispc} --target=host --target-os=windows %s -M -MT target_name 2>&1 | FileCheck %s
```

aneshlya (approved):

PR #3078, opened 2024-10-18
Fix for #3071 - inconsistent __libcpp_verbose_abort definition

### LINKED ISSUE #3071: macOS build fails with LLVM 20.0 because of `__libcpp_verbose_abort` missing `noexcept`.
Here's an error message:
```
/Users/dbabokin/ispc/dbabokin_ispc/wt/20_0_fix/src/util.cpp:52:11: error: '__libcpp_verbose_abort' is missing exception specification 'noexcept'
   52 | void std::__libcpp_verbose_abort(char const *format, ...) {
      |           ^
      |                                                           noexcept
/Users/dbabokin/llvm/bin-trunk/bin/../include/c++/v1/__verbose_abort:24:49: note: previous declaration is here
   24 | _LIBCPP_ATTRIBUTE_FORMAT(__printf__, 1, 2) void __libcpp_verbose_abort(const char* __format, ...) _NOEXCEPT;
      |                                                 ^
```

The fix is obvious - add `noexcept` to the definition of the function for LLVM >= 20.0. But more fundamental question is why '__libcpp_verbose_abort'  needs to be defined in our source at all - because with the latest dev environment (macOS 15.0.1) this function is not needed.

Here's is my investigation and conclusion (which might be wrong, so I put full investigation here):
1. This was added initially by commit https://github.com/ispc/ispc/commit/a5dfc25f0cecd07c7b5b7fc67d0e80dfbcc3eff0 for the build with LLVM 17.0.
2. This function is needed for hardened versions of some of libc++ functions - so they can terminate nicely. At the same time this function is not available in older libc++ implementations.
3. The idea to implement this function in the user code was proposed by the library documentation, see here: https://github.com/llvm/llvm-project/blob/f0dfe682bca0b39d1fd64845f1576243d00c9d59/libcxx/docs/UsingLibcxx.rst#overriding-the-default-termination-handler
4. But this was later removed by https://github.com/llvm/llvm-project/commit/8dfc67d6724eb0af5d278f4d1d5511ca9f9e039f#diff-0c8caac74425d7493f6ec01919ae921c05fcb69cb3b493bc243544bbad082eedL149 with a description how a vendor building libc++ can provide a different assertion handler.
5. One problem here is that we are not messing up with libc++ distribution, but we distribute a binary, that needs to be compatible with old macOS distributions. So I checked what macOS SDKs have this function defined as part of their `libc++.1.dylib` - basically I grepped for `__ZNSt3__122__libcpp_verbose_abortEPKcz` in different SDKs at https://github.com/alexey-lysiuk/macos-sdk and the symbols is defined starting from `13.3`, which means that if we are not defining this symbol, we'll get problem with `macOS 12 Monterey` and earlier.

So the proposal is to keep the hack, just modify it to work with LLVM 20.0+.

dbabokin:
Fixes #3071.

The issue has long investigation of the problem, and I think that the proposed fix is the right one, as we need to preserve compatibility with older macOS versions.

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -45,11 +45,14 @@
 
 #ifdef _LIBCPP_VERSION
 #if ISPC_LLVM_VERSION >= ISPC_LLVM_17_0
-// Provide own definition of std::__libcpp_verbose_abort to avoid missing
-// symbols error in macOS builds.
-// See https://libcxx.llvm.org/UsingLibcxx.html#overriding-the-default-termination-handler
-// It is not quite clear why and where this symbol is used.
-void std::__libcpp_verbose_abort(char const *format, ...) {
+// Provide own definition of std::__libcpp_verbose_abort to avoid missing symbols error on macOS with old
+// system libc++.1.dylib. The symbol is there for macOS 13 Ventura and later, but not macOS 12 and earlier.
+// See #3071 for more details.
+void std::__libcpp_verbose_abort(char const *format, ...)
+#if ISPC_LLVM_VERSION >= ISPC_LLVM_20_0
+    noexcept
+#endif
+{
     va_list list;
     va_start(list, format);
     vfprintf(stderr, format, list);


aneshlya (approved):
LGTM

Issue #3082, opened 2024-10-25
Linker errors on Windows with v1.25


johguenther:
Compiling OSPRay with v1.24 on Windows is fine, switching to v1.25.0 or v1.25.1 results in linker errors.
MSVC15, 16 and 17 and ICX:
> World.dev.obj : fatal error LNK1227: conflicting weak extern definition for '__get_system_isa'.  New default '.weak.__get_system_isa.default.__xmm@00000200000001000000004000000800' conflicts with old default '.weak.__get_system_isa.default.ISPCDevice_dummyCompute' in ISPCDevice.dev.obj 

DPCPP:
```
lld-link: error: duplicate symbol: __get_system_isa
>>> defined at modules/cpu/common/World.dev.obj
lld-link: error: duplicate symbol: __set_system_isa
>>> defined at modules/cpu/common/World.dev.obj
lld-link: error: duplicate symbol: __system_best_isa
>>> defined at modules/cpu/common/World.dev.obj
lld-link: error: duplicate symbol: .weak.__get_system_isa.default.__xmm@00000200000001000000004000000800
>>> defined at modules/cpu/common/World.dev.obj
>>> defined at modules/cpu/fb/FrameBufferDispatch.dev.obj
lld-link: error: duplicate symbol: .weak.__set_system_isa.default.__xmm@00000200000001000000004000000800
>>> defined at modules/cpu/common/World.dev.obj
>>> defined at modules/cpu/fb/FrameBufferDispatch.dev.obj
lld-link: error: duplicate symbol: .weak.__system_best_isa.default.__xmm@00000200000001000000004000000800
>>> defined at modules/cpu/common/World.dev.obj
>>> defined at modules/cpu/fb/FrameBufferDispatch.dev.obj
```

PR #3087, opened 2024-10-31
Fix dispatch link errors

aneshlya:
Fixes linker error from #3082.
Windows linker doesn’t automatically handle weak symbols across modules and throws an error if conflicting symbols are found.
So first this PR reverts change made it https://github.com/ispc/ispc/pull/2970.
Also it adds internal linkage to `__set_system_isa` and adds `static` to `__get_system_isa` (it's only called from `__set_system_isa` so let's inline it). Overall these changes restores the linkage behavior that was before https://github.com/ispc/ispc/pull/2879.


dbabokin (approved):
LGTM

PR #3089, opened 2024-11-01
CI jobs to build RK ospray on Windows and examples 

aneshlya:
Eventually we need to fix warnings in examples build and to build them with `-Werror`.


dbabokin (approved):
Looks good.

But we probably want to run this kind of tests weekly. This way we don't have surprises just before the release.

aneshlya:
RK build (rk-superbuild) runs on Pull request. 
rk-superbuild-release is for checking already released binaries so it runs on demand.

Issue #3091, opened 2024-11-04
Another --vectorcall issue in v1.25.2


jsiwek:
Related to #3068, I find that at least use of `log()` still doesn't work when using the `--vectorcall` option.  Example `test.ispc`:

```
export uniform float Round(uniform float x) { return round(x); }
export uniform float Log(uniform float x)   { return log(x); }
```

and `test.cc`:

```
#include <stdio.h>
#include "test.h"

int main(int argc, char** argv) {
  printf("Round: %f\n", ispc::Round(13.f));
  printf("Log:   %f\n", ispc::Log(13.f));
  return 0;
  }
```

compiled/ran like:

```
> ispc.exe --vectorcall --target avx2-i32x8 -h test.h -o test.ispc.o test.ispc
> cl.exe test.cc -link test.ispc.o
> test.exe
Round: 13.000000
> echo %errorlevel%
-2147483645
> ispc.exe --version
Intel(r) Implicit SPMD Program Compiler (Intel(r) ISPC), 1.25.2 (build commit dbd426565e24f200 @ 20241102, LLVM 18.1.8)
```

I show use of `round()` just to confirm #3068 fix, but seems other functions like `log()` may still be broken?

aneshlya:
Thank you very much for reporting this issue. You’re right; the previous fix was incomplete, but I hope PR https://github.com/ispc/ispc/pull/3092 fully resolves it. The tests are passing in our CI, and if you have the opportunity, could you try [this binary](https://github.com/ispc/ispc/actions/runs/11675072113/artifacts/2144173932) on your end?

Apologies for the issues with v1.25.release. We made a significant [internal design change](https://github.com/ispc/ispc/blob/main/docs/design/TargetRedesign.md) to simplify support for new targets and streamline maintenance, but, as with any major update, it came with some challenges.

In recent weeks, we’ve added several new Windows jobs to our CI to catch and address reported issues more effectively.

jsiwek:
> could you try [this binary](https://github.com/ispc/ispc/actions/runs/11675072113/artifacts/2144173932) on your end?

That works for me, thanks.

Issue #3076, opened 2024-10-18
Add CI run for `run_tests` with `--vectorcall`


nurmukhametov:
It would be good to have in CI a job with `run_tests.py -f '--vectorcall'` to catch issues like #3068.

PR #3075, opened 2024-10-18
Fix stdlib functions calling convention for --vectorcall

nurmukhametov:
This PR fixes #3068.

PR #3092, opened 2024-11-04
Fix stdlib functions calling convention for --vectorcall. Part 2

aneshlya:
Fix for https://github.com/ispc/ispc/issues/3091 and https://github.com/ispc/ispc/issues/3076.
This PR is addition to PR https://github.com/ispc/ispc/pull/3075. The problem of previous PR was that it didn't set calling convention for stdlib function calls, only for functions themselves which led to overoptimization to "unreachable".
Also we should set calling convention only to the functions **defined** in stdlib (not to all functions declared there like builtins).


dbabokin (approved):
Look good.

PR #3094, opened 2024-11-06
Prepare for v1.25.3 release

aneshlya:


dbabokin (approved):
Looks good, but I would add the lit test as well.

PR #3099, opened 2024-11-07
Fix trunk build

aneshlya:
1. Replace getPointerTo with PointerType::get
2. Update lit tests


tests/lit-tests/2954.ispc
```diff
@@ -1,6 +1,6 @@
 // RUN: %{ispc} -O2 --target=avx2-i32x8 --emit-asm --x86-asm-syntax=intel %s -o - 2>&1 | FileCheck %s
 
-// XFAIL: *
```
dbabokin, line 3: Smart! I didn't know we can do that.

We'll need to find a patch and backport it to earlier versions. [Here](https://github.com/ispc/ispc/issues/2954#issuecomment-2438529795) I have a note when it started passing to help finding the patch.

dbabokin (approved):
Looks good!

PR #3102, opened 2024-11-09
Support bool non-type template parameter

### LINKED ISSUE #3100: templates doesn't work with bool
I'm trying to do this
```C++
template <bool foo> unmasked void somecode() {
    if (foo) {
        // ..
    } else {
        // ..
    }
}

export void somecode_with_fixed_path() {
    somecode<true>();
}

export void somecode_with_fixed_path2() {
    somecode<false>();
}
```
Which doesn't work https://godbolt.org/z/EbnaoEqeM if I do something like this instead:
```C++
template <int foo> unmasked void somecode2() {
    if (foo) {
        // ...
    } else {
        // ..
    }
}

export void somecode2_with_fixed_path() {
    somecode2<0>();
}

export void somecode2_with_fixed_path2() {
    somecode2<1>();
}
```
Works fine. So I wonder if `bool` should be supported in this case?

aneshlya:
Fixes #3100

### DIFF
--- a/src/parse.yy
+++ b/src/parse.yy
@@ -2585,14 +2585,15 @@ template_type_parameter
     ;
 
 int_constant_type
-    : TOKEN_INT8  { $$ = AtomicType::UniformInt8->GetAsConstType(); }
-    | TOKEN_INT16 { $$ = AtomicType::UniformInt16->GetAsConstType(); }
-    | TOKEN_INT   { $$ = AtomicType::UniformInt32->GetAsConstType(); }
-    | TOKEN_INT64 { $$ = AtomicType::UniformInt64->GetAsConstType(); }
-    | TOKEN_UINT8 { $$ = AtomicType::UniformUInt8->GetAsConstType(); }
-    | TOKEN_UINT16{ $$ = AtomicType::UniformUInt16->GetAsConstType(); }
-    | TOKEN_UINT  { $$ = AtomicType::UniformUInt32->GetAsConstType(); }
-    | TOKEN_UINT64{ $$ = AtomicType::UniformUInt64->GetAsConstType(); }
+    : TOKEN_INT8   { $$ = AtomicType::UniformInt8->GetAsConstType(); }
+    | TOKEN_INT16  { $$ = AtomicType::UniformInt16->GetAsConstType(); }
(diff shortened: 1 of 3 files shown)


src/parse.yy
```diff
@@ -2593,6 +2593,7 @@ int_constant_type
     | TOKEN_UINT16{ $$ = AtomicType::UniformUInt16->GetAsConstType(); }
     | TOKEN_UINT  { $$ = AtomicType::UniformUInt32->GetAsConstType(); }
     | TOKEN_UINT64{ $$ = AtomicType::UniformUInt64->GetAsConstType(); }
+    | TOKEN_BOOL{ $$ = AtomicType::UniformBool->GetAsConstType(); }
```
dbabokin, line 2596:
```suggestion
    | TOKEN_BOOL  { $$ = AtomicType::UniformBool->GetAsConstType(); }
```

tests/lit-tests/func_template_nontype_10.ispc
```diff
@@ -0,0 +1,33 @@
+// Instantiation of templates with bool parameters.
+// RUN: %{ispc}  %s --emit-llvm-text --target=host --nostdlib -o - | FileCheck %s
+
+// CHECK-LABEL: define {{.*}} @somecode_with_fixed_path1
+// CHECK: call <[[WIDTH:.*]] x [[TYPE:.*]]> @somecode___Cunbtrue
+
+// CHECK-LABEL: define {{.*}} @somecode___Cunbtrue
+// CHECK: call <[[WIDTH:.*]] x [[TYPE:.*]]> @true_call___
```
dbabokin, line 8: Probably do a CHECK-NOT for the other branch?

dbabokin (approved):
Thats's was an easy fix!

Looks good!

tests/lit-tests/func_template_nontype_10.ispc
aneshlya: I slightly changed a test to validate nested templates as well.

dbabokin: I we check that one branch is there, we assume that the other one should not be there, right? Otherwise, if both are present, the check seemed to be pointless, right?

aneshlya: Fixed

dbabokin (approved):

PR #3107, opened 2024-11-12
Fix for SDE download in CI

### LINKED ISSUE #3103: SDE download in CI
I'm looking at the failed nightly testing: https://github.com/ispc/ispc/actions/runs/11763079238

The problem that SDE was not available. Looks like it's fixed now. But I noticed two things:
- [workflow for the nightly run](https://github.com/ispc/ispc/actions/runs/11763079238/workflow) is not defining `USER_AGENT` variable for the install script.
- The SDE version is not the latest, probably it's time to update it.

Looks like it's a good idea to fail in install script if the `USER_AGENT` is not defined.

aneshlya:
Fixes #3103.
Currently is being tested: 
https://github.com/ispc/ispc/actions/runs/11806913030
https://github.com/ispc/ispc/actions/runs/11806917945

### DIFF
--- a/.github/workflows/ispc-ci.yml
+++ b/.github/workflows/ispc-ci.yml
@@ -26,8 +26,8 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  SDE_MIRROR_ID: 813591
-  SDE_TAR_NAME: sde-external-9.33.0-2024-01-07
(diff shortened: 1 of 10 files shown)


dbabokin (approved):
Looks great!

PR #3108, opened 2024-11-13
CMake: enable /DEPENDENTLOADFLAG

### LINKED ISSUE #2741: Pass /DEPENDENTLOADFLAG linker flag to ISPC/ISPC Runtime link process
The flag will be available starting LLVM 18.0. This is useful to mitigate DLL preload attacks.

nurmukhametov:
This fixes #2741


aneshlya:
Looks good! Can we validate it somehow? Maybe in check dll injection workflow?

nurmukhametov:
> Looks good! Can we validate it somehow? Maybe in check dll injection workflow?

I validated it manually. During validation, I have realized that we should not apply this flag for `ispcrt` otherwise, e.g., `ze_loader.dll` has to be installed in `system32` dir.

aneshlya (approved):
LGTM

PR #3112, opened 2024-11-15
Support LLVM intrinsics with no arguments

### LINKED ISSUE #3111: Calling LLVM intrinsics with no arguments is unsupported
Consider the following code:

```ispc
uniform int64() { return @llvm.readcyclecounter(); }
```

that results in the following error:

```bash
Error: syntax error, unexpected ')'. 
uniform int64() { return @llvm.readcyclecounter(); }
    ^

Compiler returned: 1
```

nurmukhametov:
This fixes #3111

### DIFF
--- a/src/parse.yy
+++ b/src/parse.yy
@@ -569,7 +569,20 @@ intrinsic_name
     ;
 
 intrincall_expression
-    : intrinsic_name '(' argument_expression_list ')'
+    : intrinsic_name '(' ')'
+      {
+          std::string *name = $1;
+          name->erase(0, 1);
+          Symbol* sym = m->AddLLVMIntrinsicDecl(*name, nullptr, Union(@1,@3));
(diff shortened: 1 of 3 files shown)


src/parse.yy
```diff
@@ -569,7 +569,20 @@ intrinsic_name
     ;
 
 intrincall_expression
-    : intrinsic_name '(' argument_expression_list ')'
+    : intrinsic_name '(' ')'
+      {
+          std::string *name = $1;
+          name->erase(0, 1);
+          Symbol* sym = m->AddLLVMIntrinsicDecl(*name, nullptr, Union(@1,@3));
```
aneshlya, line 576: I don't see any code in `AddLLVMIntrinsicDecl` verifying that `args` in not `nullptr` before accessing it. Could you add it?

aneshlya (approved):
Looks good!

PR #3113, opened 2024-11-15
Fix declarations in stdlib includes

nurmukhametov:
This fixes some declarations across `core.isph` and `builtins.isph` removing unused/unneeded and adjusting int/uint in a few places.


stdlib/include/builtins.isph
```diff
@@ -73,52 +73,24 @@ EXT READNONE varying float16 __tan_varying_half(varying float16);
 
 // TODO: do they need to be implemented in LLVM IR?
 // AOS/SOA conversion primitives
-EXT inline void __aos_to_soa2_double64(varying double, varying double, NOALIAS varying double *uniform,
```
aneshlya, line 76: I see these functions in util.m4 but looks like they aren't used anywhere. Is it right? Can they be removed from util.m4 as well?

stdlib/include/builtins.isph
```diff
@@ -187,14 +187,14 @@ EXT varying int16 __pdivs_vi16(varying int16, varying int16);
...
 EXT varying int8 __pdivs_vi8(varying int8, varying int8);
-EXT uniform int16 __pdivus_ui16(uniform int16, uniform int16);
-EXT uniform int32 __pdivus_ui32(uniform int32, uniform int32);
-EXT uniform int64 __pdivus_ui64(uniform int64, uniform int64);
-EXT uniform int8 __pdivus_ui8(uniform int8, uniform int8);
-EXT varying int16 __pdivus_vi16(varying int16, varying int16);
-EXT varying int32 __pdivus_vi32(varying int32, varying int32);
-EXT varying int64 __pdivus_vi64(varying int64, varying int64);
-EXT varying int8 __pdivus_vi8(varying int8, varying int8);
+EXT uniform uint16 __pdivus_ui16(uniform uint16, uniform uint16);
```
aneshlya, line 190: Hmm, looks like pdivus builtins are not used anywhere as well.

dbabokin: Looks like a need a warning for unused declarations ;-)

stdlib/include/builtins.isph
nurmukhametov: I think that they might be used for targets with width 64. They are removed here because they are purely internal m4 implementations that are not exposed to stdlib.ispc.

stdlib/include/builtins.isph
aneshlya: If __pdivus_* is really not used in stdlib.ispc, it can be removed from here.

nurmukhametov: I have removed them.

stdlib/include/builtins.isph
```diff
@@ -151,100 +119,46 @@ EXT inline uniform int32 __atomic_compare_exchange_uniform_int32_global(uniform
...
-EXT varying int64 __atomic_swap_varying_int64_global(varying int64, varying int64, UIntMaskType);
-EXT inline uniform int32 __atomic_umax_uniform_uint32_global(uniform int8 *uniform, uniform int32);
-EXT inline uniform int64 __atomic_umax_uniform_uint64_global(uniform int8 *uniform, uniform int64);
-EXT varying int32 __atomic_umax_varying_uint32_global(varying int64, varying int32, UIntMaskType);
-EXT varying int64 __atomic_umax_varying_uint64_global(varying int64, varying int64, UIntMaskType);
-EXT inline uniform int32 __atomic_umin_uniform_uint32_global(uniform int8 *uniform, uniform int32);
-EXT inline uniform int64 __atomic_umin_uniform_uint64_global(uniform int8 *uniform, uniform int64);
-EXT varying int32 __atomic_umin_varying_uint32_global(varying int64, varying int32, UIntMaskType);
-EXT varying int64 __atomic_umin_varying_uint64_global(varying int64, varying int64, UIntMaskType);
+EXT inline uniform int32 __atomic_umax_uniform_uint32_global(uniform int8 *uniform, uniform uint32);
```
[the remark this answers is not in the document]
nurmukhametov: Fixed

aneshlya (approved):

PR #3114, opened 2024-11-18
ISPC intrinsics

nurmukhametov:
This adds ISPC intrinsics (llvm.ispc.*) that can be used in ISPC to provide direct mapping to some LLVM IR instructions. They are to be used to implement a common target (builtin functions) without m4 templates of LLVM IR.

I moved all code related to intrinsic processing into a new file `src/intrinsics.cpp`. Support of ISPC intrinsics is two-staged. On the first stage, parser invokes `AddLLVMIntrinsicDecl` that detects them by prefix `llvm.ispc` and create a corresponding function declaration in LLVM IR and ISPC symbol. On the second stage, they are lowered to the actual LLVM IR instructions in `LowerISPCIntrinsics.cpp` pass.


src/intrinsics.cpp
```diff
@@ -0,0 +1,558 @@
...
+    }
+    for (unsigned i = 1; i <= (unsigned)ISPCIntrinsics::stream_store; i++) {
+        if (name == ISPCIntrinsicsNames[i]) {
+            return (ISPCIntrinsics)i;
+        }
+    }
+    return ISPCIntrinsics::not_intrinsic;
+}
+
+static llvm::VectorType *lGetDoubleVectorType(llvm::Type *type) {
```
aneshlya, line 301: I think the better name would be `lGetDoubleWidthVectorType`

src/intrinsics.cpp
```diff
@@ -0,0 +1,558 @@
...
+    if (ref.starts_with("llvm.ispc.atomicrmw.")) {
+        return ISPCIntrinsics::atomicrmw;
+    }
+    if (ref.starts_with("llvm.ispc.cmpxchg.")) {
+        return ISPCIntrinsics::atomicrmw;
+    }
+    if (ref.starts_with("llvm.ispc.fence.")) {
+        return ISPCIntrinsics::fence;
+    }
+    for (unsigned i = 1; i <= (unsigned)ISPCIntrinsics::stream_store; i++) {
```
aneshlya, line 293: I don't see a check anywhere that number of `ISPCIntrinsics` should correspond to the number of `ISPCIntrinsicsNames`. I'd prefer the function which would do strict mapping between enum values and strings similar to `ISPCTargetToString`.

src/module.cpp
```diff
@@ -3220,6 +3141,12 @@ static void lSetPreprocessorOptions(const std::shared_ptr<clang::PreprocessorOpt
     opts->addMacroDef(ispc_major);
     opts->addMacroDef(ispc_minor);
 
+    char llvm_major[buf_size], llvm_minor[buf_size];
+    snprintf(llvm_major, buf_size, "LLVM_VERSION_MAJOR=%d", LLVM_VERSION_MAJOR);
+    snprintf(llvm_minor, buf_size, "LLVM_VERSION_MINOR=%d", LLVM_VERSION_MINOR);
+    opts->addMacroDef(llvm_major);
+    opts->addMacroDef(llvm_minor);
+
```
aneshlya, line 3149: It seems not to be related to the current commit and should be done separately. ispc.rst and ispc.vim should be updated accordingly.

src/intrinsics.cpp
```diff
@@ -0,0 +1,558 @@
...
+    auto p = llvm::cast<llvm::Function>(M->getOrInsertFunction(name, funcType).getCallee());
+    return p;
+}
+
+static std::vector<llvm::Type *> lDeductArgTypes(llvm::Intrinsic::ID ID, ExprList *args) {
+    std::vector<llvm::Type *> exprType;
+    // TODO: not all intrinsics are overloaded, so we need to handle them also?
+    if (llvm::Intrinsic::isOverloaded(ID)) {
+        Assert(args);
+        std::vector<int> nInits(args->exprs.size());
```
aneshlya, line 408: Here more comments about nInits are needed. Some example of deducing types would be useful.

aneshlya (commented):
I only started a review and will continue tomorrow. Here are some initial comments.

src/module.cpp
nurmukhametov: Ok, here is a separate PR https://github.com/ispc/ispc/pull/3120

src/intrinsics.cpp
nurmukhametov: I actually removed this change from this PR as it doesn't relate to ISPC intrinsics at all. I am going to move it to PR #3119

src/intrinsics.cpp
nurmukhametov: Done

src/intrinsics.cpp
```diff
@@ -162,6 +279,127 @@ Symbol *lCreateISPCSymbolForLLVMIntrinsic(llvm::Function *func, SymbolTable *sym
...
 }
 
+static ISPCIntrinsics lLookupISPCInstrinsic(const std::string &name) {
+    llvm::StringRef ref(name);
+    if (ref.starts_with("llvm.ispc.atomicrmw.")) {
+        return ISPCIntrinsics::atomicrmw;
+    }
+    if (ref.starts_with("llvm.ispc.cmpxchg.")) {
+        return ISPCIntrinsics::atomicrmw;
+    }
```
aneshlya, line 289: This should return `ISPCIntrinsics::cmpxchg`

src/intrinsics.cpp
```diff
@@ -162,6 +279,127 @@ Symbol *lCreateISPCSymbolForLLVMIntrinsic(llvm::Function *func, SymbolTable *sym
...
+        break;
+    }
+    case ISPCIntrinsics::stream_store: {
+        retType = llvm::Type::getVoidTy(*g->ctx);
+        name = ISPCIntrinsicsNames[(unsigned)ID];
+        name += "." + lGetMangledTypeStr(TYs[1], hasUnnamedType);
+        break;
+    }
+    default: {
+        Error(SourcePos(), "Unknown ISPC intrinsic \"%s\".", ISPCIntrinsicsNames[(unsigned)ID]);
```
aneshlya, line 394: This assumes ID is always valid but in case of unknown intrinsics it will be most likely invalid

src/intrinsics.cpp
```diff
@@ -162,6 +279,127 @@ Symbol *lCreateISPCSymbolForLLVMIntrinsic(llvm::Function *func, SymbolTable *sym
...
+    std::string name = {};
+    llvm::Type *retType = nullptr;
+    std::vector<llvm::Type *> TYs;
+    for (const Type *type : argTypes) {
+        TYs.push_back(type->LLVMType(g->ctx));
+    }
+    bool hasUnnamedType = false;
+    switch (ID) {
+    case ISPCIntrinsics::atomicrmw: {
+        retType = TYs[1];
```
aneshlya, line 322: Here and in similar cases below, please add validation for `TYs` size to avoid crashes

src/intrinsics.cpp
aneshlya: Please add assert for array and enum size consistency

tests/lit-tests/ispc-intrinsics.ispc
```diff
@@ -0,0 +1,78 @@
...
+    return @llvm.ispc.atomicrmw.xchg.seq_cst(ptr, val);
+}
+
+// CHECK-LABEL: @bitcast
+// CHECK-DAG:   %calltmp = tail call half @llvm.ispc.bitcast.i16.f16(i16 %x, half 0xH0000)
+ATTRS uniform float16 bitcast(uniform int16 x) {
+    return @llvm.ispc.bitcast(x, (uniform float16)0);
+}
+
+// TODO!
```
aneshlya, line 37: Missed TODO here?

tests/lit-tests/ispc-intrinsics.ispc
```diff
@@ -0,0 +1,78 @@
+// RUN: %{ispc} %s --nowrap --target=host --emit-llvm-text --enable-llvm-intrinsics -o - 2>&1 | FileCheck %s
```
aneshlya, line 1: Could you also add unknown intrinsic to the test?

src/opt/LowerISPCIntrinsics.cpp
```diff
@@ -0,0 +1,327 @@
...
+
+        auto N = lGetVecNumElements(V0);
+        llvm::Value *V = builder.CreateShuffleVector(V0, V1, lGetSequentialMask(builder, 2 * N));
+
+        return builder.CreateExtractElement(V, I);
+    }
+    return nullptr;
+}
+
+static llvm::Value *lLowerInserIntrinsic(llvm::CallInst *CI) {
```
aneshlya, line 60: lLowerInserIntrinsic->lLowerInsertIntrinsic

src/opt/LowerISPCIntrinsics.cpp
```diff
@@ -0,0 +1,327 @@
...
+    // get type with the same width as target width
+    llvm::Type *newVT = llvm::IntegerType::get(builder.getContext(), lGetVecNumElements(V));
+    llvm::Value *packed = builder.CreateBitCast(V, newVT);
+
+    // zero extend to i64
+    return builder.CreateZExt(packed, llvm::Type::getInt64Ty(builder.getContext()));
+}
+
+static bool lRunOnBasicBlock(llvm::BasicBlock &BB) {
+    // TODO: add lit tests
```
aneshlya, line 271: Looks like this TODO can be removed.

src/opt/LowerISPCIntrinsics.cpp
```diff
@@ -0,0 +1,327 @@
...
+    if (VT) {
+        // check if the vector element type is not i1
+        llvm::Type *ET = VT->getElementType();
+        if (!ET->isIntegerTy(1)) {
+            // truncate vector of i32/i16/i8 to vector of i1
+            llvm::Type *i1 = llvm::IntegerType::get(builder.getContext(), 1);
+            llvm::Type *newVT = llvm::VectorType::get(i1, lGetVecNumElements(C), false);
+            C = builder.CreateTrunc(C, newVT);
+        }
+    }
```
aneshlya, line 226: Not mandatory, but this can be extracted to separate function.

src/opt/LowerISPCIntrinsics.cpp
```diff
@@ -0,0 +1,327 @@
...
+
+static unsigned lGetVecNumElements(llvm::Value *V) {
+    auto *vecType = llvm::dyn_cast<llvm::VectorType>(V->getType());
+    Assert(vecType);
+    return vecType->getElementCount().getKnownMinValue();
+}
+
+static llvm::Value *lLowerConcatIntrinsic(llvm::CallInst *CI) {
+    llvm::IRBuilder<> builder(CI);
+    llvm::Value *V0 = CI->getArgOperand(0);
```
aneshlya, line 30: Here and in similar places below please check that intrinsic has required number of arguments. I know that we often don't check it in optimizations but it's a big new change that will be a foundation for ISPC targets, so it should work perfectly.

aneshlya (changes requested):

src/intrinsics.cpp
nurmukhametov: Done

src/intrinsics.cpp
nurmukhametov: Done

tests/lit-tests/ispc-intrinsics.ispc
nurmukhametov: Indeed, `concat` was needed in one more piece of code to be supported.

tests/lit-tests/ispc-intrinsics.ispc
nurmukhametov: I have added a separate test `ispc-intrinsics-unknown.ispc`.

src/opt/LowerISPCIntrinsics.cpp
nurmukhametov: Fixed

src/opt/LowerISPCIntrinsics.cpp
nurmukhametov: Done

src/opt/LowerISPCIntrinsics.cpp
nurmukhametov: Done

src/opt/LowerISPCIntrinsics.cpp
nurmukhametov: Done

src/intrinsics.cpp
nurmukhametov: I have removed `ISPCIntrinsicsNames` completely and streamlined logic here.

aneshlya (approved):
Looks good!

tests/lit-tests/ispc-intrinsics.ispc
```diff
@@ -0,0 +1,82 @@
...
```
dbabokin, line 1: Why do we want to test them for host width only, but not for all width that intrinsics supposed to cover? I think we want to ensure that all relevant target width are tested.

dbabokin (commented):
New `llvm.ispc.*` intrinsics are not documented anywhere. Do we want any documentation for them? User visible or not user visible? I guess we do want to document them in design documentation.

nurmukhametov:
> New `llvm.ispc.*` intrinsics are not documented anywhere. Do we want any documentation for them? User visible or not user visible? I guess we do want to document them in design documentation.

I documented them in `docs/design/ispc-intrinsics.md`:
https://github.com/ispc/ispc/blob/61318ab4a3fea80c96dc6e97d9a6f2ad2fde0aa0/docs/design/ispc-intrinsics.md

dbabokin (approved):

Issue #2966, opened 2024-08-23
Move nightly builds to github releases.


aneshlya:
Move nightly builds from appveyor to github releases to avoid limitations of appveyor traffic. 
```
{"message":"Artifacts download limit (1024 MB/day) exceeded."}
```

dbabokin:
To be more clear, we should do something like https://github.com/llvm/torch-mlir-release project, which has a "release" with nightly builds. But we should probably have just the latest, not the history with multiple artifacts.

dbabokin:
Once it's done, we should not forget to update the links in:

- [ ] README.md
- [ ] in the downloads section on the website
- [ ] fix the download location for Compiler Explorer.

PR #3118, opened 2024-11-20
Migrate appveyor jobs to Github Actions

nurmukhametov:
This relates to #2966.


aneshlya:
Great that jobs are passing!
But I think it can be combined with the jobs from ispc-ci.
So it seems that it does 3 things:
1. Wasm-enabled build (win and linux)
2. ISPC_CROSS=ON (win and lin) - this should be enabled for the builds in ispc-ci at least in "release" jobs.
3. Benchmarks running (not related to wasm, win only) - can be part of full nightly testing
4. Examples running (not related to wasm, win only) - can be part of full nightly testing

dbabokin (commented):
I'm not sure that we need to introduce one more build workflow, which just "Ex Appveyor".

The key reason for Appveyor was to have the latest build available to public download. This was the build with all targets enabled (including WASM). So enabling uploading artifacts to the single permanent location is the key for porting Appveyor workflow to Github Actions. I think this can be done as part of `ispc-ci` workflow.

To enable permanent upload location we can have a "pre-release" on Github called "latest" (or something like that) and push there for the builds in `ispc/ispc` repo when pushed to `main`.

nurmukhametov:
As requested, this PR now adds ex-AppVeyor jobs inside ISPC-CI and creates a separate workflow to build and publish trunk artifacts to GitHub pre-release. See an example job: https://github.com/nurmukhametov/ispc/actions/runs/12810517014 pre-release: https://github.com/nurmukhametov/ispc/releases/tag/trunk-artifacts

.github/workflows/ispc-ci.yml
```diff
@@ -1370,3 +1371,81 @@ jobs:
...
+        call scripts\install_emscripten.bat
+        set VSVARS="C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+        call %VSVARS%
+        cmake . -B build -Thost=x64 -G "Visual Studio 16" -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=%ISPC_HOME%\install -DISPC_PREPARE_PACKAGE=ON -DISPC_CROSS=ON -DISPC_GNUWIN32_PATH=%CROSS_TOOLS_GNUWIN32% -DWASM_ENABLED=ON -DISPC_INCLUDE_BENCHMARKS=ON
+        cmake --build build --config Release --target package --verbose
+        set PATH=%ISPC_HOME%\build\bin\Release;%PATH%
+        cd %ISPC_HOME%
+        check_isa.exe
+        ispc --support-matrix
+        python scripts\perf.py -n 1 -g "Visual Studio 16" && ^
```
aneshlya, line 1443: This command builds and runs examples and it's not related to wasm build anyhow. I would move it from here to another job (probably <os>-build-examples is the best one). I think running perf.py on Linux also makes sense.

.github/workflows/trunk-artifacts.yml
```diff
@@ -0,0 +1,115 @@
...
+
+    - name: Install Dependencies
+      run: |
+        .github/workflows/scripts/install-build-deps.sh
+
+    - name: Configure
+      run: |
+        # Setup emcc for WASM target
+        source scripts/install_emscripten.sh && emcc --version
+        cmake -B build ./ -DISPC_PREPARE_PACKAGE=ON -DCMAKE_INSTALL_PREFIX="$ISPC_HOME"/install -DISPC_CROSS=ON -DWASM_ENABLED=ON -DCMAKE_CXX_FLAGS="-Werror"
```
aneshlya, line 44: I think we should run here the command closest to our release packages (from linux-build-ispc-llvm18-release job?). Same for Windows.

.github/workflows/trunk-artifacts.yml
```diff
@@ -0,0 +1,115 @@
+# Copyright 2025, Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause
+
+# These jobs build ISPC for trunk on every push to main pushing artifacts to
+# Github pre-release.
+name: Trunk Artifacts
+
```
aneshlya, line 7: Please add concurrency group to prevent parallel releases
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true

.github/workflows/trunk-artifacts.yml
```diff
@@ -0,0 +1,115 @@
...
+      uses: actions/upload-artifact@50769540e7f4bd5e21e526ee35c689e35e0d6874 # v4.4.0
+      with:
+        name: ispc-linux
+        path: build/ispc-*
+      
+    - name: Release
+      uses: softprops/action-gh-release@c95fe1489396fe8a9eb87c0abf8aa5b2ef267fda # 2.2.1
+      with:
+        name: trunk-artifacts
+        tag_name: trunk-artifacts
```
aneshlya, line 59: Maybe use tag_name "main"?

aneshlya (changes requested):

.github/workflows/trunk-artifacts.yml
nurmukhametov: I am not sure that having a tag name (it is just a usual git tag) the same as the branch name is a good idea. What about `main-artifacts`, `main-binaries`, `main-archives`, or `main-pre-release`?

.github/workflows/trunk-artifacts.yml
nurmukhametov: Done

.github/workflows/trunk-artifacts.yml
nurmukhametov: I have fixed it

.github/workflows/ispc-ci.yml
nurmukhametov: Done

.github/workflows/trunk-artifacts.yml
aneshlya: `main-artifacts` is fine. What I'd like to make clear here is that it is latest binaries built from the main branch.

aneshlya (approved):
Looks good to me. Maybe @dbabokin may comment more since he has experience with publishing pre-release artifacts.

dbabokin (approved):
Looks good to me.

PR #3119, opened 2024-11-20
Support overloaded LLVM intrinsics

### LINKED ISSUE #3117: Error with overloaded LLVM intrinsics
ISPC is struggling to correctly use overloaded LLVM intrinsics. Consider the two following examples:

1. 
```ispc
uniform float foo(uniform float a, uniform float b) {
    return @llvm.maxnum(a, b);
}
```

Error:
```bash
Intrinsic name not mangled correctly for type arguments! Should be: llvm.maxnum.f32
ptr @llvm.maxnum.f32.f32
/usr/local/src/ispc/src/main.cpp(414): FATAL ERROR: Unhandled signal sent to process; terminating.
***
***
Compiler returned: 139
```

Compiler explorer link: https://godbolt.org/z/KWh9cjPEr

2.
```ispc
int16 foo(int16 t) {
    return @llvm.aarch64.neon.sqrdmulh(t, t);
}
```

Error:
```bash
Intrinsic name not mangled correctly for type arguments! Should be: llvm.aarch64.neon.sqrdmulh.v8i16
ptr @llvm.aarch64.neon.sqrdmulh.v8i16.v8i16
/usr/local/src/ispc/src/main.cpp(414): FATAL ERROR: Unhandled signal sent to process; terminating.
***
***
Compiler returned: 139
```

Compiler explorer link: https://godbolt.org/z/qPx14Er61

This issue was resolved for a few specific LLVM intrinsics in https://github.com/ispc/ispc/pull/3114, but more general fix is required.

nurmukhametov:
This PR supports overloaded LLVM intrinsics. 

We need to produce correct function declarations for overloaded LLVM intrinsics. ISPC used all arguments to mangling in extra suffixes (which is wrong, e.g., see #3117). We need to provide only some arguments.

This PR adds support to `<N x ptr>` ISPC type. It also introduces ISPC type that matches to `<N x i1>` for targets with `mask_bits != 1`. Both are needed to support `llvm.masked.gather` and similar. These intrinsics will be used to implement common target.

This fixes #3117.

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -7134,23 +7134,11 @@ static llvm::Value *lTypeConvAtomicOrUniformVector(FunctionEmitContext *ctx, llv
             cast = exprVal;
             break;
         case AtomicType::TYPE_INT8:
-        case AtomicType::TYPE_UINT8: {
-            llvm::Value *zero = LLVMIntAsType(0, fromType->LLVMType(g->ctx));
-            cast = ctx->CmpInst(llvm::Instruction::ICmp, llvm::CmpInst::ICMP_NE, exprVal, zero, cOpName);
-            break;
-        }
+        case AtomicType::TYPE_UINT8:
         case AtomicType::TYPE_INT16:
-        case AtomicType::TYPE_UINT16: {
-            llvm::Value *zero = LLVMIntAsType(0, fromType->LLVMType(g->ctx));
-            cast = ctx->CmpInst(llvm::Instruction::ICmp, llvm::CmpInst::ICMP_NE, exprVal, zero, cOpName);
-            break;
-        }
+        case AtomicType::TYPE_UINT16:
         case AtomicType::TYPE_INT32:
-        case AtomicType::TYPE_UINT32: {
-            llvm::Value *zero = LLVMIntAsType(0, fromType->LLVMType(g->ctx));
-            cast = ctx->CmpInst(llvm::Instruction::ICmp, llvm::CmpInst::ICMP_NE, exprVal, zero, cOpName);
-            break;
-        }
+        case AtomicType::TYPE_UINT32:
         case AtomicType::TYPE_INT64:
         case AtomicType::TYPE_UINT64: {
             llvm::Value *zero = LLVMIntAsType(0, fromType->LLVMType(g->ctx));
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -941,6 +941,9 @@ static llvm::Constant *lLLVMConstantValue(const Type *type, llvm::LLVMContext *c
             } else {
                 return LLVMBoolVector(value != 0.);
             }
+        case AtomicType::TYPE_INT1: {
+            return isUniform ? LLVMInt1(value != 0.) : LLVMInt1Vector(value != 0.);
+        }
         case AtomicType::TYPE_INT8: {
             int i = (int)value;
             Assert((double)i == value);
@@ -6028,6 +6031,7 @@ ConstExpr::ConstExpr(const ConstExpr *old, SourcePos p) : Expr(p, ConstExprID) {
 
     switch (basicType) {
     case AtomicType::TYPE_BOOL:
+    case AtomicType::TYPE_INT1:
         std::copy(old->boolVal, old->boolVal + Count(), boolVal);
         break;
     case AtomicType::TYPE_INT8:
@@ -6089,6 +6093,8 @@ llvm::Value *ConstExpr::GetValue(FunctionEmitContext *ctx) const {
         } else {
             return boolVal[0] ? LLVMTrue : LLVMFalse;
         }
+    case AtomicType::TYPE_INT1:
+        return isVarying ? LLVMInt1Vector(boolVal[0]) : LLVMInt1(boolVal[0]);
     case AtomicType::TYPE_INT8:
         return isVarying ? LLVMInt8Vector(int8Val) : LLVMInt8(int8Val[0]);
     case AtomicType::TYPE_UINT8:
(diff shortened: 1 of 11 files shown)


src/intrinsics.cpp
```diff
@@ -259,6 +259,11 @@ static const Type *lLLVMTypeToISPCType(const llvm::Type *t, bool intAsUnsigned)
         return PointerType::GetUniform(AtomicType::VaryingDouble);
     }
 
+    // vector of pointers
+    else if (t == LLVMTypes::PtrVectorType) {
```
aneshlya, line 267: Could you add `(t == LLVMTypes::PtrType)` condition here as well?

src/intrinsics.cpp
```diff
@@ -223,6 +223,8 @@ static const Type *lLLVMTypeToISPCType(const llvm::Type *t, bool intAsUnsigned)
         return intAsUnsigned ? AtomicType::VaryingUInt64 : AtomicType::VaryingInt64;
     } else if (t == LLVMTypes::MaskType) {
         return AtomicType::VaryingBool;
+    } else if (t == LLVMTypes::Int1VectorType) {
```
aneshlya, line 228: For complete picture here should be `t == LLVMTypes::Int1Type` as well.

src/expr.cpp
```diff
@@ -7144,6 +7171,7 @@ static llvm::Value *lTypeConvAtomicOrUniformVector(FunctionEmitContext *ctx, llv
             }
             cast = exprVal;
             break;
+        // TODO!: all int casts are literally same, refactor to remove copy-paste
```
aneshlya, line 7181: There is a bunch of other places in this file like `ConstExpr` or `TypeCheckExpr` where we check particular atomic type. 
I believe that `AtomicType::TYPE_INT1` should be added to all these places as well.

src/type.cpp
```diff
@@ -114,6 +114,8 @@ const Type *Type::ResolveDependenceForTopType(TemplateInstantiation &templInst)
 
 const AtomicType *AtomicType::UniformBool = new AtomicType(AtomicType::TYPE_BOOL, Variability::Uniform, false);
 const AtomicType *AtomicType::VaryingBool = new AtomicType(AtomicType::TYPE_BOOL, Variability::Varying, false);
+const AtomicType *AtomicType::UniformInt1 = new AtomicType(AtomicType::TYPE_INT1, Variability::Uniform, false);
```
aneshlya, line 117: The implementation of int1 type is not complete in comparison to other types (like int8). I understand that the reason behind it that int1 type is not explicitly visible to users and it has limited usage but it raises lots of questions - why it is added to some methods and not to others, how can we guarantee that it will not break if we add more use cases (intrinsics) that use i1 type?

src/expr.cpp
```diff
@@ -4141,6 +4141,13 @@ llvm::Value *FunctionCallExpr::GetValue(FunctionEmitContext *ctx) const {
         if (isInvoke) {
             return ctx->InvokeSyclInst(callee, ft, argVals);
         } else {
+            // if callee is llvm.masked.gather then cast with inttoptr the first arg to PtrVectorType
+            if (callee->getName().starts_with("llvm.masked.gather")) {
+                argVals[0] = ctx->IntToPtrInst(argVals[0], LLVMTypes::PtrVectorType);
+            }
+            if (callee->getName().starts_with("llvm.masked.scatter")) {
+                argVals[1] = ctx->IntToPtrInst(argVals[1], LLVMTypes::PtrVectorType);
+            }
```
aneshlya, line 4150: There is nothing wrong with the code except that it looks like a hack in this place :)
It should be done earlier when we check the function arguments (line 4111 or earlier?). Also can we do this through `TypeConvertExpr`?

src/intrinsics.cpp
nurmukhametov: Pointer to values are already covered on lines 233-264

src/expr.cpp
nurmukhametov: I have added more places with `TYPE_INT1` in the last commit.

src/expr.cpp
nurmukhametov: > There is nothing wrong with the code except that it looks like a hack in this place :) It should be done earlier when we check the function arguments (line 4111 or earlier?). Also can we do this through `TypeConvertExpr`?

4111 or earlier and `TypeConvertExpr` operates in ISPC types whereas we adjust only LLVM types here.

src/type.cpp
aneshlya: Please add your comment about VaryingInt1 type above to type.h, it'll be very helpful.

src/expr.cpp
aneshlya: I see.

src/type.cpp
```diff
@@ -257,10 +258,11 @@ bool AtomicType::IsUnsignedType() const {
 }
 
 bool AtomicType::IsSignedType() const {
-    return (basicType == TYPE_INT8 || basicType == TYPE_INT16 || basicType == TYPE_INT32 || basicType == TYPE_INT64);
+    return (basicType == TYPE_INT1 || basicType == TYPE_INT8 || basicType == TYPE_INT16 || basicType == TYPE_INT32 ||
+            basicType == TYPE_INT64);
 }
 
-bool AtomicType::IsBoolType() const { return basicType == TYPE_BOOL; }
+bool AtomicType::IsBoolType() const { return basicType == TYPE_BOOL || basicType == TYPE_BOOL; }
```
aneshlya, line 265: `|| basicType == TYPE_INT1`?

aneshlya (approved):
Please address two last minor comments and I think it's ready to go.

src/type.cpp
nurmukhametov: Fixed

src/type.cpp
nurmukhametov: Done

PR #3120, opened 2024-11-21
Add macro definitions for LLVM version

nurmukhametov:
Define `LLVM_VERSION_MAJOR` and `LLVM_VERSION_MINOR` macros. They can be used to conditionally call LLVM intrinsics that change names between LLVM releases (either new ones are added, or old ones are moved out from experimental).

This is created as separate PR as requested here https://github.com/ispc/ispc/pull/3114#discussion_r1851374275


docs/ispc.rst
```diff
@@ -667,6 +668,12 @@ now generated for min/max operations, replacing the previous use of ``fmin`` and
 The ``avx512knl-x16``, ``gen9-x8``, and ``gen9-x16`` targets are deprecated and
 will be removed in future releases.
 
+Updating ISPC Programs For Changes In ISPC 1.26.0
+-------------------------------------------------
+
+Macro definitions for LLVM version that ISPC is based on were added. Please,
+refet to `The Preprocessor`_ for more details.
```
aneshlya, line 675:
```suggestion
refer to `The Preprocessor`_ for more details.
```

docs/ispc.rst
```diff
@@ -1084,6 +1091,12 @@ preprocessor runs:
   * - ISPC_MINOR_VERSION
     - 13
     - Minor version of the ``ispc`` compiler/language
+  * - LLVM_VERSION_MAJOR
+    - 18
```
aneshlya, line 1095: Let's remove LLVM version number from the docs, we'll forget updating it for sure. Just keep the description.

aneshlya: Same for ISPC_MAJOR_VERSION/ISPC_MINOR_VERSION.

nurmukhametov: Done

docs/ispc.rst
nurmukhametov: Fixed

aneshlya (approved):

PR #3127, opened 2024-12-04
Fix error message when unknown target/os/arch is used

### LINKED ISSUE #3126: Incorrect error message when building for aarch64
If I compile with commandline (which is incorrect)
```
--target=neon-i8x4
```
I get the following error message
```
Error: Incorrect targets: neon-i8x4. Choices are: sse2-i32x4, sse2-i32x4, neon-i8x16, neon-i32x4, gen9-x8. 
Compiler returned: 255
```
But if you try (which is included in the list above)
```
--target=neon-i8x16
``` 
You will get this error
```
Error: arch = aarch64 and target = neon-i8x16 is not a valid combination. 
Compiler returned: 1
```

aneshlya:
Fixes #3126.
Now the error message also mentions a `-support-matrix` with complete info about supported targets/os/archs.

### DIFF
--- a/tests/lit-tests/3126.ispc
+++ b/tests/lit-tests/3126.ispc
@@ -1,0 +1,20 @@
+// This test verifies that when an unknown target is specified on the command line,
+// the complete list of supported targets is displayed.
+// It's impractical to validate the entire output without imposing strict conditions on the test,
+// which is unnecessary, so we'll verify only the beginning of the output when x86 targets are enabled
+// (the targets are listed in alphabetical order).
+// Additionally, confirm that the message about '--support-matrix' is printed when an unknown target/OS or target is specified.
+
+// RUN: not %{ispc} --target=unknown --nowrap %s -o - 2>&1 | FileCheck %s --check-prefix=CHECK-TARGET
+// RUN: not %{ispc} --target-os=unknown --nowrap %s -o - 2>&1 | FileCheck %s --check-prefix=CHECK-OS
+// RUN: not %{ispc} --arch=unknown --nowrap %s -o - 2>&1 | FileCheck %s --check-prefix=CHECK-ARCH
+// REQUIRES: X86_ENABLED
+
+// CHECK-TARGET: Incorrect targets: unknown. Choices are: {{avx1-i32x16, avx1-i32x4, avx1-i32x8, avx1-i64x4, avx2-i16x16, avx2-i32x16, avx2-i32x4, avx2-i32x8, .*}}. Use --support-matrix for complete information about supported targets, archs and target OSes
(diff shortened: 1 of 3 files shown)


tests/lit-tests/3126.ispc
```diff
@@ -0,0 +1,15 @@
+// RUN: not %{ispc} --target=unknown --nowrap %s -o - 2>&1 | FileCheck %s --check-prefix=CHECK-TARGET
+// RUN: not %{ispc} --target-os=unknown --nowrap %s -o - 2>&1 | FileCheck %s --check-prefix=CHECK-OS
+// RUN: not %{ispc} --arch=unknown --nowrap %s -o - 2>&1 | FileCheck %s --check-prefix=CHECK-ARCH
+// REQUIRES: ARM_ENABLED && X86_ENABLED && LINUX_HOST && !WASM_ENABLED && !XE_ENABLED
+
+// CHECK-TARGET: Incorrect targets: unknown. Choices are: avx1-i32x16, avx1-i32x4, avx1-i32x8, avx1-i64x4, avx2-i16x16, avx2-i32x16, avx2-i32x4, avx2-i32x8, avx2-i64x4, avx2-i8x32, avx2vnni-i32x16, avx2vnni-i32x4, avx2vnni-i32x8, avx512icl-x16, avx512icl-x32, avx512icl-x4, avx512icl-x64, avx512icl-x8,
```
dbabokin, line 6: I'm wondering if the sting length is the same across different systems.

tests/lit-tests/3126.ispc
```diff
@@ -0,0 +1,15 @@
+// RUN: not %{ispc} --target=unknown --nowrap %s -o - 2>&1 | FileCheck %s --check-prefix=CHECK-TARGET
+// RUN: not %{ispc} --target-os=unknown --nowrap %s -o - 2>&1 | FileCheck %s --check-prefix=CHECK-OS
+// RUN: not %{ispc} --arch=unknown --nowrap %s -o - 2>&1 | FileCheck %s --check-prefix=CHECK-ARCH
+// REQUIRES: ARM_ENABLED && X86_ENABLED && LINUX_HOST && !WASM_ENABLED && !XE_ENABLED
```
dbabokin, line 4: That's quite restrictive. Should we have more tests, which have run everywhere, but check less (i.e. not matching the whole string, but only portions of it)?

dbabokin (approved):

tests/lit-tests/3126.ispc
aneshlya: I think checking the beginning of the string with targets is sufficient. I updated the test and added comments.

dbabokin: Looks good!

PR #3128, opened 2024-12-05
Fix functions and function templates overloads

### LINKED ISSUE #3015: Templates and regular functions can't have the same name
This code produces: `Unable to find any matching overload for call to function "func"`
```
varying int32 func(varying int32 a, varying int32 b);
uniform int32 func(uniform int32 a, uniform int32 b);
template <typename T>
uniform T func(uniform T a, uniform T b, uniform int c)
{
    uniform T r = func(a, b);
    return r;
}
uniform int foo2(uniform int a, uniform int b, uniform int c) {
    return func<uniform int>(a, b, c);
}
```
Looks like templates and regular functions can't have the same name.
https://github.com/ispc/ispc/blob/main/src/expr.cpp#L8327-L8339

aneshlya:
This PR fixes #3015.
It has 3 main changes:
1. Unifies `FunctionSymbolExpr` for functions and function templates (it can be improved even better when we refactor `Symbol` and `TemplateSymbol` classes)
2. Extracts `FindBestMatchCost` from `ResolveOverloads`
3. Introduces logic how to choose between template function and function when resolving overloads.

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -1826,23 +1826,13 @@ bool lCreateBinaryOperatorCall(const BinaryExpr::Op bop, Expr *a0, Expr *a1, Exp
     }
     if (CastType<StructType>(type0) != nullptr || CastType<StructType>(type1) != nullptr) {
         std::string opName = std::string("operator") + lOpString(bop);
-        std::vector<Symbol *> funs;
-        m->symbolTable->LookupFunction(opName.c_str(), &funs);
-        if (funs.size() > 0) {
-            Expr *func = new FunctionSymbolExpr(opName.c_str(), funs, sp);
-            ExprList *args = new ExprList(sp);
-            args->exprs.push_back(arg0);
-            args->exprs.push_back(arg1);
-            op = new FunctionCallExpr(func, args, sp);
-            return abort;
-        }
-
-        // templates
+        std::vector<Symbol *> funcs;
+        bool foundAnyFunction = m->symbolTable->LookupFunction(opName.c_str(), &funcs);
         std::vector<TemplateSymbol *> funcTempls;
-        bool foundAny = m->symbolTable->LookupFunctionTemplate(opName.c_str(), &funcTempls);
-        if (foundAny && funcTempls.size() > 0) {
-            TemplateArgs templArgs;
-            FunctionSymbolExpr *functionSymbolExpr = new FunctionSymbolExpr(opName.c_str(), funcTempls, templArgs, sp);
+        bool foundAnyTemplate = m->symbolTable->LookupFunctionTemplate(opName.c_str(), &funcTempls);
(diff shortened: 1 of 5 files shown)


tests/lit-tests/func_template_12.ispc
```diff
@@ -0,0 +1,21 @@
+// RUN: %{ispc}  %s --emit-llvm-text --target=host --nostdlib -o - | FileCheck %s
```
dbabokin, line 1: Would be good to spell it out what you are trying to test here.

src/expr.cpp
```diff
@@ -9354,16 +9312,16 @@ bool FunctionSymbolExpr::ResolveOverloads(SourcePos argPos, const std::vector<co
...
     for (int i = 0; i < (int)candidateCosts.size(); ++i) {
         if (candidateCosts[i] != -1 && candidateCosts[i] < bestMatchCost) {
             bestMatchCost = candidateCosts[i];
         }
     }
+
     // None of the candidates matched
     if (bestMatchCost == (1 << 30)) {
-        return lReportErrorUnableToFindOverload(matches, argTypes, argCouldBeNULL, name, pos);
+        return bestMatchCost;
```
nurmukhametov, line 9324: As I understand the sanitizer output, the memory leaks happens here for memory allocated on line 9310 `int *cost = new int[argTypes.size()];`

nurmukhametov (changes requested):

src/expr.cpp
```diff
@@ -1826,32 +1826,21 @@ bool lCreateBinaryOperatorCall(const BinaryExpr::Op bop, Expr *a0, Expr *a1, Exp
...
-        std::vector<Symbol *> funs;
-        m->symbolTable->LookupFunction(opName.c_str(), &funs);
-        if (funs.size() > 0) {
-            Expr *func = new FunctionSymbolExpr(opName.c_str(), funs, sp);
-            ExprList *args = new ExprList(sp);
-            args->exprs.push_back(arg0);
-            args->exprs.push_back(arg1);
-            op = new FunctionCallExpr(func, args, sp);
-            return abort;
-        }
```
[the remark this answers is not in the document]
aneshlya: There was a duplicate code (same is on lines 1832-1841) - one for functions and another for function templates. Now since we're passing both functions and function templates to FunctionSymbolExpr constructor we can keep only one.

src/expr.cpp
aneshlya: Fixed

tests/lit-tests/func_template_12.ispc
aneshlya: Added.

nurmukhametov (approved):
LGTM!

dbabokin (approved):

PR #3129, opened 2024-12-06
Add common targets

nurmukhametov:
This PR adds generic targets, including their definitions and corresponding build commands/targets. It introduces builtins/common.ispc (empty at the moment). These targets will be used in target traversal during hierarchical target builtin linking.

### DIFF
--- a/builtins/generic.ispc
+++ b/builtins/generic.ispc
@@ -1,0 +1,8 @@
+// Copyright (c) 2025, Intel Corporation
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+// This file provides the target-independent definitions for builtin functions.
+
+// Empty for now.
+;
--- a/cmake/GenerateBuiltins.cmake
+++ b/cmake/GenerateBuiltins.cmake
@@ -1,5 +1,5 @@
 #
-#  Copyright (c) 2018-2024, Intel Corporation
+#  Copyright (c) 2018-2025, Intel Corporation
 #
 #  SPDX-License-Identifier: BSD-3-Clause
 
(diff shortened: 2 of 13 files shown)


aneshlya:
Not sure if it's ready for review but one though regarding naming:

- if common targets are supposed to be accessible for user, it's better to use the word "generic" since it emphasizes that the target designed to work across multiple ISAs
- if they are internal to ISPC - "common" is a good word since it's baseline for implementation.

nurmukhametov:
> Not sure if it's ready for review but one though regarding naming:
> 
> * if common targets are supposed to be accessible for user, it's better to use the word "generic" since it emphasizes that the target designed to work across multiple ISAs
> * if they are internal to ISPC - "common" is a good word since it's baseline for implementation.

I am not sure if they are supposed to be accessible to users, but I actually prefer `generic` because `common` is already used in builtins and can be quite confusing in some contexts.

cmake/GenerateBuiltins.cmake
```diff
@@ -503,3 +503,117 @@ function (generate_builtins)
...
+    set(include ${CMAKE_CURRENT_SOURCE_DIR}/stdlib/include)
+
+    string(REPLACE "-" "_" target_ ${target})
+    set(name "builtins_target_${target_}_${bit}bit_${os}")
+    set(input_ispc builtins/generic.ispc)
+
+    if (${target} MATCHES "x86_64")
+        set(arch "x86_64")
+    elseif (${target} MATCHES "aarch64")
+        set(arch "aarch64")
```
aneshlya, line 517: I don't quite understand if generic targets are supposed to work on x86 and arm or not? Could you describe the logic regarding arch, please?

cmake/GenerateBuiltins.cmake
```diff
@@ -503,3 +503,117 @@ function (generate_builtins)
...
+            "generic-x86_64-i32x4"
+            "generic-x86_64-i32x8"
+            "generic-x86_64-i32x16"
+            "generic-x86_64-i64x4"
+        )
+    endif()
+
+    if (ARM_ENABLED)
+      list(APPEND TARGET_LIST
+          "generic-aarch64-i32x4"
```
aneshlya, line 602: Could you please add two other arm targets: neon-i8x16 and neon-i16x8. As was commented in https://github.com/ispc/ispc/pull/3121, we'd like to keep these targets and implement them using generic target.

cmake/GenerateBuiltins.cmake
nurmukhametov: I have removed arch from target names. We have now `generic-<mask_size>x<width>` with any --arch value.

cmake/GenerateBuiltins.cmake
nurmukhametov: They should have been supported now.

scripts/bitcode2cpp.py
```diff
@@ -44,14 +44,15 @@ def bundle_header(src, output):
             outfile.write("0x00 };\n\n")
             outfile.write(f"int {name}_length = {length};\n")
 
+ARCH_CHOICES = ['i686', 'x86', 'x86_64', 'arm', 'armv7', 'arm64', 'aarch64', 'wasm32', 'wasm64', 'xe64']
```
aneshlya, line 47: Looking into this I think we should reorganize arm architectures in ispc.

- "arm" = "armv7" should be deprecated
- "aarch32" ( as armv8 in 32 execution mode) should be added.

It's a note to myself, no need to add it to PR.

aneshlya (approved):
Looks good!

nurmukhametov:
I added two lit-tests

aneshlya:
Eventually we may want to hide generic targets from users. We discussed two ways to do this:
1. introduce a separate `--target-dev` option and allow generic targets to be used only with it.
2. allow them with slim binary only.

aneshlya (approved):

cmake/GenerateBuiltins.cmake
```diff
@@ -503,3 +503,125 @@ function (generate_builtins)
         ${DISPATCH_BUILTIN_CPP_FILES} ${COMMON_BUILTIN_CPP_FILES} ${TARGET_BUILTIN_CPP_FILES})
     add_dependencies(builtin builtins-cpp)
 endfunction()
+
+function (generate_generic_target_builtin_1 ispc_name target arch bit os)
```
dbabokin, line 507: Why the name has suffix `_1`?

Also it wouldn't hurt to document the function. I know, other functions are not documented, but I always struggled with understanding what they do without comments.

cmake/GenerateBuiltins.cmake
```diff
@@ -503,3 +503,125 @@ function (generate_builtins)
...
+    )
+
+    list(APPEND GENERIC_TARGET_BC_FILE ${bc})
+    set(GENERIC_TARGET_BC_FILE ${GENERIC_TARGET_BC_FILE} PARENT_SCOPE)
+
+    list(APPEND GENERIC_TARGET_CPP_FILE ${cpp})
+    set(GENERIC_TARGET_CPP_FILE ${GENERIC_TARGET_CPP_FILE} PARENT_SCOPE)
+endfunction()
+
+function (generate_generic_builtins ispc_name)
```
dbabokin, line 564: Same - I would really appreciate the comments for the function as a whole.

builtins/generic.ispc
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2024, Intel Corporation
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+// This file provides the target-independent definitions for builtin functions.
+
+// Empty for now.
+void foo() {}
```
dbabokin, line 8: Do we want that popular function name here? Would it clash with user `foo()`? As an alternative, I would suggest just `;`, which is a valid empty file.

src/ispc.cpp
```diff
@@ -1245,6 +1245,111 @@ Target::Target(Arch arch, const char *cpu, ISPCTarget ispc_target, PICLevel picL
     // Check default LLVM generated targets
     bool unsupported_target = false;
     switch (m_ispc_target) {
+    // TODO: figure out maskingFree and hasFeatures based on arch and CPU
```
dbabokin, line 1248: I do suggest putting a comment briefly defining what "generic" means, as it used to mean very different thing - the target that generated C++.

So something like "generic targets are the base targets that supposed to work on all hardware and serve as a baseline for implementing more fine-tuned targets"

cmake/GenerateBuiltins.cmake
```diff
@@ -503,3 +503,125 @@ function (generate_builtins)
...
+    list(APPEND os_list)
+    if (ISPC_WINDOWS_TARGET)
+        list(APPEND os_list "windows")
+    endif()
+    if (ISPC_UNIX_TARGET)
+        list(APPEND os_list "unix")
+    endif()
+
+    list(APPEND TARGET_LIST
+        "i1x4"
```
dbabokin, line 574: Hmm.... We do want a square matrix here. That's one of the goals. They might be dead slow, but should work.

cmake/GenerateBuiltins.cmake
nurmukhametov: I think I had initially two nested ones. Anyway, I have removed the suffix.

cmake/GenerateBuiltins.cmake
nurmukhametov: I have added some comments.

builtins/generic.ispc
nurmukhametov: Right, it works fine with `;`.

src/ispc.cpp
nurmukhametov: Done

cmake/GenerateBuiltins.cmake
```diff
@@ -503,3 +503,133 @@ function (generate_builtins)
         ${DISPATCH_BUILTIN_CPP_FILES} ${COMMON_BUILTIN_CPP_FILES} ${TARGET_BUILTIN_CPP_FILES})
     add_dependencies(builtin builtins-cpp)
 endfunction()
+
+# Define custom commands to generate bitcode for a specific generic target and
+# a corresponding cpp file. All they are generated from builtins/generic.ispc
```
dbabokin, line 508: I suggest clarifying that it's not just cpp, but a "cpp wrapper for bitcode". The reason I'm picking about, generic targets used to mean "target that produces C++ code", so disambiguating this important for anyone familiar with old thing.

cmake/GenerateBuiltins.cmake
dbabokin: I think this is the key issue that need to be ironed.

nurmukhametov: What do you mean?

cmake/GenerateBuiltins.cmake
nurmukhametov: Done

cmake/GenerateBuiltins.cmake
nurmukhametov: I am not entirely sure why we need to include them in this PR at this moment. If necessary, they can be added at a later stage. Additionally, I believe that the mask width is an implementation detail and should not be exposed to users.

aneshlya: I agree. We first need to make this work with existing masks/width and test it. Adding SVE targets depends on generic targets and we can't spend time right now to support (AND TEST) the whole square matrix. Eventually it should be done.

dbabokin: Ok, let's move on.

Though I see a lot of value in the square matrix.

dbabokin (approved):

PR #3130, opened 2024-12-06
Remove llvm::Type::X86_MMXTyID

aneshlya:
Remove mentions of llvm::Type::X86_MMXTyID: for LLVM 20.0 since it was removed from trunk
https://github.com/ispc/ispc/actions/runs/12194482841/job/34026470500#step:6:281


dbabokin (approved):
LGTM

Issue #2951, opened 2024-08-13
Utilization of `SDOT`/`UDOT` for `dot4add_u8i8*` on ARM


Wunkolo:
This is a follow-up on #2736 to capture interest on an ARM implementation.

On ARM there is the [UDOT instruction](https://developer.arm.com/documentation/100069/0609/A64-SIMD-Vector-Instructions/UDOT--vector-) and [SDOT instruction](https://developer.arm.com/documentation/100069/0609/A64-SIMD-Vector-Instructions/SDOT--vector-) that would be a great candidate to implement the newly added `dot4add_u8i8*` functions.

aneshlya:
Yes, we need to add target-specific implementations for ARM (and Xe). We've implemented `dot*` functions to evaluate how users will interact with them, assess their convenience, and determine any additional features needed in ISPC to facilitate easier dot products.

Could you share your experience if you've already used the `dot*` functions?

Wunkolo:
> Could you share your experience if you've already used the dot* functions?

I haven't utilized it just yet, but I currently have an intent to use these 8-bit dot-product functions to accelerate [image convolution](https://en.wikipedia.org/wiki/Kernel_(image_processing)#Convolution) and [other image operations](https://github.com/Wunkolo/qAverageColor?tab=readme-ov-file#avx512-vnni).

Wunkolo:
I have a proof-of-concept UDOT utilization here now, showing my particular use-case.
https://github.com/Wunkolo/qAverageColor/commit/afefda7592a2025c39b30b62851bb2847380f8ac

aneshlya:
I looked closely to `UDOT`/`SDOT` instructions and they use mixed width vectors.
For example:
```
 <4 x i32> @llvm.aarch64.neon.sdot.v4i32.v16i8(<4 x i32> %a, <16 x i8> %b, <16 x i8> %c)
```
**Return Type**: <4 x i32>
The result is a 4-element vector of 32-bit integers.
**Operands**:
%a: <4 x i32>
This is the initial vector accumulator, where the dot product results will be accumulated.
%b: <16 x i8>
The first operand vector for the dot product. It consists of 16 signed 8-bit integers.
%c: <16 x i8>
The second operand vector for the dot product. It also consists of 16 signed 8-bit integers.

It doesn't directly map to ISPC because ISPC currently does not support mixed-width operands.

[...]

PR #3138, opened 2024-12-13
Dot product on ARM

aneshlya:
Fixes https://github.com/ispc/ispc/issues/2951

### DIFF
--- a/stdlib/include/builtins.isph
+++ b/stdlib/include/builtins.isph
@@ -309,11 +309,15 @@ EXT inline READNONE varying int32 __insert_int32(varying int32, uniform int32, u
 EXT inline READNONE varying int64 __insert_int64(varying int64, uniform int32, uniform int64);
 EXT inline READNONE varying int8 __insert_int8(varying int8, uniform int32, uniform int8);
 
-// dor product
+// dot product
 EXT READNONE varying int32 __dot2add_i16packed_sat(varying uint32, varying uint32, varying int32);
 EXT READNONE varying int32 __dot2add_i16packed(varying uint32, varying uint32, varying int32);
 EXT READNONE varying int32 __dot4add_u8i8packed_sat(varying uint32, varying uint32, varying int32);
 EXT READNONE varying int32 __dot4add_u8i8packed(varying uint32, varying uint32, varying int32);
+EXT READNONE varying uint32 __dot4add_u8u8packed_sat(varying uint32, varying uint32, varying uint32);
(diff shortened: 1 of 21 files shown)


tests/lit-tests/aarch64_dotprod.ispc
```diff
@@ -0,0 +1,30 @@
+// Test checks emitted code for ARM (aarch64) dot product instructions.
```
nurmukhametov, line 1: Do we need a test to ensure that udot/sdot instructions are not generated on CPUs that do not support them?

nurmukhametov (approved):
LGTM!

stdlib/include/builtins.isph
```diff
@@ -309,11 +309,15 @@ EXT inline READNONE varying int32 __insert_int32(varying int32, uniform int32, u
 EXT inline READNONE varying int64 __insert_int64(varying int64, uniform int32, uniform int64);
 EXT inline READNONE varying int8 __insert_int8(varying int8, uniform int32, uniform int8);
 
-// dor product
+// dot product
 EXT READNONE varying int32 __dot2add_i16packed_sat(varying uint32, varying uint32, varying int32);
 EXT READNONE varying int32 __dot2add_i16packed(varying uint32, varying uint32, varying int32);
 EXT READNONE varying int32 __dot4add_u8i8packed_sat(varying uint32, varying uint32, varying int32);
 EXT READNONE varying int32 __dot4add_u8i8packed(varying uint32, varying uint32, varying int32);
+EXT READNONE varying uint32 __dot4add_u8u8packed_sat(varying uint32, varying uint32, varying uint32);
```
dbabokin, line 317: 32 bit saturation doesn't make sense for 8 bit data types. 8*8 gives 16 bit result. So 16 bit saturation does make sense, but not 32 bit.

dbabokin: I haven't realized that it accumulates with 32 bit accumulator. So saturation does make sense, but only on the last step `acc + tmp1234`

tests/lit-tests/aarch64_dotprod.ispc
aneshlya: Added

stdlib/include/builtins.isph
aneshlya: Removed unnecessary saturations

dbabokin (approved):

PR #3140, opened 2024-12-16
Generic target implementation

nurmukhametov:
This PR adds an implementation of `generic.ispc` with quite a lot of auxiliary changes.

This depends on #3129


cmake/GenerateBuiltins.cmake
```diff
@@ -503,3 +503,186 @@ function (generate_builtins)
...
+    )
+
+    list(APPEND GENERIC_TARGET_BC_FILE ${bc})
+    set(GENERIC_TARGET_BC_FILE ${GENERIC_TARGET_BC_FILE} PARENT_SCOPE)
+
+    list(APPEND GENERIC_TARGET_CPP_FILE ${cpp})
+    set(GENERIC_TARGET_CPP_FILE ${GENERIC_TARGET_CPP_FILE} PARENT_SCOPE)
+endfunction()
+
+function (generate_generic_stdlib_1 ispc_name target arch bit os)
```
aneshlya, line 564: We have separated logic for generating stdlib and builtins for other targets between Stdlib.cmake and GenerateBuiltins.cmake. Would it be feasible to do the same for generic targets?

aneshlya (commented):
It looks good! I think it can be merged when all TODOs are resolved.

cmake/GenerateBuiltins.cmake
nurmukhametov: I have refactored CMake code working with generic builtins and stdlib files.

builtins/generic.ispc
```diff
@@ -4,5 +4,2221 @@
...
+EXT varying int32 __psubs_vi32(varying int32 x, varying int32 y) { return @llvm.ssub.sat(x, y); }
+EXT varying int64 __psubs_vi64(varying int64 x, varying int64 y) { return @llvm.ssub.sat(x, y); }
+EXT uniform uint16 __psubus_ui16(uniform uint16 x, uniform uint16 y) { return @llvm.usub.sat(x, y); }
+EXT uniform uint32 __psubus_ui32(uniform uint32 x, uniform uint32 y) { return @llvm.usub.sat(x, y); }
+EXT uniform uint64 __psubus_ui64(uniform uint64 x, uniform uint64 y) { return @llvm.usub.sat(x, y); }
+EXT uniform uint8 __psubus_ui8(uniform uint8 x, uniform uint8 y) { return @llvm.usub.sat(x, y); }
+EXT varying uint16 __psubus_vi16(varying uint16 x, varying uint16 y) { return @llvm.usub.sat(x, y); }
+EXT varying uint32 __psubus_vi32(varying uint32 x, varying uint32 y) { return @llvm.usub.sat(x, y); }
+EXT varying uint64 __psubus_vi64(varying uint64 x, varying uint64 y) { return @llvm.usub.sat(x, y); }
+EXT varying uint8 __psubus_vi8(varying uint8 x, varying uint8 y) { return @llvm.usub.sat(x, y); }
```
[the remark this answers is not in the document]
nurmukhametov: I am not quite sure because some calls to these functions in `stdlib.ispc` are not guarded by `__have_saturating_arithmetic ` preficate, e.g., https://github.com/ispc/ispc/blob/d6eca64dae5c669a8d90ad83260c1a0714f393cc/stdlib/stdlib.ispc#L4921

builtins/generic.ispc
```diff
@@ -4,5 +4,2221 @@
...
+unmasked int64 __insert(int64 v, uniform int32 i, uniform int64 val) { return @llvm.ispc.insert(v, i, val); }
+unmasked uint8 __insert(uint8 v, uniform int32 i, uniform uint8 val) { return @llvm.ispc.insert(v, i, val); }
+unmasked uint16 __insert(uint16 v, uniform int32 i, uniform uint16 val) { return @llvm.ispc.insert(v, i, val); }
+unmasked uint32 __insert(uint32 v, uniform int32 i, uniform uint32 val) { return @llvm.ispc.insert(v, i, val); }
+unmasked uint64 __insert(uint64 v, uniform int32 i, uniform uint64 val) { return @llvm.ispc.insert(v, i, val); }
+unmasked float16 __insert(float16 v, uniform int32 i, uniform float16 val) { return @llvm.ispc.insert(v, i, val); }
+unmasked float __insert(float v, uniform int32 i, uniform float val) { return @llvm.ispc.insert(v, i, val); }
+unmasked double __insert(double v, uniform int32 i, uniform double val) { return @llvm.ispc.insert(v, i, val); }
+
+unmasked uniform bool __extract(bool v, uniform int32 i) { return @llvm.ispc.extract(v, i); }
```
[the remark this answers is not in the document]
nurmukhametov: I think it is fine.

builtins/generic.ispc
```diff
@@ -4,5 +4,2292 @@
...
+#elif TARGET_WIDTH == 32
+#define DOUBLE_WIDTH 64
+#elif TARGET_WIDTH == 64
+#define DOUBLE_WIDTH 128
+#else
+#error Unknown value of TARGET_WIDTH
+#endif
+
+// Macros for attributes
+#define NOESCAPE __attribute__((noescape))
```
dbabokin, line 32: I think it worth going to long prefixed names for macros in stdlib. As in case of name collision we redefine it and undef in the end of the file.

I think something like `ISPC_STDLIB_NOESCAPE` should be better. Same for other macros, which are implementation detail.

dbabokin (approved):
LGTM

builtins/generic.ispc
nurmukhametov: These macro definitions are internal to this file. The undef at the end emphasizes this, and it is actually unnecessary unless someone decides to include stdlib.ispc somewhere. Because of this, there is no point in adding extra prefixes that would create visual noise across the file while being virtually useless at the same time.

aneshlya: @nurmukhametov, as we discussed it would be good to change macro names in *.isph files using longer prefixes. Do you want to do it as part of this PR or create another one? If you want to do in a separate one, feel free to merge this one.

PR #3142, opened 2024-12-16
Fix IITDescs traversing

### LINKED ISSUE #3141: FATAL ERROR for deinterleave2 intrinsic
The following code fails with the fatal error:

```ispc
#if LLVM_VERSION_MAJOR >= 19
#define DEINTERLEAVE2 llvm.vector.deinterleave2
#else
#define DEINTERLEAVE2 llvm.experimental.vector.deinterleave2
#endif

void __deinterleave2(uniform float<8> val) {
    @DEINTERLEAVE2(val);
}
```

```bash
$ ispc --enable-llvm-intrinsics t.ispc
/root/ispc/src/main.cpp(414): FATAL ERROR: Unhandled signal sent to process; terminating.
```

Compiler explorer link: https://godbolt.org/z/9bdPzhf3e

nurmukhametov:
Struct and Vector are required to skip next entries also.

This PR fixes #3141

### DIFF
--- a/tests/lit-tests/3141.ispc
+++ b/tests/lit-tests/3141.ispc
@@ -1,0 +1,20 @@
+// RUN: not %{ispc} --nowrap --target=host --enable-llvm-intrinsics --emit-llvm-text -o - %s 2>&1 | FileCheck %s
+
+// REQUIRES: LLVM_17_0+
+
+#if LLVM_VERSION_MAJOR >= 19
+#define DEINTERLEAVE2 llvm.vector.deinterleave2
+#else
+#define DEINTERLEAVE2 llvm.experimental.vector.deinterleave2
+#endif
+
+// This checks that logic for deducting LLVM type for overloaded LLVM
+// intrinsics works.
+// CHECK-NOT: FATAL ERROR
+
+// This error is expected because the logic of mapping LLVM function signatures
+// to ISPC ones does not support the struct return type.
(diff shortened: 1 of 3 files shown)


aneshlya (approved):
LGTM!

tests/lit-tests/3141.ispc
```diff
@@ -0,0 +1,20 @@
+// RUN: not %{ispc} --nowrap --target=host --enable-llvm-intrinsics --emit-llvm-text -o - %s 2>&1 | FileCheck %s
+
+// REQUIRES: LLVM_17_0+
+
+#if LLVM_VERSION_MAJOR >= 19
```
dbabokin, line 5: Do we really define `LLVM_VERSION_MAJOR` for tests? I believe we don't.

dbabokin (commented):
Looks good, but I believe the test needs to be fixed.

Though, to make it more usable for intrinsic users, we should probably expose LLVM version used so developers can rely on some intrinsics existence.

tests/lit-tests/3141.ispc
nurmukhametov: We expose LLVM versions (PR #3120) for users, so this macro is defined.

aneshlya: I think we can proceed with this PR and then do renaming in a separate one.

PR #3144, opened 2024-12-17
Enable lower ispc intrinsics pass

nurmukhametov:
This is a part of #3140

### DIFF
--- a/tests/lit-tests/ispc-intrinsics.ispc
+++ b/tests/lit-tests/ispc-intrinsics.ispc
@@ -1,15 +1,41 @@
-// RUN: %{ispc} %s --nowrap --target=avx512skx-x4 --emit-llvm-text --enable-llvm-intrinsics -o - 2>&1 | FileCheck %s
-// RUN: %{ispc} %s --nowrap --target=avx512skx-x8 --emit-llvm-text --enable-llvm-intrinsics -o - 2>&1 | FileCheck %s
-// RUN: %{ispc} %s --nowrap --target=avx512skx-x16 --emit-llvm-text --enable-llvm-intrinsics -o - 2>&1 | FileCheck %s
-// RUN: %{ispc} %s --nowrap --target=avx512skx-x32 --emit-llvm-text --enable-llvm-intrinsics -o - 2>&1 | FileCheck %s
-// RUN: %{ispc} %s --nowrap --target=avx2-i8x32 --emit-llvm-text --enable-llvm-intrinsics -o - 2>&1 | FileCheck %s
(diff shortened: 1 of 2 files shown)


tests/lit-tests/ispc-intrinsics.ispc
```diff
@@ -1,15 +1,15 @@
...
-// RUN: %{ispc} %s --nowrap --target=avx512skx-x32 --emit-llvm-text --enable-llvm-intrinsics -o - 2>&1 | FileCheck %s
-// RUN: %{ispc} %s --nowrap --target=avx2-i8x32 --emit-llvm-text --enable-llvm-intrinsics -o - 2>&1 | FileCheck %s
-// RUN: %{ispc} %s --nowrap --target=avx2-i16x16 --emit-llvm-text --enable-llvm-intrinsics -o - 2>&1 | FileCheck %s
-// RUN: %{ispc} %s --nowrap --target=avx2-i32x4 --emit-llvm-text --enable-llvm-intrinsics -o - 2>&1 | FileCheck %s
-// RUN: %{ispc} %s --nowrap --target=avx2-i32x8 --emit-llvm-text --enable-llvm-intrinsics -o - 2>&1 | FileCheck %s
-// RUN: %{ispc} %s --nowrap --target=avx2-i32x16 --emit-llvm-text --enable-llvm-intrinsics -o - 2>&1 | FileCheck %s
-// RUN: %{ispc} %s --nowrap --target=avx2-i64x4 --emit-llvm-text --enable-llvm-intrinsics -o - 2>&1 | FileCheck %s
-// RUN: %{ispc} %s --nowrap --target=sse4.2-i8x16 --emit-llvm-text --enable-llvm-intrinsics -o - 2>&1 | FileCheck %s
-// RUN: %{ispc} %s --nowrap --target=sse4.2-i16x8 --emit-llvm-text --enable-llvm-intrinsics -o - 2>&1 | FileCheck %s
+// RUN: %{ispc} %s --nowrap --target=avx512skx-x4 --emit-llvm-text --enable-llvm-intrinsics --debug-phase=11:12 -o - 2>&1 | FileCheck %s
```
dbabokin, line 1: I'm not sure I fully understand the test changes. Are you checking explicitly the output after calling the lowering pass? Why keeping the output of the final IR then (i.e. `--emit-llvm-text`)?

Also if you rely on the phase number, you should pin it in the PM. Otherwise someone will eventually change this uncontiously by adding or removing passes before.

And one more thing, you should probably verify in the tests that the pass is working for all opt levels.

nurmukhametov: Indeed, no reason to keep the final IR. It is supposed to check the one pass output only (LowerISPCIntrinsics).

I have pinned the number, added tests for `-O1`/`-O2`/`-O3`.

aneshlya (approved):
Looks good!

PR #3147, opened 2024-12-17
Remove non-opaque ptr mode

### LINKED ISSUE #3146: Remove non-opaque pointer mode
All supported LLVM versions [16, 19] have opaque pointer mode enabled by default. Since LLVM 17, only opaque pointers are supported. There is no need to have opaque pointer flag in cmake, CI and tests.

nurmukhametov:
This PR removes non-opaque pointer mode support.

This fixes #3146.


dbabokin (commented):
Should we deprecate LLVM 16 support? I think opaque ptr mode doesn't work well for LLVM 16, so that's why we have it OFF by default.

dbabokin:
Looking further at the logic for default setting for LLVM 16, I'm confused. Is it really on or off?

aneshlya:
> Should we deprecate LLVM 16 support? I think opaque ptr mode doesn't work well for LLVM 16, so that's why we have it OFF by default.

We'll deprecate it when get confirmation that it's not used by UE. Opaque pointers works fine with LLVM 16. There were enabled by default in that version.

dbabokin:
> We'll deprecate it when get confirmation that it's not used by UE. Opaque pointers works fine with LLVM 16. There were enabled by default in that version.

Ok, sounds good! I was confused by the logic and comments in `CMakeLists.txt`.

I also rely on your @aneshlya review of the implementation in this PR.

aneshlya (approved):
Please remove opaque pointers in one more place: https://github.com/ispc/ispc/blob/main/tests/CMakeLists.txt#L32 and it's good to go.

aneshlya (approved):

PR #3149, opened 2024-12-17
Do not generate LLVM IR if preprocessor only mode

### LINKED ISSUE #3148: IR is generated in the preprocessor only mode
In `-E` mode, ISPC should not generate IR and run the optimization pipeline.

Consider the following command:

```bash
$ ispc t.ispc -E -o - --debug-phase=0:last
```

The expected output should not contain debug dumps of LLVM passes.

This was broken in PR #2743

nurmukhametov:
This PR fixes #3148


dbabokin (approved):
LGTM

aneshlya (approved):

Issue #3150, opened 2024-12-17
Assertion failed with `--nocpp`


nurmukhametov:
Consider the following code:

```ispc
int foo() { return 42; }
```

ISPC asserts with `--nocpp` flag. It happened after PR #2743 .

```bash
$ ispc t.ispc --emit-llvm-text -o - --nocpp --nowrap
Error: Assertion failed (/usr/local/src/ispc/src/ctx.cpp:3014): "maskedStoreFunc != nullptr". 
***
*** Please file a bug report at https://github.com/ispc/ispc/issues
*** (Including as much information as you can about how to reproduce this error).
*** You have apparently encountered a bug in the compiler that we'd like to fix!
***
/usr/local/src/ispc/src/main.cpp(414): FATAL ERROR: Unhandled signal sent to process; terminating.
Program terminated with signal: SIGSEGV
Compiler returned: 139
```

Compiler explorer link: https://godbolt.org/z/qhvjdeWo5

PR #3151, opened 2024-12-17
Deprecate --nocpp command line flag

nurmukhametov:
This PR partially addresses #3150

### DIFF
--- a/docs/ispc.rst
+++ b/docs/ispc.rst
@@ -674,6 +674,8 @@ Updating ISPC Programs For Changes In ISPC 1.26.0
 Macro definitions for LLVM version that ISPC is based on were added. Please,
 refer to `The Preprocessor`_ for more details.
 
+``--nocpp`` command line flag is deprecated and will be removed in future.
+
 Getting Started with ISPC
 =========================
 
@@ -1080,8 +1082,7 @@ The Preprocessor
 
 ``ispc`` automatically runs the C preprocessor on your input program before
 compiling it.  Thus, you can use ``#ifdef``, ``#define``, and so forth in
-your ispc programs.  (This functionality can be disabled with the ``--nocpp``
-command-line argument.)
+your ispc programs.
 
 A number of preprocessor symbols are automatically defined before the
 preprocessor runs:
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -119,7 +119,6 @@ static void lPrintVersion() {
     printf("    [--no-omit-frame-pointer]\t\tDisable frame pointer omission. It may be useful for profiling\n");
     printf("    [--nostdlib]\t\t\tDon't make the ispc standard library available\n");
     printf("    [--no-pragma-once]\t\t\tDon't use #pragma once in created headers\n");
-    printf("    [--nocpp]\t\t\t\tDon't run the C preprocessor\n");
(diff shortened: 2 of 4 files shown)


dbabokin (commented):
Why do we want to remove this flag? Why not fixing it?

dbabokin:
> > Why do we want to remove this flag? Why not fixing it?
> 
> It contradicts the current design, which relies on internal headers.
> 
> I also assume this flag was supposed to be used in tests only, but it is not actually used a lot. That's why I think it is easier to just deprecate and eventually remove it rather than fix it.

My understanding is that you can manually run preprocessor separately and run the compiler separately. I.e. run `-E` and then `--nocpp`. It might be for debugging purposes, not for testing. I guess that's the main use case. If we can fix that, I think we should fix that just to look like C++ compiler from the prospective of using `-E`/`--nocpp`. The question is how easy to make it work with new stdlib logic. If it's really messy, we can remove it.

Though we should probably give more detailed explanation in the deprecation notice.

nurmukhametov:
> My understanding is that you can manually run preprocessor separately and run the compiler separately. I.e. run `-E` and then `--nocpp`. It might be for debugging purposes, not for testing. I guess that's the main use case.

This use-case, i.e., something like this:

```bash
$ ispc -E test.ispc -o - | ispc --nocpp - --emit-llvm-text -o -
```

works with reasonably same ISPC compiler.

At the same time, some tests that can be compiled before PR #2743 are not compiled now (like the test in this PR).

dbabokin:
> works with reasonably same ISPC compiler.

It should work for the same compiler in this use-case scenario. It's for the user to debug what compiler does with the source code and try to understand the problems in user code. It's not to debug compiler itself.

> At the same time, some tests that can be compiled before PR #2743 are not compiled now (like the test in this PR).

So, the question is how difficult to make it work as previously. If it's not straightforward, I'm ok with depreciation. But if we can make it work relatively easy, we should keep it.

aneshlya:
> So, the question is how difficult to make it work as previously. If it's not straightforward, I'm ok with depreciation. But if we can make it work relatively easy, we should keep it.

It's not straightforward, that's why this PR is here in the first place.
Also this option is broken (like in the test in this PR) in the last release and no one complained :)

dbabokin:
> It's not straightforward, that's why this PR is here in the first place. Also this option is broken (like in the test in this PR) in the last release and no one complained :)

Ok, then let's go with this PR.

dbabokin (approved):

docs/ispc.rst
```diff
@@ -674,6 +674,8 @@ Updating ISPC Programs For Changes In ISPC 1.26.0
 Macro definitions for LLVM version that ISPC is based on were added. Please,
 refer to `The Preprocessor`_ for more details.
 
+``--nocpp`` command line flag is deprecated and will be removed in future.
+
```
aneshlya, line 678: There is a mention of `--nocpp` flag in Preprocessor section - I think it can be removed.

nurmukhametov: Done

aneshlya: I don't see ispc.rst in the modified files anymore.

nurmukhametov: Right, it has to be now.

aneshlya (approved):

Issue #2860, opened 2024-05-19
Ambiguous use of overloaded function with template types.


MkazemAkhgary:
following code does not compile even with the cast to `T`. I get this error `Ambiguous use of overloaded function "rcp". `

```cpp
template<typename TVec, typename T>
inline TVec FastDiv(T Lhs, TVec Rhs)
{
    TVec Result = { Lhs * rcp((T)Rhs.X), Lhs * rcp((T)Rhs.Y) };
    return Result;
}
```

but the following code compiles fine

```cpp
template<typename TVec, typename T>
inline TVec FastDiv(T Lhs, TVec Rhs)
{
    T RhsX = Rhs.X, RhsY = Rhs.Y;
    TVec Result = { Lhs * rcp(RhsX), Lhs * rcp(RhsY) };
    return Result;
}
```

PR #3154, opened 2024-12-19
Add lit tests for #2860

aneshlya:
This is a lit test for https://github.com/ispc/ispc/issues/2860.
The issue itself was unintentionally fixed in v1.25.3.


dbabokin (commented):
Should we instantiate the template to make sure that it really works?

aneshlya:
> Should we instantiate the template to make sure that it really works?

Added instantiation

PR #3162, opened 2025-01-08
Support deprecated attribute

### LINKED ISSUE #2941: Support deprecated attribute
I propose to support the following attribute for functions in two formats:

```ispc
__attribute__((deprecated)) void foo();
__attribute__((deprecated("reason"))) void func();
```

When such functions called, the compiler warning should be emitted:
```ispc
void f() {
    foo(); // Warning: function `foo` is deprecated.
    func(); // Warning: function `func` is deprecated, reason.
}
```

nurmukhametov:
Two variants are supported:
`__attribute__((deprecated))`
`__attribute__((deprecated("reason")))`

This fixes #2941

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -1,5 +1,5 @@
 /*
-  Copyright (c) 2010-2024, Intel Corporation
+  Copyright (c) 2010-2025, Intel Corporation
 
   SPDX-License-Identifier: BSD-3-Clause
 */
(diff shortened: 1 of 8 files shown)


src/expr.cpp
```diff
@@ -3981,6 +3981,40 @@ void SelectExpr::Print(Indent &indent) const {
 ///////////////////////////////////////////////////////////////////////////
 // FunctionCallExpr
 
+void lWarnDeprecatedFunction(Symbol *sym, SourcePos pos, const std::vector<const Type *> &argTypes) {
+    if (!sym) {
```
aneshlya, line 3985: I would also check that `sym` is a `Function`

aneshlya (approved):
Looks good

PR #3165, opened 2025-01-09
Tune -O1 pipeline for code size

### LINKED ISSUE #3164: Optimize -O1 pipeline for code size
The current O1 optimization pipeline in ISPC is nearly identical to O2, except that coherent control flow is disabled. To better tailor O1 for size optimization, we should:

- Adjust settings for the Function Inliner and Loop Unrolling passes to prioritize size optimization, aligning them with LLVM's -Os behavior.
- Pass the -Os flag to the code generation.

These changes would significantly benefit mobile platforms.

aneshlya:
Fixes #3164.
Surprisingly ISPC was using optLevel 1 by default. It didn't matter because we didn't pass optlevel to optimization passes but it matters now with the changes in this PR.

1. Change default optLevel to 2
2. Pass `CodegenOptLevel::Default` to codegen when -O1 is selected (it corresponds to LLVM -Os)
3. Pass optLevel to Loop unrolling passes preventing them from aggressing unrolling
4. Get InlineParams from optLevel and sizeOptlevel and pass to Inliner pass
5. Add `optSize` function attribute when -O1 is passed. It's needed for loop unrolling passes.

Note: pragma has priority over optimization level. This behavior is aligned with LLVM.

### DIFF
--- a/src/ispc.cpp
+++ b/src/ispc.cpp
@@ -2240,11 +2240,16 @@ Target::Target(Arch arch, const char *cpu, ISPCTarget ispc_target, PICLevel picL
             // Set Optimization level for llvm codegen based on Optimization level
             // requested by user via ISPC Optimization Flag. Mapping is :
             // ISPC O0 -> Codegen O0
-            // ISPC O1,O2,O3,default -> Codegen O3
+            // ISPC O1 -> Codegen Default (-Os)
+            // ISPC O2,O3,default -> Codegen O3
             CodegenOptLevel cOptLevel = CodegenOptLevel::Aggressive;
             switch (g->codegenOptLevel) {
             case Globals::CodegenOptLevel::None:
                 cOptLevel = CodegenOptLevel::None;
+                break;
+
+            case Globals::CodegenOptLevel::Default:
+                cOptLevel = CodegenOptLevel::Default;
                 break;
 
             case Globals::CodegenOptLevel::Aggressive:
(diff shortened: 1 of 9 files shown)


nurmukhametov (commented):
Does changing the default optimization level to 2 influence O2/O3 performance?

aneshlya:
> Does changing the default optimization level to 2 influence O2/O3 performance?

No, changing default doesn't affect performance:
1. we didn't pass optLevel in individual passes before
2. we passed CodegenOptLevel::Aggressive to codegen for O1, O2 and O3

Now with default optimization level 2:
1. we pass optLevel to Unrolling and Inlining passes but it corresponds to default that is used there if you don't pass any parameters (2).
2. we still pass CodegenOptLevel::Aggressive to codegen

What may change is performance with O3 since unrolling and inlining will use aggressive thresholds.

nurmukhametov (approved):

src/ispc.cpp
```diff
@@ -2292,6 +2297,10 @@ Target::Target(Arch arch, const char *cpu, ISPCTarget ispc_target, PICLevel picL
         for (auto const &f_attr : m_funcAttributes) {
             fattrBuilder->addAttribute(f_attr.first, f_attr.second);
         }
+        // This attribute required for LoopUnroll passes
```
dbabokin, line 2300:
```suggestion
        // This attribute is required for LoopUnroll passes
```

dbabokin (approved):
LGTM

Issue #3084, opened 2024-10-27
Crash when using sizeof etc in size_t global variables (!)


Hugobros3:
I'm running into what seems like https://github.com/ispc/ispc/issues/2181, but with a different diagnostic. At the top-level, I have something like:
```
size_t i = sizeof(int);
//int does the same!
```
and I get
```
Global variable initializer type does not match global variable type!
ptr @i
LLVM ERROR: Broken module found, compilation aborted!
/usr/local/src/ispc/src/main.cpp(414): FATAL ERROR: Unhandled signal sent to process; terminating.
***
*** Please file a bug report at https://github.com/ispc/ispc/issues
*** (Including as much information as you can about how to reproduce this error).
*** You have apparently encountered a bug in the compiler that we'd like to fix!
***
Program terminated with signal: SIGSEGV
Compiler returned: 139
```

**This sample does compile correctly if I use `--addressing=64`**

pbrubaker:
Okay, this is weird.

https://ispc.godbolt.org/z/bvjdqb7Ph

Issue #2775, opened 2024-02-27
Sizeof of varying types


nurmukhametov:
The following code

```ispc
typedef struct { uniform int<2> x; } SV;

typedef struct { uniform int x[2]; } SA;

unmasked int foo()
{	
    int s1 = sizeof(SV);
    int s2 = sizeof(SA);
    return s1 + s2;
}
```

fails to compile with the error:

```bash
Stored value type does not match pointer operand type!
  store <8 x i64> <i64 ptrtoint (i32 16 to i64), i64 ptrtoint (i32 16 to i64), i64 ptrtoint (i32 16 to i64), i64 ptrtoint (i32 16 to i64), i64 ptrtoint (i32 16 to i64), i64 ptrtoint (i32 16 to i64), i64 ptrtoint (i32 16 to i64), i64 ptrtoint (i32 16 to i64)>, <8 x i32>* %5, align 64, !dbg !24049
 <8 x i64>Stored value type does not match pointer operand type!
  store <8 x i64> <i64 ptrtoint (i32 8 to i64), i64 ptrtoint (i32 8 to i64), i64 ptrtoint (i32 8 to i64), i64 ptrtoint (i32 8 to i64), i64 ptrtoint (i32 8 to i64), i64 ptrtoint (i32 8 to i64), i64 ptrtoint (i32 8 to i64), i64 ptrtoint (i32 8 to i64)>, <8 x i32>* %6, align 64, !dbg !24050
 <8 x i64>LLVM ERROR: Broken module found, compilation aborted!
/usr/local/src/ispc/src/main.cpp(404): FATAL ERROR: Unhandled signal sent to process; terminating.
```

https://ispc.godbolt.org/z/enxbGYxaj

This was introduced somewhere in between ISPC v1.12.0 and v1.13.0.

nurmukhametov:
Since ISPC v1.24.0, there is no crash anymore but generated code doesn't look right:

```llvm
define <8 x i32> @foo___(<8 x i32> %__mask) local_unnamed_addr {
allocas:
  ret <8 x i32> undef
}
```

The shorter reproducer of similar issue:

```ispc
int foo() {	
    int s2 = sizeof(int);
    return s2;
}
```

aneshlya:
This is a duplicate of https://github.com/ispc/ispc/issues/2181

pbrubaker:
#3166 was fixed, closing this as well.

Issue #2181, opened 2021-10-14
Miscompiled sizeof code


larsbohn:
Hi!

There seems to be an issue when using sizeof, causing a fatal error:

https://ispc.godbolt.org/z/4ecPEM6Tx

```
typedef float<3> vec3;

export void func()
{
    const int test = sizeof(vec3);
}

```

Will cause:
```
ispc: /usr/local/src/llvm/llvm-13.0/llvm/lib/IR/Constants.cpp:2198: static llvm::Constant *llvm::ConstantExpr::getPtrToInt(llvm::Constant *, llvm::Type *, bool): Assertion `C->getType()->isPtrOrPtrVectorTy() && "PtrToInt source must be pointer or pointer vector"' failed.
/home/appveyor/projects/ispc/src/main.cpp(394): FATAL ERROR: Unhandled signal sent to process; terminating.
```

Hugobros3:
I can't repro on ISPC 1.25. Ran into another similar bug and opened a fresh issue.

aneshlya:
It doesn't crash since v1.24.0 but it doesn't produce the right result: https://ispc.godbolt.org/z/sxr4baxsb for varying value.
If result of sizeof is saved to uniform variable, it works fine.

pbrubaker:
Being explicit with the variability solves the codegen issue.

https://ispc.godbolt.org/z/vWo1vKPsT

```
typedef float<3> vec3;

export void func(uniform int* uniform a)
{
    const int test = (varying int) sizeof(vec3);
    a[programIndex] = test;
}
```

```
func___un_3C_uni_3E_:                   # @func___un_3C_uni_3E_
        vbroadcastss    ymm1, dword ptr [rip + .LCPI1_0] # ymm1 = [1.34524653E-43,1.34524653E-43,1.34524653E-43,1.34524653E-43,1.34524653E-43,1.34524653E-43,1.34524653E-43,1.34524653E-43]
        vmaskmovps      ymmword ptr [rdi], ymm0, ymm1
        vzeroupper
        ret
```

aneshlya:
Copying @pbrubaker comments from #3084 here:

https://ispc.godbolt.org/z/bvjdqb7Ph

```
unmasked void funcFail0(uniform int* uniform a)
{
    size_t test = sizeof(int);
    a[programIndex] = test;
}

unmasked void funcFail1(uniform int* uniform a)
{
    varying size_t test = (varying size_t) sizeof(int);
    a[programIndex] = test;
[...]

PR #3166, opened 2025-01-09
Fix crash when using sizeof as global variable initializer

aneshlya:
Fixes crash from #3084.

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -7929,7 +7929,8 @@ static llvm::Constant *lConvertPointerConstant(llvm::Constant *c, const Type *co
 
     // Handle conversion to int and then to vector of int or array of int
     // (for varying and soa types, respectively)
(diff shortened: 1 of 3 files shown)


nurmukhametov (approved):
LGTM!

dbabokin (approved):
Looks good.

One question though - do you really need to check IR and not just check that it's not crashing? This way the test will work on any platform.

aneshlya:
It's not necessary to check IR for this particular fix, but we have some issues with sizeof like https://github.com/ispc/ispc/issues/2775 and https://github.com/ispc/ispc/issues/2181 where the code produced doesn't look right. So i want to ensure here that it's right at least for global variables initialization.
Probably splitting a test into two is a good idea.

dbabokin:
Iʼve noticed that `tests/lit-tests/3084-1.ispc` started failing with LLVM trunk: https://github.com/ispc/ispc/actions/runs/12731369278/job/35488256887

Itʼs because it has the following syntax now: ` @test31 = local_unnamed_addr global <8 x i32> splat (i32 32)`

PR #3169, opened 2025-01-13
Fix leak of file handle in DispatchHeaderInfo

nurmukhametov:
This fixes the issue of leaking a file handle (DHI->file) if the file was opened but the compilation was terminated earlier for some reason, e.g., at https://github.com/ispc/ispc/blob/c9ee084affb75700d495a600f52b5ae7c2ff2173/src/module.cpp#L3776

### DIFF
--- a/src/module.cpp
+++ b/src/module.cpp
@@ -2737,8 +2737,15 @@ struct ispc::DispatchHeaderInfo {
     bool Emit4;
     bool Emit8;
     bool Emit16;
-    FILE *file;
-    const char *fn;
+    FILE *file = nullptr;
+    const char *fn = nullptr;
+
+    ~DispatchHeaderInfo() {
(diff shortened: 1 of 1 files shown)


dbabokin (commented):
The scope of `DHI` will end at line 3942, that's where it will be deallocated. Is it the intent? I.e. It will be deallocated much later than before the changes.

nurmukhametov:
> The scope of `DHI` will end at line 3942, that's where it will be deallocated. Is it the intent? I.e. It will be deallocated much later than before the changes.

I have indeed changed the live scope of the file handle because I thought that it didn't really matter. If it does, then the current version should preserve the original behavior.

dbabokin (approved):
LGTM

aneshlya (approved):

Issue #3008, opened 2024-09-16
Hierarchical linking of builtin files


nurmukhametov:
Implement the final step of the `target redesign` as described [here](https://github.com/ispc/ispc/blob/main/docs/design/TargetRedesign.md#hierarchical-implementation-of-builtin-functions).

nurmukhametov:
Resolved by #3171

PR #3171, opened 2025-01-14
Hierarchical builtins linking

nurmukhametov:
This PR implements #3008 and depends on #3140.

It introduces a hierarchical approach to traverse and link target-specific builtins. A target hierarchy map is added, which is traversed to get the parent (more generic) target for the current one. During traversal, target-specific builtins are linked if needed until all builtins are resolved.

Future work may include uncommenting and testing additional target mappings as needed.


src/builtins.cpp
```diff
@@ -434,26 +434,155 @@ bool lStartsWithLLVM(llvm::StringRef name) {
...
+    }
+    return ISPCTarget::none;
+}
+
+// Check if the module has used but unresolved symbols from the given set of symbols.
+bool lHasUnresolvedSymbols(llvm::Module *module, const llvm::StringSet<> &symbols) {
+    for (llvm::Function &F : module->functions()) {
+        auto name = F.getName().str();
+        if (F.isDeclaration() && !F.use_empty() && symbols.count(name)) {
+            return true;
```
aneshlya, line 509: I think it would be helpful if you pass one more parameter to this function `std::string& unresolvedSymbol` and set it `unresolvedSymbol = name;` here. It can be used in a error message  "Unresolved symbols in target bitcode".
But it's up to you.

src/builtins.cpp
```diff
@@ -434,26 +434,155 @@ bool lStartsWithLLVM(llvm::StringRef name) {
...
             targetBuiltins.insert(name);
         }
     }
+}
 
-    targetBCModule->setDataLayout(g->target->getDataLayout()->getStringRepresentation());
-    targetBCModule->setTargetTriple(module->getTargetTriple());
+llvm::Module *lGetTargetBCModule(ISPCTarget target) {
+    const BitcodeLib *targetBCLib = g->target_registry->getISPCTargetLib(target, g->target_os, g->target->getArch());
+    Assert(targetBCLib);
```
aneshlya, line 526: Assert is good but I think error message would be beneficial here too especially in ispc-slim mode.
``` 
   if (!targetBCLib) {
        Error(SourcePos(), "Failed to get target bitcode library");
        return nullptr;
    }
```

aneshlya (approved):
Looks good to me!

aneshlya:
I think we should disable warnings in generic.ispc so it doesn't produces lot's of messages during ispc build as here: https://github.com/ispc/ispc/actions/runs/12788320421/job/35649369303?pr=3171#step:5:510

nurmukhametov:
> I think we should disable warnings in generic.ispc so it doesn't produces lot's of messages during ispc build as here: https://github.com/ispc/ispc/actions/runs/12788320421/job/35649369303?pr=3171#step:5:510

```bash
builtins/generic.ispc:723:128: Performance Warning: Division with varying 
        integer types is very inefficient. 
EXT varying int8 __avg_up_int8(varying int8 a, varying int8 b) { return ((int16)a + (int16)b + 1) / 2; }
```

These warnings actually highlight the real issue in the implementation. I followed the implementation from util.m4 here: https://github.com/ispc/ispc/blob/b5fbcd0f6552be657c7d73bcaeead679a5b128c6/builtins/util.m4#L7332.

I don't understand why ashr wasn't used there in the first place.

Although we don't have an arithmetic shift to the right in the language, I can create an `llvm.ispc.ashr` intrinsic that is mapped to the ashr LLVM IR instruction and use it here instead of division by 2. I suppose that the generated code will be better and the warnings will be eliminated.

nurmukhametov:
> Although we don't have an arithmetic shift to the right in the language, I can create an `llvm.ispc.ashr` intrinsic that is mapped to the ashr LLVM IR instruction and use it here instead of division by 2. I suppose that the generated code will be better and the warnings will be eliminated.

After additional thought, I concluded that it is fine to use even logical shifts in these functions, see https://github.com/ispc/ispc/blob/6e1beb51f5518daf847042b1db30709c711d2009/builtins/generic.ispc#L726

dbabokin (approved):

aneshlya (approved):
Looks good!

Issue #3167, opened 2025-01-12
i16 and i8 targets are missing for the aarch64 architecture


csmulhern:
I would have expected `neon-i16x8` and `neon-i8x16` to work with aarch64, and for "double-pumped" targets to exist as well (`neon-i16x16` and `neon-i8x32`).

aneshlya:
You can see the current support matrix for neon targets here: 
https://github.com/ispc/ispc/blob/main/docs/ispc.rst#selecting-the-compilation-target

We're working on extending neon support.

PR #3176, opened 2025-01-15
Remove Armv7 and update neon-i8x16 and neon-i16x8 targets for ARMv8 (32-bit and 64-bit)

aneshlya:
Partially fixes https://github.com/ispc/ispc/issues/3167

This PR introduces breaking changes on ARM platform
1. `arm` architecture is remapped from armv7 (32-bit) to armv8 (32-bit)
2. ARMv7 CPU definitions (cortex-a9 and cortex-a15) are removed.
3. `neon-i8x16` and `neon-i16x8` targets are updated to be used on armv8 (both 32-bit and 64-bit). Previously these targets only supported 32-bit env.

On the way I fixed a bunch of typos and TODOs across neon targets.


cmake/CommonStdlibBuiltins.cmake
```diff
@@ -125,6 +125,8 @@ function (generate_stdlib_or_target_builtins func ispc_name CPP_LIST BC_LIST)
             # Not all targets have 64bit
             disp_target_stdlib(${func} ${ispc_name} neon-i32x4 64 ${os} ${CPP_LIST} ${BC_LIST})
             disp_target_stdlib(${func} ${ispc_name} neon-i32x8 64 ${os} ${CPP_LIST} ${BC_LIST})
+            disp_target_stdlib(${func} ${ispc_name} neon-i8x16 64 ${os} ${CPP_LIST} ${BC_LIST})
+            disp_target_stdlib(${func} ${ispc_name} neon-i16x8 64 ${os} ${CPP_LIST} ${BC_LIST})
```
nurmukhametov, line 129: It looks like now we can move them to the loop above like this:
```cmake
disp_target_stdlib(${func} ${ispc_name} ${target} 64 ${os} ${CPP_LIST} ${BC_LIST})
```

nurmukhametov (changes requested):

cmake/CommonStdlibBuiltins.cmake
aneshlya: Fixed

nurmukhametov (approved):
LGTM!

PR #3177, opened 2025-01-16
Add ISPC_ prefix to macro definitions in core.isph

nurmukhametov:
This is needed to avoid possible name conflicts with user's ones as requested here https://github.com/ispc/ispc/pull/3140#discussion_r1911764916


aneshlya (approved):
LGTM!

dbabokin (approved):
LGTM.

We are not fully following "big languages" practice to start internal stdlib names with double underscore prefix, but this is probably is good enough.

PR #3180, opened 2025-01-18
Check for incompatible attributes before setting optsize

aneshlya:
Fix setting of `optsize` attribute. Currently failing on print tests https://github.com/ispc/ispc/actions/runs/12824288745/job/35760635474 due to incompatible `optnone` and `optsize` attributes.


nurmukhametov (approved):
LGTM!

dbabokin (approved):
LGTM

Issue #3182, opened 2025-01-21
Enable ARM testing using new Github ARM runners


aneshlya:
https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

PR #3183, opened 2025-01-21
Enable ARM testing using new Github ARM runners

aneshlya:
Fixes https://github.com/ispc/ispc/issues/3182.
Plus fixed logic of skipping tests in run_tests.py (now the tests which are supposed to be skipped on generic targets only are skipped everywhere).
Also I pinned the version of actionlint.


.github/workflows/scripts/install-build-deps-aarch64.sh
dbabokin, line 1: With shared runners, do we want different versions of `install-build-deps` script for different architectures? Cross compiler will be different, but we can install all of them or do that based on the architecture.

.github/workflows/ispc-ci.yml
```diff
@@ -53,7 +55,9 @@ jobs:
     runs-on: ubuntu-22.04
     outputs:
       tests_matrix_targets: ${{ steps.set-flow.outputs.matrix }}
```
dbabokin, line 57: Should we rename add `_x86` suffix here?

.github/workflows/ispc-ci.yml
```diff
@@ -53,7 +55,9 @@ jobs:
...
+      tests_matrix_targets_arm: ${{ steps.set-flow.outputs.matrix_arm }}
       tests_optsets: ${{ steps.set-flow.outputs.optsets }}
+      tests_optsets_array: ${{ steps.set-flow.outputs.optsets_array }}
```
dbabokin, line 60: ARM and x86 paths are different. One uses alloy, another runs `run_tests.py` directly. Should we unify them?

aneshlya: Right, to be honest, I'm not sure. I'd rather avoid adding ARM support to `alloy.py`—in fact, I believe we should consider deprecating it. Using `run_tests.py` is preferable as it allows for the creation of smaller, more focused tasks to test specific configurations more efficiently. But currently, `alloy.py` is more convenient for x86 targets due to its built-in SDE logic.

dbabokin (approved):
Feel free to merge, when you think the changes are ready.

.github/workflows/scripts/install-build-deps-aarch64.sh
aneshlya: Done.

.github/workflows/ispc-ci.yml
aneshlya: Done

PR #3186, opened 2025-01-23
Generic target and neon perf

nurmukhametov:
This PR improves the implementation of generic builtins, allowing for broader reuse across all neon targets.

The reuse of the generic implementation improves performance on the GameDev benchmark, making it approximately 3% faster on average. Some tests are up to 30% faster, while a few are around 10% slower.

This needs to be merged after #3190.


aneshlya (approved):
Looks good to me!

aneshlya:
Would be great to have an average perf improvement on neon targets in PR description (for future release notes).

aneshlya (approved):

PR #3187, opened 2025-01-23
Remove knl target

nurmukhametov:
This removes `avx512knl-x16` target fixing #2867.


docs/ispc.rst
dbabokin, line 1: Add a removal notice for "what's new section" for the upcoming release.

dbabokin (approved):
Looks good.

The irony here is that ISPC was created in the first place to support Knights family (Larrabee at the time).

docs/ispc.rst
aneshlya: It's already in `Updating ISPC Programs For Changes In ISPC 1.26.0`

PR #3190, opened 2025-01-24
Set CPU for generic targets

nurmukhametov:
This PR should repair full tests run for generic targets.

The job result is here: https://github.com/nurmukhametov/ispc/actions/runs/12949417763

This PR sets CPU in CPU_x86_64 for generic targets when `--cpu` is not explicitly specified on command line. Otherwise, the oldest CPU model supported by LLVM is picked in backend that results in very old CPUs in case of `i686-*` triple (when ISPC compiles with `--arch=x86`).

### DIFF
--- a/src/ispc.cpp
+++ b/src/ispc.cpp
@@ -2116,6 +2116,15 @@ Target::Target(Arch arch, const char *cpu, ISPCTarget ispc_target, PICLevel picL
 
     if (CPUID == CPU_None) {
         cpu = a.GetDefaultNameFromType(CPUfromISA).c_str();
+        std::string cpu_string = cpu;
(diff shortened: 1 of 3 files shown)


dbabokin:
What is we compile it on ARM?

nurmukhametov:
> What is we compile it on ARM?

I had the perception that ARM compilation should pick the CPU earlier, as seen here: https://github.com/ispc/ispc/blob/81b1fa9a7f62308b34226498811bbd73a7d68269/src/ispc.cpp#L2112 However, I have added an additional check for x86* to avoid unforeseen errors for other architectures when/if they appear.

dbabokin (approved):

Issue #3143, opened 2024-12-17
'no platform load command found' warning when linking with Xcode 16.1


alanmikhak:
I see this issue on Apple M1 with Xcode 16.1 (16B40) and ispc-1.16.1 even if I specify --target-os=macos. It seems ispc-1.16.1 doesn't inject any platform load command into the resulting .o file. Has this issue been resolved in newer versions of ispc?

Here is how I compile the source code with ispc-1.16.1 on macos
% ~/tools/intel.com/ispc-1.16.1/bin/ispc --arch=x86_64 --target-os=macos -O2 cvTimexMethods.ispc -o x86_64/cvTimexMethods_ispc.o
% ~/tools/intel.com/ispc-1.16.1/bin/ispc --arch=aarch64 --target-os=macos --target=neon -O2 cvTimexMethods.ispc -o arm64/cvTimexMethods_ispc.o

Examine the ispc built .o with vtool
% cd x86_64
% vtool -arch x86_64 -show cvTimexMethods_ispc.o
cvTimexMethods_ispc.o (architecture x86_64):

By comparison, .o build with clang++ has a platform load command.
% vtool -arch x86_64 -show cvTimexMethods_cl.o
cvTimexMethods_cl.o (architecture x86_64):
Load command 1
cmd LC_BUILD_VERSION
cmdsize 24
platform MACOS
minos 13.3
sdk 15.1
ntools 0

alanmikhak:
I reproduced this issue with ispc-1.25.3 and Xcode 16.1 (16B40) on the same Apple M1 machine with either Release or Debug builds for either x86_64 or aarch64.

% ~/tools/intel.com/ispc-1.25.3/bin/ispc --arch=x86_64 --target-os=macos --target=sse4-i32x4 --nowrap -O2 cvTimexMethods.ispc -o x86_64/cvTimexMethods_ispc.o
% ~/tools/intel.com/ispc-1.25.3/bin/ispc --arch=aarch64 --target-os=macos --target=neon-i32x4 --nowrap -O2 cvTimexMethods.ispc -o arm64/cvTimexMethods_ispc.o

% cd x86_64
% vtool -arch x86_64 -show cvTimexMethods_ispc.o 
cvTimexMethods_ispc.o (architecture x86_64):

% cd ../arm64
% vtool -arch arm64 -show cvTimexMethods_ispc.o
cvTimexMethods_ispc.o (architecture arm64):

dbabokin:
[...]

PR #3191, opened 2025-01-24
Add a flag to set min os version on darwin platforms

aneshlya:
This PR introduces option to set min OS version for Darwin platforms.
Fixes: https://github.com/ispc/ispc/issues/3143

### DIFF
--- a/docs/ispc.rst
+++ b/docs/ispc.rst
@@ -701,6 +701,10 @@ recovery. Such users must update their code to handle ``SIGILL`` instead. The
 change improves predictability and eliminates reliance of the dispatcher on the
 C standard library.
 
+The ``--darwin-version-min`` option was added to specify the minimum deployment
+target version for macOS and iOS applications addressing the new linker
+introduced in Xcode 15.0 that issues a warning when no version is provided.
+
 Getting Started with ISPC
 =========================
 
@@ -915,6 +919,12 @@ that removes unused sections: ``--gc-sections`` for ``GNU ld`` and ``/OPT:REF``
 for ``MSVC link.exe``. On macOS, this flag does not have any effect (as in
 clang) because dead stripping ``-dead_strip`` for ``ld64`` works differently.
 The ``-fno-function-sections`` disables this behavior.
+
+The ``--darwin-version-min`` option was added to specify the minimum deployment
+target version for macOS and iOS applications addressing the new linker
+introduced in Xcode 15.0 that issues a warning when no version is provided.
+The version should be specified in the format ``[major.minor]``. When an empty
+string ("") is passed, no minimum version will be specified in the output binary.
 
 Selecting The Compilation Target
 --------------------------------
--- a/src/ispc.cpp
+++ b/src/ispc.cpp
@@ -2351,6 +2351,7 @@ std::string Target::SupportedCPUs() {
 
 std::string Target::GetTripleString() const {
     llvm::Triple triple;
+    llvm::VersionTuple darwinVersionMin = g->darwinVersionMin;
     switch (g->target_os) {
     case TargetOS::windows:
         if (m_arch == Arch::x86) {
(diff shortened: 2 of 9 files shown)


nurmukhametov (approved):
LGTM!

src/main.cpp
```diff
@@ -67,6 +67,8 @@ static void lPrintVersion() {
 #endif
     printf("    [--cpu=<type>]\t\t\tAn alias for [--device=<type>] switch\n");
     printf("    [-D<foo>]\t\t\t\t#define given value when running preprocessor\n");
+    printf("    [--darwin-version-min=<major.minor>]\tSet the minimum macOS/iOS version required for the "
```
dbabokin, line 70: I'm wondering if we should put it under `ISPC_MACOS_TARGET || ISPC_IOS_TARGET`. Any thoughts?

tests/lit-tests/darwin-minos-ver.ispc
```diff
@@ -0,0 +1,15 @@
+// The test checks that the triple contain minimum OS version if provided.
+
+// RUN: %{ispc} %s --nostdlib --target-os=macos --emit-llvm-text --nowrap --target=host -o - | FileCheck %s --check-prefix=CHECK-MACOS-DEFAULT
```
dbabokin, line 3: We advertise it for macOS/iOS, but test it for macOS only. Should we do that for iOS?..

Though that might be not straightforward, as we don't enable iOS in default CI builds. We should probably test it at least locally.

tests/lit-tests/darwin-minos-ver.ispc
```diff
@@ -0,0 +1,15 @@
...
+// RUN: %{ispc} %s --nostdlib --target-os=macos --emit-llvm-text --nowrap --target=host --darwin-version-min=10.12 -o - | FileCheck %s --check-prefix=CHECK-MACOS-VER
+// RUN: not %{ispc} %s --nostdlib --target-os=ios --emit-llvm-text --nowrap --target=host --darwin-version-min=a.b -o - 2>&1 | FileCheck %s --check-prefix=CHECK-ERROR-VER
```
dbabokin, line 5: Should we test a more complex / realistic case here? I.e. build ispc object file, build c++ object file, link and check 1) no warnings 2) supported os in the binary?

To check the supported OS we can do (that's what Claude suggests, we should use something similar):
```
otool -l <binary> | grep -A 4 "LC_VERSION_MIN_MACOSX"
# or for newer binaries:
otool -l <binary> | grep -A 4 "LC_BUILD_VERSION"
```

dbabokin (commented):
We should definitely add a section to the documentation explaining this problem that we are trying to solve and what this options does.

src/ispc.cpp
```diff
@@ -2443,7 +2444,17 @@ std::string Target::GetTripleString() const {
...
-        triple.setOS(llvm::Triple::OSType::MacOSX);
+        // If nothing is specified in command line - set default
+        if (darwinVersionMin.empty()) {
+            darwinVersionMin = (m_arch == Arch::x86_64) ? llvm::VersionTuple(10, 12) : llvm::VersionTuple(11, 0);
+        }
+        if (darwinVersionMin != darwinUnspecifiedVersion) {
+            triple.setOSName(llvm::Triple::getOSTypeName(llvm::Triple::OSType::MacOSX).str() +
+                             darwinVersionMin.getAsString());
+        } else {
+            // If empty string was specified for --darwin-version-min, do not et any version in the triple
```
dbabokin, line 2455: "do no Set"

docs/ispc.rst
```diff
@@ -701,6 +701,10 @@ recovery. Such users must update their code to handle ``SIGILL`` instead. The
 change improves predictability and eliminates reliance of the dispatcher on the
 C standard library.
 
+The ``--darwin-version-min`` option was added to specify the minimum deployment
+target version for macOS and iOS applications which is required by new linker
```
dbabokin, line 705: It's not required by new linker, but the new linker will issue a warning for that.

dbabokin (approved):
LGTM

src/main.cpp
aneshlya: Added

tests/lit-tests/darwin-minos-ver.ispc
aneshlya: Done

src/ispc.cpp
aneshlya: Fixed

docs/ispc.rst
aneshlya: Fixed

dbabokin (approved):
LGTM

PR #3192, opened 2025-01-25
Update .gitignore

pbrubaker:
Update .gitignore to include ./CMakeFiles/* which is populated when building with superbuild.


pbrubaker:
The `${ISPC_HOME}/CMakeFiles` directory seems to be created even if the build is executed inside `${ISPC_HOME}/build`  Perhaps there is a better solution here that I'm not seeing?

aneshlya:
> The `${ISPC_HOME}/CMakeFiles/*` directory seems to be created even if the build is executed inside `${ISPC_HOME}/build` Perhaps there is a better solution here that I'm not seeing?

@nurmukhametov, do you have any recommendations?

nurmukhametov:
> The `${ISPC_HOME}/CMakeFiles/*` directory seems to be created even if the build is executed inside `${ISPC_HOME}/build` Perhaps there is a better solution here that I'm not seeing?

I am sure that this directory is not created when superbuild is executed inside `${ISPC_HOME}/build`.

dbabokin:
> > The `${ISPC_HOME}/CMakeFiles/*` directory seems to be created even if the build is executed inside `${ISPC_HOME}/build` Perhaps there is a better solution here that I'm not seeing?
> 
> I am sure that this directory is not created when superbuild is executed inside `${ISPC_HOME}/build`.

Does it depend on the command that is used to do the superbuild?

pbrubaker:
Okay, I'm going to withdraw this for now.  I am almost certain that I was calling build from within `${ISPC_HOME}/builds` and `${ISPC_HOME}/build-ispc-stage2/src/ispc-stage2-build` and I was surprised to see these files show as created in GitHub desktop.  However I just tried doing two clean builds and I can't replicate it.  I am not sure what happened.

PR #3196, opened 2025-01-28
dispatcher: remove dependency on C abort

### LINKED ISSUE #3163: ISPC multiple runtime libc dependencies on Windows
ISPC seems to depend on multiple libc versions for different things.

1. Doing `ispc --help` from the regular command line produces errors that `MSVCP140.dll` is missing, as well as `VCRUNTIME140.dll` and `VCRUNTIME140_1.dll`. Running the same command from Visual Studio command line removes those errors. I do not think those dependencies are necessary either way.

2. Compiling object file is impossible without libc, for example `ucrt`.

ISPC program compiled to sse2 and avx2:
```c
export void data_double(uniform int data[], uniform int data_count)
{
 foreach (i = 0 ... data_count)
 {
  data[i] *= 2;
 }
 assert(0);
}
```
C program:
```c
#pragma comment(linker, "/subsystem:console")
#pragma comment(linker, "/entry:enter")
#pragma comment(lib, "kernel32")
#pragma comment(lib, "ucrt") // this is necessary
#include "demo.ispc.h"

void enter(void)
{
 int data[20] = {0};
 for (int i = 0; i < 20; i++)
 {
  data[i] = i;
 }
 data_double(data, 20);
}
```
With `ucrt` removed, errors are produced:
```
demo.ispc.obj : error LNK2019: unresolved external symbol abort referenced in function __set_system_isa
demo.ispc_sse2.obj : error LNK2001: unresolved external symbol abort
demo.ispc_avx2.obj : error LNK2001: unresolved external symbol abort
demo.ispc_sse2.obj : error LNK2019: unresolved external symbol puts referenced in function data_double___un_3C_uni_3E_uni_sse2
demo.ispc_avx2.obj : error LNK2001: unresolved external symbol puts
```

`__set_system_isa` seems to correspond to dynamic dispatch of the program between architectures, the rest are due to `assert` being present.

In my opinion, libc is an unnecessary dependency, as printing procedures could be implemented without libc dependency trivially, and dynamic dispatch can use a Window intrinsic to get CPU features. I know that libc is basically necessary on Linux, as it acts as a platform layer of a kind, but on Windows the platform layer are all the standard dlls, such as `kernel32` and `user32` and the rest. `ucrt` is also present, but only since Windows 10, so using it would break compatibility with earlier versions.

Whatever is the decision of the team, I'm fine with it - whether you decide to stick with a bunch of different libcs, or use `ucrt`, or remove the dependency altogether. It's merely inconvenient that I will have to work around it by doing my own dispatch, since in my program I do not link to any libc, to avoid having a user environment dependency.

nurmukhametov:
Replace the dependency on the C abort function with a custom implementation (`__terminate_now`), which directly invokes the `ud2` instruction.

This change introduces a breaking change for compiled user programs. Specifically, when no ISA is detected, the behavior changes from triggering an abort to executing an illegal instruction, meaning a different signal is raised (`SIGABRT` → `SIGILL`). User code that installs signal handlers for `SIGABRT` and processes them in any way will break and require modifications.

ISPC does not appear to guarantee any specific behavior for this case, so this change is likely acceptable. However, it is beneficial to document this behavior explicitly.

This fixes #3163 (at least the most important its part).

### DIFF
--- a/builtins/dispatch.c
+++ b/builtins/dispatch.c
@@ -21,9 +21,13 @@
 #error "Either REGULAR or MACOS macro need to defined"
 #endif
 
-void abort();
-
 static int __system_best_isa = -1;
+
+void __terminate_now() {
+    // Terminate execution using x86 UD2 instruction. UD2 raises an invalid
+    // opcode exception, ensuring program termination.
+    __asm__ __volatile__("ud2");
+}
 
 static void __cpuid(int info[4], int infoType) {
     __asm__ __volatile__("cpuid" : "=a"(info[0]), "=b"(info[1]), "=c"(info[2]), "=d"(info[3]) : "0"(infoType));
@@ -146,7 +150,7 @@ static int __get_system_isa() {
         if (skx) {
             return 7; // SKX
         } else if (knl) {
-            abort(); // KNL
+            __terminate_now(); // KNL
         }
         // If it's unknown AVX512 target, fall through and use AVX2
         // or whatever is available in the machine.
@@ -167,9 +171,12 @@ static int __get_system_isa() {
         return 1; // SSE4.1
     } else if (sse2) {
         return 0; // SSE2
-    } else {
-        abort();
     }
+
+    __terminate_now();
+
+    // This is actually unreachable, but needed to suppress the warning.
+    return -1;
 }
 
 void __set_system_isa() {
--- a/docs/ispc.rst
+++ b/docs/ispc.rst
@@ -693,6 +693,13 @@ deprecated. It leads to a warning when the function is called.
 
 The target ``avx512knl-x16`` was removed.
 
+The release introduces a breaking change by altering the behavior of user
+programs when no supported ISA is detected in the auto-dispatch code. It
+replaces the previous ``SIGABRT`` signal with ``SIGILL``. This affects users
+who rely on ``SIGABRT`` in their signal handlers for error handling or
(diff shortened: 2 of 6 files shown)


aneshlya (approved):
Looks good!

dbabokin (approved):
The changes look good.

Though it's difficult to say that I'm completely happy with the changes. I.e. I don't think that I'd implemented it differently, given the problems that we are trying to solve here. But from external observer's prospective (who doesn't know all the issues that we are trying to solve here), it does look strange.

It does look strange, because:
(1) we should operate under the assumption that we interoperate with C/C++ program, and hence libc should not be an issue. I know that it is exactly an issue here, but still.
(2) we push strange behavior to all platforms just because of weird Windows problem. Though it still might be better to be consistent across the platforms.

What are possible alternatives here, especially if we were designing the compiler from scratch?
(1) Have "bare metal" mode of operation - no dependencies for the generated code from any APIs/libraries?
(2) an option to use user-provided alternatives for APIs/library calls that we use (might be part of "bare metal" solution.
(3) have a wrapper for all our library calls, possibly? A-la `ispc-c` lib and have different implementation for different requirements. This is probably the most portable solution, especially if we want to execute on different types of devices (like CPU, GPU, etc). But this is the least convenient at it one more entity for the use to worry about.

Anyway, all this is to brainstorm on possible catches with the current solution and existing alternatives.

Current PR looks good, please proceed with this fix.

PR #3198, opened 2025-01-30
Use target specific intrinsics for masked load/store operations for int8/int16 types

### LINKED ISSUE #3181: Poor code generation when explicitly handling instances (but not with foreach)
I'm working with a kernel that does selective updates. A reduced version would look like:

```c
void update(uniform int8 values[16], bool upd, uniform int8 newVal) {
    foreach(index = 0 ... 16) {
        values[index] = select(upd, newVal, values[index]);
    }
}
```
On `avx512icl-x16`, this compiles to a nice 6-instruction function.

However, when using an explicit `for` loop with `programCount`/`programIndex`, like the commented-out loop header in the linked Godbolt snippet, the code seemingly de-vectorizes, and turns into what I can only describe as a monstrosity, involving some 70 instructions and a ton of branches.

https://godbolt.org/z/MEWdvo7vh

However, since I would like to use the explicit for in this kernel (for unrelated reasons), this is kind of annoying. Is there a way to let ispc generate the good code in that scenario? The performance guide suggests it should be possible:

> Sometimes the running program instances will access a linear sequence of memory locations; this happens most frequently when array indexing is done based on the built-in programIndex variable. In many of these cases, the compiler is also able to detect this case and then do a vector load. For example, given:

>> for (int i = programIndex; i < count; i += programCount)
>>   // process array[i];

> Regular vector loads and stores are issued for accesses to array[i].

aneshlya:
This PR introduces using of target-specific instructions for masked load/store operations for int8/int16 types on avx512 where is directly maps to HW instructions. These instructions are covered by `avx512_bw` and `avx512_vl` features which are available starting skx (but were not available in knl)

These implementations still performs better then corresponding generic targets. However eventually we should replace remaining masked load/store functions coming from utils to generic implementation which produces much better code.

Fixes #3181.

### DIFF
--- a/builtins/target-avx512-common-16.ll
+++ b/builtins/target-avx512-common-16.ll
@@ -1,4 +1,4 @@
-;;  Copyright (c) 2015-2024, Intel Corporation
+;;  Copyright (c) 2015-2025, Intel Corporation
 ;;
 ;;  SPDX-License-Identifier: BSD-3-Clause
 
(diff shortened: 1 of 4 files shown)


nurmukhametov (approved):
LGTM!

dbabokin (approved):
LGTM!

PR #3199, opened 2025-01-30
Add definitions for LLVM 20 

aneshlya:
LLVM 20 has branched out.
This PR also fixes build with LLVM trunk and couple of lit-tests related to conversion of `nocapture` to `captures(none)`.


dbabokin (approved):
Change looks good, but would be good to add 20.1 definition to alloy and superbuild, so we can test it locally before LLVM is released.

dbabokin (approved):

PR #3200, opened 2025-01-31
Add lanemask benchmark

nurmukhametov:
Add benchmark to test `lanemask` function. 

It also fixes loop unroll test and `run_tests.py`.


aneshlya (approved):
Looks good! Do you have any performance results?

nurmukhametov:
> Looks good! Do you have any performance results?

I used the lanemask benchmark to measure the performance impact of changing `__movmsk` implementation of `neon-i32x4` from this one (link) to that one (link). The results are as follows:

```bash
lanemask_test/256_median                    -0.1496         -0.1496           102            87           102            87
lanemask_test/4096_median                   -0.1391         -0.1391          1593          1372          1593          1371
lanemask_test/32768_median                  -0.1321         -0.1321         13107         11375         13104         11373
lanemask_test/1048576_median                -0.0814         -0.0815        519976        477640        519812        477465
```

It matches the average improvements across GameDev benchmarks.

I used the ArrayMean test to estimate the performance difference between ISPC and C++. For this test, `neon-i32x4` is about 2x slower than C++ with `-O3 -ffast-math`, but the double-pumped target `neon-i32x8` is on par.

PR #3201, opened 2025-02-03
Fix generic targets with `-O0`

nurmukhametov:
This PR fixes the tests (https://github.com/ispc/ispc/actions/runs/13108110736). Nightly runs revealed that for generic/neon targets, the `__masked_blend_store` implementation from https://github.com/ispc/ispc/commit/1f1a117743cf5110e7b109b4cd475f1ef83a6bac is effectively recursive for `-O0`. This happens because ISPC generates masked stores even inside blocks where the mask is certainly `all_on` (this occurs when the first IR is generated). Later on, ISPC relies on optimizations to replace such masked stores with usual ones.

To fix that, introduce `llvm.ispc.blend_store` and use it for `__masked_store_blend` implementation.


aneshlya (approved):
Looks good!

PR #3203, opened 2025-02-04
Fix attributes for functions returning pointers

### LINKED ISSUE #3202: Fatal error on (I think) pointer-to-pointer in ISPC 1.25.3
This contrived program causes ISPC 1.25.3 to crash:

```ispc
uniform int** f(uniform int** p) { }
```

```
$ ispc bug.ispc
Warning: No output file or header file name specified. Program will be compiled and
        warnings/errors will be issued, but no output will be generated.
Warning: No --target specified on command-line. Using default system target
        "avx2-i32x8".
/build/source/src/module.cpp:1242: Assertion failed: "decl && decl->functionParams.size() == nArgs".
***
***
/build/source/src/main.cpp(414): FATAL ERROR: Unhandled signal sent to process; terminating.
Aborted (core dumped)
```

This is as small as I could make the program. The problem is unrelated to the empty function body.

nurmukhametov:
This PR fixes #3202 and a broader issue with functions attributes that were not working for functions returning pointer types.


aneshlya (approved):
Looks good!

dbabokin (approved):

PR #3204, opened 2025-02-04
Fix predefined macro values

### LINKED ISSUE #2167: Incomplete preprocessor documentation
The documentation regarding the preprocessor capabilities is incomplete. Testing all macros defined at https://en.cppreference.com/w/cpp/preprocessor/replace (as well as the C language version) with ISPC 1.16 (https://ispc.godbolt.org/), I got the following results:

#### Defined in `.ispc` file:

- `__VA_ARGS__`: C++11/C99: Very useful for macro definitions.
- `__VA_OPT__`: C++20: Also very useful.
- `__STDC__`, `__STDC_HOSTED__`
`__FILE__`, `__LINE__`, `__DATE__`, `__TIME__`
- `__STDC_UTF_16__`, `__STDC_UTF_32__`: Do not seem to be useful, but still defined. Should these be defined at all?

#### NOT defined in `.ispc` file:

- `__STDC_VERSION__`: Defines the underlying C standard. Might be helpful to clarify if ISPC is like C89 or C99 + extensions. It is a mixture of both, but specifying a clear minimum standard could be helpful.
- All other macros from https://en.cppreference.com/w/cpp/preprocessor/replace

For instance, Open3D (https://github.com/isl-org/Open3D) already uses `__VA_ARGS__` to simulate templates via macros. Declaring this macro as well as the other working ones as officially supported would be very helpful for all users.

nurmukhametov:
This PR fixes #2167 by updating the documentation about predefined macros and fixing the values for some of them.

### DIFF
--- a/docs/ispc.rst
+++ b/docs/ispc.rst
@@ -86,6 +86,7 @@ Contents:
   + `Selecting The Compilation Target`_
   + `Selecting 32 or 64 Bit Addressing`_
   + `The Preprocessor`_
+  + `Pragma Directives`_
   + `Debugging`_
   + `Optimization Settings`_
   + `Other ways of passing arguments to ISPC`_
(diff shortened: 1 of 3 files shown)


docs/ispc.rst
```diff
@@ -1176,6 +1177,51 @@ preprocessor runs:
...
+    void test_va_args() {
+        PRINT("% % %\n", 0, 1, 2);
+        // you can't call PRINT with just string because of trailing comma in macro expansion
+        // PRINT("Hello, World!\n"); is compilation error
+        // call EPRINT with __VA_OPT__(,) instead
+        EPRINT("Hello, World!\n");
+    }
+
+
+pragma Directives
```
aneshlya, line 1222: Starting from lowercase looks strange. I think it's ok to have `Pragma Directives` as section title or `Use of #pragma in ISPC`

aneshlya (approved):

docs/ispc.rst
```diff
@@ -1176,6 +1177,51 @@ preprocessor runs:
...
+
+Variadic macros are supported in ``ispc``. The ``__VA_ARGS__`` and
+``__VA_OPT__`` macros are defined inside a variadic macro definition.
+
+``__VA_ARGS__`` is a variable argument list macro that represents the arguments
+after the last named argument.
+
+``__VA_OPT__(...)`` is a function macro that expands to its argument if the
+variable argument has any tokens, but if the variable argument does not have
+any tokens, the __VA_OPT__ expands to nothing.
```
dbabokin, line 1204:
```suggestion
any tokens, the ``__VA_OPT__`` expands to nothing.
```

dbabokin (approved):

PR #3209, opened 2025-02-06
Skip 2954.ispc

nurmukhametov:
This test passes when built with LLVM 20.0 or with LLVM 18/19 built with the backport patch llvm_patches/18_1_19_1_InstCombine-foldVecExtTruncToExtElt-extend.patch It is better to skip it as UNSUPPORTED for the release because many package maintainers will not have older LLVMs patched.

We also need to update LLVM 18/19 prebuilt archives to enable this test in future.

### DIFF
--- a/tests/lit-tests/2954.ispc
+++ b/tests/lit-tests/2954.ispc
@@ -1,6 +1,11 @@
 // RUN: %{ispc} -O2 --target=avx2-i32x8 --emit-asm --x86-asm-syntax=intel %s -o - 2>&1 | FileCheck %s
 
-// XFAIL: !LLVM_20_0+
+// This test passes when built with LLVM 20.0 or with LLVM 18/19 built with the
+// backport patch llvm_patches/18_1_19_1_InstCombine-foldVecExtTruncToExtElt-extend.patch
+// It is better to skip it as UNSUPPORTED for the release because many package
+// maintainers will not have older LLVMs patched.
+
+// UNSUPPORTED: !LLVM_20_0+
 
 // REQUIRES: X86_ENABLED
 


aneshlya (approved):

jcaesar:
You silently moved v.1.26.0 based on this, right? It's not a big deal, but I do would like to say hi and that it's noticeable downstream, e.g. https://github.com/NixOS/nixpkgs/pull/379990

aneshlya:
Hi @jcaesar, thanks for bringing this up. We typically don’t consider the creation of a release tag as a "public" release. Due to infrastructure issues, we had to move it once more, but we won’t do so in the future. I agree that this is a bad practice, and we appreciate your patience.

nurmukhametov:
> You silently moved v.1.26.0 based on this, right? It's not a big deal, but I do would like to say hi and that it's noticeable downstream, e.g. [NixOS/nixpkgs#379990](https://github.com/NixOS/nixpkgs/pull/379990). From that point of view, calling it 1.26.1 or so would have been favorable.

We haven't done a new release yet. This can be easily verified by checking the release page or by making a request to this API: https://api.github.com/repos/ispc/ispc/releases/latest. Both clearly show that the current latest release is v1.25.3, not v1.26.0. How and why did you decide that we made a new release to bump the version and update it in `NixOS`?

jcaesar:
I'm not maintainer of that package, nor have I merged the version bump, nor am I a frequent contributor to nixpkgs, so take this with a grain of salt:

Nixpkgs has >100.000 packages. To keep that all reasonably up to date, a bot auto-opens PRs like [the one updating ispc](https://github.com/NixOS/nixpkgs/pull/379816). From there, the maintainers may apply some scrutiny, but if I were to find a commit saying [Bump version to X](https://github.com/ispc/ispc/commit/edb4ac7363aa12761057227af4ea2a912bc54866) and a tag vX on that, I'd assume a release was made. After all, as pointed out, moving tags around is bad practice. There are quite a few [downstreams](https://repology.org/project/ispc/versions) packaging ispc, so some coming to that same conclusion is likely. Gentoo e.g. also went [that way](https://github.com/gentoo/gentoo/commit/96456da43c4f4775ce55b0efae4b847542402aae).

Also, if I were to scrutinize, I might start wondering: I've seen upstreams silently stopping doing GH releases, and I've seen projects that delete and re-create GH releases it with a different commit. There is no perfect defense here. If you can, please don't make it plain and obvious what a release is. There should be other ways to get GHA moving.

(Also, sorry for usurping this PR for the discussion. I might want to open an issue on nixpkgs instead.)

nurmukhametov:
> Nixpkgs has >100.000 packages. To keep that all reasonably up to date, a bot auto-opens PRs like [the one updating ispc](https://github.com/NixOS/nixpkgs/pull/379816).

Thank you, I understand better now what happened. I just couldn't believe that the maintainer of ISPC didn't know the usual release process of our compiler.

> There are quite a few [downstreams](https://repology.org/project/ispc/versions) packaging ispc, so some coming to that same conclusion is likely. Gentoo e.g. also went [that way](https://github.com/gentoo/gentoo/commit/96456da43c4f4775ce55b0efae4b847542402aae). 

I would say the same wrong conclusion.

>  After all, as pointed out, moving tags around is bad practice.

I totally agree. 

The problem here is that, as it currently works, we can start building release archives only after we have made the release tag. I am not sure why the internal CI and ISPC build system rely on the presence of the tag, but it is so. In my opinion, there are two principal ways of fixing this and avoiding any tag moves around in the future:

1. Keep the current build/CI system as is. Tag something like vN-rc1 first, then build release archives, test them, and if everything is fine, tag the same SHA as vN and repeat the release build, then publish. If something is wrong, then we need to create another tag, vN-rc2, and repeat the process.
2. Modify CMake/CI to build release archives from virtually any SHA (probably providing extra options like the release version or deducing it from the common.h). Having this, we can build release archives from any SHA, test them, and if everything is fine, tag this SHA as vN and publish them. If something is wrong, then we just need to fix it, rebuild a new SHA, test it, and tag it.

Approach 1 is simpler and uses the existing system but involves redundant builds and can lead to a few extra tags (will the maintainers package them?). Approach 2 offers greater flexibility and efficiency but requires changes to the build/CI system.

> (Also, sorry for usurping this PR for the discussion. ...)

It is fine. I am also use this PR for slightly unrelated topics.

alexfmpe:
https://github.com/ispc/ispc/pull/3382 now implements approach 2 right?

PR #3214, opened 2025-02-10
Add release pipeline

### LINKED ISSUE #3216: Add aarch64 release pipeline
Use public arm runner for it.

nurmukhametov:
It can build release archives for both Linux aarch64 and x86_64. This fixes #3216

### DIFF
--- a/.github/workflows/release-pipeline.yml
+++ b/.github/workflows/release-pipeline.yml
@@ -1,0 +1,75 @@
+# Copyright 2025, Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause
+
+name: Release Pipeline
+
+permissions: read-all
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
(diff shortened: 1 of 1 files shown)


dbabokin (commented):
I think we should make it work for regular nightly / weekly builds. This is not much different. To do that we should treat `main` a bit differently (actually not just `main`, but anything not matching `vX.X.X`) - use different path to the Dockerfile and pass `SHA=main` argument.

dbabokin:
How long does this pipeline runs on x86 and arm? I guess with current default of 4-core machines, this should run below 6-hours threshold.

nurmukhametov:
> How long does this pipeline runs on x86 and arm? I guess with current default of 4-core machines, this should run below 6-hours threshold.

It is ~1.5h (aarch64) and ~3h (x86), e.g., here: https://github.com/nurmukhametov/ispc/actions/runs/13203846598

Issue #1192, opened 2016-04-01
Update wiki


michaeljs1990:
Is it still recommended to use llvm 3.3 it seems like the entire wiki is a few years out of date.

dbabokin:
Yes, wiki is a little bit out of date. Recommended version of LLVM is 3.8.

RoyiAvital:
Are there plans to update the Wiki?

It would be great to have Users Samples section there.

dbabokin:
Yes, we still plan to update the wiki, it's just not the top priority task at the moment.

I'm not sure what User Samples section you'd like to see on the wiki. I think examples from the package do pretty good job, plus examples from the documentation. If you feel that we have not enough examples covering your problem domain, you are welcome to contribute examples, which you find relevant.

aneshlya:
@nurmukhametov , I think it can be closed as completed.

PR #3215, opened 2025-02-10
Quick start build scripts

nurmukhametov:
This PR adds build scripts (a bash script scripts/quick-start-build.sh for Linux/macOS and a PowerShell script scripts/quick-start-build.ps1 for Windows). These scripts can be used to build ISPC with prebuilt Clang/LLVM fetched from https://github.com/ispc/ispc.dependencies/releases. They serve as convenient entry points for external developers wanting to contribute to ISPC.

To use them, simply run them from the bash/PowerShell powered command line. There is an optional command line argument to specify the LLVM version to download. For example, `./quick-start-build.sh 19` will download LLVM version 19, unpack it, and build ISPC with this LLVM version. These scripts assume that prerequisites are already installed. The PowerShell script requires `7z` to be installed to unpack archives. The bash script requires `grep`, `sed`, `tar`, and `curl` to be installed.


nurmukhametov:
This relates to #1192

aneshlya:
I haven’t reviewed `./quick-start-build.sh` yet, but I’m not a fan of adding new shell scripts like `run_targets.sh` and `run_native_targets.sh` as wrappers around `run_tests.py`. They seem like a non-cross-platform reincarnation of `alloy.py`. I would like these scripts as one Python script.
Ideally, we should deprecate `alloy.py` and replace it with multiple smaller Python scripts. I’m not a fan of the current `alloy.py` approach, where everything is crammed into a single script—it's too large, making it difficult to understand and use effectively.
1. LLVM build using superbuild
2. ISPC build (like `quick-start-build.sh`)
3. Running tests (like your `run_tests.sh` here) + SDE setup
4. [optional] Running examples
5. [optional] Running benchmarks

I believe that it should be easy to translate `./quick-start-build.sh` into Python too using LLMs but if it requires too much effort we can go with sh/ps scripts at least for upcoming hackathon and then replace it.

nurmukhametov:
> I believe that it should be easy to translate `./quick-start-build.sh` into Python too using LLMs but if it requires too much effort we can go with sh/ps scripts at least for upcoming hackathon and then replace it.

I am not a big fan of using Python as a scripting language for such tasks, but it is here. A job run is here: https://github.com/nurmukhametov/ispc/actions/runs/13311888247

pbrubaker:
This is one of the things we discussed in Slack and I appreciate you adding this.

I am also not really a fan of using Python for this, and would have preferred bash/batch but I think it's reasonable to assume most people that will be building this will have Python installed.

scripts/quick-start-build.py
```diff
@@ -0,0 +1,511 @@
...
+
+    Examples:
+        # Build with default LLVM 18
+        $ python build.py
+
+        # Build with specific LLVM version
+        $ python build.py 17
+
+        # Build with custom LLVM location
+        $ LLVM_HOME=/path/to/llvm python build.py
```
pbrubaker, line 396: I would like to see an example Windows command line here.

Also `build.py` was deleted.  These should be `quick-start-build.py.`

scripts/quick-start-build.py
```diff
@@ -0,0 +1,511 @@
...
+
+    Build Configurations:
+        Windows:
+            - Build Type: RelWithDebInfo
+            - Binary Location: build-{version}/bin/RelWithDebInfo/ispc
+
+        Other Platforms:
+            - Build Type: Debug
```
pbrubaker, line 404: This is a little confusing, how can you specify the build type?

scripts/quick-start-build.py
```diff
@@ -0,0 +1,511 @@
...
+            - Binary Location: build-{version}/bin/ispc
+
+    Notes:
+        - Automatically determines optimal parallel build count
```
pbrubaker, line 408: Does it do so based on logical core count or physical core count?

scripts/quick-start-build.py
```diff
@@ -0,0 +1,511 @@
...
+            - If LLVM archive detection fails and ARCHIVE_URL is not set
+            - If CMake configuration fails
+        OSError: For filesystem operations or subprocess failures
+
+    Examples:
+        # Build with default LLVM 18
+        $ python build.py
+
+        # Build with specific LLVM version
+        $ python build.py 17
```
pbrubaker, line 393: How are point releases specified?

pbrubaker:
I think we should also update the [Development Guide](https://github.com/ispc/ispc/wiki/ISPC-Development-Guide) with modern build instructions along with this change.

pbrubaker (changes requested):
I've mentioned the changes inline.

scripts/quick-start-build.py
nurmukhametov: Point releases are not specified

scripts/quick-start-build.py
nurmukhametov: Logical cores

scripts/quick-start-build.py
nurmukhametov: No, you can't.

nurmukhametov:
> I've mentioned my proposed changes inline. I also think we should update the [Development Guide](https://github.com/ispc/ispc/wiki/ISPC-Development-Guide) to reflect the modern build instructions.

I will update after this PR is merged. The WIP version of the updated development guide is here (it has not been updated with changes about the Python script).

aneshlya (approved):

scripts/quick-start-build.py
pbrubaker: We might want to mention that but I don't think it's a big deal.

scripts/quick-start-build.py
pbrubaker: I think we may want to add this.  Possibly outside the scope of this commit unless it's quick to add.  If not maybe create a new issue for that?  I can see it being really useful.

scripts/quick-start-build.py
pbrubaker: I would note this in the comments.  This has bit me before with superbuild and LLVM 19.  This is slightly different behavior to that.

pbrubaker:
> > I've mentioned my proposed changes inline. I also think we should update the [Development Guide](https://github.com/ispc/ispc/wiki/ISPC-Development-Guide) to reflect the modern build instructions.
> 
> I will update after this PR is merged. The WIP version of the updated development guide is here (it has not been updated with changes about the Python script).

Would you please send me a link to the WIP version of the dev guide?

pbrubaker (approved):
LGTM aside from the comment changes and what you want to do about specifying build type.

nurmukhametov:
> > > I've mentioned my proposed changes inline. I also think we should update the [Development Guide](https://github.com/ispc/ispc/wiki/ISPC-Development-Guide) to reflect the modern build instructions.
> > 
> > 
> > I will update after this PR is merged. The WIP version of the updated development guide is here (it has not been updated with changes about the Python script).
> 
> Would you please send me a link to the WIP version of the dev guide?

I forgot to attach the link in my original message, here it is : https://github.com/nurmukhametov/ispc/wiki/ISPC-Development-Guide

PR #3228, opened 2025-02-14
Max/min/abs for short vectors

nurmukhametov:
This PR implements `max`, `min` and `abs` functions for short vector types. It addresses issue #2707.

This function is implemented in `stdlib.isph` as template function.

Initially, it was only possible to call it like this:

```ispc
max<int, 3>(a, b); // where a and b have uniform or varying int<3> type.
```

After #3243, proper types are deduced from the passed arguments in order to choose/instantiate the correct function, i.e., it is possible to call it like this:

```ispc
max(a, b);
```


aneshlya:
My first question is why only `max` function is chosen to be implemented? Plus it should be described in the documentation.
Also I'm not sure if it is the right moment to merge it. As we discussed today it will be much more useful if type deduction for short vectors is in place.

nurmukhametov:
> My first question is why only `max` function is chosen to be implemented? 

Why not? It was explicitly requested in #2707. All of them will be implemented more or less in the same manner. Therefore, I think it would be better to agree first on the principal way of implementing them, and then all other implementations can follow the same approach. I don't think it is worth implementing all functions only to start a review about the real matter.

> Also I'm not sure if it is the right moment to merge it. As we discussed today it will be much more useful if type deduction for short vectors is in place.

It would certainly be more convenient. However, the lack of template argument deduction for this particular case (#3229) is a separate and independent issue from the task of implementing standard library functions for short vectors.

dbabokin (commented):
I agree that we can move forward with a single library function implementation - this way it's easier to iron the rough edges before rolling out the implementation to larger list of functions. But would be good to build a list of functions that we are going to implement by the next release.

For the type deduction - we should also target this for the next release and consider it coupled with this change. There might be practical consequences of that - i.e. we must ensure that combined solution works as expected. I do expect some catches here.

For the current changes - we do miss doc update. Spelling out what exactly we implemented to the user is the first "sanity check". And my first question here is - do we intend to limit the set of type that we accept? Are we going to handle `bool`? Are we going to handle user types? How are we going to communicate about unsupported types?

aneshlya:
As we discussed, functional tests may not be necessary. We don’t need to verify that the `max` function itself works correctly—we already have tests for that. Instead, we need to ensure that the template implementation for short vectors resolves to the max function with the expected types. I believe Lit tests that check the AST should be sufficient for this.

nurmukhametov:
> As we discussed, functional tests may not be necessary. We don’t need to verify that the `max` function itself works correctly—we already have tests for that. Instead, we need to ensure that the template implementation for short vectors resolves to the max function with the expected types. I believe Lit tests that check the AST should be sufficient for this.

I have removed them and added a few more lit tests.

nurmukhametov:
> I agree that we can move forward with a single library function implementation - this way it's easier to iron the rough edges before rolling out the implementation to larger list of functions. But would be good to build a list of functions that we are going to implement by the next release.

I think we need to implement at least the following:
- min
- max
- abs

We may also add all element-wise functions, such as:
- round
- floor
- ceil
- trunc
- rcp
- rcp_fast
- sqrt
- rsqrt
- ldexp
- frexp
- sin
- asin
- cos
- acos
- sincos
- tan
- atan
- atan2
- exp
- log
- pow
- fmod
- clamp

Their implementation would probably be quite straightforward, with writing lit tests being the most time-consuming part.

> For the current changes - we do miss doc update. Spelling out what exactly what we implemented to the user is the first "sanity check". 

I have updated docs.

> And my first question here is - do we intend to limit the set of types that we accept? 

The template itself does not limit the set of types it accepts. The only limitation is the actual max functions defined earlier and implemented in the standard library.

> Are we going to handle `bool`? 

It depends on the function. There is no `max(bool, bool)` in the standard library, so we don't support it, and `max` for short vectors also doesn't support `bool` types.

> Are we going to handle user types? 

No. It could be possible if we allowed users to manually include `stdlib.isph` after their own definitions and implementations.

> How are we going to communicate about unsupported types?

The usual compiler error occurs as it does now when a user tries to call any standard library function for unsupported types. See `max-vec-4.ispc` as an example.

tests/lit-tests/max-vec.ispc
```diff
@@ -0,0 +1,51 @@
+// This test checks max functions for short vectors. They are now defined in standard library.
+
+// RUN: %{ispc} -O2 --vectorcall --target=avx512icl-x4 --emit-asm --x86-asm-syntax=intel %s -o - 2>&1 | FileCheck %s
```
aneshlya, line 3: Is `vectorcall` essential for this test?

docs/ispc.rst
```diff
@@ -4602,16 +4607,74 @@ standard types.  These functions also map to corresponding intrinsic functions.
...
+    uniform unsigned int8 max(uniform unsigned int8 a, uniform unsigned int8 b);
+    uniform int8 max(uniform int8 a, uniform int8 b);
+    uniform unsigned int16 max(uniform unsigned int16 a, uniform unsigned int16 b);
+    uniform int16 max(uniform int16 a, uniform int16 b);
+    uniform unsigned int max(uniform unsigned int a, uniform unsigned int b);
+    uniform int max(uniform int a, uniform int b);
+    uniform unsigned int64 max(uniform unsigned int64 a, uniform unsigned int64 b);
+    uniform int64 max(uniform int64 a, uniform int64 b);
+
+A maximum function also supports short vector types with basic types that are
```
aneshlya, line 4660: I would still add `min` function for short vectors as well because having just `max` is a little strange.

docs/ispc.rst
```diff
@@ -715,6 +715,11 @@ Updating ISPC Programs For Changes In ISPC 1.27.0
 Added cross lane operations for unsigned types: ``broadcast``, ``rotate``, ``shift``,
 and ``shuffle``.
 
+``max`` function for short vectors of basic types was added to the standard
```
aneshlya, line 718: I'd write here that it was added experimentally and will be extended with other functions which are used often with short vectors later.
Do you plan to add other functions for the next release?

tests/lit-tests/max-vec.ispc
nurmukhametov: Yes, otherwise on Windows it would be different asm.

docs/ispc.rst
nurmukhametov: Ok, I'll do it.

nurmukhametov: I have added `min`, and `abs` as well.

docs/ispc.rst
nurmukhametov: I think that adding more or less all element-wise functions needs to be done for the next release.

nurmukhametov:
The `min` function revealed an interesting miscompilation caused by the current implementation `./tests/func-tests/funcptr-uniform-4.ispc`,  here https://github.com/ispc/ispc/blob/81aab5cf9601ed95422b51f57e284d3eca3a78d7/tests/func-tests/funcptr-uniform-4.ispc#L15

```bash
Compilation of test ./tests/func-tests/funcptr-uniform-4.ispc failed            
/home/runner/work/ispc/ispc/tests/func-tests/funcptr-uniform-4.ispc:15:32: 
        Error: syntax error, unexpected ';', expecting '('. 
    uniform FuncType func = min;
                               ^
```

I created the issue #3253

docs/ispc.rst
```diff
@@ -715,6 +716,24 @@ Updating ISPC Programs For Changes In ISPC 1.27.0
 Added cross lane operations for unsigned types: ``broadcast``, ``rotate``, ``shift``,
 and ``shuffle``.
 
+The ``max``, ``min`` and ``abs`` functions for short vectors of basic types
+have been added to the standard library. They support both uniform and varying
+short vector types for all basic types supported by the corresponding standard
+functions, i.e., signed and unsigned integer types and floating-point types.
+
+It makes possible to find the maximum value of two short vector types with the
```
aneshlya, line 724: Seems like this sentence can be removed.

aneshlya:
> I created the issue https://github.com/ispc/ispc/issues/3253
Yeah, it was never supported. Do you want to fix it before merging this PR?

nurmukhametov:
> > I created the issue #3253

> Yeah, it was never supported. Do you want to fix it before merging this PR?

#3253 is fixed in #3293.

aneshlya:
Could you remove merge commit and rebase on top of trunk?

nurmukhametov:
> Could you remove merge commit and rebase on top of trunk?

Done.

PR #3233, opened 2025-02-17
Permit smearing for uniform Enum values

### LINKED ISSUE #2285: Enum field in struct causes crash
#### Steps to reproduce 
Compile the following code (enum.ispc):

```
enum material_type {
  DIFFUSE = 0,
};

struct sphere {
  material_type type;
};

export uniform float render(uniform sphere spheres[]) {
  sphere current_sphere = spheres[0]; // this the offensive line. If I add "uniform" to the type of current_sphere, compilation succeeds.
  return 0.0f;
}
```

with

```
ispc enum.ispc --target=avx2
```

#### Current behaviour
The compilation fails with the error:

```
/usr/local/src/ispc/src/expr.cpp:6536: Assertion failed: "CastType<PointerType>(type) != NULL".
***
***
/usr/local/src/ispc/src/main.cpp(396): FATAL ERROR: Unhandled signal sent to process; terminating.
Aborted (core dumped)
```

#### Expected behaviour
Compilation succeeds.

#### Other comment
In the above example it may seem that I could make `spheres` uniform but in my actual code, I get different indices depending on the gang's previous computation.

FWIW, I'm working around the issue by using `int` instead of an enum, i.e.

```
#define  DIFFUSE 0

struct sphere {
  int type;
};
```

#### Environment
Ubuntu 20.04 with ispc 1.17.0:

```
$ ispc --version
Intel(r) Implicit SPMD Program Compiler (Intel(r) ISPC), 1.17.0 (build commit 7ad8429369a4d5ce @ 20220128, LLVM 12.0.1)
```

nurmukhametov:
This PR fixes #2285

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -7309,7 +7309,7 @@ static llvm::Value *lUniformValueToVarying(FunctionEmitContext *ctx, llvm::Value
                                               CastType<AtomicType>(type), pos);
     }
 
-    Assert(CastType<PointerType>(type) != nullptr);
+    Assert(CastType<PointerType>(type) != nullptr || CastType<EnumType>(type) != nullptr);
     return ctx->SmearUniform(value);
 }
 
--- a/tests/lit-tests/2285.ispc
+++ b/tests/lit-tests/2285.ispc
@@ -1,0 +1,22 @@
+// RUN: %{ispc} -O2 %s -o - --target=host --emit-llvm-text --nowrap | FileCheck %s
+
+enum material_type {
+  DIFFUSE = 0,
+};
+
+struct sphere {
+  material_type type;
+};
+
+// CHECK-NOT: FATAL ERROR:
+export uniform float render(uniform sphere spheres[]) {
+    // this the offensive line. If I add "uniform" to the type of current_sphere, compilation succeeds.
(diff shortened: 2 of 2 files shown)


aneshlya:
Maybe we should add processing of Enum type to `lTypeConvAtomicOrUniformVector`? it may benefit cases for short vectors of enum types. But I'm OK with this fix too.

aneshlya:
@nurmukhametov, any thoughts on this? 

> Maybe we should add processing of Enum type to lTypeConvAtomicOrUniformVector? it may benefit cases for short vectors of enum types.

Or do you want to proceed with the current fix for now?

nurmukhametov:
> @nurmukhametov, any thoughts on this?
> 
> > Maybe we should add processing of Enum type to lTypeConvAtomicOrUniformVector? it may benefit cases for short vectors of enum types.
> 

I have no thoughts, except that this function is already quite large, and it is not entirely clear to me whether enums are systematically treated as atomic in the ISPC codebase.

> Or do you want to proceed with the current fix for now?

I would prefer to proceed with the current fix.

aneshlya (approved):
Looks good to me since short vectors of enum types are not used often.

PR #3235, opened 2025-02-17
Skip applying the patch that is already in 19.1.7

nurmukhametov:
This patch was not present in version 19.1.1.

Job run https://github.com/ispc/ispc/actions/runs/13367866339/job/37329619775


aneshlya (approved):
LGTM!

Issue #2349, opened 2022-07-18
Develop a comprehensive API reference for the standard library.


pbrubaker:
Currently there is no comprehensive reference for the standard library.  Many of the functions are listed in the user guide, but it would be nice to have an API level documentation (doxygen or similar?) of the included functions.

Also see: #2276

aneshlya:
Hey @pbrubaker ,
I'm trying to understand is the separate API reference really needed. We're committed to keeping our User Guide up-to-date, so if any information is missing, it's considered a bug. Our goal is to consolidate all the relevant details in one place.

pbrubaker:
Hey @aneshlya.  I created this issue in response to `max-cura's` comments on [this other issue.](https://github.com/ispc/ispc/issues/2276#issue-1163405257) 

Something as simple a list of all the declared functions would be a big help.  I think putting that in the User Guide would be exceptional.

pbrubaker:
Hey @aneshlya.  I created this issue in response to `max-cura's` comments on [this other issue.](https://github.com/ispc/ispc/issues/2276#issue-1163405257) 

I think this could be something as simple as a list of all the declared functions.  That would have helped in that case.  I don't think we need full Doxygen documentation.  Putting that list in the User Guide would be exceptional.

An argument could be made that people should just look at the code, but I think putting it on the website makes it more accessible.

aneshlya:
I think one of the best places to see the list of declared stdlib functions is https://github.com/ispc/ispc/blob/main/stdlib/include/stdlib.isph.

aneshlya:
> I echo that. I think we explicitly point out to the `stdlib.isph` file in the overview of the stdlib in the documentation.

We don't but it's a good idea: https://github.com/ispc/ispc/pull/3239

pbrubaker:
Yeah, that would be great!

One of the tasks for the editor plugins is that we need to automatically parse that with a regex and generate a text file containing a list of function prototypes.  Those are loaded and used in a dictionary for auto complete.  If you did want to automate this for the generation of a list for the website it will soon be possible to share or copy that code.

PR #3239, opened 2025-02-18
Minor docs and CI update 

aneshlya:
1. Mention `stdlib.isph` as a reference for standard library functions, fixes https://github.com/ispc/ispc/issues/2349
2. Replace appveyor with trunk-releases in CI workflows.


dbabokin (approved):
LGTM

Issue #3229, opened 2025-02-14
Improve implicit template argument deduction


nurmukhametov:
At the moment, it is not possible to call the following template function `max` without explicitly specifying the template arguments.

```ispc
template <typename T, uint N> uniform T<N> max(uniform T<N> a, uniform T<N> b) {
    uniform T<N> result;
    foreach (i = 0 ... N) {
        result[i] = max(a[i], b[i]);
    }
    return result;
}

uniform int<4> foo(uniform int<4> a, uniform int<4> b) {
    return max<int, 4>(a, b);
}

uniform int<4> boo(uniform int<4> a, uniform int<4> b) {
    return max(a, b);
// <source>:14:12: Error: Unable to find any matching overload for call to function "max". 
//         Passed types: (uniform int32<4>, uniform int32<4>)     
}
```

https://ispc.godbolt.org/z/bMe6PoKr1

PR #3243, opened 2025-02-20
Support argument deduction for short vector in templates

nurmukhametov:
This PR implements #3229

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -9011,6 +9011,43 @@ static std::pair<std::string, const Type *> lDeduceParam(const Type *paramType, 
     return std::pair<std::string, const Type *>("", nullptr);
 }
 
+bool lCheckDeductionCompatibility(TemplateInstantiation &inst, const std::pair<std::string, const Type *> &deduction,
+                                  SourcePos pos) {
+    if (deduction.second == nullptr) {
+        // Deduction failed
+        return false;
+    }
+
+    // Check if this deduction is compatible with previously deduced arguments
+    const Type *previousDeductionResult = inst.InstantiateType(deduction.first);
+
+    if (previousDeductionResult == nullptr) {
+        // This template parameter was deducted for the first time. Add it to the map.
+        inst.AddArgument(deduction.first, TemplateArg(deduction.second, pos));
+        return true;
+    }
+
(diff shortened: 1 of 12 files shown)


src/expr.cpp
```diff
@@ -9011,14 +9011,77 @@ static std::pair<std::string, const Type *> lDeduceParam(const Type *paramType,
...
+        // First, deduce the element type T from the argument type.
+        const Type *argElemType = argType->GetElementType();
+        auto deduction = lDeduceParam(elemType, argElemType);
+        if (!lCheckDeductionCompatibility(inst, deduction, pos)) {
+            // Deduction failed, skip to the next candidate.
+            return false;
+        }
+
+        // Then deduce the count N from the argument type.
+        // TODO!: Can argument contain not a constant count?
```
aneshlya, line 9067: No. As far as I remember now we support integer constants and enums only. I don't expect that we will support anything here that is is not const expression known in compile time.

tests/lit-tests/3229.ispc
```diff
@@ -0,0 +1,23 @@
+// Check argument type deduction for template functions with short vector types
```
aneshlya, line 1: I also suggest to check short vectors of enums and bools and deduction in nested templates.

src/expr.cpp
```diff
@@ -9011,14 +9011,77 @@ static std::pair<std::string, const Type *> lDeduceParam(const Type *paramType,
...
+        int elemCount = argType->GetElementCount();
+        ConstExpr *expr = new ConstExpr(AtomicType::UniformInt32, elemCount, pos);
+        // TODO!: do we need to check here that this argument is not already deduced?
```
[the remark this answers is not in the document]
nurmukhametov: I have added `bool lCheckDeductionCompatibility(TemplateInstantiation &inst, const std::pair<std::string, const Expr *> &deduction...`

tests/lit-tests/3229.ispc
nurmukhametov: I have added tests for bool and nested templates, but it looks like short vectors with enums as the type or size are not supported.

aneshlya: Ah, ok. It should be easy to add it but it is out of scope of this PR.

PR #3245, opened 2025-02-24
Fix address calculations involving unsigned indexes

### LINKED ISSUE #3242: Address calculations with int vs uint
Consider this:

https://godbolt.org/z/sebW8W5xP

`
varying int fetch_with_uint(uniform int* data, varying uint offset_uint) {
    return data[offset_uint];
}

varying int fetch_with_sint(uniform int* data, varying int offset_sint) {
    return data[offset_sint];
}
`

The resulting assembly is in both cases the same - in both cases the result of the address calculation seems to be that the offset overflows at 2^31.

This is quite different to regular C where the uint address calculation does not overflow until 2^31.

This is on purpose or can this be fixed?

Regards,
Lars

nurmukhametov:
This PR fixes #3242

### DIFF
--- a/docs/ispc.rst
+++ b/docs/ispc.rst
@@ -736,6 +736,11 @@ vectors:
 Attribute ``aligned(N)`` was added to specify alignment of variables and struct
 types.
 
+Accessing an array using an unsigned index and pointer arithmetic with unsigned
+offsets was flawed due to overflow caused by sign extension when promoting the
+index to pointer size. This issue has been resolved, and the compiler now
+generates correct code for such cases involving unsigned integer types.
+
 Getting Started with ISPC
 =========================
 


src/ctx.cpp
```diff
@@ -1979,13 +1979,26 @@ llvm::Instruction *FunctionEmitContext::ZExtInst(llvm::Value *value, llvm::Type
     return inst;
 }
 
+/** Extend both uniform and varying integer values to 64 bits. */
+llvm::Value *lExtendIntTo64(FunctionEmitContext *ctx, llvm::Value *value, const Type *indexType,
+                            const llvm::Twine &name = "") {
+    bool valueIsVarying = llvm::isa<llvm::VectorType>(value->getType());
```
aneshlya, line 1985: We have this function `IsVaryingAtomicOrUniformVectorType`. Is it what you want to check here?

aneshlya:
This change should be described in documentation.

nurmukhametov:
> This change should be described in documentation.

Done

aneshlya (approved):
Looks good to me!

dbabokin (commented):
One big thing that needs to be explored and most likely changed here.

We do multiply (or shift) first and only then the extension.  Comment in the code points out that it's more efficient - probably because it was causing better code gen with indexing being part of the addressing in the instruction encoding. But there's a problem with that. For int (i.e. 4 byte size element type), it will destroy 2 upper bits from index doing 32 bit shift left. So to enable full 32 bit index, we need to extend (sign/zero) first and then do multiply/shift.

I my understanding is right, it's a bigger problem than just wrong sign extension here.

This needs investigation, but it does look wrong.

nurmukhametov:
> We do multiply (or shift) first and only then the extension. Comment in the code points out that it's more efficient - probably because it was causing better code gen with indexing being part of the addressing in the instruction encoding. But there's a problem with that. For int (i.e. 4 byte size element type), it will destroy 2 upper bits from index doing 32 bit shift left. So to enable full 32 bit index, we need to extend (sign/zero) first and then do multiply/shift.

We do shift/multiplication but it is a left shift (<< 2), so it destroys the lower bits (the least significant) rather than the upper ones.

dbabokin:
> We do shift/multiplication but it is a left shift (<< 2), so it destroys the lower bits (the least significant) rather than the upper ones.

It destroys upper 2 bits of the index. As it's a 32 bit operation, the upper two bits are not shifted to 64 bit value, they are destroyed.

nurmukhametov:
> > We do shift/multiplication but it is a left shift (<< 2), so it destroys the lower bits (the least significant) rather than the upper ones.
> 
> It destroys upper 2 bits of the index. As it's a 32 bit operation, the upper two bits are not shifted to 64 bit value, they are destroyed.

Argh, indeed. I see now that integer overflow may occur. However, it seems like a separate issue from #3242, and fixing it will probably affect performance (probably noticeably).

dbabokin:
> Argh, indeed. I see now that integer overflow may occur. However, it seems like a separate issue from #3242, and fixing it will probably affect performance (probably noticeably).

It worth experimenting with it. It may or may not lead to worse perf. I think it should not be a big deal with modern ISA. And fixing sign/zero extension without fixing the order will not help much. I suggest to at least evaluate a complete fix here. Implementing it is not much work, I think.

nurmukhametov:
> > Argh, indeed. I see now that integer overflow may occur. However, it seems like a separate issue from #3242, and fixing it will probably affect performance (probably noticeably).
> 
> It worth experimenting with it. It may or may not lead to worse perf. I think it should not be a big deal with modern ISA. And fixing sign/zero extension without fixing the order will not help much. I suggest to at least evaluate a complete fix here. Implementing it is not much work, I think.

After reviewing this code and consulting the documentation, I conclude that this is not a bug but intentional.

Under the default addressing mode (32), the function `FunctionEmitContext::applyVaryingGep` first truncates the index to a 32-bit vector type to perform multiplication in 32 bits, and then extends the result for pointer arithmetic, which is clearly intentional. Moreover, the documentation states that the default addressing mode (32) doesn't support addressing more than 4GB of memory: https://ispc.github.io/ispc.html#selecting-32-or-64-bit-addressing. The performance guide also includes a section on the performance impact of using 64-bit addressing mode: https://ispc.github.io/perfguide.html#avoid-64-bit-addressing-calculations-when-possible.

PR #3247, opened 2025-02-25
Fix quick-start-build.py

nurmukhametov:
This PR fixes the `quick-start-build.py` script by correcting the `asset_name` for Windows and adding a help message. It also pins action versions.


aneshlya (approved):
Looks good!

Issue #3249, opened 2025-02-25
Fix failing release pipelines


aneshlya:
Scan release binaries with cve-bin-tool https://github.com/ispc/ispc/actions/runs/13531141441 and hardening check are failing with release archives.

PR #3250, opened 2025-02-25
Add LLVM 19 to full ISPC testing + minor CI fixes in release workflows

### LINKED ISSUE #3248: Add LLVM 19 into ispc-ci testing
Add LLVM 19 into ispc-ci testing and consider switching to LLVM 19 for the next release.

aneshlya:
Tests run: https://github.com/ispc/ispc/actions/runs/13531551449
Fixes #3248 , #3249

### DIFF
--- a/.github/workflows/ispc-ci.yml
+++ b/.github/workflows/ispc-ci.yml
@@ -164,6 +164,9 @@ jobs:
           - version: "18.1"
             full_version: "18.1.8"
             short_version: 18
+          - version: "19.1"
+            full_version: "19.1.7"
+            short_version: 19
     env:
       LLVM_VERSION: ${{ matrix.llvm.version }}
       LLVM_TAR: llvm-${{ matrix.llvm.full_version }}-ubuntu22.04-Release+Asserts-x86.arm.wasm.tar.xz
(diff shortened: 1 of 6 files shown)


dbabokin (approved):
LGTM

nurmukhametov (approved):

aneshlya:
We hit the limit of strategy expansion so I'm splitting `win-test` and `lin-test` jobs to two: one for testing with LLVM version used for release and newer versions and another for testing with older LLVM versions.

.github/workflows/ispc-ci.yml
```diff
@@ -320,14 +323,105 @@ jobs:
...
+        path: tests/fail_db.txt
+
+    - name: Upload alloy logs
+      uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
+      if: failure()
+      with:
+        name: alloy_results.llvm${{matrix.version}}.${{matrix.arch}}.${{matrix.target}}
+        path: alloy_results_*
+
+  linux-test-prior:
```
nurmukhametov, line 417: I don't really like this approach. This workflow file is already huge, and this only makes it bigger. What about making it reusable with the LLVM version as a parameter and creating separate workflows for each LLVM version that call this one? Basically, the same approach we use for rebuild-llvm jobs.

nurmukhametov (changes requested):

.github/workflows/ispc-ci.yml
aneshlya: I created reusable workflow for test jobs. I'd like to do the same for build but I'll do it in another PR for easier review.
The full run is here: https://github.com/ispc/ispc/actions/runs/13713528891

nurmukhametov: It is really nice. It is much more convenient to navigate the job list now.

nurmukhametov (approved):
LGTM!

Issue #3095, opened 2024-11-07
Support int8/int16 for packed_store_active/packed_load_active 


ColonelPhantom:
Needing to generate variable output can be a common pattern (see e.g. [the FAQ](https://ispc.github.io/faq.html#how-can-a-gang-of-program-instances-generate-variable-amounts-of-output-efficiently)).

I would think that instructions like VPCOMPRESS can make this easier and more efficient, since results can go directly where they should and there is no expensive 'post-processing' step involving expensive scans and memory copies.

For reference, the FAQ code (modified to actually put things in `outLocal`) is:
```c
uniform int func(uniform float outArray[], ...) {
   int numOut = ...;  // figure out how many to be output
   float outLocal[MAX_OUT]; // staging area

   // each program instance in the gang puts its results in
   //  outLocal[0], ..., outLocal[numOut-1]
   for (...) {
      if (condition) {
         outLocal[numOut++] = value;
      }
   }

   int startOffset = exclusive_scan_add(numOut);
   for (int i = 0; i < numOut; ++i)
       outArray[startOffset + i] = outLocal[i];
   return reduce_add(numOut);
}
```

With a `compress_store` intrinsic, one could instead write something like:

```c
uniform int func(uniform float outArray[], ...) {
    uniform int numOut = 0;
    for (...) {
        if (condition) {
            numOut += compress_store(&outArray[numOut], value);
        }
    }
    return numOut;
}
```

assuming that `compress_store` returns the number of written elements, and I didn't get any pointer semantics wrong :)

This also changes the order in which elements are written, but I would consider the compress approach to have a more 'logical' result, since something like `foreach (i = 0 ... n) { compress_store(...); }` would also put `i` into the array in order.

I feel like in-register compress would be kind of weird in ISPC, since it would also need to modify the mask and as such mess up the 'scalar illusion', but compress-store can be perfectly explained in ISPC semantics (with `foreach_active`) and does not affect the mask in any way.

ColonelPhantom:
Actually, it seems like it's been there since the beginning, under the name `packed_store_active` (and `packed_load_active`). It also uses the VPCOMPRESS intrinsics.

However, it does seem these are limited to int32 and int64. I think it would be nice if all datatypes were supported.

dbabokin:
Yep, we have these library function, check out https://ispc.github.io/ispc.html#packed-load-and-store-operations for more details. Note that we have also `packed_store_active2()` version, that is a faster version, see the documentation for details.

[...]

PR #3252, opened 2025-02-26
Add missing types for packed_load/packed_store

aneshlya:
This PR adds support for signed/unsigned int8, int16, float16, float and double into `packed_load`/`packed_store`.
Fixes https://github.com/ispc/ispc/issues/3095.

### DIFF
--- a/builtins/generic.ispc
+++ b/builtins/generic.ispc
@@ -1847,11 +1847,39 @@ EXT void __scatter_factored_base_offsets64_double(uniform int8 *uniform base, va
 }
 
 // packed active stores
+EXT PackedStoreResultType __packed_store_activei8(uniform int8 *uniform ptr, varying int8 val, UIntMaskType mask) {
+    @llvm.masked.compressstore(val, ptr, mask);
+    return __popcnt_int64(__cast_mask_to_i64(mask));
+}
+EXT PackedStoreResultType __packed_store_activei16(uniform int8 *uniform ptr, varying int16 val, UIntMaskType mask) {
+    @llvm.masked.compressstore(val, ptr, mask);
+    return __popcnt_int64(__cast_mask_to_i64(mask));
+}
 EXT PackedStoreResultType __packed_store_activei32(uniform int8 *uniform ptr, varying int32 val, UIntMaskType mask) {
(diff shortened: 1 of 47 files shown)


builtins/generic.ispc
```diff
@@ -1863,8 +1891,33 @@ EXT PackedStoreResultType __packed_store_active2i64(uniform int8 *uniform ptr, v
...
+EXT PackedStoreResultType __packed_store_active2f32(uniform int8 *uniform ptr, varying float val, UIntMaskType mask) {
+    @llvm.masked.compressstore(val, ptr, mask);
+    return __popcnt_int64(__cast_mask_to_i64(mask));
+}
+EXT PackedStoreResultType __packed_store_active2f64(uniform int8 *uniform ptr, varying double val, UIntMaskType mask) {
+    @llvm.masked.compressstore(val, ptr, mask);
+    return __popcnt_int64(__cast_mask_to_i64(mask));
+}
 // packed active loads
+EXT uniform int32 __packed_load_activei8(uniform int8 *uniform from, uniform int8 *uniform to, UIntMaskType mask) {
```
nurmukhametov, line 1907: With so many implementations, it would be better to define a template function `__packed_load_active` and call it in all `__packed_load_activeTYPE` functions. Unfortunately, we can't call intrinsics directly from template functions, so we need to create overloaded functions for `@llvm.masked.expandload`. See the change here: https://github.com/nurmukhametov/ispc/commit/43e4e2398e43b7321de5ecb84a4b76ea6dd74155. The same can be done with `__packed_store_active`.

nurmukhametov (changes requested):

builtins/generic.ispc
aneshlya: I agree that for `packed_load` it's beneficial - it removes code duplication. However for 2-lines code implementation for `packed_store` it seems as overkill. I tried it and `packed_store` implementation has optimized out https://github.com/aneshlya/ispc/commit/3b229495e5ec3b204a6983b789263f5d7225475f.

nurmukhametov: Yes, for short functions, this might not make sense and could result in more code being written.

nurmukhametov (approved):

Issue #478, opened 2013-04-18
Compiler error for extern "C" struct


Martinsos:
simple.cpp:
    #include "simple_ispc.h"
    extern "C" struct Test {
        int b;
    };
    int main() {
        ispc::simple();
    }

simple.ispc:
    extern "C" struct Test;
    export void simple() {
        Test t;
    }

compile command:
ispc simple.ispc -o simple_ispc.o -h simple_ispc.h

Error message:
func.cpp(423): FATAL ERROR: Function verificication failed

---

**\* Please file a bug report at https://github.com/ispc/ispc/issues
**\* (Including as much information as you can about how to reproduce this error).
**\* You have apparently encountered a bug in the compiler that we'd like to fix!

---

main.cpp(204): FATAL ERROR: Unhandled signal sent to process; terminating.

ispc version: 1.3.0
machine: Intel(R) Core(TM)2 Quad CPU    Q6600  @ 2.40GHz

aneshlya:
Here is a repro: https://godbolt.org/z/991h19EEd

aneshlya:
@nurmukhametov , can it be closed?

nurmukhametov:
Yes, sure. It is fixed by https://github.com/ispc/ispc/pull/3255

PR #3255, opened 2025-02-28
Report compilation errors about incomplete types

### LINKED ISSUE #2844: Incomplete struct types
The following examples are crashed:

```ispc
struct S1 { struct S2 a; };
struct S1 func() {
    struct S1 x;
    return x;
}
```

```ispc
struct Sx bar() {}
```

```ispc
int boo(struct Sy) {}
```

```ispc
int foo() { struct Sx x; }
```

I guess the reason for all of them is same. By some reason, ISPC doesn't check the usage of incomplete structs types on AST level. So, LLVM IR is incorrect:
```llvm-ir
Cannot allocate unsized type
  %7 = alloca %v4_varying_Sx, align 8
```

This example is compiled but shouldn't. It has to generate the compiler error "error: field 'a' has incomplete type 'S2'"
```ispc
struct S1 { struct S2 a; };
```

nurmukhametov:
This fixes #2844 and #478.

This change adds error reports for incomplete types in various situations:
- return values of functions
- parameters of functions
- variable declaration and initialization

### DIFF
--- a/src/func.cpp
+++ b/src/func.cpp
@@ -206,6 +206,21 @@ Function::Function(Symbol *s, Stmt *c, Symbol *ms, std::vector<Symbol *> &a)
 }
 
 void Function::typeCheckAndOptimize() {
+    const FunctionType *funcType = GetType();
+    const Type *retType = GetType()->GetReturnType();
+    if (!retType->IsCompleteType()) {
+        Error(funcType->GetSourcePos(), "return type is an incomplete type: %s", retType->GetString().c_str());
+    }
+
+    for (int i = 0; i < funcType->GetNumParameters(); i++) {
+        const Type *paramType = funcType->GetParameterType(i);
+        if (!paramType->IsCompleteType()) {
+            const SourcePos &pos = funcType->GetParameterSourcePos(i);
+            const std::string &paramName = funcType->GetParameterName(i);
+            Error(pos, "parameter '%s' is an incomplete type: %s", paramName.c_str(), paramType->GetString().c_str());
+        }
+    }
+
     if (code != nullptr) {
         debugPrintHelper(DebugPrintPoint::Initial);
 
--- a/src/type.cpp
+++ b/src/type.cpp
@@ -266,6 +266,8 @@ bool AtomicType::IsBoolType() const { return basicType == TYPE_BOOL || basicType
 
 bool AtomicType::IsConstType() const { return isConst; }
 
+bool AtomicType::IsCompleteType() const { return true; }
+
 const AtomicType *AtomicType::GetAsUnsignedType() const {
     if (IsUnsignedType() == true) {
         return this;
@@ -742,6 +744,8 @@ bool TemplateTypeParmType::IsSignedType() const { return false; }
 
 bool TemplateTypeParmType::IsConstType() const { return isConst; }
 
+bool TemplateTypeParmType::IsCompleteType() const { return false; }
+
 const Type *TemplateTypeParmType::GetBaseType() const { return this; }
 
 const Type *TemplateTypeParmType::GetAsVaryingType() const {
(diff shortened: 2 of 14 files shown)


src/type.cpp
```diff
@@ -1503,6 +1511,8 @@ bool ArrayType::IsBoolType() const { return false; }
 
 bool ArrayType::IsConstType() const { return child ? child->IsConstType() : false; }
 
+bool ArrayType::IsCompleteType() const { return GetBaseType()->IsCompleteType(); }
```
aneshlya, line 1514: We have an unsized arrays which I think are also incomplete types.

tests/lit-tests/2844-1.ispc
```diff
@@ -0,0 +1,14 @@
...
+// RUN: not %{ispc} --target=host --nowrap --nostdlib %s -o %t.o 2>&1 | FileCheck %s
+
+// CHECK: Error: return type is an incomplete type: varying struct Sx
+struct Sx bar() {}
+
+// CHECK-NOT: Error: return type is an incomplete type: varying struct Sy
+struct Sy foo();
+
+// CHECK-NOT: Error: return type is an incomplete type: varying struct Sz
+struct Sz func();
```
aneshlya, line 12: What is the difference between these two?

tests/lit-tests/2844-3.ispc
```diff
@@ -0,0 +1,17 @@
+// Check that we catch incomplete struct types in variable definitions
+
+// RUN: not %{ispc} --target=host --nowrap --nostdlib %s -o %t.o 2>&1 | FileCheck %s
+
+// CHECK: Error: type struct 'Sx' is incomplete and cannot be initialized
+// CHECK: Error: variable 'x' has initializer but incomplete struct type
+int foo() { struct Sx x = { 0 }; }
+// CHECK: Error: type struct 'Sy' is incomplete and cannot be declared
+// CHECK: Error: variable 'y' has incomplete struct type and cannot be defined
```
aneshlya, line 9: Can be here a case when one error is reported and the other is not? Feels like we only need on error message instead of two. It would be the same behavior as in C++.

src/func.cpp
```diff
@@ -206,6 +206,21 @@ Function::Function(Symbol *s, Stmt *c, Symbol *ms, std::vector<Symbol *> &a)
...
+    const Type *retType = GetType()->GetReturnType();
+    const FunctionType *funcType = GetType();
+    if (!retType->IsCompleteType()) {
+        Error(funcType->GetSourcePos(), "return type is an incomplete type: %s", retType->GetString().c_str());
+    }
+
+    for (int i = 0; i < funcType->GetNumParameters(); i++) {
+        const Type *paramType = funcType->GetParameterType(i);
+        const SourcePos &pos = funcType->GetParameterSourcePos(i);
+        const std::string &paramName = funcType->GetParameterName(i);
```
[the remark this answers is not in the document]
nurmukhametov: Done

src/type.cpp
nurmukhametov: I think that, generally, unsized arrays should not be considered incomplete types unless they are the only members of a struct.

tests/lit-tests/2844-1.ispc
nurmukhametov: No idea, I have removed the second copy.

tests/lit-tests/2844-3.ispc
nurmukhametov: I have removed the first report.

aneshlya (approved):
Looks good to me!

src/type.cpp
```diff
@@ -742,6 +744,8 @@ bool TemplateTypeParmType::IsSignedType() const { return false; }
 
 bool TemplateTypeParmType::IsConstType() const { return isConst; }
 
+bool TemplateTypeParmType::IsCompleteType() const { return false; }
```
dbabokin, line 747: I don't really understand how this works with templates / template paramters.

Would be good to have a test that checks that the same errors are properly reported when the same happens for function instantiation - i.e. wrt function parameters and return types.

nurmukhametov: I have added a test to check template instantiation with an incomplete struct type.

dbabokin: The tests look good, thank you!

dbabokin (approved):
LGTM

PR #3256, opened 2025-03-01
Enable all x86 targets in targetParentMap

aneshlya:
This PR enables hierarchical processing for all our targets. 
AVX10 targets are planned to be based on this hierarchy so it's a prerequisite for AVX10 support.
To check that hierarchical processing works I'm removing extract/insert and bitcast functions from util.m4 so they should be used from generic.target.


src/builtins.cpp
```diff
@@ -419,62 +419,81 @@ bool lStartsWithLLVM(llvm::StringRef name) { return name.starts_with("llvm."); }
...
+    {ISPCTarget::xehpg_x8, ISPCTarget::generic_i1x8},
+    {ISPCTarget::xehpg_x16, ISPCTarget::generic_i1x16},
+    {ISPCTarget::xehpc_x16, ISPCTarget::generic_i1x16},
+    {ISPCTarget::xehpc_x32, ISPCTarget::generic_i1x32},
+    {ISPCTarget::xelpg_x8, ISPCTarget::generic_i1x8},
+    {ISPCTarget::xelpg_x16, ISPCTarget::generic_i1x16},
+    {ISPCTarget::xe2hpg_x16, ISPCTarget::generic_i1x16},
+    {ISPCTarget::xe2hpg_x32, ISPCTarget::generic_i1x32},
+    {ISPCTarget::xe2lpg_x16, ISPCTarget::generic_i1x16},
+    {ISPCTarget::xe2lpg_x32, ISPCTarget::generic_i1x32}
```
nurmukhametov, line 496: If we want to use generic targets for `xe`, then we need to support building them for `xe` arch in this file: https://github.com/ispc/ispc/blob/main/cmake/GenericTargets.cmake.

aneshlya: I decided to limit the change for x86 targets only since on other targets I see many issues that require careful investigation. Right now I don't want to do this, and prefer to enable x86 targets only for further avx10 support.

aneshlya: Full testing is here: https://github.com/aneshlya/ispc/actions/runs/13709157474

aneshlya:
I've split fmod-[float|double]-varying tests into two to reduce stack usage with `-O0`. They should pass now with generic targets as well.

PR #3266, opened 2025-03-04
Fix generic-i1x64 target

### LINKED ISSUE #3123: Long short vector
Consider the following short vector declaration:

```ispc
uniform int<127> data;
```

that fails to compile for `avx512skx-x64` with the error:

```bash
<source>:1:13: Error: syntax error, unexpected unsigned int8  constant. 
uniform int<128> data; 
Compiler returned: 1
```

Compiler explorer link https://ispc.godbolt.org/z/dbv6j3eaq

aneshlya:
This PR fixes generic_i1x64 target so it can be used in targets' hierarchy.
Fixes #3123

### DIFF
--- a/builtins/generic.ispc
+++ b/builtins/generic.ispc
@@ -1381,7 +1381,7 @@ EXT uniform uint64 __cast_mask_to_i64(UIntMaskType mask) { return @llvm.ispc.pac
 // mask operations
 EXT uniform int64 __movmsk(UIntMaskType mask) { return __cast_mask_to_i64(mask); }
 EXT uniform bool __any(UIntMaskType mask) { return __cast_mask_to_i64(mask) != 0; }
-EXT uniform bool __all(UIntMaskType mask) { return __cast_mask_to_i64(mask) == (1ULL << TARGET_WIDTH) - 1; }
+EXT uniform bool __all(UIntMaskType mask) { return __cast_mask_to_i64(mask) == ~0ULL >> (64 - TARGET_WIDTH); }
 EXT uniform bool __none(UIntMaskType mask) { return __cast_mask_to_i64(mask) == 0; }
 
 // general gather functions
--- a/cmake/GenericTargets.cmake
+++ b/cmake/GenericTargets.cmake
@@ -40,12 +40,12 @@ function (generate_generic_builtins ispc_name)
         list(APPEND os_list "unix")
     endif()
 
-    # "generic-i1x64" is not supported due to #3123
     list(APPEND TARGET_LIST
         "i1x4"
         "i1x8"
         "i1x16"
         "i1x32"
+        "i1x64"
         "i8x16"
         "i8x32"
         "i16x8"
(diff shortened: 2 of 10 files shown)


aneshlya:
Full run is here: https://github.com/aneshlya/ispc/actions/runs/13664154208

builtins/generic.ispc
```diff
@@ -1381,7 +1381,15 @@ EXT uniform uint64 __cast_mask_to_i64(UIntMaskType mask) { return @llvm.ispc.pac
...
 EXT uniform int64 __movmsk(UIntMaskType mask) { return __cast_mask_to_i64(mask); }
 EXT uniform bool __any(UIntMaskType mask) { return __cast_mask_to_i64(mask) != 0; }
-EXT uniform bool __all(UIntMaskType mask) { return __cast_mask_to_i64(mask) == (1ULL << TARGET_WIDTH) - 1; }
+EXT uniform bool __all(UIntMaskType mask) {
+#if TARGET_WIDTH == 64
+    //  Shifting a 64-bit value left by 64 positions is undefined behavior
+    //  so check mask explicitely
+    return __cast_mask_to_i64(mask) == 0xFFFFFFFFFFFFFFFF;
+#else
+    return __cast_mask_to_i64(mask) == (1ULL << TARGET_WIDTH) - 1;
```
dbabokin, line 1390: What about this? 😎 
```cpp
~(~0ULL << N);
```

aneshlya: `~(~0ULL << TARGET_WIDTH);` causes warning `Binary expression with type "const uniform unsigned int64" can't represent value.`
But `~0ULL >> (64 - TARGET_WIDTH);` works without warning.

dbabokin: Heh, right, warnings is correct, but we consciously doing what we are doing.

`~0ULL >> (64 - TARGET_WIDTH);` looks good.

cmake/GenericTargets.cmake
```diff
@@ -46,6 +46,7 @@ function (generate_generic_builtins ispc_name)
         "i1x8"
         "i1x16"
         "i1x32"
+        "i1x64"
```
nurmukhametov, line 49: Please, remove the comment above reffering to #3123

tests/lit-tests/3123.ispc
```diff
@@ -0,0 +1,6 @@
+// RUN: %{ispc} --target=generic-i1x64 --nostdlib %s --nowrap -o %t.o 2>&1 | FileCheck %s --allow-empty
+// CHECK-NOT: Error: syntax error
+
+// REQUIRES: X86_ENABLED && !MACOS_HOST
+
+uniform int<127> data;
```
nurmukhametov, line 6: `128`, `127` fits in singed int8.

nurmukhametov (changes requested):

cmake/GenericTargets.cmake
aneshlya: Done

tests/lit-tests/3123.ispc
aneshlya: Oh, right. I didn't notice it.

nurmukhametov (approved):
LGTM!

Issue #3098, opened 2024-11-07
Fix warnings in examples and build them with -Werror in CI


aneshlya:
ISPC examples from the examples/cpu folder should compile without warnings and must be built in CI with the -Werror flag to enforce this requirement.

yi2cao:
PR link: https://github.com/ispc/ispc/pull/3268

Summary of the PR 3268: fix the warnings in examples/cpu and update CI workflow to enforce -Werror flag for examples build
There are four warnings detected for the examples:
1. in examples/cpu/deferred, the compiler warning arises from the use of a reinterpret_cast to convert the address of a local variable into a float pointer. A common and safe solution is to avoid the reinterpret_cast altogether and use a union for type punning.
2. in examples/cpu/gmres: the warning appears because the variable 'error' is set (in the else branch of the first check) but is never actually used in any decision-making or output. It was fixed by removing the unused 'error' variable.
3. in examples/common/tasksys.cpp, the "unused-function" warning for the inline function lAtomicAdd() arises on macOS. It was fixed by marking it as [[maybe-unused]]
4. in examples/cpu/gmres/main.cpp, MSVC requires a different format specifier for printing a size_t value than Linux and macOS. A macro was added in debug.h to assign different format specifiers for different environments.

Additionally, CMakeLists.txt for examples build was updated to define the C++17 standard

Finally, The ispc-ci.yml workflow was updated to enforce -Werror flag for examples build

PR #3268, opened 2025-03-05
#3098 fix the warnings in examples/cpu and update CI workflow to enforce -Werror flag for examples build

yi2cao:
fix the warnings in examples/cpu and update CI workflow to enforce -Werror flag for examples build
there are two warnings detected for the examples/cpu
1. in examples/cpu/deferred, the compiler warning arises from the use of a reinterpret_cast to convert the address of a local variable into a float pointer. A common and safe solution is to avoid the reinterpret_cast altogether and use a union for type punning.
2. in examples/cpu/gmres: the warning appears because the variable 'error' is set (in the else branch of the first check) but is never actually used in any decision-making or output. It was fixed by removing the unused 'error' variable

Finally, The ispc-ci.yml workflow was updated to enforce -Werror flag for examples build

### DIFF
--- a/examples/common/tasksys.cpp
+++ b/examples/common/tasksys.cpp
@@ -1,5 +1,5 @@
 /*
-  Copyright (c) 2011-2023, Intel Corporation
+  Copyright (c) 2011-2025, Intel Corporation
 
   SPDX-License-Identifier: BSD-3-Clause
 */
@@ -326,7 +326,7 @@ static int32_t lAtomicCompareAndSwap32(volatile int32_t *v, int32_t newValue, in
 #endif // ISPC_IS_WINDOWS
 }
 
-static inline int32_t lAtomicAdd(volatile int32_t *v, int32_t delta) {
+[[maybe_unused]] static inline int32_t lAtomicAdd(volatile int32_t *v, int32_t delta) {
 #ifdef ISPC_IS_WINDOWS
     return InterlockedExchangeAdd((volatile LONG *)v, delta) + delta;
 #else
--- a/examples/cpu/deferred/dynamic_c.cpp
+++ b/examples/cpu/deferred/dynamic_c.cpp
@@ -1,5 +1,5 @@
 /*
-  Copyright (c) 2011-2023, Intel Corporation
+  Copyright (c) 2011-2025, Intel Corporation
 
   SPDX-License-Identifier: BSD-3-Clause
 */
@@ -331,8 +331,14 @@ static inline float half_to_float_fast(uint16_t h) {
     uint32_t xm = ((uint32_t)hm) << 13;
 
     uint32_t bits = (xs | xe | xm);
-    float *fp = reinterpret_cast<float *>(&bits);
-    return *fp;
+
+    // Use a union for safe type punning.
+    union {
+        uint32_t u;
+        float f;
+    } conv;
+    conv.u = bits;
+    return conv.f;
 }
 
 static void ShadeTileC(int32_t tileStartX, int32_t tileEndX, int32_t tileStartY, int32_t tileEndY, int32_t gBufferWidth,
--- a/examples/cpu/gmres/debug.h
+++ b/examples/cpu/gmres/debug.h
@@ -1,5 +1,5 @@
 /*
-  Copyright (c) 2012-2023, Intel Corporation
+  Copyright (c) 2012-2025, Intel Corporation
 
   SPDX-License-Identifier: BSD-3-Clause
 */
@@ -8,11 +8,21 @@
 #define __DEBUG_H__
 
 #include <cassert>
+#include <cstdio>
 
 /**************************************************************\
 | Macros
 \**************************************************************/
 #define DEBUG
+
+// Define the correct format specifier for size_t
+#ifdef _MSC_VER
+  // MSVC: use '%zu'
+  #define SIZE_T_FORMAT "%zu"
+#else
+  // Linux/macOS: use '%lu'
+  #define SIZE_T_FORMAT "%lu"
+#endif
 
 #ifdef DEBUG
 #define ASSERT(expr) assert(expr)
--- a/examples/cpu/gmres/mmio.c
+++ b/examples/cpu/gmres/mmio.c
@@ -357,13 +357,12 @@ char *mm_typecode_to_str(MM_typecode matcode) {
     char buffer[MM_MAX_LINE_LENGTH];
     char *types[4];
     char *mm_strdup(const char *);
-    int error = 0;
 
     /* check for MTX type */
     if (mm_is_matrix(matcode))
         types[0] = MM_MTX_STR;
     else
-        error = 1;
+        return NULL;
 
     /* check for CRD or ARR matrix */
     if (mm_is_sparse(matcode))
--- a/.github/workflows/ispc-ci.yml
+++ b/.github/workflows/ispc-ci.yml
@@ -1404,7 +1404,7 @@ jobs:
     - name: Build examples
       run: |
         mkdir examples-build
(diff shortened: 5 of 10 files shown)


aneshlya:
Please do the following:
1. update copyright in examples/cpu/deferred/dynamic_c.cpp to 2025
2. split your commit into two: 1. Fixed warnings in ISPC examples 2. CI: build examples with -Werror

I'm not sure that all warnings are fixed (but will see it in CI). Here is what I see in CI on main branch: https://github.com/ispc/ispc/actions/runs/13648097998/job/38151760629

examples/cpu/gmres/mmio.c
```diff
@@ -357,13 +357,14 @@ char *mm_typecode_to_str(MM_typecode matcode) {
     char buffer[MM_MAX_LINE_LENGTH];
     char *types[4];
     char *mm_strdup(const char *);
-    int error = 0;
+    //int error = 0;
```
nurmukhametov, line 360: Feel free to remove old code.

aneshlya:
Not all warning has been fixed:
Please check `macos-example-build` job: https://github.com/ispc/ispc/actions/runs/13665740155/job/38216300142

examples/cpu/deferred/dynamic_c.cpp
```diff
@@ -331,8 +331,16 @@ static inline float half_to_float_fast(uint16_t h) {
     uint32_t xm = ((uint32_t)hm) << 13;
 
     uint32_t bits = (xs | xe | xm);
-    float *fp = reinterpret_cast<float *>(&bits);
-    return *fp;
+
+    //float *fp = reinterpret_cast<float *>(&bits);
+    //return *fp;
```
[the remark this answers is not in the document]
aneshlya: This comment still remains actual.

aneshlya:
Please enable treating warnings as errors in [win-build-examples](https://github.com/ispc/ispc/actions/runs/13665740155/job/38216644336) job as well. As you can see there are some warnings that has to be fixed. 
To enable `-Werror` mode on Windows, please use `-Wx` flag like: `-DCMAKE_C_FLAGS="/WX" -DCMAKE_CXX_FLAGS="/WX"`

aneshlya (changes requested):

examples/cpu/deferred/dynamic_c.cpp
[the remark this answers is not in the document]
yi2cao: I fix the warning for macos-build-examples. It looks like AppleClang 15.0.0 on macOS is more aggressive: it sees that within tasksys.cpp the function lAtomicAdd is not called and issues a -Wunused-function warning. So I added an attribute to explicitly tell the compiler that the function is used. 

I also enable treating warnings as errors in [win-build-examples] job. 

My local system is Linux only. It's hard to test examples on Windows and MacOS in my local. Please run the CI again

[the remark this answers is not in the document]
nurmukhametov: I am puzzled how `lAtomicAdd` can be detected as unused, although suppressing this warning with an attribute seems fine to me.

nurmukhametov (approved):
Unfortunately, syntax `__attrubute__((used))` is not supported by MSVC (failed job [here](https://github.com/ispc/ispc/actions/runs/13697197484/job/38302294618), compiler explorer [link](https://godbolt.org/z/91cbo8ojE)).

nurmukhametov (changes requested):
Moreover, it looks like there are more warnings on macos and windows builds: https://github.com/ispc/ispc/actions/runs/13697197488/job/38303140142?pr=3268

yi2cao:
> Moreover, it looks like there are more warnings on macos and windows builds: https://github.com/ispc/ispc/actions/runs/13697197488/job/38303140142?pr=3268

Yes, this warning is for examples/cpu/gmres/matrix.h. The windows compiler issues an "unused private field" warning for the DenseMatrix class because its private member shared_ptr is never actually referenced after being initialized. 
So, I remove it from the DenseMatrix class

yi2cao:
> Unfortunately, syntax `__attrubute__((used))` is not supported by MSVC (failed job [here](https://github.com/ispc/ispc/actions/runs/13697197484/job/38302294618), compiler explorer [link](https://godbolt.org/z/91cbo8ojE)).

Since  `__attrubute__((used))` is not supported by MSVC, I changed it to [[maybe_unused]]. This corrected syntax is standard and should work on Linux (with GCC), macOS (with Clang), and MSVC (with C++17 support).

examples/common/tasksys.cpp
```diff
@@ -326,7 +326,7 @@ static int32_t lAtomicCompareAndSwap32(volatile int32_t *v, int32_t newValue, in
 #endif // ISPC_IS_WINDOWS
 }
 
-static inline int32_t lAtomicAdd(volatile int32_t *v, int32_t delta) {
+static inline int32_t lAtomicAdd(volatile int32_t *v, int32_t delta) [[maybe_unused]] {
```
aneshlya, line 329:
```suggestion
[[maybe_unused]] static inline int32_t lAtomicAdd(volatile int32_t *v, int32_t delta) {
```

yi2cao: unfortunately, [[maybe_unused]] is not supported by MSVC.
Also, another warning was detected on windows-build-example.
Failed job here: https://github.com/ispc/ispc/actions/runs/13711393211/job/38354732674.

I will work on it

nurmukhametov: It looks like `[[maybe unused]]` is supported by MSVC only when `/std:c++17` is passed. This seems correct since it is indeed C++17 syntax, as noted here: https://en.cppreference.com/w/cpp/language/attributes/maybe_unused. It seems we need to define the C++17 standard in the CMakeLists.txt of the examples, as it is done here: https://github.com/ispc/ispc/blob/f3ad7f936c498262ba7769acd8e6ad231d5454e3/CMakeLists.txt#L14.

yi2cao: I update the CMakeLists.txt to define the C++17 standard. I re-push the changes to the PR. Hope this works

yi2cao: update C++17 standard doesn't solve the [[maybe_unused]] support issue. I am suppressing the "unused-function" warning on macOS.

aneshlya: Please don't suppress it. It seems that the standard was overridden later:
https://github.com/ispc/ispc/blob/main/examples/cpu/cmake/AddISPCExampleModern.cmake#L31-L33
https://github.com/ispc/ispc/blob/main/examples/cpu/cmake/AddISPCExampleLegacy.cmake#L75-L77.
Feel free to remove these lines.

yi2cao: Ok, I will re-run the CI without suppressing the "unuse-function" warning after removing thoese lines.

Btw, when suppressing the warning, it passed all CI workflow

xiaohanIntel:
Fixes #3268 (comment for issue mention)

examples/common/tasksys.cpp
yi2cao: OK, it looks good after removing those lines. Now, [[maybe_unused]] sign works for all example builds. 
All warnings have been solved

examples/cpu/gmres/debug.h
```diff
@@ -8,12 +8,22 @@
...
 \**************************************************************/
 #define DEBUG
 
+// Define the correct format specifier for size_t
+#ifdef _MSC_VER
+  // MSVC: use '%zu'
+  #define SIZE_T_FORMAT "%zu"
+#else
+  // Linux/macOS: use '%lu'
+  #define SIZE_T_FORMAT "%lu"
```
aneshlya, line 24: Just a note: I think that "%zu" would work on Linux as well.

PR #3269, opened 2025-03-05
Add -dD and -dM compile flags for ispc ( Closes #3205 )

### LINKED ISSUE #3205: Support `-dD` and `-dM` flags
As discussed in issue #2167, it would be good to support `-dD` and `-dD` command line flags. See comments  https://github.com/ispc/ispc/issues/2167#issuecomment-2631786461, https://github.com/ispc/ispc/issues/2167#issuecomment-2630832796 and https://github.com/ispc/ispc/issues/2167#issuecomment-2631005883 for more context and sample test cases.

xiaohanIntel:
Added -dD and -dM options for ispc and tests. 
(Commit clean, no squash merge needed)

Implementation:
Flags macthing clang pre-process options refer to https://clang.llvm.org/doxygen/classclang_1_1PreprocessorOutputOptions.html:
-dM: ShowMacros only - Find customized macros, and no functions
-E: ShowCPP and ShowComments - Find function but no macros definition
-dD: ShowCPP, ShowMacros, and ShowComments - Find both macros definition and function.
Functionality difference: Unlike clang -dD or -dM must be used in addition with -E flag, both flags for ispc in this implementation makes the "-E" implicit. This is an easier implementation and I didn't find the benefit to follow clang. Please leave your thoughts if you prefer the clang way.

Test:
Covered by functions and macros search in results.
In discussion https://github.com/ispc/ispc/issues/2167#issuecomment-2631786461, ideally, the test should make sure "all the expected macros are added and nothing else is defined". This is not doable because ispc can add macros during runtime, by using g->addMacroDef(<DEF>). Thus, it is not easy to design a golden result to cover the exact necessary ispc macros.
Thus, the detection of macros only covers "finding all the user's macros" instead.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -71,6 +71,7 @@ static void lPrintVersion() {
     printf("    [--darwin-version-min=<major.minor>]Set the minimum macOS/iOS version required for the "
            "deployment.\n");
 #endif
+    printf("    [-dD]\t\t\t\tPrint macro definitions in addition to the preprocessor result\n");
     printf("    [--dev-stub <filename>]\t\tEmit device-side offload stub functions to file\n");
     printf("    ");
     char cpuHelp[2048];
@@ -78,6 +79,7 @@ static void lPrintVersion() {
              Target::SupportedCPUs().c_str());
     PrintWithWordBreaks(cpuHelp, 16, TerminalWidth(), stdout);
     printf("    [--dllexport]\t\t\tMake non-static functions DLL exported.  Windows target only\n");
+    printf("    [-dM]\t\t\t\tPrint macro definitions for the preprocessor result\n");
     printf("    [--dwarf-version={2,3,4,5}]\t\tGenerate source-level debug information with given DWARF version "
            "(triggers -g).  It forces the usage of DWARF debug info on Windows target\n");
     printf("    [-E]\t\t\t\tRun only the preprocessor\n");
@@ -821,6 +823,15 @@ int main(int Argc, char *Argv[]) {
         } else if (!strcmp(argv[i], "-E")) {
             g->onlyCPP = true;
             ot = Module::CPPStub;
+            // g->preprocessorOutputType is initialized as "Cpp" automatically
+        } else if (!strcmp(argv[i], "-dM")) {
+            g->onlyCPP = true;
+            ot = Module::CPPStub;
+            g->preprocessorOutputType = Globals::PreprocessorOutputType::MacrosOnly;
+        } else if (!strcmp(argv[i], "-dD")) {
+            g->onlyCPP = true;
+            ot = Module::CPPStub;
+            g->preprocessorOutputType = Globals::PreprocessorOutputType::WithMacros;
         } else if (!strcmp(argv[i], "--emit-asm")) {
             ot = Module::Asm;
         } else if (!strcmp(argv[i], "--emit-llvm")) {
--- a/src/module.cpp
+++ b/src/module.cpp
@@ -358,7 +358,7 @@ int Module::preprocessAndParse() {
 
     initCPPBuffer();
 
-    const int numErrors = execPreprocessor(filename, bufferCPP->os.get());
+    const int numErrors = execPreprocessor(filename, bufferCPP->os.get(), g->preprocessorOutputType);
     errorCount += (g->ignoreCPPErrors) ? 0 : numErrors;
 
     if (g->onlyCPP) {
@@ -2996,11 +2996,29 @@ static void lInitializePreprocessor(clang::Preprocessor &PP, clang::Preprocessor
     PP.setPredefines(std::move(PredefineBuffer));
 }
 
-static void lSetPreprocessorOutputOptions(clang::PreprocessorOutputOptions *opts) {
-    // Don't remove comments in the preprocessor, so that we can accurately
-    // track the source file position by handling them ourselves.
-    opts->ShowComments = 1;
-    opts->ShowCPP = 1;
+// Initilialize PreprocessorOutputOptions with ISPC PreprocessorOutputType
+// Return if the output can be treated as preprocessor output
+static bool lSetPreprocessorOutputOptions(clang::PreprocessorOutputOptions *opts,
+                                          Globals::PreprocessorOutputType preprocessorOutputType) {
+    switch (preprocessorOutputType) {
+    case Globals::PreprocessorOutputType::Cpp:
+        // Don't remove comments in the preprocessor, so that we can accurately
+        // track the source file position by handling them ourselves.
+        opts->ShowComments = 1;
+        opts->ShowCPP = 1;
+        return true;
+    case Globals::PreprocessorOutputType::WithMacros:
+        opts->ShowCPP = 1;
+        opts->ShowMacros = 1;
+        opts->ShowComments = 1;
+        return true;
+    case Globals::PreprocessorOutputType::MacrosOnly:
+        opts->ShowMacros = 1;
+        return false;
+    }
+
+    // Undefined PreprocessorOutputType
+    return false;
 }
 
 static void lSetHeaderSeachOptions(const std::shared_ptr<clang::HeaderSearchOptions> opts) {
@@ -3184,7 +3202,8 @@ static void lSetPreprocessorOptions(const std::shared_ptr<clang::PreprocessorOpt
 
 static void lSetLangOptions(clang::LangOptions *opts) { opts->LineComment = 1; }
 
-int Module::execPreprocessor(const char *infilename, llvm::raw_string_ostream *ostream) const {
+int Module::execPreprocessor(const char *infilename, llvm::raw_string_ostream *ostream,
+                             Globals::PreprocessorOutputType preprocessorOutputType) const {
     clang::FrontendInputFile inputFile(infilename, clang::InputKind());
     llvm::raw_fd_ostream stderrRaw(2, false);
 
@@ -3207,17 +3226,17 @@ int Module::execPreprocessor(const char *infilename, llvm::raw_string_ostream *o
 
     // Create and initialize PreprocessorOutputOptions
     clang::PreprocessorOutputOptions preProcOutOpts;
-    lSetPreprocessorOutputOptions(&preProcOutOpts);
+    bool isPreprocessedOutput = lSetPreprocessorOutputOptions(&preProcOutOpts, preprocessorOutputType);
 
     // Create and initialize HeaderSearchOptions
     const std::shared_ptr<clang::HeaderSearchOptions> hdrSearchOpts = std::make_shared<clang::HeaderSearchOptions>();
     lSetHeaderSeachOptions(hdrSearchOpts);
 
-    // Create and initializer PreprocessorOptions
+    // Create and initialize PreprocessorOptions
     const std::shared_ptr<clang::PreprocessorOptions> preProcOpts = std::make_shared<clang::PreprocessorOptions>();
     lSetPreprocessorOptions(preProcOpts);
 
-    // Create and initializer LangOptions
+    // Create and initialize LangOptions
     clang::LangOptions langOpts;
     lSetLangOptions(&langOpts);
 
@@ -3235,9 +3254,9 @@ int Module::execPreprocessor(const char *infilename, llvm::raw_string_ostream *o
     clang::TrivialModuleLoader modLoader;
     clang::Preprocessor prep(preProcOpts, diagEng, langOpts, srcMgr, hdrSearch, modLoader);
 
-    // intialize preprocessor
+    // Initialize preprocessor
     prep.Initialize(*tgtInfo);
-    prep.setPreprocessedOutput(preProcOutOpts.ShowCPP);
+    prep.setPreprocessedOutput(isPreprocessedOutput);
     lInitializePreprocessor(prep, *preProcOpts, fileMgr, srcMgr);
 
     // do actual preprocessing
--- a/tests/lit-tests/3205.ispc
+++ b/tests/lit-tests/3205.ispc
@@ -1,0 +1,37 @@
+// Ensure that output(-dM) can output the embedded predefined ISPC macros with empty input
+// RUN: %{ispc} -dM --outfile=emptydm.dM.cpp -
+// RUN: FileCheck %s --input-file=emptydm.dM.cpp --check-prefix=EMPTYDM
+// EMPTYDM: ISPC
+
+// Ensure that the output(-dM) includes macros but doesn't include functions
+// -dm itself doesn't enable "ShowComments" so only need to check for 1 count.
+// RUN: %{ispc} %s -dM --outfile=%t.dM.cpp
+// RUN: FileCheck %s --input-file=%t.dM.cpp --check-prefix=DM
+// DM: MyDescriptor
+// DM: MyType
+// DM: MyValue
+// DM-NOT: my_func
+
+// Ensure that output(-E) has functions but doesn't include macros
+// -E has "ShowComments", so need to count lit comments
+// RUN: %{ispc} %s -E --outfile=%t.E.cpp
+// RUN: FileCheck %s --input-file=%t.E.cpp --check-prefix=E
+// E: CHECK-COUNT-3: MyDescriptor
+// E: CHECK-COUNT-3: MyType
+// E: CHECK-COUNT-3: MyValue
+// E: CHECK-COUNT-4: my_func
+
+// Ensure that output(-dD) has both functions and macros.
+// -dD has "ShowComments", so need to count lit comments
+// RUN: %{ispc} %s -dD --outfile=%t.dD.cpp
+// RUN: FileCheck %s --input-file=%t.dD.cpp --check-prefix=DD
+// DD: CHECK-COUNT-4: MyDescriptor
+// DD: CHECK-COUNT-4: MyType
+// DD: CHECK-COUNT-4: MyValue
+// DD: CHECK-COUNT-4: my_func
+
+#define MyType float
+#define MyDescriptor const
+#define MyValue 100
+
+MyType my_func(MyDescriptor MyType _value) { return _value + MyValue; }
--- a/src/ispc.h
+++ b/src/ispc.h
@@ -891,6 +891,11 @@ struct Globals {
 
     enum pragmaUnrollType { none, nounroll, unroll, count };
 
+    /** Preprocessor Output Types, process with -E/-dD/-dM options */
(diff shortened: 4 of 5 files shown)


nurmukhametov:
This fixes #3205

src/module.cpp
```diff
@@ -2996,13 +2997,6 @@ static void lInitializePreprocessor(clang::Preprocessor &PP, clang::Preprocessor
     PP.setPredefines(std::move(PredefineBuffer));
 }
 
-static void lSetPreprocessorOutputOptions(clang::PreprocessorOutputOptions *opts) {
```
nurmukhametov, line 2999: Let's keep this function and move the switch/case statements inside it. Also, please make it return a bool value to pass as an argument to prep.setPreprocessedOutput.

src/module.cpp
```diff
@@ -3231,21 +3241,20 @@ int Module::execPreprocessor(const char *infilename, llvm::raw_string_ostream *o
...
     lInitializePreprocessor(prep, *preProcOpts, fileMgr, srcMgr);
 
-    // do actual preprocessing
     diagPrinter.BeginSourceFile(langOpts, &prep);
-    clang::DoPrintPreprocessedInput(prep, ostream, preProcOutOpts);
+    clang::DoPrintPreprocessedInput(prep, cppbuffer, preProcOutOpts);
     diagPrinter.EndSourceFile();
 
-    // deallocate some objects
+    // Cleanup
```
nurmukhametov, line 3257: Why do we need unnecessary changes in comments? It makes the change larger and noisier. Fixing typos or language is fine.

tests/lit-tests/3205.ispc
```diff
@@ -0,0 +1,38 @@
+// Ensure that the output(-dM) includes macros but doesn't include functions
```
nurmukhametov, line 1: > In discussion https://github.com/ispc/ispc/issues/2167#issuecomment-2631786461, ideally, the test should make sure "all the expected macros are added and nothing else is defined". This is not doable because ispc can add macros during runtime, by using g->addMacroDef().

As I understand from this comment: https://github.com/ispc/ispc/issues/2167#issuecomment-2631786461, it is intended to check the embedded predefined ISPC macros, not the user-defined ones. Therefore, an additional empty test with `ispc -o - -E -dM` could cover this.

nurmukhametov (changes requested):
I slightly prefer matching GCC/Clang behavior, but overall, I don't see a big issue with your implementation. Let's ask @aneshlya and @dbabokin; they may have stronger opinions about this.

In general, this PR looks good and implements the `-dD`/`-dM` functionality as expected. After addressing minor changes, it will be ready to go.

tests/lit-tests/3205.ispc
```diff
@@ -0,0 +1,38 @@
...
+
+
+#define MyType float
+#define MyDescriptor const
+#define MyValue 100
+
+MyType my_func(MyDescriptor MyType _value)
+{
+	return _value + MyValue;
+}
```
aneshlya, line 38: This code is not formatted in a right way. We don't enforce formatting in lit-tests but it would be nice to run clang-format on this function.

aneshlya:
> I slightly prefer matching GCC/Clang behavior, but overall, I don't see a big issue with your implementation. Let's ask @aneshlya and @dbabokin; they may have stronger opinions about this.

I'm OK with this implementation.

src/main.cpp
```diff
@@ -821,6 +823,14 @@ int main(int Argc, char *Argv[]) {
         } else if (!strcmp(argv[i], "-E")) {
             g->onlyCPP = true;
             ot = Module::CPPStub;
```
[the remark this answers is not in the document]
xiaohanIntel: Well.. That is tricky. Basically, we need this value always be CPP for all flags.. Like if -E is not there, Cpp should be set. Otherwise will crash on some platform. Thus, I initialized this value in Struct Gobals instead.
If we want to assign this value again in main, I would say this is not the right place. Maybe initialize in the very beginning of main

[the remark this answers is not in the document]
aneshlya: OK, just add a comment about it then.

src/module.cpp
xiaohanIntel: Agreed. This change was warned by AI.. like complains "this is the only none cap comments" lol
Yeah I would agree this is not necessary

src/module.cpp
xiaohanIntel: @aneshlya  Do you agree on this? if so it looks weird like
switch(..){
case A:
lSetPreprocessorOutputOptions();
break;
case B: 
.showMacros=xx
break;
case C:
.showMacros=xx
.showCpp=yy
.showComments=zz
break;
}

I find use "lSetPreprocessorOutputOptions" is harder to read and understand and seems a little meaningless. Thoughts?

aneshlya: Aleksei's idea is to have the whole switch statement inside `lSetPreprocessorOutputOptions` function which totally makes sense to me
```
static void lSetPreprocessorOutputOptions(/*pass correct params here*/). 
    switch (preprocessorOutputType) {
    case Globals::PreprocessorOutputType::Cpp:
        // Don't remove comments in the preprocessor, so that we can accurately
        // track the source file position by handling them ourselves.
        preProcOutOpts.ShowComments = 1;
        preProcOutOpts.ShowCPP = 1;
        break;
    case Globals::PreprocessorOutputType::WithMacros:
        preProcOutOpts.ShowCPP = 1;
        preProcOutOpts.ShowMacros = 1;
        preProcOutOpts.ShowComments = 1;
        break;
    case Globals::PreprocessorOutputType::MacrosOnly:
        preProcOutOpts.ShowMacros = 1;
        break;
    }
}

Comment (nurmukhametov) [reply src/module.cpp:2999]: > Aleksei's idea is to have the whole switch statement inside `lSetPreprocessorOutputOptions` function which totally makes sense to me
> 
> ```
> static bool lSetPreprocessorOutputOptions(/*pass correct params here*/). 

Correct, this is exactly I have in my mind.

File: tests/lit-tests/3205.ispc (hunk 4 above)
Comment (xiaohanIntel) [reply tests/lit-tests/3205.ispc]: resolved

File: src/main.cpp (hunk 5 above)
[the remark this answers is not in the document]
Comment (xiaohanIntel) [reply src/main.cpp:825]: added

File: src/module.cpp (hunk 1 above)
Comment (xiaohanIntel) [reply src/module.cpp:2999]: reverted to the original format

File: tests/lit-tests/3205.ispc (hunk 3 above)
Comment (xiaohanIntel) [reply tests/lit-tests/3205.ispc:6]: added as below (-E was implicit)

// Ensure that output(-dM) can output the embedded predefined ISPC macros with empty input
// RUN: %{ispc} -dM --outfile=emptydm.dM.cpp -
// RUN: FileCheck %s --input-file=emptydm.dM.cpp --check-prefix=EMPTYDM
// EMPTYDM: ISPC

Comment (nurmukhametov) [reply tests/lit-tests/3205.ispc:6]: If you created a new test, it seems you forgot to add it to this PR.

File: src/module.cpp [hunk 6]
```diff
@@ -3207,17 +3222,17 @@ int Module::execPreprocessor(const char *infilename, llvm::raw_string_ostream *o
 
     // Create and initialize PreprocessorOutputOptions
     clang::PreprocessorOutputOptions preProcOutOpts;
-    lSetPreprocessorOutputOptions(&preProcOutOpts);
+    lSetPreprocessorOutputOptions(&preProcOutOpts, preprocessorOutputType);
```
nurmukhametov, line 3225: Please, return `bool` value here and pass it to `prep.setPreprocessedOutput` below, i.e.:
```c++
bool isPreprocessedOutput = lSetPreprocessorOutputOptions(&preProcOutOpts, preprocessorOutputType);
...
prep.setPreprocessedOutput(isPreprocessedOutput);
```

src/module.cpp
```diff
@@ -1576,6 +1576,7 @@ bool Module::writeOutput(OutputType outputType, OutputFlags flags, const char *o
         return writeHostStub(outFileName);
     case CPPStub:
         return writeCPPStub(outFileName);
+
```
nurmukhametov, line 1579: Please, remove this new line addition.

nurmukhametov (changes requested):

tests/lit-tests/3205.ispc
xiaohanIntel: check the line 1- line 5 of 3205.ispc


37 changes: 37 additions & 0 deletions37  
[tests/lit-tests/3205.ispc](https://github.com/ispc/ispc/pull/3269/files#diff-5132ad70cdc63f0dba77cf8a60c45af340205086709ac877bd4df235c428e33e)
Viewed
Original file line number	Diff line number	Diff line change
@@ -0,0 +1,37 @@
// Ensure that output(-dM) can output the embedded predefined ISPC macros with empty input
// RUN: %{ispc} -dM --outfile=emptydm.dM.cpp -
// RUN: FileCheck %s --input-file=emptydm.dM.cpp --check-prefix=EMPTYDM
// EMPTYDM: ISPC

xiaohanIntel: same source file

src/module.cpp
xiaohanIntel: honestly didn't get why this change.. since "setPreprocessedOutput" is purely relying on if it is including showCPP or not.. but fine if you like it

xiaohanIntel: done

src/module.cpp
xiaohanIntel: done

src/module.cpp
nurmukhametov: > honestly didn't get why this change.. since "setPreprocessedOutput" is purely relying on if it is including showCPP or not.. but fine if you like it

It wasn't so before. The previous edition of your PR relied on some combination of these flags.

tests/lit-tests/3205.ispc
nurmukhametov: > check the line 1- line 5 of 3205.ispc
> 
> 37 changes: 37 additions & 0 deletions37 [tests/lit-tests/3205.ispc](https://github.com/ispc/ispc/pull/3269/files#diff-5132ad70cdc63f0dba77cf8a60c45af340205086709ac877bd4df235c428e33e) Viewed Original file line number Diff line number Diff line change @@ -0,0 +1,37 @@ // Ensure that output(-dM) can output the embedded predefined ISPC macros with empty input // RUN: %{ispc} -dM --outfile=emptydm.dM.cpp - // RUN: FileCheck %s --input-file=emptydm.dM.cpp --check-prefix=EMPTYDM // EMPTYDM: ISPC

I see it now, but it checks only one predefined ISPC macro. 

The comment #2167 (comment) and my previous message requested predefined macros (plural). The point of this test is to verify that ISPC actually defines the macros we claim to define in the documentation here  https://ispc.github.io/ispc.html#the-preprocessor

nurmukhametov:
> seems github doesn't allow me to reply directly under your comment complaining missing tests, replying it here. @nurmukhametov Go to the source file and check line 1 to line 6

I have answered in the thread here https://github.com/ispc/ispc/pull/3269#discussion_r1983730664.

tests/lit-tests/3205.ispc
xiaohanIntel: are you expected me to design a test to test everything there? I am feeling it is out of the scope.
So you want to check all these macros in a lit test? It doesn't sound like a lit test thing with -dM empty input.

I would prefer an implementation like below:
write an executable task and try to use every macro you claimed in your tests. This is better than using lit tests like "check-count". 

I want to point out that this feels out of the scope of adding -dD/-dM ticket.

"We can check specific macros in lit test with -E without executing the binary." <- this sounds more reasonable to match your expectation. 

Could you give me an example of how you want it to be implemented? Does it mean that every macro not listed in this website https://ispc.github.io/ispc.html#the-preprocessor should not be there? Is that the expectation on all of the OS or environment?

xiaohanIntel: This won't work.
For example, even you miss "ISPC", ISPC_TARGET_NEON is still include "ISPC" string

nurmukhametov: > This won't work. For example, even you miss "ISPC", ISPC_TARGET_NEON is still include "ISPC" string

// CHECK: ISPC$
Will this work?

xiaohanIntel: I have found a workaround to handle macro detections like this

```
// Ensure that the empty input can somehow
// RUN: %{ispc} -dM --outfile=empty.dM.cpp -
// RUN: FileCheck %s --input-file=empty.dM.cpp --check-prefix=EmptyDM
// EmptyDM-DAG: #define ISPC{{($|[ ])}}
// EmptyDM-DAG: #define ISPC_TARGET_{{(NEON|SSE2|SSE4|AVX|AVX2|AVX512KNL|AVX512SKX|AVX512SPR)}}{{($|[ ])}}
// EmptyDM-DAG: #define ISPC_POINTER_SIZE{{($|[ ])}}
// EmptyDM-DAG: #define ISPC_MAJOR_VERSION{{($|[ ])}}
// EmptyDM-DAG: #define ISPC_MINOR_VERSION{{($|[ ])}}
// EmptyDM-DAG: #define LLVM_VERSION_MAJOR{{($|[ ])}}
// EmptyDM-DAG: #define LLVM_VERSION_MINOR{{($|[ ])}}
// EmptyDM-DAG: #define PI{{($|[ ])}}
// EmptyDM-DAG: #define TARGET_WIDTH{{($|[ ])}}
// EmptyDM-DAG: #define TARGET_ELEMENT_WIDTH{{($|[ ])}}
// EmptyDM-DAG: #define ISPC_UINT_IS_DEFINED{{($|[ ])}}
// EmptyDM-DAG: #define ISPC_ATTRIBUTE_SUPPORTED{{($|[ ])}}
// EmptyDM-DAG: #define ISPC_FP16_SUPPORTED{{($|[ ])}}
// EmptyDM-DAG: #define ISPC_FP64_SUPPORTED{{($|[ ])}}
// EmptyDM-DAG: #define ISPC_LLVM_INTRINSICS_ENABLED{{($|[ ])}}
// EmptyDM-DAG: #define INT8_MIN{{($|[ ])}}
// EmptyDM-DAG: #define INT16_MIN{{($|[ ])}}
// EmptyDM-DAG: #define INT32_MIN{{($|[ ])}}
// EmptyDM-DAG: #define INT64_MIN{{($|[ ])}}
// EmptyDM-DAG: #define INT8_MAX{{($|[ ])}}
// EmptyDM-DAG: #define INT16_MAX{{($|[ ])}}
// EmptyDM-DAG: #define INT32_MAX{{($|[ ])}}
// EmptyDM-DAG: #define INT64_MAX{{($|[ ])}}
// EmptyDM-DAG: #define UINT8_MAX{{($|[ ])}}
// EmptyDM-DAG: #define UINT16_MAX{{($|[ ])}}
// EmptyDM-DAG: #define UINT32_MAX{{($|[ ])}}
// EmptyDM-DAG: #define UINT64_MAX{{($|[ ])}}
// EmptyDM-DAG: #define FLT16_MIN{{($|[ ])}}
// EmptyDM-DAG: #define FLT_MIN{{($|[ ])}}
// EmptyDM-DAG: #define DBL_MIN{{($|[ ])}}
// EmptyDM-DAG: #define FLT16_MAX{{($|[ ])}}
// EmptyDM-DAG: #define FLT_MAX{{($|[ ])}}
// EmptyDM-DAG: #define DBL_MAX{{($|[ ])}}
```

aneshlya: I see that  ISPC_TARGET is defined correctly, it is ` ISPC_TARGET_AVX2VNNI` on your machine.
Maybe this would help?
`// EmptyDM-DAG: #define ISPC_TARGET_{{(NEON|SSE2|SSE4|AVX|AVX2|AVX2VNNI|AVX512KNL|AVX512SKX|AVX512SPR)}}{{($|[ ])}}`

If `ISPC_FP16_SUPPORTED` is not defined it's a bug.
Ideally could you create one passing test with all the macros that are correctly defined and another one with the macros that are not defined and marked it as `// XFAIL: *`? 
If it's too much that instead of 2nd test just create a new issue.

xiaohanIntel: I disagree. If there is a mismatch, then the first thing to do is to find if the document needs to be changed first. If we reach an agreement on the document mismatch first, and correct that properly, then we should consider to follow that with a test coverage.
 
I am deciding to stop here for my hackathon. Feeling sick already. Feel free to close it on your side if you think the work is not solid.
In your original github issues, it is not mentioning the docs of macro definitions at all; and now since there is a mismatch in the doc, I am asked to handle those mismatches.
Then after that, what about
"ISPC_MATH_LIB_.."
"#define UIntMaskType unsigned int32
#define UniformMaskType uniform int32
#define assert(x) __assert(#x, x)"?
 Those are not following the discussion results:
" all the expected macros are added and **nothing else** is defined"

No mentioning that we have tons of operating systems on your test env with no error messages reported. I cannot find a good way to give a good handling. Only way to do that is to do github pushes 100 times to find the right combination to pass CI with a solid result?

As a part time event it starts to block my daily work by "Accessing Github Api limit" and my predicted bandwidth.

aneshlya: Thank you for all the effort you've put into this. We truly appreciate the time and energy you’ve dedicated. If you're deciding to stop here, we completely respect that, and we’ll finish it on our side.

Compilers are expected to support a broad range of requirements across many environments, which naturally increases the number of factors to consider. This is a challenge in compiler development, and we understand how frustrating it can be, especially in a time-constrained setting like a hackathon.

aneshlya (approved):
The implementation looks good to me. We can update the lit test with all required definitions in a separate PR.

tests/lit-tests/3205.ispc
xiaohanIntel: I personally do not agree with you on this.
I am adding a test for "add -dD and -dM" feature, and I am not adding a validation pipeline to verify the functionality of Test if the Macro output matches expectation should be a different story.

xiaohanIntel: and if there is a test covering the macro mismatch, then it should be named as "3204.ispc" instead

xiaohanIntel: oh I didn't receive notifications of approve. I did that right after I saw "we'll finish on our side". Maybe when you did that it didn't reflect on my page. I am not seeing that I am pretty sure.
Feel free to merge it if needed. I was closing it just want to show that I will stop taking ARs to let you know no need to spend time on giving me advices. Take it easy

nurmukhametov (approved):
LGTM!

Issue #3227, opened 2025-02-11
Explore Python to ISPC bindings


aneshlya:
Explore how to bind ISPC functions to Python, using [this discussion](https://github.com/ispc/ispc/discussions/2898) as a starting point.

Expected Outcomes:

- [ ] A tutorial explaining the process of binding ISPC functions to Python.
- [ ] A new test driver (replacing the current scripts/run_tests.py) that directly calls ISPC functions from Python. This should eliminate the need for the C++ test harness (main() in test_static.cpp).

sdkoppen:
I am inclined to use native ctypes bindings as @[iperov](https://github.com/iperov)
iperov suggested in the linked discussion. I've used ctypes bindings in other scripts and generally agree with reducing dependencies. I'm not sure if we want to worry about type info.

My suggestion:

* Write a ispc_bindings.py file that implements the ispc -> Python bindings. I haven't decided where this will live in the repo yet.
* Update the scripts/run_tests.py to use ispc_bindings.py.
* Delete test_static.cpp
* Cleanup all of the code that compiles test_static.cpp.

Thoughts from project owners?

aneshlya:
I recommend trying both approaches to fully evaluate what is the best. Here are some thoughts on the tradeoffs:

`nanobind` has certain advantages that might be particularly relevant for AI workflows:

1. DLPack integration - Nanobind enables zero-copy data exchange between ML frameworks.
2. Performance - usually nanobind calls are faster (at ~60ns vs ctypes at ~230ns). This difference becomes significant in AI applications with many function calls.

sdkoppen:
I have spent the day reading the discussion, reading through the documentation for the various referenced solutions ([ctypes](https://docs.python.org/3/library/ctypes.html), [ctypesgen](https://github.com/ctypesgen/ctypesgen), [nanobind](https://github.com/wjakob/nanobind), and [litgen](https://github.com/pthom/litgen), and testing the frameworks. Throughout the day, I kept finding myself confused on what the goal was, and was unsettled by it. The problem as described seemed straightforward: create Python bindings for C++ functions compiled by ispc. It seemed simple, but I couldn't keep the goal in focus.

[...]

PR #3271, opened 2025-03-05
 [WIP] ISPC ctypes PoC

sdkoppen:
For issue https://github.com/ispc/ispc/issues/3227

This is a basic working example of binding ispc functions with ctypes. It can be greatly expanded, but I wanted to start a PR to see if this is in the right direction.

I'm also exploring nanobind but haven't been able to get it fully working yet. I plan to work on the nanobind PoC for the rest of the Hackathon.

Example running it:

```
~> ./TestRunner.py --ispc_bin ../../../build-ispc/bin/ispc
Loading config file '/home/sean/Storage/Programming/hackathon/my_ispc2/tests/python/config.toml'
Running cmd '/home/sean/Storage/Programming/hackathon/build-ispc/bin/ispc --pic --woff /home/sean/Storage/Programming/hackathon/my_ispc2/tests/python/swizzle-3.ispc -o /home/sean/Storage/Programming/hackathon/my_ispc2/tests/python/build/swizzle-3.ispc.o --arch=x86-64 --target=sse4-i32x4 -O2 -h /home/sean/Storage/Programming/hackathon/my_ispc2/tests/python/build/swizzle-3.ispc.h'
Running cmd '/home/sean/Storage/Programming/hackathon/build-ispc/bin/ispc --pic --woff /home/sean/Storage/Programming/hackathon/my_ispc2/tests/python/gs-double-improve-progindex.ispc -o /home/sean/Storage/Programming/hackathon/my_ispc2/tests/python/build/gs-double-improve-progindex.ispc.o --arch=x86-64 --target=sse4-i32x4 -O2 -h /home/sean/Storage/Programming/hackathon/my_ispc2/tests/python/build/gs-double-improve-progindex.ispc.h'
Running cmd 'clang++ -O2 -m64 /home/sean/Storage/Programming/hackathon/my_ispc2/tests/python/top_bind.cpp /home/sean/Storage/Programming/hackathon/my_ispc2/tests/python/build/swizzle-3.ispc.o /home/sean/Storage/Programming/hackathon/my_ispc2/tests/python/build/gs-double-improve-progindex.ispc.o -o ./py_bound_test.so -fPIC -shared'
Loading library '/home/sean/Storage/Programming/hackathon/my_ispc2/tests/python/py_bound_test.so'
Running test 'swizzle3'
Test 'swizzle3' passed.
Running test 'gb_double_improve_progindex'
Test 'gb_double_improve_progindex' passed.
```

### DIFF
--- a/tests/python/README
+++ b/tests/python/README
@@ -1,0 +1,25 @@
+Issue 3227 (https://github.com/ispc/ispc/issues/3227) asked to explore Python bindings in ispc.
+
+This implementation uses the built-in ctypes module to show an example how to call ispc code
+from Python. Consider this a Proof Of Concept (PoC) with room for improvement.
+
+Design decisions:
+* Uniqify test function names in .ispc code to avoid name collision when generating the shared object.
+  This, along with removing the inlcude for test_static.cpp, necessitated creating a copy of the test
+  files. I don't like the redundnacy but I did not want to break functionality of existing tests. Future
+  efforts can clean this if needed.
+* One shared object is generated that contains all enabled tests. This reduces the number of dlls/
+  shared objects loaded during the Python run code. Per-test shared objects could be enabled if desired,
+  it just means a bit more Python code to generate the objects and load them.
+* Only tested on Linux on a single target. Will need more extensive testing and support for
+  other platforms. Mostly affects ispc/clang compile when generating the shared object.
+* Implement the TEST_SIG define from test_static.cpp through a toml configuration file.
+* Mirrored test_static.cpp as closely as possible in TestRunner.py.
+* top_bind.cpp implements functions that ispc needs for launch and should be linked into all generated
+  shared object files.
+
+Limitations:
+* It only works on Linux and one target.
+* You have to run TestRunner.py from tests/python or it will fail.
+* Adding new tests requires copying the test code, removing the test_static.cpp include, and uniqifying
+  function names
--- a/.gitignore
+++ b/.gitignore
@@ -19,4 +19,5 @@ examples/*/test
 *.swp
 check_isa.exe
 build*
+tests/python/*.so
 
--- a/examples/python_bindings/ctypes_example.txt
+++ b/examples/python_bindings/ctypes_example.txt
@@ -1,0 +1,43 @@
+This example demonstrates how to use ctypes to call a single ISPC function in Python.
+
+First, save this code to sample.ispc. This function takes in a pointer to an array of floats and modifies the array.
+
+task void f_v(uniform float RET[]) {
+    float<3> a = {1, 10, 100};
+    float<6> b = a.zxyyxz;
+
+    RET[0] = b.x + 2 * b.y + 3 * b.z - 3 * b[3] - 2 * b[4] - b[5];
+    RET[1] = b.x + 2 * b.y + 3 * b.z - 3 * b[3] - 2 * b[3] - b[4];
+}
+
+
+Next, we need to compile this function into a shared object named sample.so that ctypes can call into.
+To compile this function using ISPC, run this command, modifying the flags for your target architecture.
+
+ispc --pic --woff sample.ispc -o sample.ispc.o --arch=x86-64 --target=sse4-i32x4 -O2 -h sample.ispc.h
+clang++ -O2 -m64 sample.ispc.o -o ./sample.so -fPIC -shared
+
+
+Finally, use ctypes in Python to call the f_v function defined in sample.so:
+
+from ctypes import c_float, POINTER, CDLL
+
+def f_v_test():
+    # Load the shared object that contains the ispc function we want to call (f_v)
+    lib = CDLL("sample.so")
+
+    # Define the input argument types and the return type (None = void)
+    lib.f_v.argtypes = [POINTER(c_float)]
+    lib.f_v.restype = None
+
+    # Initialize an array 256 floats. Set the default value of each to -1e20
+    array_size = 256
+    returned_result = (c_float * array_size)(-1e20)
+    lib.f_v(returned_result)
+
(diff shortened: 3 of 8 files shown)


tests/python/README
```diff
@@ -0,0 +1,25 @@
...
+from Python. Consider this a Proof Of Concept (PoC) with room for improvement.
+
+Design decisions:
+* Uniqify test function names in .ispc code to avoid name collision when generating the shared object.
+  This, along with removing the inlcude for test_static.cpp, necessitated creating a copy of the test
+  files. I don't like the redundnacy but I did not want to break functionality of existing tests. Future
+  efforts can clean this if needed.
+* One shared object is generated that contains all enabled tests. This reduces the number of dlls/
+  shared objects loaded during the Python run code. Per-test shared objects could be enabled if desired,
+  it just means a bit more Python code to generate the objects and load them.
```
aneshlya, line 13: ISPC has > 1500 tests and I don't think we will be willing to describe each test in TestRunner.py. However ISPC uses only ~10 test signatures declared in https://github.com/ispc/ispc/blob/main/tests/test_static.isph. I think it makes sense to add support for all signatures from test_static.isph and generate a shared object for each test.
With this approach we would be able to run all the tests easily and you don't need to duplicate them.

aneshlya:
Basically it's right direction and it would be good PoC for the test driver (if the comment above is fixed). Even if it's not finished we'll count it for the hackathon.

**But before that** we'd like to see just two examples (ctypes, nanobind) demonstrating how to do bindings for our users.
Feel free to use any AI tool for this.
I've got [this](https://claude.site/artifacts/28bb785a-edf2-4b4c-810c-941db74ebbbe) for ctypes with Claude and it seems OK, you just need to check that it compiles and works as expected:
Here is a description for Readme file:
```
This example demonstrates how to use ctypes to call a single ISPC function that takes three different types of inputs:

Float arrays (the x and y coordinates of points)
A custom structure (the Transform struct with scaling, translation, and rotation parameters)
A constant value (the strength parameter that controls transformation intensity)

The example includes:

An ISPC implementation that applies geometric transformations to points:

Scaling on both x and y axes
Rotation around the origin
Translation with a configurable strength factor


A Python wrapper that:

Defines a matching ctypes Structure for the ISPC struct
Sets up the proper function signature
Handles conversion between NumPy arrays and C pointers
Passes the structure by reference using ctypes.byref()


A NumPy reference implementation to verify correctness and compare performance

To use this example:

Save the ISPC code to a file named point_transform.ispc
Compile it using the provided commands for your platform
Run the Python script to transform points and see the performance comparison

This pattern can be adapted for other scenarios where you need to pass complex data structures between Python and ISPC for high-performance computing tasks.
```
Feel free to put it in our examples folder (to python_bindings subfolder).

xiaohanIntel:
@aneshlya  Thanks for reviewing for this PR. I am now telling my teammates only to "open" a PR when they think it is the right time to ask for review.. but obviously this is still a poc without code cleaning ups.. should we convert that to "draft" instead? so that his work on this ticket can be silent without triggering the build or review requests?

aneshlya:
> @aneshlya Thanks for reviewing for this PR. I am now telling my teammates only to "open" a PR when they think it is the right time to ask for review.. but obviously this is still a poc without code cleaning ups.. should we convert that to "draft" instead? so that his work on this ticket can be silent without triggering the build or review requests?

That's OK. We'd like to check big changes like this at the start to ensure the right approach. You can add [WIP] label at the beginning of the PR's name.

sdkoppen:
> **But before that** we'd like to see just two examples (ctypes, nanobind) demonstrating how to do bindings for our users. Feel free to use any AI tool for this. I've got [this](https://claude.site/artifacts/28bb785a-edf2-4b4c-810c-941db74ebbbe) for ctypes with Claude and it seems OK, you just need to check that it compiles and works as expected: Here is a description for Readme file:
> 
> Feel free to put it in our examples folder (to python_bindings subfolder).

I can do this.

I spent most of Monday/Tuesday trying to enable nanobind. I realized I likely wouldn't have anything done by the end of the week if I stuck with it, so I moved to the ctypes example.

Two Nanobind hurtles:

1. Nanobind requires the user to write C++ code that defines the bound function to their custom wrapper functions. I was able to get one of the functions bound, but it had a float instead of a float pointer as a function signature, and I couldn't figure out how to make it a pointer. The nanobind framework is large and the documentation is lacking in some areas. My lack of experience with ispc and Nanobind meant experimenting with the framework alone would likely take the week.
2. Nanobind uses cmake to build its shared object. I'm unfamiliar with Cmake and how to integrate the nanobind commands into the ispc system.

tests/python/README
aneshlya: I’d expect the new test runner to be quite similar to `run_tests.py`, just without the `test_static.cpp` layer (something like top_bind.cpp should be fine). The issue with adding an extra config is that every new test would need to be manually added to the config, which won’t be practical until the process is fully automated and embedded in our practice. Also, handling a dozen test signatures feels much simpler to me than managing over a thousand tests. I see the advantages of your approach (and we may use it in a long term), but in the short term, I believe mirroring the `run_tests.py` logic is the better path forward.

sdkoppen:
I had no success getting nanobind to work. I pushed an example as requested for ctypes. I'm sorry I was not able to complete the entire ticket. Hopefully this provides a baseline for others to start from in future work.

aneshlya:
Thanks for your contribution. This is a great foundation for our future work on this topic.
We can't merge it in a current state but I'd like it to be linked with the issue so whoever next will work on it will see your changes.

aneshlya:
I'm closing this PR and open https://github.com/ispc/ispc/pull/3299 as continuation of the effort.

PR #3274, opened 2025-03-07
Refactor CompileAndOutput

nurmukhametov:
This PR primarily refactors the `Module::CompileAndOutput` function. This change splits the code into several smaller functions, extracting the logic of creating the dispatcher from the target module generation. It also aims to avoid copy-pasting and reduces the number of arguments passed across Module methods.

Two additional changes:
1. Manage memory allocated for targets and modules using smart pointers. This fixes a potential memory leak on early exits.
1. Set the actual target and corresponding module for dispatcher code generation. Previously, the default target for the most common ISA was used.


src/module.cpp
```diff
@@ -3849,6 +3849,37 @@ int lWriteMultiTargetHeaders(Module *module, Target *gTarget, DispatchHeaderInfo
     return 0;
 }
 
+int lFinilizeDispatchModule(Module *module, llvm::Module *dispatchModule, Arch arch, const char *cpu,
```
aneshlya, line 3852: lFinilizeDispatchModule -> lFinalizeDispatchModule

aneshlya:
In the commit message:
`Before, ISPC generated dispatcher for the default target for the most
common ISA, i.e., for --target=sse4-i32x8,avx2-i32x8 it generated
dispatcher code for sse4-i32x4.` do you mean `sse4-i32x4` or `sse4-i32x8`?

nurmukhametov:
> In the commit message: `Before, ISPC generated dispatcher for the default target for the most common ISA, i.e., for --target=sse4-i32x8,avx2-i32x8 it generated dispatcher code for sse4-i32x4.` do you mean `sse4-i32x4` or `sse4-i32x8`?

I mean `sse-i32x4`. The previous logic was as follows:
- Save all ISAs (SSE4, AVX2).
- Iterate over them to find the most common one (SSE4).
- Get the default target for this ISA using this function (sse4-i32x4): https://github.com/ispc/ispc/blob/58ed350034584f55f6c7647769cca8adff316f81/src/ispc.cpp#L2603
- Use it for the rest of the dispatcher generation.

aneshlya:
> I mean sse-i32x4

OK, I couldn't believe my eyes :) Interesting that no one created issue for that. Could you cover it by a test?

nurmukhametov:
> > I mean sse-i32x4
> 
> OK, I couldn't believe my eyes :) Interesting that no one created issue for that. Could you cover it by a test?

I am not entirely sure whether this should be possible at all. I don't understand why ISPC allows varying types to be used in extern function declarations. For example:

```ispc
extern "C" int same(int x) { return x; }
```

When compiled with `--target=sse4-i32x4,avx2-i32x8`, it produces the following:

```llvm-ir
# in the dispatcher module
declare <8 x i32> @foo_sse4(<8 x i32>)
declare <8 x i32> @foo_avx2(<8 x i32>)
define <8 x i32> @foo(<8 x i32>) {
...
  %ret = call <8 x i32> @foo_sse4(<8 x i32> %0)
...
# in sse4 module
define <4 x i32> @foo_sse4(<4 x i32> %x) {
# in avx2 module
define <8 x i32> @foo_avx2(<8 x i32> %x) {
```

aneshlya:
>  I don't understand why ISPC allows varying types to be used in extern function declarations. 

I think it generates warning in this case.

Could you check what was behavior with --target=sse4-i32**x8**,avx2-i32x8 before and now?

aneshlya:
Could you please explain how did you change the behavior with globals in multi-target compilation and also add it to PR description?

nurmukhametov:
> Could you please explain how did you change the behavior with globals in multi-target compilation and also add it to PR description?

I didn't change it; I preserved it. I only documented it.

aneshlya:
> No, it is not: https://ispc.godbolt.org/z/dGTEKjzsP

Ah, right. So the warning is only for global variables.

> Yes, I will check it.

I checked it and it seems the same.

aneshlya (approved):
LGTM!

Issue #3270, opened 2025-03-05
Add AVX10.2 support to ISPC


aneshlya:
AVX10 support is available starting LLVM 20 and can be enabled in ISPC.
There can be two vector width variants: 256 or 512 and CPUID flags can tell which one is available on particular device.

The proposal is to introduce new targets with the following names:
`avx10.2_256-x4`, `avx10.2_256-x8`, `avx10.2_256-x16`
`avx10.2_512-x4`, `avx10.2_512-x8`, `avx10.2_512-x16`

pbrubaker:
Can this be closed now @aneshlya?

PR #3277, opened 2025-03-07
Initial support of AVX10.2

aneshlya:
This PR adds definitions of AVX10.2 targets and processes CPUID flags for it.
AVX10 targets are based on generic targets.
Target-specific implementations and enabling in testing will be done in a future PR.
Should be merged after https://github.com/ispc/ispc/pull/3256.

Related to https://github.com/ispc/ispc/issues/3270


dbabokin (approved):
Looks good, but I would suggest also adding x32 and x64 targets right now, instead of postponing it for later time.

I have two reasons for that:
- they are useful for AI workloads with 16 and 8 nit wide data types.
- it's a simpler mental model for support matrix when it's just a square support matrix.

Downside is, of course, more testing.

dbabokin (approved):
Updated changes look good.

nurmukhametov (approved):
LGTM!

PR #3280, opened 2025-03-08
Enable CMake option to disable QUIET for superbuild's FetchContent

### LINKED ISSUE #3265: Add switch to disable `QUIET` on `FetchContent` in `CMakeLists.txt`
There are a few places in CMakeLists.txt where QUIET is specified for FetchContent .  It would be nice if there was a switch to disable that.  The first couple times I ran it I thought it was hung while pulling LLVM source.

Commenting those out was helpful to see what (if anything) was happening behind the scenes.

I can see this being particularly helpful for those with slow internet connections.

jmbuzano:
[bug:3265] Enable CMake option to disable QUIET for superbuild's FetchContent calls Adding an option to superbuild's CMake file to be able to turn ON and OFF the QUIET setting when calling FetchContent_Populate.

Fixes #3265


aneshlya (approved):
Looks good to me

PR #3288, opened 2025-03-17
Check ISA consolidation

### LINKED ISSUE #3284: Consolidate get_system_isa code
There are three copies of virtually the same code in the project:

- `builtin/dispatch.c:__get_system_isa` returns an int.
- `tools/check_isa.cpp:lGetSystemISA` returns an ISPCtarget.
- `src/ispc.cpp:lGetSystemISA` returns a string with the ISA.

We need to consider consolidating all of them in one place.

nurmukhametov:
This PR fixes #3284

### DIFF
--- a/src/isa.h
+++ b/src/isa.h
@@ -1,0 +1,272 @@
+/*
+  Copyright (c) 2025, Intel Corporation
+
+  SPDX-License-Identifier: BSD-3-Clause
+*/
+
+#ifndef ISA_H
+#define ISA_H
+
+enum ISA {
+    INVALID = -1,
+
+    SSE2 = 0,
+    SSE41 = 1,
+    SSE42 = 2,
+    AVX = 3,
+    AVX11 = 4,
+    AVX2 = 5,
+    AVX2VNNI = 6,
+    KNL_AVX512 = 7,
+    SKX_AVX512 = 8,
+    ICL_AVX512 = 9,
+    SPR_AVX512 = 10,
+    AVX10_2_512 = 11,
+
+    COUNT
+};
+
+// Define UNUSED_ATTR macro based on language standard and compiler support
+#if defined(__cplusplus) && __cplusplus >= 201703L
+// C++17 or newer
+#define UNUSED_ATTR [[maybe_unused]]
+#elif defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L
+// C23 or newer
+#define UNUSED_ATTR [[maybe_unused]]
+#elif defined(__clang__) || defined(__GNUC__)
+// Clang/GCC specific attribute
+#define UNUSED_ATTR __attribute__((unused))
+#else
+// No attribute support - define to nothing
+#define UNUSED_ATTR
+#endif
+
+#ifndef MACOS
+// MACOS macro can be defined when we are compiling dispatch.c for macOS.
+// In other cases, we need to define it manually if we are compiling for macOS.
+#if defined(__APPLE__)
+#define MACOS
+#endif
+#endif // !MACOS
+
+#if defined(__i386__) || defined(__x86_64__) || defined(_M_IX86) || defined(_M_X64)
+
+#if defined(_MSC_VER)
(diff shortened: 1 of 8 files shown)


aneshlya (commented):
I forgot to click `Submit review` yesterday :) The comments are the same as we discussed today on the call so I'm marking them as resolved.

src/ispc.cpp
```diff
@@ -260,160 +204,37 @@ static ISPCTarget lGetSystemISA() {
...
+    case Target::ISA::SKX_AVX512:
+        return ISPCTarget::avx512skx_x16;
+    case Target::ISA::ICL_AVX512:
+        return ISPCTarget::avx512icl_x16;
+    case Target::ISA::SPR_AVX512:
+        return ISPCTarget::avx512spr_x16;
+    case Target::ISA::AVX10_2_512:
+        return ISPCTarget::avx10_2_512_x16;
+    default:
+        Error(SourcePos(), "Detected unsupported KNL ISA. Exiting.");
```
aneshlya, line 234: Default error message should be updated.

tools/check_isa.cpp
```diff
@@ -10,246 +10,63 @@
...
-            // So far, so good.  AVX2?
-            if (avx2) {
-                return "AVX2 (codename Haswell)";
-            } else {
-                // Ivy Bridge specific target was deprecated in ISPC, but
-                // no harm detecting it in standalone tool.
-                return "AVX1.1 (codename Ivy Bridge)";
-            }
+    enum ISA isa_id = get_x86_isa();
+    const char *isa = isa_strings[isa_id];
```
aneshlya, line 61: If `get_x86_isa()` returns INVALID (-1), this will cause an array index out of bounds error.

src/isa.h
```diff
@@ -0,0 +1,256 @@
...
+    int rEAX = 0, rEDX = 0;
+    __asm__ __volatile__(".byte 0x0f, 0x01, 0xd0" : "=a"(rEAX), "=d"(rEDX) : "c"(0));
+    return rEAX;
+#endif // !defined(_MSC_VER)
+}
+
+static int __os_has_avx_support() { return (xgetbv() & 6) == 6; }
+
+static int __os_has_avx512_support() {
+#if !defined(MACOS)
```
aneshlya, line 65: `MACOS` is passed as compiler flag when we compile dispatch.c. But when isa.h is included into ispc.cpp should we use `ISPC_HOST_IS_APPLE`? I'd double check other macros too.

src/ispc.cpp
nurmukhametov: Done

tools/check_isa.cpp
nurmukhametov: Yes, indeed. Fixed

src/isa.h
nurmukhametov: > `MACOS` is passed as compiler flag when we compile dispatch.c. But when isa.h is included into ispc.cpp should we use `ISPC_HOST_IS_APPLE`? I'd double check other macros too.

Right, I have decided not to rely on `ISPC_HOST_IS_APPLE` here, but to check `__APPLE__` directly, as they are virtually the same.

aneshlya (approved):
Looks good!. I'd like @dbabokin to review it as well.

src/isa.h
```diff
@@ -0,0 +1,272 @@
+/*
+  Copyright (c) 2025, Intel Corporation
+
+  SPDX-License-Identifier: BSD-3-Clause
+*/
+
+#ifndef ISA_H
```
dbabokin, line 7: What about `#pragma once`?

src/isa.h
```diff
@@ -0,0 +1,272 @@
...
+#define ISA_H
+
+enum ISA {
```
dbabokin, line 10: Do we want it as `enum class`?

dbabokin (approved):
Looks good overall.

src/isa.h
nurmukhametov: It is not part of the C or C++ standard.

Issue #1417, opened 2019-02-02
Reductions for int8/int16


JeffRous:
int8 and int16 should have reduce_max, reduce_min and reduce_equal available for signed and unsigned types. This could be used to do reductions on pixel channels of a packed buffer. Inputs should match outputs for max and min.

Ex:
uniform int8 reduce_min(int8 a)
uniform int16 reduce_min(int16 a)
uniform int8 reduce_max(int8 a)
uniform int16 reduce_max(int16 a)
uniform bool reduce_equal(int8 a)
uniform bool reduce_equal(int16 a)
uniform bool reduce_equal(int8 v, uniform int8 * uniform sameval)
uniform bool reduce_equal(int16 v, uniform int16 * uniform sameval)

PR #3289, opened 2025-03-17
Add missing stdlib reduction functions for int8/int16 types

aneshlya:
Fixes [#1417.](https://github.com/ispc/ispc/issues/1417)

This PR adds missing `reduce_min/reduce_max` and `reduce_equal` stdlib functions for signed and unsigned int8/int16 types.
The generic implementation for `reduce_min/reduce_max` produces the best results utilizing `vphminposuw` instruction on avx512 so it doesn't require target-specific implementation.

Additionally the PR adds missing `exclusive_scan` functions for int8/int16 types for completeness.
Now all cross-operation functions has variants for int8/int16 types.


stdlib/stdlib.ispc
```diff
@@ -1044,10 +1044,37 @@ __declspec(safe, cost1) static inline uniform double select(uniform bool cond, u
 
 __declspec(safe) static inline uniform int16 reduce_add(int8 x) { return __reduce_add_int8(__mask ? x : (int8)0); }
 
+__declspec(safe) static inline uniform int8 reduce_min(int8 v) {
+    // Set values for non-running lanes to the maximum int8 value so
+    // they don't affect the result.
+    int8 int8_max = 0x7F;
+    return __reduce_min_int8(__mask ? v : int8_max);
```
nurmukhametov, line 1051: I think we'd better use `INT8_MAX` here and in similar places, it should be defined for `stdlib.ispc` from here https://github.com/ispc/ispc/blob/ee408969519b859d065c27fca9adcc9b1e5cd4f8/stdlib/include/core.isph#L160

aneshlya: Updated

nurmukhametov (approved):
LGTM!

Issue #3258, opened 2025-03-02
FATAL ERROR: Unhandled signal sent to process; terminating.


illua1:
Code bellow cause an compiler internal error:
```C
struct float3
{
    float data[3];
};

export uniform float3 float3_dot_product(const uniform float values[][3],
                                         const uniform float factors[],
                                         const uniform int count)
{
  uniform float3 accumulator;
  accumulator.data[0] = 0.0f;
  accumulator.data[1] = 0.0f;
  accumulator.data[2] = 0.0f;
  foreach (i = 0 ... count) {
    const uniform float value[3] = values[i];
    const uniform float factor = factors[i];
    accumulator.data[0] += value[0] * factor;
    accumulator.data[1] += value[1] * factor;
    accumulator.data[2] += value[2] * factor;
  }
  return accumulator;
}
```

Used ispc version:
```
Intel(r) Implicit SPMD Program Compiler (Intel(r) ISPC), 1.26.0 (build commit af616443634432c6 @ 20250207, LLVM 18.1.8)
Supported Visual Studio versions: Visual Studio 2015 and later.
```

Compilation command:
`ispc --target=sse4.2-i32x8 "fmm_math.ispc" -o "fast_math_ispc.obj" --header-outfile="fast_math.h" --addressing=64`

nurmukhametov:
A crash should not have happened here, but it seems to me that you are thinking of `i` incorrectly because `i` is a `varying` variable. Reading from a uniform array `factors` with a `varying` index `i` results in a `varying float` value, not a `uniform float` one.

PR #3291, opened 2025-03-18
Report an error when casting to unsized array

### LINKED ISSUE #1234: Compiler crash when casting to unsized array
Using release 1.9.1 on Windows x64.

```
void test1()
{
    float values[4];
    (float [])values;
}
```

Compiling this gives

```
Invalid bitcast
  %values_load_bitcast = bitcast [4 x <8 x float>] %values_load to [0 x <8 x float>]
LLVM ERROR: Broken function found, compilation aborted!
main.cpp(235): FATAL ERROR: Unhandled signal sent to process; terminating.
***
***
```

Additionally doing instead

```
void test2()
{
    float values[4];
    (float [])values + 1;
}
```

gives

```
Error: Illegal to take address of non-lvalue or function.
    (float [])values + 1;
              ^^^^^^

expr.cpp:168: Assertion failed: "addr != NULL".
***
***
```

aneshlya:
This PR fixes a crash on malformed code.
Fixes #1234 and #3258.

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -457,6 +457,11 @@ static bool lDoTypeConv(const Type *fromType, const Type *toType, Expr **expr, b
             // the case of different element counts should have returned
             // successfully earlier, yes??
             AssertPos(pos, toArrayType->GetElementCount() != fromArrayType->GetElementCount());
+            if (toArrayType->IsUnsized() && !failureOk) {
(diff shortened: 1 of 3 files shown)


nurmukhametov (approved):
LGTM!

aneshlya:
The fix is also resolves crash from #3258 so I'm adding lit test for this.

PR #3292, opened 2025-03-19
Fix warnings

nurmukhametov:
This PR fixes warnings produced by static analysis tools after #3274 and #3283


aneshlya (approved):
Looks good to me!

Issue #3253, opened 2025-02-27
Template Instantiation on Function Pointer Assignment


nurmukhametov:
Consider the following code:
```ispc
template <typename T> T same(T x) { return x; }

typedef float (*FuncType)(float);

float foo(float a) {
    uniform FuncType func = same;
    return func(a);
}
```

It fails to compiler with the following error:
```bash
<source>:6:33: Error: syntax error, unexpected ';', expecting '('. 
    uniform FuncType func = same;
                                ^
```

Compiler explorer link: https://ispc.godbolt.org/z/xzT9oo3oe

It seems that the instantiation of templates for such a case is not supported.

PR #3293, opened 2025-03-19
Support instantiation of template function on pointer assignment

nurmukhametov:
This implements https://github.com/ispc/ispc/issues/3253

### DIFF
--- a/tests/lit-tests/ptr-assign-inst.ispc
+++ b/tests/lit-tests/ptr-assign-inst.ispc
@@ -1,0 +1,42 @@
+// This test checks that we can properly instantiate a template function at
+// the moment when we assign its address to a function pointer.
+
+// RUN: %{ispc} --target=host --nostdlib %s --emit-llvm-text -o - 2>&1 | FileCheck %s
+
+float same_exp(float x);
+uniform int same_exp(uniform int x);
+uniform double same_exp(uniform double x);
+
+typedef float (*FuncType)(float);
+typedef uniform int (*FuncType2)(uniform int);
+
+template <typename T> T same(T x) { return same_exp(x); }
+
+// CHECK-LABEL: @foo___vyf
+// CHECK-DAG: call <{{.*}}> @same_exp___vyf({{.*}})
+float foo(float a) {
+    uniform FuncType func_ptr = same;
+    return func_ptr(a);
+}
+
+// CHECK-LABEL: @foo2___uni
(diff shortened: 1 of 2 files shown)


tests/lit-tests/ptr-assign-inst.ispc
```diff
@@ -0,0 +1,26 @@
...
+
+typedef float (*FuncType)(float);
+typedef uniform int (*FuncType2)(uniform int);
+
+template <typename T> T same(T x) { return same_exp(x); }
+
+// CHECK-LABEL: @foo___vyf
+// CHECK-DAG: call <{{.*}}> @same_exp___vyf({{.*}})
+float foo(float a) {
+    uniform FuncType func_ptr = same;
```
aneshlya, line 17: Looks like the intention here is to check `varying float` case so `uniform` keyword is not needed?

tests/lit-tests/ptr-assign-inst.ispc
```diff
@@ -0,0 +1,26 @@
+// This test checks that we can properply instantiate a template function at
+// the moment when we assign its address to a function pointer.
+
+// RUN: %{ispc} --target=host --nostdlib %s --emit-llvm-text -o - 2>&1 | FileCheck %s
+
+float same_exp(float x);
+uniform int same_exp(uniform int x);
+
+typedef float (*FuncType)(float);
+typedef uniform int (*FuncType2)(uniform int);
```
aneshlya, line 10: So looks like you're adding support only for one case and the code below would not compile?
```
typedef uniform double (FuncType3)(uniform double);
uniform double foo3(uniform double a) {
    uniform* FuncType3 func_ptr = same;
    return func_ptr(a);
}

tests/lit-tests/ptr-assign-inst.ispc
nurmukhametov: This case tests instantiation with a varying float argument, creating a uniform pointer to this instantiation. Although, I have also added a test with a varying function pointer.

tests/lit-tests/ptr-assign-inst.ispc
nurmukhametov: The syntax is incorrect. For example, consider `const *int x`. Therefore, it should be `uniform FuncType3 *uniform func_ptr`, and it works as demonstrated by the test case I have added.

aneshlya (approved):
Looks good!

PR #3303, opened 2025-03-31
Fix static analysis warnings

nurmukhametov:
This PR fixes a few warnings that were introduced in #3298 and few others.

### DIFF
--- a/src/type.cpp
+++ b/src/type.cpp
@@ -2516,8 +2516,9 @@ const ReferenceType *ReferenceType::GetAsConstType() const {
     }
 
     if (asOtherConstType == nullptr) {
-        asOtherConstType = createWithBaseType(targetType->GetAsConstType());
-        CastType<ReferenceType>(asOtherConstType)->asOtherConstType = this;
+        const ReferenceType *t = createWithBaseType(targetType->GetAsConstType());
+        t->asOtherConstType = this;
+        asOtherConstType = t;
     }
     return CastType<ReferenceType>(asOtherConstType);
 }
@@ -2532,8 +2533,9 @@ const ReferenceType *ReferenceType::GetAsNonConstType() const {
     }
 
     if (asOtherConstType == nullptr) {
-        asOtherConstType = createWithBaseType(targetType->GetAsNonConstType());
-        CastType<ReferenceType>(asOtherConstType)->asOtherConstType = this;
+        const ReferenceType *t = createWithBaseType(targetType->GetAsNonConstType());
+        t->asOtherConstType = this;
+        asOtherConstType = t;
     }
     return CastType<ReferenceType>(asOtherConstType);
 }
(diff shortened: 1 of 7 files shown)


src/type.cpp
```diff
@@ -2505,7 +2505,7 @@ const ReferenceType *ReferenceType::GetAsConstType() const {
 
     if (asOtherConstType == nullptr) {
         asOtherConstType = createWithBaseType(targetType->GetAsConstType());
-        CastType<ReferenceType>(asOtherConstType)->asOtherConstType = this;
+        static_cast<const ReferenceType *>(asOtherConstType)->asOtherConstType = this;
```
aneshlya, line 2508: I don't quite understand this fix. 
`static_cast` can lead to undefined behavior if the object isn't actually of type `ReferenceType`. 
Maybe just check that it's not null?
```
    const ReferenceType* refType = CastType<ReferenceType>(asOtherConstType);
    if (refType != nullptr) {
        refType->asOtherConstType = this;
    }
```

aneshlya: Also I didn't notice it before but after refactoring in type.cpp there are many places when `static_cast` is used. For consistency and clarity, I think we should stick to using `CastType` (or remove it completely?). This would eliminate confusion about when to use `static_cast` versus `CastType`. Let me know what do you think.

src/type.cpp
```diff
@@ -2519,8 +2520,8 @@ const ReferenceType *ReferenceType::GetAsNonConstType() const {
     }
 
     if (asOtherConstType == nullptr) {
-        asOtherConstType = createWithBaseType(targetType->GetAsNonConstType());
-        CastType<ReferenceType>(asOtherConstType)->asOtherConstType = this;
+        asOtherConstType = createWithBaseType(targetType->GetAsConstType());
```
aneshlya, line 2523: GetAsNonConstType()?

src/type.cpp
nurmukhametov: > I don't quite understand this fix. `static_cast` can lead to undefined behavior if the object isn't actually of type `ReferenceType`.

But this object is definitely of type `ReferenceType`. I have rewritten it to make it clear.

nurmukhametov: > Also I didn't notice it before but after refactoring in type.cpp there are many places when `static_cast` is used. For consistency and clarity, I think we should stick to using `CastType` (or remove it completely?). This would eliminate confusion about when to use `static_cast` versus `CastType`. Let me know what do you think.

`CastType` and `static_cast` are used in different scenarios. It is acceptable to use `static_cast` when we are certain that a pointer is of a specific type. This occurs, for instance, with `createWith...` methods that return a base type; if we call them from `AtomicType` methods, we know for sure that we have constructed `AtomicType` objects although 
 `createWithAilgnment` returns `Type*`.  `CastType`, on the other hand, is a tool for verifying that a `Type` pointer is indeed of a particular derived type when it could be anything

aneshlya (approved):
LGTM!

src/type.cpp
aneshlya: Looks very clear now and thanks for explanation!

PR #3304, opened 2025-03-31
Deduplicate avx512 builtins

nurmukhametov:
Remove duplication of built-in function from avx512 targets. After enabling hierarchical linking for all x86 targets, we can rely on avx512icl to provide all the necessary built-ins for avx512spr, and avx512skx for avx512icl.

I am also unsure whether we need to remove the `avx512-common*.ll` files or not. They are no longer shared between multiple files; they are now only included in avx512skx-xN.ll.


aneshlya:
Could you add a link to a full run with avx512 targets?

aneshlya (approved):

PR #3305, opened 2025-03-31
Reusable workflow for ISPC build

aneshlya:
This PR introduces reusable workflow for ISPC build on different platforms similar to reusable.ispc.test introduced in https://github.com/ispc/ispc/pull/3250

Full run is here: https://github.com/ispc/ispc/actions/runs/14181517913


nurmukhametov (approved):
LGTM!

PR #3306, opened 2025-03-31
Fix constructor of clang::HeaderSearch for LLVM trunk

aneshlya:
A small fix for ISPC build with LLVM trunk: https://github.com/ispc/ispc/actions/runs/14165935006/job/39690190700


nurmukhametov (approved):
LGTM!

PR #3308, opened 2025-04-01
Update osPresets

aneshlya:
1. Update L0 version to avoid problems like this: https://github.com/ispc/ispc/actions/runs/14202438838/job/39792275414#step:4:107
2. Remove `SPIRV_TRANSLATOR_BRANCH` - it's not used anywhere, I'll update internal scripts accordingly.


nurmukhametov (approved):
LGTM!

PR #3309, opened 2025-04-02
Improve the performance of masked_load and masked_store for AVX-512 targets (32- and 64-wide).

nurmukhametov:
This PR drops the implementation of maked stores and loads from `util.m4` (LLVM IR generic ones) in favor of generic ones from based on LLVM intrinsics from `generic.ispc` for targets `avx512*-x32` and `avx512*-x64`.

It improves performance of `08_masked_load_store` on order of magnitude, i.e., around ~10x times.

This PR changesthe implementation of masked stores and loads from `util.m4` (the generic LLVM IR versions) in favor of the generic versions based on LLVM intrinsics from `generic.ispc` for the targets avx512*-x32 and avx512*-x64.

This change improves the performance of `08_masked_load_store` by an order of magnitude (approximately ~10x).

### DIFF
--- a/builtins/target-avx512skx-x32.ll
+++ b/builtins/target-avx512skx-x32.ll
@@ -1,4 +1,4 @@
-;;  Copyright (c) 2020-2024, Intel Corporation
+;;  Copyright (c) 2020-2025, Intel Corporation
 ;;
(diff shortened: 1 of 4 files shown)


aneshlya (approved):
Great perf result! Looks good!

PR #3311, opened 2025-04-02
Updates to AVX10.2 target

### LINKED ISSUE #3294: Remove references to AVX10.2 256-bit
Remove references to AVX10.2 256-bit since it was removed from the spec: https://cdrdv2.intel.com/v1/dl/getContent/784343

Now we can follow the naming scheme started for avx512 targets:
`avx10.2dmr-x[WIDTH]`

aneshlya:
Fixes #3294.
1. Remove references to AVX10.2 256-bit and rename targets appropriately. We don't need to have cpu name in the target name anymore (according to spec and I hope it'll not change).
2. Add avx10.2 testing to CI

> A developer can expect that Intel AVX10 Version N+1 will include all the features and capabilities included in Version N. With the stated goal of minimizing developer impact, a new version of the Intel AVX10 ISA can be expected to include a significant suite of
new instructions and capabilities, delivering sufficient additional value to justify the associated software enablement effort. I


aneshlya:
Full nightly run is here: https://github.com/ispc/ispc/actions/runs/14231318872
We can't enable testing with LLVM 20 yet, we should update it to version 20.1.2 first in CI to include the latest fixes.

aneshlya:
< I think that the preprocessor section in docs needs to be updated with avx10.2 target macro here

Not sure we need to do it now. When ISPC built with LLVM < 20, there is no preprocessor definition and no avx10.2 targets available. I think it makes sense to add when we build officially with LLVM 20. Plus update vim.contrib.

< Do we need to resolve it in this PR?

Yes, last commit ensures that avx10.2 are not available to user before LLVM 20.

nurmukhametov:
> > I think that the preprocessor section in docs needs to be updated with avx10.2 target macro here
> 
> Not sure we need to do it now. When ISPC built with LLVM < 20, there is no preprocessor definition and no avx10.2 targets available. I think it makes sense to add when we build officially with LLVM 20. Plus update vim.contrib.

Ok.

nurmukhametov (approved):
LGTM!

PR #3312, opened 2025-04-02
Update LLVM 20 version to 20.1.2

aneshlya:
LLVM update required to enable avx10.2 targets testing with LLVM 20.


nurmukhametov (approved):
LGTM!

PR #3319, opened 2025-04-08
Enable AVX10.2 targets in testing with LLVM 20

aneshlya:
Full run is here: https://github.com/ispc/ispc/actions/runs/14345758419


nurmukhametov (approved):
LGTM!

PR #3320, opened 2025-04-09
Use PointerType::get with LLVMContext over Type

aneshlya:
This PR switches from typed pointers to more generic pointer types. This corresponds to the latest changes in LLVM trunk.


nurmukhametov (approved):
LGTM!

Issue #2292, opened 2022-04-05
Varying operation in masked control flow


Yakokse:
When calling the program below I expected the division to only apply to active program instances, however it seems to be calculated for all of them, and therefore results in a floating point exception (SIGFPE), since the other program instances have the divisor 0. 

```c
export void test(){
  varying int x = 0;
  varying int y = 0;
  foreach_active(i){
    if(i==0){
      x = 1;
      y = 1;
      print("%\n", x/y);
    }
  }
}
```

Expected output: 1
Actual output: floating point exception

dbabokin:
Related issues: #2169, #708.

It was not fixed, as the fix hurts performance. But I think we should go forward with the fix.

Hugobros3:
This affects us and this is not a satisfactory resolution. Could the slow-path be at least an option ?

nurmukhametov:
Fixed by #3322

Issue #2169, opened 2021-09-27
Integer division crashes


stotko:
The following code (simplified version of Open3D's tensor division op introduced [here](https://github.com/isl-org/Open3D/pull/4107)) crashes for `int32_t`, `uint32_t`, `int64_t`, `uint64_t` on AVX2 (`avx2-i32x8`), but works for all other types.

```
/* Div.ispc */

struct Array {
    void* data_ptr_;
    int64 byte_size_;
};

struct State {
    Array input0_;
    Array input1_;
    Array output_;
};

static inline uint8* GetDataPtr(const uniform Array* const uniform array,
                                int64 workload_idx) {
    return (uint8*)(array->data_ptr_) + workload_idx * array->byte_size_;
}

#define TEMPLATE(T)                                                     \
    static inline void DoDiv(int64 idx, uniform State* uniform state) { \
        const T* lhs = (const T*)GetDataPtr(&(state->input0_), idx);    \
        const T* rhs = (const T*)GetDataPtr(&(state->input1_), idx);    \
        T* dst = (T*)GetDataPtr(&(state->output_), idx);                \
        *dst = *lhs / *rhs;                                             \
    }

#pragma ignore warning(perf)
TEMPLATE(int32)

#undef TEMPLATE

export void DoDivKernel(uniform int64 N, uniform State* uniform state) {
    foreach (i = 0...N) { DoDiv(i, state); }
}
```
```
/* Div.cpp */

#include <vector>
#include "Div_ispc.h"

int main() {
    using T = int32_t;

    std::vector<T> a = {4, 6, 8, 10, 12, 14};
    std::vector<T> b = {2, 3, 4, 5, 6, 7};

    ispc::State state;

    state.input0_.data_ptr_ = a.data();
    state.input0_.byte_size_ = sizeof(T);

    state.input1_.data_ptr_ = b.data();
    state.input1_.byte_size_ = sizeof(T);

    state.output_.data_ptr_ = a.data();
    state.output_.byte_size_ = sizeof(T);

    ispc::DoDivKernel(a.size(), &state);

    if (a != std::vector<T>({2, 2, 2, 2, 2, 2})) {
        return 1;
    }
    
    return 0;
}
```

If `T` is set to either of the affected types, a `floating point exception` error is raised. All other types, including `int16_t`, etc., work fine. Doing some quick checks, other ISAs like SSE4 are likely also affected. But I have not exhaustively tested all of them.

[...]

Issue #708, opened 2014-01-06
Bad generated code for integer div (crash)


ccoustet:
The emitted code does 4 (or 8, or 16) idiv, as for an entire gang, regardless of the number of program instances.

The following chunk of code illustrates the problem:

```c
export void test(uniform int nb)
{
    foreach(i = 0 ... nb)
    {
        // foreach (j = nb+1 ... 1)
        float j = nb + 1 - i;
        float k;
        if (j > 0)
            k = nb / j;
        else
            k = -1;
        print("%\n", k);
    }
    foreach(i = 0 ... nb)
    {
        // foreach (j = nb+1 ... 1)
        int j = nb + 1 - i;
        int k;
        if (j > 0)
            k = nb / j;
        else
            k = -1;
        print("%\n", k);
    }
}
```

test(5) results in (win7 x86, VS2013 Debug):
[0.833333,1.000000,1.250000,1.666667]
[2.500000,((0.000000)),((0.000000)),((0.000000))]
[0,1,1,1]
<exception: divide by zero>

nurmukhametov:
Fixed by #3322

PR #3322, opened 2025-04-09
Perform DIV and MOD binary operators only for active lanes

### LINKED ISSUE #1415: Integer divide crash with AVX
When compiling with AVX or AVX2 targets, the following crashes with integer divide by zero. SSE2 and SSE4 targets work as expected.

Compiled with 1.10.0 on Windows using -O2 and --target=avx or --target=avx

C++:
```
#include <cstdio>
#include "ISPCDivTest_ispc.h"

int main()
{
	const int Iterations = 16;
	int Dst[Iterations] = { 0 };
	int Src0[Iterations] = { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15 };
	int Src1[Iterations] = { 15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0 };

	ispc::Divi(Dst, Src0, Src1, Iterations);

	for (int i = 0; i < Iterations; i++)
	{
		printf("Dst[%d] = %d\n", i, Dst[i]);
	}
}
```

ISPC:
```
export void Divi(uniform int Dst[], uniform int Src0[], uniform int Src1[], uniform int Count)
{
	foreach(i = 0 ... Count)
	{
		bool bDivByZero = Src1[i] == 0;
		Dst[i] = bDivByZero ? 0 : Src0[i] / Src1[i];
	}
}
```

nurmukhametov:
This PR fixes #1415, #708, #2169, #2292.


dbabokin:
Perf?

nurmukhametov:
> Perf?

I haven't noticed a measurable impact. In cases where not all lanes are active, it should be faster than before.

src/ctx.cpp
```diff
@@ -1668,6 +1767,13 @@ llvm::Value *FunctionEmitContext::BinaryOperator(llvm::Instruction::BinaryOps in
     int arraySize = lArrayVectorWidth(type);
 
     if (arraySize == 0) {
+        if (ispcType && ispcType->IsVaryingType() &&
+            (inst == llvm::Instruction::SDiv || inst == llvm::Instruction::UDiv || inst == llvm::Instruction::SRem ||
+             inst == llvm::Instruction::URem)) {
+            // TODO!: make sure we don't call it for uniform vectors
```
aneshlya, line 1773: Do you want to address this TODO in this PR?

src/ctx.h
```diff
@@ -408,9 +409,20 @@ class FunctionEmitContext {
         See the Expressions section of docs/ispc.rst for more information on this
         optimization.
     */
-    llvm::Value *BinaryOperator(llvm::Instruction::BinaryOps inst, llvm::Value *v0, llvm::Value *v1,
+    llvm::Value *BinaryOperator(llvm::Instruction::BinaryOps inst, llvm::Value *v0, llvm::Value *v1, const Type *type,
```
aneshlya, line 412: Seems like you only need information about the variability of type for division operators, right?
Maybe you can just check `llvm::isa<llvm::VectorType>(v0->getType()) ` inside `BinaryOperator`?

The changes in this PR regarding the usage of the function are inconsistent. In some cases, a `nullptr` is passed (which makes sense when the type isn't related to division). In others, a non-null type is used even when it's unrelated to division—why include it there at all? And in some places, the code passes `type0` and then `type1`, even though they should be the same for binary operators.

src/ctx.cpp
```diff
@@ -1656,8 +1639,124 @@ static int lArrayVectorWidth(llvm::Type *t) {
...
+// it. We need to perform division only for the active lanes. It is done by
+// calling builtin __idiv function. If all lanes are active, we can still use
+// the instction to divide the entire vector, which can be useful for
+// optimization when we divide compile time constants, e.g, programIndex
+// similar to the tests/lit-tests/2468.ispc.
+llvm::Value *FunctionEmitContext::VectorIntDivision(llvm::Instruction::BinaryOps inst, llvm::Value *v0, llvm::Value *v1,
+                                                    const llvm::Twine &name) {
+    // When handling vector division, we need to create a branching structure:
+    // If fullmask is all ON, use direct SDiv instruction
+    // Otherwise, use the __idiv intrinsic which handles masked division
```
aneshlya, line 1706: Could you remind me if we perform reciprocal optimization somewhere later? Like where the divisor has the same value across all vector lanes, we can just compute rcp and multiply.

tests/lit-tests/1415.ispc
```diff
@@ -0,0 +1,53 @@
+// This test checks that generated code does not try to delete to zero.
```
aneshlya, line 1: divide by zero?

benchmarks/02_medium/test03_int_div.cpp
```diff
@@ -0,0 +1,208 @@
...
+WARM_UP_RUN();
+
+// Minimum size is maximum target width * 4, i.e. 64*4 = 256.
+// 256 * sizeof (int) = 1kb - expected to reside in L1
+// 256 * sizeof (int) << 4 = 16kb - expected to reside in L1
+// 256 * sizeof (int) << 7 = 128kb - expected to reside in L2
+// 256 * sizeof (int) << 12 = 4 Mb - expected to reside in L3.
+#define ARGS Arg(256)->Arg(256 << 4)->Arg(256 << 7)->Arg(256 << 12)
+
+const float eps = 0.00001f;
```
aneshlya, line 24: This is not used anywhere

src/ctx.h
nurmukhametov: > Seems like you only need information about the variability of type for division operators, right? Maybe you can just check `llvm::isa<llvm::VectorType>(v0->getType()) ` inside `BinaryOperator`?

Yes, I only need variability, but the suggested check gives incorrect answers for uniform short vectors because they are also represented as vectors in LLVM IR. This is why I need to pass around the proper ISPC type to check variability.

> The changes in this PR regarding the usage of the function are inconsistent. In some cases, a `nullptr` is passed (which makes sense when the type isn't related to division). In others, a non-null type is used even when it's unrelated to division—why include it there at all? And in some places, the code passes `type0` and then `type1`, even though they should be the same for binary operators.

In many cases, it is unclear to me what the ISPC type of the operands is. Moreover, in the current state of this function, the type parameter is used only for SDiv, UDiv, SRem, and URem operators. For any other operators, such as the frequent calls for And, Xor, and Or, it is not used, so passing a nullptr seems acceptable. I understand that this approach is somewhat unusual, but I haven't found a better solution to provide variability for the subset of required operators.

src/ctx.cpp
nurmukhametov: I have fixed the case when a divisor is a constant, and added the corresponding test `1415-2.ispc`, and benchmarks (fastdiv_mask).

src/ctx.h
nurmukhametov: I have added a comment regarding this to the function declaration.

tests/lit-tests/1415.ispc
aneshlya: There is still a "delete to zero." in a bunch of lit tests.

aneshlya:
Looks good!

illua1:
Wonder does this is by default on all platforms or only for whose support masked execution?

tests/lit-tests/1415.ispc
nurmukhametov: Indeed, fixed.

aneshlya (approved):

PR #3324, opened 2025-04-09
Fix skipping tests and fix hardening check workflow

aneshlya:
- Fix hardening check workflow: https://github.com/ispc/ispc/actions/runs/14367775288/job/40284675177
- Reorder skipping rules in some tests to skip them on GPUs.


nurmukhametov (approved):
LGTM!

PR #3327, opened 2025-04-14
Fix dependencies for subprojects (examples/benchmarks)

nurmukhametov:
Fix missing dependencies in ISPC components for builds with ISPC_SLIM_BINARY=ON. This issue could lead to build failures when using Ninja.


aneshlya (approved):
LGTM!

Issue #2583, opened 2023-07-24
conversion of uniform float[] to int8* to float* results in compiler crash, ispc version 1.20


karna48:
I have tried to replicate bahaviour of cudaMallocPitch and access the aligned columns:
```
task void sum_M_float_task(
    uniform float v_out[], 
    uniform float M[],
    uniform int M_pitch,
    uniform int N, 
    uniform int chunk_size)
{
    uniform int col = taskIndex0 * chunk_size;

    uniform float *v_out_chunked = new float [chunk_size];

    foreach(row = 0 ... N) {
        varying float m = ((uniform float *)((uniform int8 *)M + col * M_pitch))[row];
```

M_pitch being the aligned/padded size of the column of a 2D matrix.

The compiler crashed:
```
../src/test_first_try.ispc:42:28: Error: Assertion failed (/usr/local/src/ispc/src/expr.cpp:4522): "baseSym != nullptr". 
        varying float m = (((uniform float *)((uniform int8 *)M + col * M_pitch))[row]);
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

***
*** Please file a bug report at https://github.com/ispc/ispc/issues
*** (Including as much information as you can about how to reproduce this error).
*** You have apparently encountered a bug in the compiler that we'd like to fix!
***
/usr/local/src/ispc/src/main.cpp(385): FATAL ERROR: Unhandled signal sent to process; terminating.
Aborted (core dumped)
make: *** [Makefile:27: test_first_try.ispc.o] Error 134
```

aneshlya:
Hi @karna48, thanks for reporting this issue. Compiler should not crash in this case. 
Seems that splitting the expression helps:
https://ispc.godbolt.org/z/vj5Y7Eren

BTW do you expect to have uniform pointer or varying?
```uniform int8 *``` is a varying pointer to uniform int8, ```uniform float *``` is a varying pointer to uniform float.
If you need a uniform pointer, please specify it explicitly:
https://ispc.godbolt.org/z/c1cMYshKP
On this example you can notice the difference with ```-O0```.

aneshlya:
> Still the compiler frontend should be able to handle a 2 pointer casts and one index operation.
Sure, that's a bug.

aneshlya:
Fixed by #3329

PR #3329, opened 2025-04-15
Allow lvalue to be a pointer arithmetic

### LINKED ISSUE #2496: Accessing members of expression involving pointer arithmetic triggers internal error
```c
struct S {
    uint32 a;
};
S v;

void foo(int i) {
    // fine
    S* p = (S*) (&v) + i;
    uint32 a = p->a;
    // borked
    uint32 a_ = (&v + i)->a;
}
```
The first case performs pointer arithmetic and stores the intermediary value before accessing the struct member using the arrow operator. The second example does the same thing but without an intermediary value, which triggers an assert in the compiler:

```
<source>:11:17: Error: Assertion failed 
        (/usr/local/src/ispc/src/expr.cpp:5331): "baseSym != NULL". 
    uint32 a_ = (&v + i)->a;
                ^^^^^^^^^^^
```

The problem is also present if you dereference the result first and then access the field
```c
void foo2(int i) {
    // all well and good
    S* p = (S*) (&v) + i;
    uint32 a = (*p).a;
    // oopsie
    uint32 a_ = (*(&v + i)).a;
}
```

[Godbolt example link](https://godbolt.org/z/PoKPWs1E7)

aneshlya:
This fixes #2496, #2583.

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -4911,11 +4911,6 @@ llvm::Value *IndexExpr::GetValue(FunctionEmitContext *ctx) const {
         mask = LLVMMaskAllOn;
     } else {
         Symbol *baseSym = GetBaseSymbol();
-        if (llvm::dyn_cast<FunctionCallExpr>(baseExpr) == nullptr && llvm::dyn_cast<BinaryExpr>(baseExpr) == nullptr &&
-            llvm::dyn_cast<SelectExpr>(baseExpr) == nullptr) {
-            // Don't check if we're doing a function call or pointer arith or select
-            AssertPos(pos, baseSym != nullptr);
-        }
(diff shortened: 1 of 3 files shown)


nurmukhametov (approved):
LGTM!

aneshlya:
After analyzing another issue, I believe we can safely remove the assertion—it’s overly defensive.
The `lMaskForSymbol` function already handles a null `baseSym`, and `baseSym` isn’t used elsewhere in this context.
A similar issue was fixed a while ago for `AssignExpr`, which actually added logic to handle a null `baseSym` in `lMaskForSymbol`:
https://github.com/ispc/ispc/commit/11033e108e557a0198e0cf4117eb72e1f431c63e

nurmukhametov (approved):

PR #3331, opened 2025-04-17
Fix logic in treating pointers for IndexExpr lvalue

### LINKED ISSUE #2186: Fatal error with casts and indexing
Hi!

```
typedef float<4> vec4;

struct MyStruct
{
    float* data;
};

vec4 GetData(MyStruct& from, int index)
{
    return ((vec4*)from.data)[index];
}
```
causes
```
Error: Assertion failed 
 (/home/appveyor/projects/ispc/src/ctx.cpp:2819): "maskedStoreFunc != 
        NULL". 
    return ((vec4*)from.data)[index];
```

https://ispc.godbolt.org/z/es7oPG73K

On current trunk.

aneshlya:
Fixes #2186 .
this PR aligns logic between `IndexExpr::GetLValueType()` and `IndexExpr::GetLValue()`.
The new code checks if `baseExprType` is a pointer first, which matches how `GetLValue` branches its logic

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -5130,18 +5130,16 @@ const Type *IndexExpr::GetLValueType() const {
 
     // Find the type of thing that we're indexing into
     const Type *elementType = nullptr;
-    const SequentialType *st = CastType<SequentialType>(baseExprLValueType->GetBaseType());
-    if (st != nullptr) {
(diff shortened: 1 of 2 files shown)


nurmukhametov (approved):
LGTM!

PR #3332, opened 2025-04-18
Fix features for ARM arch

aneshlya:
One lit test is failing on current trunk because the features returned from `getExtensionFeatures` and `getFPUFeatures` for `arm` architecture started to conflict. This PR merges features prioritizing enabled features. For example if one list has `-neon` and another `+neon`, the resulting list will have `+neon`.


nurmukhametov (approved):
LGTM!

PR #3333, opened 2025-04-21
Add testing on Windows ARM runners

aneshlya:
Github recently added support of Windows ARM runners. This PR adds testing on Windows ARM and slightly refactors reusable test workflow.
Full run is here: https://github.com/ispc/ispc/actions/runs/14581259840


nurmukhametov (approved):
LGTM!

PR #3336, opened 2025-04-23
Update for LLVM 20

nurmukhametov:
This PR updates:
- The default LLVM version to 20.1.2
- XE dependencies to compatible versions
- lit-tests accordingly to LLVM IR changes
- CI to use LLVM 20 instead of LLVM 18

Additionally, it disables the `SPV_KHR_untyped_pointers` extension, as it is unsupported by the current ocloc version.

Furthermore, it reverts an upstream patch that resulted in a performance regression on GameDev benchmarks using AVX2. This issue may be fixed later, you can track it via the upstream issue at https://github.com/llvm/llvm-project/issues/135998.


aneshlya (approved):
Looks good! As the next step we should update Dockerfiles to use LLVM 20.

PR #3337, opened 2025-04-23
Add support for Apple framework builds

Treata11:


aneshlya:
Hi @Treata11, please rebase your change on top of the `main` to enable CI runs.

aneshlya:
The CI is failing on ispcrt jobs. 
@Treata11, are you interested in completing this PR? Could you provide some brief description of the change and motivation behind it? I assume you're adding build capabilities to create an Apple private framework from ispc that will be used internally within your own apps, not intended for external distribution? The details will help us to assist you better and proceed with the change.

Treata11:
Hi @aneshlya 

> The CI is failing on ispcrt jobs.

I couldn't figure out why... I successfully built the project on my mac.

-

>  are you interested in completing this PR?

Of course!

- 

> Could you provide some brief description of the change and motivation behind it? I assume you're adding build capabilities to create an Apple private framework from ispc that will be used internally within your own apps, not intended for external distribution?

My changes would allow building Framework artifacts for Apple platforms (iOS-derived devices most importantly) that are intended to be used in XCode projects. This doesn't mean that the frameworks can't be used for external distribution; as a matter of fact, it's the best solution to build downstream projects for an iOS-derived targets. 

-

> The details will help us to assist you better and proceed with the change. 

I opened the PR as a draft since GitHub didn't allow me to fetch upstream changes due to conflicts. I will provide a full description of all the changes once the tasks are complete.

aneshlya:
The CI jobs, which are failing, build ispcrt separately from ispc. So from what I see `ISPC_FRAMEWORK ` is applicable to ispcrt only, not to ispc, so you should move `configure_ispc_framework` to ispcrt's CMakelists.txt. Also I think it should be called only on Apple platforms.
```
if (APPLE)
  configure_ispc_framework(${PRNAME})
endif()
```

aneshlya:
Also consider moving `ISPC_FRAMEWORK` option to ispcrt and renaming it to `ISPCRT_FRAMEWORK`.

aneshlya:
Hi @Treata11 ,
Just to let you know, we're approaching the release and if you want to have the changes in the source code tagged for the release - it's a good time to complete this PR. However we'll not modify our release binaries for this release (`ISPC_FRAMEWORK` will stay off for release builds).

aneshlya:
So all these
```
check_isa, ispc, ispc-opt, ispc-slim
```
are binaries, not libraries. You may see `add_library` commands in the root `CMakeLists.txt` file but they are all `OBJECT` libraries rather than a traditional static or shared library. Object libraries don't produce actual library files (like .a or .so files); instead, they compile source files into object files that can be used by other targets, in our case `add_executable`. So the only "real" libraries in ISPC distribution are ISPCRT libraries only.

Treata11:
> Just to let you know, we're approaching the release and if you want to have the changes in the source code tagged for the release - it's a good time to complete this PR. 

I'd like to; but ispc depends on LLVM which unfortunately, doesn't have iOS builds deployed in upstream. 
I either have to open a similar PR on LLVM & handle iOS compilations there which is going to be very time consuming considering the scale of the project, or I have to rely on some third-party scripts to build it for iOS.

I've already tried the latter, but there were some discrepancies between the iOS builds & macOS one. So, I firstly have to figure that out.

I can continue & finish this PR without LLVM too, but I wouldn't be able to confirm that it actually works!

PR #3340, opened 2025-04-24
Disable DIV and MOD for active lanes only for XE targets

nurmukhametov:
After PR #3322, a number of tests started to fail. This PR fixes the issue by ensuring it is not applied to XE targets.


aneshlya (approved):
LGTM!

PR #3343, opened 2025-04-28
Enable wasm targets in targetParentMap 

nurmukhametov:
This PR makes wasm target to be based on the generic built-in implementation. It also adds the CI job to run tests for wasm targets. This adds also auxiliary CI changes.


aneshlya (approved):
Looks good!

PR #3344, opened 2025-04-28
Fix fmod for float16 on platforms without native half support

aneshlya:
This PR updates fmod-float16 tests to use the same implementation for `fmod` operation as in stdlib. Looks like LLVM 18 preserves precision better and LLVM 20 performs more aggressive optimizations so the difference between `x* 1/y` and `x*rcp(y)` is quite significant on the platforms without native half support.


nurmukhametov (approved):
Looks good, but I would prefer to have the submodule update as a separate commit.

aneshlya:
> Looks good, but I would prefer to have the submodule update as a separate commit.

Right, submodules changed was added by accident. Updated.

PR #3351, opened 2025-04-30
Update dockerfiles

nurmukhametov:
This PR updates LLVM versions in some dockers that still use it as an explicit argument.

It also fixes python dependencies issues.

### DIFF
--- a/docker/ubuntu/18.04/cpu_ispc_build/Dockerfile
+++ b/docker/ubuntu/18.04/cpu_ispc_build/Dockerfile
@@ -1,9 +1,9 @@
 #
-#  Copyright (c) 2017-2024, Intel Corporation
+#  Copyright (c) 2017-2025, Intel Corporation
 #
 #  SPDX-License-Identifier: BSD-3-Clause
 
-ARG LLVM_VERSION=17.0
+ARG LLVM_VERSION=20.1
 
 FROM ubuntu:18.04 AS llvm_build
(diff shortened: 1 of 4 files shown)


docker/ubuntu/18.04/cpu_ispc_build/Dockerfile
```diff
@@ -1,9 +1,9 @@
 #
-#  Copyright (c) 2017-2024, Intel Corporation
+#  Copyright (c) 2017-2025, Intel Corporation
 #
 #  SPDX-License-Identifier: BSD-3-Clause
 
-ARG LLVM_VERSION=17.0
+ARG LLVM_VERSION=20.1
```
aneshlya, line 6: It seems that we just forgot to remove LLVM version when we were moving to superbuild. Let's just remove it (since it's coming from presets) to avoid these updates in the future.

aneshlya: Please remove LLVM version and feel free to merge the PR.

aneshlya (approved):

docker/ubuntu/18.04/cpu_ispc_build/Dockerfile
nurmukhametov: It looks like we use `LLVM_VERSION` parameter of ubuntu 22.04 Dockerfile. Same parameter of ubuntu 18.04 docker can be removed.

PR #3357, opened 2025-05-07
lex.ll: fix token types emitted for identifiers

### LINKED ISSUE #3355: Template function names clobber variable names
Consider the following example:

```ispc
template <typename T>
uniform int exp(T x);

uniform int bar(uniform int x) {
  uniform int exp = 9;
  return exp;
}
```

which fails to compile with the following error:
```bash
<source>:6:10: Error: Ambiguous use of overloaded function "exp". 
  return exp;
```

Compiler explorer link: https://ispc.godbolt.org/z/nx9YPE4or

This means that https://github.com/ispc/ispc/pull/3228 will result in compilation errors for any user code that uses variables with names `abs`, `max` or `min`.

nurmukhametov:
Prioritize TOKEN_IDENTIFIER over TOKEN_TEMPLATE_NAME. In case, when both variable and template have same name it allows to shadow template with local variable.

This PR fixes #3355

### DIFF
--- a/src/lex.ll
+++ b/src/lex.ll
@@ -1,5 +1,5 @@
 /*
-  Copyright (c) 2010-2024, Intel Corporation
+  Copyright (c) 2010-2025, Intel Corporation
 
   SPDX-License-Identifier: BSD-3-Clause
 */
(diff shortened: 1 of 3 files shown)


aneshlya (approved):
Looks good!

PR #3364, opened 2025-05-08
stdlib: add more short vec math functions

nurmukhametov:
This PR supports short vector types for more functions in the standard library. It addresses issue https://github.com/ispc/ispc/issues/2707.

These element wise functions are added:
- round
- floor
- ceil
- trunc
- rcp
- rcp_fast
- sqrt
- rsqrt
- sin
- asin
- cos
- acos
- tan
- atan
- exp
- log
- atan2
- pow

These function is implemented in `stdlib.isph` as template functions.


aneshlya (approved):
Looks good!

PR #3365, opened 2025-05-08
Fix assert if intrinsics is not supported

### LINKED ISSUE #3360: Assert when using LLVM intrinsic for wrong target
This is a very minor issue, but if you are doing this
```ispc
int16 mul_16x8_high(int16 a, int16 b) {
    return @llvm.x86.ssse3.pmul.hr.sw.128(a, b);
}
```
Compile with `-O2 --target=neon-i8x16 --enable-llvm-intrinsics` (notice we are compiling for arm while the intrinsic is x86/x64 only) You will get this output
```
<source>:2:12: Error: LLVM intrinsic "ssse3.pmul.hr.sw.128" supported 
        only on "x86" target architecture. 
    return @llvm.x86.ssse3.pmul.hr.sw.128(a, b);
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

/usr/local/src/ispc/src/intrinsics.cpp:680: Assertion failed: "funcDecl != nullptr".
***
***
/usr/local/src/ispc/src/main.cpp(419): FATAL ERROR: Unhandled signal sent to process; terminating.
Compiler returned: 139
```
https://godbolt.org/z/xYdEToazs

aneshlya:
The assert `Assert(funcDecl != nullptr);` is overprotective. Just return `nullptr` if intrinisc is not supported, the result of this function is checked further in parser.
Fixes #3360

### DIFF
--- a/src/intrinsics.cpp
+++ b/src/intrinsics.cpp
@@ -641,7 +641,9 @@ Symbol *Module::AddLLVMIntrinsicDecl(const std::string &name, ExprList *args, So
     }
 
     llvm::Function *funcDecl = lGetIntrinsicDeclaration(module, name, args, pos);
-    Assert(funcDecl != nullptr);
+    if (funcDecl == nullptr) {
+        return nullptr;
+    }
 
     Symbol *funcSym = lCreateISPCSymbolForLLVMIntrinsic(funcDecl, symbolTable);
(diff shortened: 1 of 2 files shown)


nurmukhametov (approved):
LGTM!

PR #3366, opened 2025-05-08
Add const qualifier PeepholePass for LLVM trunk

aneshlya:


aneshlya:
This change is included to https://github.com/ispc/ispc/pull/3367

PR #3367, opened 2025-05-09
Fix unsized globals and incomplete types error reports

nurmukhametov:
This PR fixes error reporting for unsized globals and incomplete type in function arguments. It mostly caused by changes in upstream. It doesn't allow to create unsized global variables on IR level anymore.

### DIFF
--- a/src/func.cpp
+++ b/src/func.cpp
@@ -1139,6 +1139,20 @@ Symbol *FunctionTemplate::LookupInstantiation(const TemplateArgs &tArgs) {
 
 Symbol *FunctionTemplate::AddInstantiation(const TemplateArgs &tArgs, TemplateInstantiationKind kind, bool isInline,
                                            bool isNoinline) {
+    for (size_t i = 0; i < tArgs.size(); i++) {
+        TemplateArg tArg = tArgs[i];
+        if (tArg.IsType()) {
+            const Type *argType = tArg.GetAsType();
+            if (argType && !argType->IsCompleteType()) {
+                Symbol *arg = args[i];
+                Error(arg->pos,
+                      "Template parameter '%s' with type '%s' can't be instantiated with incomplete type '%s'\n",
+                      arg->name.c_str(), arg->type->GetString().c_str(), argType->GetString().c_str());
+                return nullptr;
+            }
+        }
+    }
+
     const TemplateParms *typenames = GetTemplateParms();
     Assert(typenames);
     TemplateInstantiation templInst(*typenames, tArgs, isInline, isNoinline);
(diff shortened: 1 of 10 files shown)


src/func.cpp
```diff
@@ -1139,6 +1139,20 @@ Symbol *FunctionTemplate::LookupInstantiation(const TemplateArgs &tArgs) {
 
 Symbol *FunctionTemplate::AddInstantiation(const TemplateArgs &tArgs, TemplateInstantiationKind kind, bool isInline,
                                            bool isNoinline) {
+    for (size_t i = 0; i < tArgs.size(); i++) {
+        TemplateArg tArg = tArgs[i];
+        if (tArg.IsType()) {
+            const Type *argType = tArg.GetAsType();
+            if (!argType->IsCompleteType()) {
```
aneshlya, line 1146: Just in case let's do `if (argType && !argType->IsCompleteType())`

tests/lit-tests/2844-2.ispc
```diff
@@ -2,10 +2,10 @@
 
 // RUN: not %{ispc} --target=host --nowrap --nostdlib %s -o %t.o 2>&1 | FileCheck %s
 
-// CHECK: Error: parameter 'a' is an incomplete type: varying struct Sy
+// CHECK: Error: parameter 0 ('a') has incomplete type 'varying struct Sy'
 int bar(struct Sy a) {}
 
-// CHECK-NOT: Error: parameter 'b' is an incomplete type: varying struct Sx
+// CHECK Error: parameter 0 ('b') has incomplete type 'varying struct Sx'
```
aneshlya, line 8: //CHECK: Error:

src/module.cpp
```diff
@@ -1075,6 +1075,16 @@ void Module::AddFunctionDeclaration(const std::string &name, const FunctionType
     auto [name_pref, name_suf] = functionType->GetFunctionMangledName(false);
     std::string functionName = name_pref + name + name_suf;
 
+    for (int i = 0; i < functionType->GetNumParameters(); i++) {
+        const Type *t = functionType->GetParameterType(i);
+        if (!t->IsCompleteType()) {
+            const SourcePos paramPos = functionType->GetParameterSourcePos(i);
+            const std::string paramName = functionType->GetParameterName(i);
+            Error(paramPos, "parameter %d ('%s') has incomplete type '%s'", i, paramName.c_str(),
```
aneshlya, line 1083: I think just reporting of parameter name with its source pos is sufficient, no index is needed. It looks too details as some debug info.

aneshlya (approved):
Looks good

aneshlya (approved):

PR #3368, opened 2025-05-09
Improve error reports for intrinsics

### LINKED ISSUE #3116: Dump unrepresentable LLVM types for intrinsics
The following code
```ispc
int16 test(int16 t) {
    return @llvm.x86.ssse3.pmul.hr.sw(t, t);
}
```
fails to compile for target `avx2-i32x8` with the error message that is not helpful for users:

```bash
Error: Return type not representable for Intrinsic llvm.x86.ssse3.pmul.hr.sw. 
<source>:2:5: Error: Must provide return value for return statement 
        for non-void function. 
    return @llvm.x86.ssse3.pmul.hr.sw(t, t);
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Compiler returned: 1
```

It would be good to dump type that is not representable. It can help to realize that the actual problem here is that user wanted to use another intrinsic @llvm.x86.ssse3.pmul.hr.sw.**128**.

Compiler explorer link: https://godbolt.org/z/eMGcYzMxY

nurmukhametov:
Report unrepresentable LLVM IR types and fix reporting candidates when ISPC is unable to find the proper overload for a function call.

This PR fixes #3116

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -9621,10 +9621,11 @@ int FunctionSymbolExpr::computeOverloadCost(const FunctionType *ftype, const std
     return costSum;
 }
 
-static bool lReportErrorUnableToFindOverload(std::vector<Symbol *> &matches, const std::vector<const Type *> &argTypes,
+static bool lReportErrorUnableToFindOverload(std::vector<Symbol *> &candidates,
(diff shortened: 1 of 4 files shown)


aneshlya (approved):
LGTM!

Issue #3354, opened 2025-05-05
Compiler crash with consitional expressions and 2 levels deep templates


clusty:
The following code crashes the ispc compiler:
It seems that the combo of 2 level deep tremplated functions and conditional expressions don't mix well.

```
ispc-problem/problem.ispc:4:12: Error: Assertion failed (/usr/local/src/ispc/src/ctx.cpp:3314): "m->errorCount > 0". 
    return b1 ? 1 : 2;
           ^^

***
*** Please file a bug report at https://github.com/ispc/ispc/issues
*** (Including as much information as you can about how to reproduce this error).
*** You have apparently encountered a bug in the compiler that we'd like to fix!
***
/usr/local/src/ispc/src/main.cpp(419): FATAL ERROR: Unhandled signal sent to process; terminating.
Abort (core dumped)
```

workaround is with top replace ternary with if
Here is my simplified code that reproduces the problem:

```ispc
template<bool b1>
inline float DoStuff()
{
    return b1 ? 1 : 2;
} 


template<bool b1>
inline float KernelInternal()
{
    float<3> rgb = {0,0,0};
    if (b1)
    {              
        return 2 * DoStuff<b1>() + 1;
    }

    return -1;
}

export void Kernel()                                                                     
{
    KernelInternal<true>();
}

```

aneshlya:
Confirmed. Thanks for reporting this.
https://ispc.godbolt.org/z/Ks79oezf1

aneshlya:
This is actually a duplicate of https://github.com/ispc/ispc/issues/2825. I'm adding priority to https://github.com/ispc/ispc/issues/2825 and closing this one.
There is a very embarrassing logic in current implementation but basically if you add dummy template argument, it would work
https://ispc.godbolt.org/z/3qcd3ocWd

clusty:
My workaround included getting rid of the conditional: a regular if works.

PR #3370, opened 2025-05-10
Fix nested template without arguments

### LINKED ISSUE #2825: Compiler crash when nested template function doesn't use templated argument
This issue was discovered in https://github.com/ispc/ispc/pull/2790#issuecomment-2032512084.
Nested template (`foo2`) without templated arguments https://ispc.godbolt.org/z/fe9szMa4v - breaks.
Nested template (`foo2`) withtemplated arguments https://ispc.godbolt.org/z/eK8j1bqcq - works.

When function has templated arguments, we check it in ResolveOverloads https://github.com/ispc/ispc/blob/main/src/expr.cpp#L8900 and return earlier before doing `AddInstantiation`. If not, we continue executing the function and create a new instantiation, but it's too earlier for that.

aneshlya:
Fixes #2825 and #3354.

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -9709,13 +9709,36 @@ bool FunctionSymbolExpr::ResolveOverloads(SourcePos argPos, const std::vector<co
     }
     triedToResolve = true;
 
+    // Check for dependent arguments
+    bool hasDependentArgs = false;
     for (auto argType : argTypes) {
-        if (argType->IsDependent()) {
-            unresolvedButDependent = true;
+        if (argType && argType->IsDependent()) {
+            hasDependentArgs = true;
             break;
         }
     }
-    if (unresolvedButDependent) {
+
+    // Check for dependent template arguments
+    bool hasDependentTemplateArgs = false;
+    for (const auto &arg : templateArgs) {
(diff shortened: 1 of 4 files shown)


dbabokin (approved):
LGTM

src/expr.cpp
```diff
@@ -9708,13 +9708,36 @@ bool FunctionSymbolExpr::ResolveOverloads(SourcePos argPos, const std::vector<co
...
+    bool hasDependentTemplateArgs = false;
+    for (const auto &arg : templateArgs) {
+        if (arg.IsType()) {
+            if (arg.GetAsType() && arg.GetAsType()->IsDependent()) {
+                hasDependentTemplateArgs = true;
+                break;
+            }
+        } else if (arg.IsNonType()) {
+            const ConstExpr *ce = arg.GetAsConstExpr();
+            // template argument is nor resolved yet
```
nurmukhametov, line 9730: not

nurmukhametov (approved):
LGTM

src/expr.cpp
aneshlya: Fixed

PR #3382, opened 2025-05-15
CI: change release pipeline parameter

nurmukhametov:
It can point to any repo reference now (SHA or branch name or tag). Define ISPC version and the corresponding docker based on the content of the file `common/version.h`.


aneshlya (approved):
LGTM!

PR #3384, opened 2025-05-15
Fix Python/ISPC examples for WIndows

aneshlya:
This PR fixes location of dll in examples build directory (it should be in the same folder as python script) and adds `--dllexport` flag.


dbabokin (approved):
LGTM

Issue #3388, opened 2025-05-18
FATAL ERROR: Unhandled signal sent to process; terminating


illua1:
Extra dereference in aggregate initialization produce internal error:

```ISPC
export void test(const uniform float a[])
{
    const uniform float v = {a[0][0]};
}
```

```

<source>:3:30: Error: Trying to index into non-array, vector, or 
        pointer type "const uniform float". 
    const uniform float v = {a[0][0]};
                             ^^^^^^^

/usr/local/src/ispc/src/main.cpp(419): FATAL ERROR: Unhandled signal sent to process; terminating.
***
*** Please file a bug report at https://github.com/ispc/ispc/issues
*** (Including as much information as you can about how to reproduce this error).
*** You have apparently encountered a bug in the compiler that we'd like to fix!
***
Program terminated with signal: SIGSEGV
```

Code example on godbolt: https://godbolt.org/z/5zPaz3xzn

zephyr111:
@aneshlya : I think we can close this issue since the fix has been merged.

aneshlya:
Fixed by https://github.com/ispc/ispc/pull/3389

PR #3296, opened 2025-03-24
Report error when pointer types are used with unsupported binary operators

nurmukhametov:
This fixes #3295

PR #3389, opened 2025-05-18
Fix the crash of #3388

zephyr111:
This PR fix the crash of #3388 in a naive way:
ISPC crashes because `exprs[0]` is null hence the segmentation fault. If we add a check then this does not cause any segfault on the example. That being said, this is a naive fix because I do not know if having a `nullptr` in `exprs` is actually normal. If not, then another more complex fix is needed.

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -4532,7 +4532,7 @@ static std::pair<llvm::Constant *, bool> lGetExprListConstant(const Type *type, 
     bool isVaryingInit = false;
     bool isNotValidForMultiTargetGlobal = false;
     if (exprs.size() == 1 && (CastType<AtomicType>(type) != nullptr || CastType<EnumType>(type) != nullptr ||
-                              CastType<PointerType>(type) != nullptr)) {
+                              CastType<PointerType>(type) != nullptr) && exprs[0] != nullptr) {
         if (isStorageType) {
             return exprs[0]->GetStorageConstant(type);
         } else {
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -4532,7 +4532,10 @@ static std::pair<llvm::Constant *, bool> lGetExprListConstant(const Type *type, 
     bool isVaryingInit = false;
     bool isNotValidForMultiTargetGlobal = false;
     if (exprs.size() == 1 && (CastType<AtomicType>(type) != nullptr || CastType<EnumType>(type) != nullptr ||
-                              CastType<PointerType>(type) != nullptr) && exprs[0] != nullptr) {
+                              CastType<PointerType>(type) != nullptr)) {
+        if (exprs[0] == nullptr) {
+            return std::pair<llvm::Constant *, bool>(nullptr, false);
+        }
         if (isStorageType) {
             return exprs[0]->GetStorageConstant(type);
         } else {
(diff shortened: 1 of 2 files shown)


illua1:
Feels like you should be able to check the fact of error in nested expression without using ambiguous null check.

src/expr.cpp
```diff
@@ -4530,7 +4530,7 @@ static std::pair<llvm::Constant *, bool> lGetExprListConstant(const Type *type,
     bool isVaryingInit = false;
     bool isNotValidForMultiTargetGlobal = false;
     if (exprs.size() == 1 && (CastType<AtomicType>(type) != nullptr || CastType<EnumType>(type) != nullptr ||
-                              CastType<PointerType>(type) != nullptr)) {
+                              CastType<PointerType>(type) != nullptr) && exprs[0] != nullptr) {
         if (isStorageType) {
```
nurmukhametov, line 4534: I would check the corner case for `exprs[0]` inside `if` statement and return `std::pair<llvm::Constant *, bool>(nullptr, false);` if it is `nullptr`. Otherwise, we might start exploring some states within this function that weren't possible before the fix.

nurmukhametov (changes requested):
Thank you for investigating this crash!

Please add a regression test using the example from the reported issue to the `tests/lit-tests` folder. You can do this in a similar way to PR #3296.

zephyr111:
I wanted to check if the `nullptr` was legit before applying the proposed modification. 
It turns out it clearly looks legit to me. Indeed, when the type error ("Trying to index into non-array, vector, or pointer type [...]") is reported in `IndexExpr::TypeCheck` at `expr.cpp:5209`, a `nullptr` is returned from the function (which is often normal in compilers when such an error happens). This function is called from `lTypeCheckNode` at `ast.cpp:337`, called from `ispc::WalkAST` at `ast.cpp:323` itself called from another `ispc::WalkAST` at `ast.cpp:282`. The later execution the following statement:

```cpp
for (unsigned int i = 0; i < el->exprs.size(); ++i) {
    el->exprs[i] = (Expr *)WalkAST(el->exprs[i], preFunc, postFunc, data);
}
```

This is where the expression (i.e. `el->exprs[i]`) is set to `nullptr`.

This makes sense: the expression is set to `nullptr` during the semantic analysis (walking though the AST) due to a type error and it is then read again when the IR code generation step happens (more specifically `ispc::AST::GenerateIR` calling indirectly `DeclStmt::EmitCode` then calling indirectly `ispc::ExprList::GetStorageConstant` and then `lGetExprListConstant`).

I did not found special error-related function/attribute to know whether there is an error on the target `eList` and `type` variable so I guess this is the way to handle errors in ISPC so far.

Returning `std::pair<llvm::Constant *, bool>(nullptr, false);` makes indeed more sense regarding the rest of the code and the calling function (safer alternative).

aneshlya:
Hi @zephyr111 ,
Thanks a lot for your analysis. Please apply all the suggestions that @nurmukhametov pointed out and it's ready to go. Don't forget to apply `clang-format`  to changed files.

zephyr111:
It should be fine now.

I spent some time trying to understand why it was not working with `--nowrap` (which is apparently not documented): the pattern was not found due to sneaky hidden line endings.

PR #3390, opened 2025-05-21
Split module.cpp to three separate files

nurmukhametov:
Split `module.cpp` source file into three separate files:

- `preprocessor.cpp` contains all functions related to creation and configuration of the preprocessor,
- `emit_headers.cpp` contains all code that is related generation of C/C++ headers and stub files,
- `module.cpp` contains the remaining functionality.


aneshlya (approved):
Looks good!

Issue #2792, opened 2024-03-12
Error when compiling for AVX512 target  with 'disable-zmm' param


petrsm:
When I try to compile some of our code for AVX512 with 'disable-zmm' parameter, I get following error:

"LLVM ERROR: Don't know how to split the result of this operator !"

Compiling without this parameter works fine. I was just curious if this option has (positive) impact
on performance of generated code.

Repro case attached.

[repro.zip](https://github.com/ispc/ispc/files/14578461/repro.zip)

nurmukhametov:
The minimized reproducer is:
```ispc
void foo(uniform float output[], float a, float b)
{
   float m = max(a, b);
   output[programIndex] = m;
}
```
Compile with: `ispc --target=avx512skx-x16 --opt=disable-zmm test.ispc --emit-obj -o test.o`

aneshlya:
This is a duplicate of https://github.com/ispc/ispc/issues/1875. We should use generic llvm intrinsics instead of target-specific on avx512skx. Make sense to update it after generic target is in place.

Issue #1875, opened 2020-09-17
Compiler error with --opt=disable-zmm on trunk


JeffRous:
Error:
SplitVectorResult #0: t18: v16f32 = llvm.x86.avx512.max.ps.512 TargetConstant:i64<6877>, t13, t15, TargetConstant:i32<4>, stdlib.ispc:1813:12 @[ example.ispc:5:21 ]

LLVM ERROR: Do not know how to split the result of this operator!

Compiler returned: 1

Reproducer: 
https://ispc.godbolt.org/z/fc1xr1

Options: --target=avx512skx-i32x16 -O2 --opt=disable-zmm
``` ispc
export void Max(uniform float Result[], const uniform float Source1[], const uniform float Source2[], const uniform int Iterations)
{
    foreach(i = 0 ... Iterations)
    {
        Result[i] = max(Source1[i], Source2[i]);
    }
}
```

DeepakRajendrakumaran:
https://godbolt.org/z/nYsYsE

Looks like we'll need to fix this in ISPC. Currently, for disabling zmm use, we set "min-legal-vector-width"="256"".
We expect llvm to split intrinsics like 'llvm.masked.load.v16f32.p0v16f32'. 

But this approach doesn't work for 'llvm.x86.avx512.max.ps.512'. Might have to go through other intrinsics starting with 'llvm.x86'

dbabokin:
Ouch, this is ``avx512skx-i32x16`` with ``disable-zmm``. We'll need to modify stdlib then... Basically this means that ``disable-zmm`` in this case requires different target. This is not urgent according to Jeff, I suggest adding this to new std lib implementation requirement that I'm working on and solving more generally than just patching it.

DeepakRajendrakumaran:
One option could be to switch to the more generic 'llvm.maximum'. I'll wait till new std implementation is done.

dbabokin:
That's very good point. But we do have more builtins, which use avx512 specific instrinsics.

DeepakRajendrakumaran:
I had talked to Craig and he made some interesting comments about difference in behavior with NaN for 'llvm.x86.avx512.max.ps.512' vs llvm.maximum vs llvm.maxnum. for floats

llvm.snum and llvm.unum should handle int

aneshlya:
This can be fixed after https://github.com/ispc/ispc/pull/3140 is merged.

PR #3394, opened 2025-05-23
[WIP] Fix--opt=disable-zmm for avx512* targets

nurmukhametov:
Supported targets:
- generic-i32x16
- generic-i1x16
- generic-i1x32
- generic-i1x64

Also report warnings that --opt=disable-zmm is unsupported or erroneous for targets that support avx512. It was implemented only for x16 wide targets, but not every program could be compiled because builtins use LLVM intrinsics that require zmm registers. For x32 and x64 targets, it was not implemented at all.

On the contrary, builtins of generic targets do not use target specific builtins so we can use them with --opt=disable-zmm and the proper CPUs, e.g., --opt=disable-zmm --target=generic-i1x16 --cpu=skx

This addresses issue #1875 and #2792.

### DIFF
--- a/src/ispc.cpp
+++ b/src/ispc.cpp
@@ -2190,6 +2190,54 @@ Target::Target(Arch arch, const char *cpu, ISPCTarget ispc_target, PICLevel picL
         cpu = a.GetDefaultNameFromType(CPUID).c_str();
     }
 
+    if (g->opt.disableZMM) {
+        std::string target_string = ISPCTargetToString(m_ispc_target);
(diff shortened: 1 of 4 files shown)


nurmukhametov:
> I recommend splitting this PR into two parts:
> 
> * Add disable-zmm to generic targets. Update the option’s description in both the help text and documentation to reflect this change appropriately.

This part is in the new PR #3397 

> * Address the issue with AVX-512 targets. There are a few possible solutions:
> 

This part will be addressed here.

PR #3396, opened 2025-05-27
Fix build with the trunk LLVM

nurmukhametov:
The related upstream change is 62fd4d18e52e
Their consolidated IIT descriptors for fractions of vector length under OneNthEltsVecArgument IIT descriptor


dbabokin (approved):
LGTM

aneshlya (approved):
Looks good!

Issue #2168, opened 2021-09-24
Missing math functions


stotko:
In contrast to `isnan`, the math functions `isinf` and `isfinite` are not defined. There might be further missing functions which I have not tested. Providing overloads for `float` and `double`of these two would be very helpful. 

One concrete application could be a vectorized version of the unary element-wise operations of Open3D's Tensor class:
- `isinf`: https://github.com/isl-org/Open3D/blob/8cc39dd9abc3ca95f171af4647d172052f66d862/cpp/open3d/core/kernel/UnaryEWCPU.cpp#L111
- `isfinite`: https://github.com/isl-org/Open3D/blob/8cc39dd9abc3ca95f171af4647d172052f66d862/cpp/open3d/core/kernel/UnaryEWCPU.cpp#L116

nurmukhametov:
@zephyr111 has submitted the PR implementing this issue

PR #3416, opened 2025-06-07
Add the isinf and isfinite standard functions

zephyr111:
## Description
This add the two basic `isinf` and `isfinite` math functions requested in #2168. They behave like in C/C++.

## Related Issue

## Checklist

### DIFF
--- a/docs/ispc.rst
+++ b/docs/ispc.rst
@@ -4945,6 +4945,36 @@ The ``isnan()`` functions test whether the given value is a floating-point
     bool isnan(double v)
     uniform bool isnan(uniform double v)
 
+The ``isinf()`` and ``isfinite()`` functions test whether the given value is
+a floating-point infinity (please note that a "not a number" value is neither 
+considered finite nor infinite):
+
+::
+
+    bool isinf(float16 v)
+    uniform bool isinf(uniform float16 v)
+    bool isinf(float v)
+    uniform bool isinf(uniform float v)
+    bool isinf(double v)
+    uniform bool isinf(uniform double v)
+
+::
+
+    bool isfinite(float16 v)
+    uniform bool isfinite(uniform float16 v)
+    bool isfinite(float v)
+    uniform bool isfinite(uniform float v)
+    bool isfinite(double v)
+    uniform bool isfinite(uniform double v)
+
+The isinf and isfinite functions also support short vector types with the
+basic types listed above.
+
+::
+
+    template <typename T, uint N> T<N> isinf(T<N> v)
+    template <typename T, uint N> T<N> isfinite(T<N> v)
+
 
 A number of functions are also available for performing operations on 8- and
 16-bit quantities; these map to specialized instructions that perform these
--- a/stdlib/stdlib.ispc
+++ b/stdlib/stdlib.ispc
@@ -1728,6 +1728,40 @@ __declspec(safe, cost1) static inline uniform bool isnan(uniform double v) { ret
 
 __declspec(safe, cost1) static inline bool isnan(double v) { return v != v; }
 
+__declspec(safe, cost1) static inline uniform bool isinf(uniform float16 v) { return intbits(abs(v)) == 0x7C00; }
+
+__declspec(safe, cost1) static inline bool isinf(float16 v) { return intbits(abs(v)) == 0x7C00; }
+
+__declspec(safe, cost1) static inline uniform bool isinf(uniform float v) { return intbits(abs(v)) == 0x7F800000; }
+
+__declspec(safe, cost1) static inline bool isinf(float v) { return intbits(abs(v)) == 0x7F800000; }
+
+__declspec(safe, cost1) static inline uniform bool isinf(uniform double v) {
+    return intbits(abs(v)) == 0x7FF0000000000000;
+}
+
+__declspec(safe, cost1) static inline bool isinf(double v) { return intbits(abs(v)) == 0x7FF0000000000000; }
+
+__declspec(safe, cost1) static inline uniform bool isfinite(uniform float16 v) {
+    return (intbits(v) & 0x7C00) != 0x7C00;
+}
+
+__declspec(safe, cost1) static inline bool isfinite(float16 v) { return (intbits(v) & 0x7C00) != 0x7C00; }
+
+__declspec(safe, cost1) static inline uniform bool isfinite(uniform float v) {
+    return (intbits(v) & 0x7F800000) != 0x7F800000;
+}
+
+__declspec(safe, cost1) static inline bool isfinite(float v) { return (intbits(v) & 0x7F800000) != 0x7F800000; }
+
+__declspec(safe, cost1) static inline uniform bool isfinite(uniform double v) {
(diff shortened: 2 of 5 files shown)


nurmukhametov (changes requested):
Thank you for implementing `isinf` and `isfinite`!

Since these functions are element-wise we need to support them for `uniform T<N>` and `varying T<N>` types.
The following code should be added to the end of `stdlib/include/stdlib.isph`:
```ispc
SHORT_VEC_ARG1(isinf)
SHORT_VEC_ARG1(isfinite)
```

stdlib/stdlib.ispc
```diff
@@ -1728,6 +1728,40 @@ __declspec(safe, cost1) static inline uniform bool isnan(uniform double v) { ret
 
 __declspec(safe, cost1) static inline bool isnan(double v) { return v != v; }
 
+__declspec(safe, cost1) static inline uniform bool isinf(uniform float16 v) { return intbits(abs(v)) == 0x7C00; }
```
aneshlya, line 1731: Is there a specific reason why you implemented `inf` functions using `abs` and `isfinite` using masking out the sign bit?
`abs()` may introduce unnecessary instructions.

zephyr111: > abs() may introduce unnecessary instructions.

It did not in my tests on x86-64/AVX2 and on ARM/Neon (on 32-bit floats): it only performed a mask. 

[Here](https://ispc.godbolt.org/z/bE3j1bsMx) are results on Godbolt. We can see the x86-64 assembly masks out the sign bit in both case because `vcmpneq_oqps` can be used on AVX2 (but AFAIK we cannot specify how to handle NaN values directly in ISPC, and AFAIK on Neon neither).

On ARM, LLVM is smart enough so to use the `abs` instruction instead. This avoid to use a constant so it put less pressure on registers. For, `isfinite`, it looks like LLVM produce a quite complicated assembly code and I wonder if it is optimal (it looks like we cannot reinterpret floats as integer on Neon based on the generated code).

There is a catch though: I assumed the code is compiled with optimizations. In `-O0`, the version with `abs` is indeed not great, though the assembly code is horrible in all functions.

Overall, I mainly used `abs` (instead of a mask) for sake of clarity in `isinf` (and also to mitigate a bit code duplication). I did not use it for `isfinite` because I could not find a way to use it efficiently (that is with only 2 instructions in loops). I am not strongly against using masking for both.

Please note that `isinf` is not the exact opposite of `isfinite` because of `NaN` value rejected by both functions. This is why the implementation is not symmetric.

zephyr111: By the way, please note that AVX-512 (and possibly SVE) contains an instruction for that: `vfpclassps`. That being said, AFAIK, we cannot use it directly in ISPC for now (this certainly require to write a new builtin intrinsic for all targets), so I left this improvement for future work. Some math functions can also benefit a bit from this (e.g. `exp` and `log`).

docs/ispc.rst
```diff
@@ -4814,6 +4814,28 @@ The ``isnan()`` functions test whether the given value is a floating-point
...
+considered finite nor infinite):
+
+::
+
+    bool isinf(float16 v)
+    uniform bool isinf(uniform float16 v)
+    bool isinf(float v)
+    uniform bool isinf(uniform float v)
+    bool isinf(double v)
+    uniform bool isinf(uniform double v)
```
nurmukhametov, line 4959: Please, add docs entries for shor-vector versions as well:
```ispc
template <typename T, uint N> T<N> isinf(T<N> v)
```

stdlib/stdlib.ispc
nurmukhametov: > By the way, please note that AVX-512 (and possibly SVE) contains an instruction for that: `vfpclassps`. That being said, AFAIK, we cannot use it directly in ISPC for now (this certainly require to write a new builtin intrinsic for all targets), so I left this improvement for future work. Some math functions can also benefit a bit from this (e.g. `exp` and `log`).

Apparently, LLVM has [llvm.is.fpclass](https://llvm.org/docs/LangRef.html#llvm-is-fpclass-intrinsic) intrinsic. It should be quite straightforward now to write new builtin functions in `builtins/generic.ispc` using it. Something like this should work:
```ispc
EXT uniform bool __isinf(uniform float v) { return @llvm.is.fpclass(v, 2) || @llvm.is.fpclass(v, 9); }
EXT bool __isinf(float v) { return @llvm.is.fpclass(v, 2) || @llvm.is.fpclass(v, 9); }
```

However, I didn't manage to generate `vfpclassps` asm intruction using these intrinsics, link to the compiler explorer: https://ispc.godbolt.org/z/vzME3nfP4

aneshlya: > Apparently, LLVM has [llvm.is.fpclass](https://llvm.org/docs/LangRef.html#llvm-is-fpclass-intrinsic) intrinsic. It should be quite straightforward now to write new builtin functions in builtins/generic.ispc using it.

This intrinsic was introduced in LLVM 18 (we still support LLVM 17). We can switch to it when we remove LLVM 17 support and figure out the code generation. Or we can use avx512-specific intrinsic: https://ispc.godbolt.org/z/Y8hsooY8c

As for current implementation I see the code difference between abs/mask implementation on doubles for sse2 target:
https://ispc.godbolt.org/z/zejP17oof.

zephyr111: > As for current implementation I see the code difference between abs/mask implementation on doubles for sse2 target: https://ispc.godbolt.org/z/zejP17oof.

The code was too small to be actually correct so I found it suspicious. Indeed, the sse2 target is actually sse2-i32x4 so there is 4 double to compute stored in 2 registers and 1 (fused) mask register. Here the compiler does a shuffle so to only compare the upper part so the lower-part is discarded. THis is a correct compilation of the input code. The problem is that the input mask is actually incorrect in `isinf_v2` is must be `0x7FFFFFFFFFFFFFFF` instead of `0x7FF0000000000000` because NaN must be rejected by `isinf`. 

See a comparison on [Godbolt](https://ispc.godbolt.org/z/r3339s7qs). The correct version is `isinf_v3` and the generated code is identical to `isinf_v1`.

This clever operation is interestingly not applied on `isfinite` in AVX2 while the compiler could certainly use the same trick (though it is more complicated due to lanes).

zephyr111: Regarding the fpclass instrinsics, if I understand correctly the situation, the generic intrinsic is currently useless in our case (not really optimized) so we need the avx-512 specific one which is already available in LLVM 17 (so there is no need for an update). If we use the later, then we should implement the alternative ISPC intrinsic for other target (e.g. AVX2 and Neon). This might be simple, but not trivial either.
Thus, I think we should first merge a version without such intrinsics and then open a ticket for future improvement (referencing this discussion). With some luck, the generic LLVM intrinsic might be optimized meanwhile.

aneshlya: Agree. I think there is one last doc-related comment in this review that should be resolved and it's good to go.

PR #3417, opened 2025-06-09
Clean up avx512 builtins

nurmukhametov:
## Description

Remove from the `avx512-*.ll` functions that are defined in the same way as the generic implementations.

## Related Issue

## Checklist


aneshlya:
How did you find all the functions that should be removed? Did you compare code produced between generic and avx512 targets for every builtin?

nurmukhametov:
> How did you find all the functions that should be removed? Did you compare code produced between generic and avx512 targets for every builtin?

I don't think that's all the functions. I have just come across them by basically comparing the generated assembly or looking into the implementation. There are definitely others. Yes, I compared the generated assembly code for the standard library functions between the generic and avx512 targets.

PR #3423, opened 2025-06-11
Add intrinsic headers and corresponding microbenchmarks.

aneshlya:
## Description
Add intrinsic headers and corresponding microbenchmarks.

## Related Issue

## Checklist


CMakeLists.txt
```diff
@@ -919,6 +923,11 @@ if (ISPC_SLIM_BINARY)
     install(DIRECTORY ${BITCODE_FOLDER} DESTINATION share)
 endif()
 
+# Install intrinsics headers
+foreach (header ${INTRINSICS_HEADERS})
+    install (FILES "stdlib/include/${header}" DESTINATION include/intrinsics)
```
nurmukhametov, line 928: Do we really need to place them it the `intrinsics` subdirectory?

aneshlya: Since our `include` directory will contain header files for both the compiler and the ISPC runtime, I think it would be a good idea to separate them into different subdirectories. If we introduce standard library headers (e.g., for tensors), they could probably live directly in the top-level include folder. But for everything that isn’t part of the "core" compiler, it makes sense to place them in separate directories. What do you think?

nurmukhametov: I agree that it makes sense.

Issue #3418, opened 2025-06-09
Missing short-vector functions


zephyr111:
Hello, 

I saw that some short-vector functions are apparently missing in the file "stdlib.ispc".
This includes:
- math/numerical functions: `fmod`, `isnan`, `clamp`
- low-level FP manipulation functions: `frexp` and `ldexp`
- reduction functions: `reduce_add`, `reduce_max`, `reduce_min`
- random-related functions (like `frandom`)
- classical SIMD functions: `select`, `broadcast`, `shift`, `all`, `any`, `popcnt`, etc. (some like `rotate` and `shuffle`, are probably not relevant)
- other special functions (probably not relevant): `extract`, `insert`, `half_to_float`, `half_to_float_fast`, `rsqrt_fast`, `float16bits`/`floatbits`/`doublebits`/`intbits` (I am not sure `extract`, `insert` are relevant since direct indexing is apparently possible on short vectors)

Which ones should be added? I think the first category is the minimum needed.
What do you think about this?

nurmukhametov:
> * math/numerical functions: `fmod`, `isnan`, `clamp`
> Which ones should be added? I think the first category is the minimum needed. What do you think about this?

It seems to me that I overlooked these element-wise functions. They are needed to be implemented.

> reduction functions: reduce_add, reduce_max, reduce_min
> classical SIMD functions: select, broadcast, shift, all, any, popcnt, etc. (some like rotate and shuffle, are probably not relevant)

Regarding other functions with cross-lane operations, it is not quite clear to me what semantics they need to have.

zephyr111:
@nurmukhametov I closed this issue since I think the most important things has been merged. Feel free to reopen it if there is anyone requesting short-vectors functions for frexp/ldexp operations, random ones, etc.
PS: should this be added to the planning?

nurmukhametov:
> [@nurmukhametov](https://github.com/nurmukhametov) I closed this issue since I think the most important things has been merged. Feel free to reopen it if there is anyone requesting short-vectors functions for frexp/ldexp operations, random ones, etc. PS: should this be added to the planning?

It is fine with me. I don't think we need frexp/ldexp or random ones until someone actually wants to use them.

nurmukhametov:
[...]

PR #3425, opened 2025-06-14
Add missing short vector functions and update the doc

zephyr111:
## Description

This PR addresses the issue #3418.

It add the most important missing short vector math functions: `fmod`, `isnan`, `clamp`, `rsqrt_fast`.
I did not add `frexp` and `ldexp` yet because they are rarely used functions and the parameters are pointers so we cannot directly use the already defined macros (this means we need to add dozens of lines of specialized code for them). I think that it may be better just to wait for a user request for this. Very few people should probably need them.
Regarding random functions, it would be nice to have the short vector variants but I am not sure this is straightforward to add the short vector variants (because of the state and pointers).

I updated the doc to add these functions and also clean some parts of the doc. There was some missing variants, some mistakes on types (e.g. `float` vs `float16`). Please note that some functions are not supported for all types despite what was actually stated in the doc.

## Related Issue

## Checklist

### DIFF
--- a/tests/lit-tests/stdlib-short-vec-funcs-3.ispc
+++ b/tests/lit-tests/stdlib-short-vec-funcs-3.ispc
@@ -1,5 +1,5 @@
-// This test just checks that abs and similar functions with short vectors are
-// compiled without errors for all types and reasonalbe vector sizes.
+// This test just checks that the target math functions with short vectors are
+// compiled without errors for all types and reasonable vector sizes.
 
 // RUN: %{ispc} --target=host -o %t.o %s 2>&1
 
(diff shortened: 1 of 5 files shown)


nurmukhametov (commented):
It is also worth checking that the compilation for these functions is OK in lit-tests, as it was done for others, for example [here](https://github.com/ispc/ispc/blob/main/tests/lit-tests/stdlib-short-vec-funcs-3.ispc).

zephyr111:
I adapted the short-vector functions for `isnan`, `isinf` and `isfinite` because they did not returned the correct type (I did not see the short-vector macro needed to be different for `isinf`/`isfinite` in the previous PR).
I also added lit-tests.

PR #3426, opened 2025-06-14
Zen4/Zen5 support

zephyr111:
## Description

This PR add the support of Zen4 and Zen5.
Zen4 is apparently [supported since LLVM 16](https://www.phoronix.com/news/AMD-Zen-4-Znver4-LLVM-Added), while Zen5 is [supported since LLVM 19~20](https://www.phoronix.com/news/LLVM-Clang-Znver5-Merged). Since the minimum [LLVM version supported by ISPC](https://github.com/ispc/ispc/wiki/ISPC-Development-Guide) is supposed to be the 18.1 I only guarded Zen5 changes with preprocessor's conditions. I used [this Wikipedia page](https://fr.wikipedia.org/wiki/Advanced_Vector_Extensions) as a reference to add the new support.

I got some questions though:
- Why `CPU_ICL` references `avx512skx_x16` and not `avx512icl_x16` [here](https://github.com/ispc/ispc/blob/b8b8bf8b685852f6719abbc2d47e9d11651730c9/src/ispc.cpp#L982)?
- I guess "avx_vnni" means VNNI for AVX2 only, so without a 512-bit support (only 256-bit), while "avx512_vnni" means VNNI for AVX-512 only, so without a 256-bit support (only 512-bit); is this right here?
- Why `compat[CPU_ZNVER2]` don't include `CPU_ZNVER1` [here](https://github.com/ispc/ispc/blob/b8b8bf8b685852f6719abbc2d47e9d11651730c9/src/ispc.cpp#L657)? Same question with `compat[CPU_ZNVER3]` and `CPU_ZNVER2`.
- Out of curiosity, is this OK to remove the ISPC_LLVM_18_1 conditionals since the minimum LLVM version required by ISPC is the 18.1?

## Checklist

### DIFF
--- a/tests/lit-tests/cpus_x86.ispc
+++ b/tests/lit-tests/cpus_x86.ispc
@@ -22,6 +22,7 @@
 //; RUN: %{ispc} %s -o %t.o --nostdlib --target=sse2-i32x4 --cpu=znver1
 //; RUN: %{ispc} %s -o %t.o --nostdlib --target=sse2-i32x4 --cpu=znver2
 //; RUN: %{ispc} %s -o %t.o --nostdlib --target=sse2-i32x4 --cpu=znver3
+//; RUN: %{ispc} %s -o %t.o --nostdlib --target=sse2-i32x4 --cpu=znver4
 //; RUN: %{ispc} %s -o %t.o --nostdlib --target=sse2-i32x4 --cpu=ps5
 //; RUN: %{ispc} %s -o %t.o --nostdlib --target=sse2-i32x4 --cpu=skx
 //; RUN: %{ispc} %s -o %t.o --nostdlib --target=sse2-i32x4 --cpu=icelake-client
--- a/src/ispc.cpp
+++ b/src/ispc.cpp
@@ -357,10 +357,14 @@ typedef enum {
 #if ISPC_LLVM_VERSION >= ISPC_LLVM_20_0
     CPU_DMR,
 #endif
-    // Zen 1-2-3
+    // Zen1 to Zen5
     CPU_ZNVER1,
     CPU_ZNVER2,
     CPU_ZNVER3,
+    CPU_ZNVER4,
+#if ISPC_LLVM_VERSION >= ISPC_LLVM_20_0
+    CPU_ZNVER5,
+#endif
 
 // FIXME: LLVM supports a ton of different ARM CPU variants--not just
 // listed below.  We should be able to handle any of them that also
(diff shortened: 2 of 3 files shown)


nurmukhametov:
> * Why `CPU_ICL` references `avx512skx_x16` and not `avx512icl_x16` [here](https://github.com/ispc/ispc/blob/b8b8bf8b685852f6719abbc2d47e9d11651730c9/src/ispc.cpp#L982)?

It looks like a bug.

> * I guess "avx_vnni" means VNNI for AVX2 only, so without a 512-bit support (only 256-bit), while "avx512_vnni" means VNNI for AVX-512 only, so without a 256-bit support (only 512-bit); is this right here?

Yes. It is correct that we shouldn't apply `avx_vnni` for zen4.

> * Why `compat[CPU_ZNVER2]` don't include `CPU_ZNVER1` [here](https://github.com/ispc/ispc/blob/b8b8bf8b685852f6719abbc2d47e9d11651730c9/src/ispc.cpp#L657)? Same question with `compat[CPU_ZNVER3]` and `CPU_ZNVER2`.

No idea.

> * Out of curiosity, is this OK to remove the ISPC_LLVM_18_1 conditionals since the minimum LLVM version required by ISPC is the 18.1?

It is still 17 https://github.com/ispc/ispc/blob/b8b8bf8b685852f6719abbc2d47e9d11651730c9/src/ispc_version.h#L24

tests/lit-tests/cpus_x86.ispc
```diff
@@ -22,6 +22,7 @@
 //; RUN: %{ispc} %s -o %t.o --nostdlib --target=sse2-i32x4 --cpu=znver1
 //; RUN: %{ispc} %s -o %t.o --nostdlib --target=sse2-i32x4 --cpu=znver2
 //; RUN: %{ispc} %s -o %t.o --nostdlib --target=sse2-i32x4 --cpu=znver3
+//; RUN: %{ispc} %s -o %t.o --nostdlib --target=sse2-i32x4 --cpu=znver4
```
aneshlya, line 25: Could you add check of `znver5` to https://github.com/ispc/ispc/blob/main/tests/lit-tests/cpus_x86_llvm20.ispc?

aneshlya:
> Why compat[CPU_ZNVER2] don't include CPU_ZNVER1 [here](https://github.com/ispc/ispc/blob/b8b8bf8b685852f6719abbc2d47e9d11651730c9/src/ispc.cpp#L657)? Same question with compat[CPU_ZNVER3] and CPU_ZNVER2.

This looks like a bug to me.

zephyr111:
I applied the fixes. 
Regarding `avx512icl_x16`, I also fixed this. I saw that this impacts `m_hasIntelVNNI`. IceLake should AFAIK support AVX-512 VNNI but not AVX-2 bit one, right? I wonder if `m_hasIntelVNNI` might enable 256-bit instructions and causes issue, especially if `disableZMM` is set. I do not have an IceLake CPU to test that carefully.
Note that the same thing is true on Zen4.

aneshlya (approved):
LGTM!

aneshlya:
> I wonder if m_hasIntelVNNI might enable 256-bit instructions and causes issue, especially if disableZMM is set. I do not have an IceLake CPU to test that carefully.

`avx512icl-x16` target might produce 512-bit instructions if `disableZMM` is set. But it is not a problem for this PR. It is known issue and it's spread across all avx512-x16 targets.

zephyr111:
Ok. I reverted the modification about `ISPCTarget::avx512icl_x16` (and applied a minor fix on Zen5). I will make another PR on the matter once this one will be merged.

PR #3427, opened 2025-06-15
Fix #2321: reports an error when accessing the component w of a short vector of size 3

### LINKED ISSUE #2321: No error is produced when reading from the `w` component in a `float<3>` datatype.
```
varying float<3> col = { 0, 0, 0 };
uniform float <3> reduce_col;
reduce_col.x = reduce_add( col.x );
reduce_col.y = reduce_add( col.y );
reduce_col.z = reduce_add( col.z );
reduce_col.w = reduce_add( col.w );
```

This currently does not produce an error.  I expect attempting to access the `w` component of a three component type to be an error.

zephyr111:
## Description

This PR fix #2321 as pointed out in the name. 
The newly reported errors for the code in the issue are now the following with this PR:

```ispc
/tmp/test.ispc:23:5: Error: Vector element identifier "w" is invalid for type "uniform float<3>" (out of bounds). 
    reduce_col.w = reduce_add(col.w);
    ^^^^^^^^^^^^

/tmp/test.ispc:23:31: Error: Vector element identifier "w" is invalid for type "varying float<3>" (out of bounds). 
    reduce_col.w = reduce_add(col.w);
                              ^^^^^
```

## Related Issue

## Checklist

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -5674,6 +5674,9 @@ llvm::Value *VectorMemberExpr::GetValue(FunctionEmitContext *ctx) const {
             int idx = lIdentifierToVectorElement(identifier[i]);
             if (idx == -1) {
                 Error(pos, "Invalid swizzle character '%c' in swizzle \"%s\".", identifier[i], identifier.c_str());
+            } else if (exprVectorType != nullptr && idx >= exprVectorType->GetElementCount()) {
+                Error(pos, "Invalid swizzle character '%c' for type \"%s\" (out of bounds).", identifier[i],
+                      exprVectorType->GetString().c_str());
             }
 
             indices.push_back(idx);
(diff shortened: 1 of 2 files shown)


src/expr.cpp
```diff
@@ -5737,11 +5740,14 @@ llvm::Value *VectorMemberExpr::GetValue(FunctionEmitContext *ctx) const {
 }
 
 int VectorMemberExpr::getElementNumber() const {
-    int elementNumber = lIdentifierToVectorElement(identifier[0]);
-    if (elementNumber == -1) {
+    int idx = lIdentifierToVectorElement(identifier[0]);
+    if (idx == -1) {
         Error(pos, "Vector element identifier \"%s\" unknown.", identifier.c_str());
+    } else if (exprVectorType != nullptr && idx >= exprVectorType->GetElementCount()) {
```
nurmukhametov, line 5746: I am wondering if we need to check this earlier, for example, during the construction of `VectorMemberExpr` with an incorrect subscript. In other words, does ISPC report errors for all other possible uses, e.g., like this: `sizeof(col.w)`?

aneshlya: Right, other error messages are constructed here: https://github.com/ispc/ispc/pull/3427/files#diff-080c8b0513a71a985e8647431edc401ac005e46ce039b830be9dc20fd073ff1fR5815

zephyr111: I agree that `sizeof(col.w)` was an issue and checking this earlier seems a good idea. That being said, I am not convinced calling `Error` in a constructor is a great idea. In fact, this is apparently never done in this file. It might have side effects.
Here, since there only one missing check, I found this a slightly better just to call `getElementNumber` after the object has been created and return `nullptr` if this is not fine (so to avoid error code replication). This matches well with the logic of `StructType`.
Ideally, It would be better not to create the `VectorMemberExpr` and check this just before.
What do you think about this?

PS: I added the `sizeof(col.w)` in the tests.

nurmukhametov: > I am not convinced calling `Error` in a constructor is a great idea. 

It looks like there are already quite a lot of errors reported from constructors but I don't have a strong opinion on this.

> I agree that sizeof(col.w) was an issue ...
> PS: I added the `sizeof(col.w)` in the tests.

Thanks for fixing this!

aneshlya (approved):

PR #3428, opened 2025-06-15
Fix #3234: sizeof of reference to an array correctly returns the sizeof the referenced array

### LINKED ISSUE #3234: Incorrect sizeof for reference to array
Consider the following code:

```ispc
uniform int foo(uniform int (&a)[7]) {
    return sizeof(a);
}
```

Compiler explorer link: https://ispc.godbolt.org/z/sq19rhdfs

At the moment, `foo` returns 8 (the pointer size), whereas it is supposed to return the size of the array (which is 28).

zephyr111:
## Description

This PR fixes #3234. If the target type is a reference, then the size of the target referenced type is extracted rather than the size of the reference.

## Related Issue

## Checklist

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -8703,6 +8703,10 @@ llvm::Value *SizeOfExpr::GetValue(FunctionEmitContext *ctx) const {
         return nullptr;
     }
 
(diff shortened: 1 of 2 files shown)


nurmukhametov (approved):
LGTM!

PR #3429, opened 2025-06-15
Fix ispc#3145: enable void functions to return expression of void type

### LINKED ISSUE #3145: Error: Can't return non-void type "void" from void
The following code produces an error: `Error: Can't return non-void type "void" from void`

```ispc
void foo() {
    return foo();
}
```

```bash
$ ispc t.ispc
<source>:2:12: Error: Can't return non-void type "void" from void 
        function. 
    return foo();
           ^^^^^

Compiler returned: 1
```

I suppose it shouldn't happen because void is a void type.

Compiler explorer link: https://godbolt.org/z/avK9Pr4oE

zephyr111:
## Description

This PR fixes #3145.
I do not think there is a need for lit-tests here since I just made the error clearer and it appears in the same exact case anyway (only the content of the message changes -- and it is specialized regarding the use-case). Do you agree?

## Related Issue
- [X] Linked to relevant issue(s)

## Checklist

### DIFF
--- a/src/ctx.cpp
+++ b/src/ctx.cpp
@@ -1283,7 +1283,15 @@ void FunctionEmitContext::CurrentLanesReturned(Expr *expr, bool doCoherenceCheck
         if (expr != nullptr) {
             const Type *exprType = expr->GetType();
             Assert(exprType);
-            Error(expr->pos, "Can't return non-void type \"%s\" from void function.", exprType->GetString().c_str());
+
+            if (exprType->IsVoidType()) {
+                // Force function calls instructions to be emitted even though
+                // the llvm::Value result is unused.
+                expr->GetValue(this);
+            } else {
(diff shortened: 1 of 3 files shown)


aneshlya:
I totally agree with comment https://github.com/ispc/ispc/issues/3145#issuecomment-2976440458. I think we should support `return` statement in `void` functions, not just improve error message. Even with proposed error message, this behavior is confusing comparing to C/C++.

zephyr111:
It should be fine now.
I added some tests since it is not that trivial now. Better be safe than sorry.

aneshlya (approved):
Looks good!

nurmukhametov (changes requested):
Excellent! Could you please address the formatting issues? Once that's done, we'll be ready to proceed with the merge.

nurmukhametov (approved):
LGTM!

Issue #3220, opened 2025-02-10
Crash when accessing return value wrapped in parenthesis


asbott:
The compiler does not handle the case of dereferencing ( `.` ) return values from structs. It either gives a compile error, or crashes if the function call is wrapped in `()`.

Minimal repro:

```
struct uint2 {
    uint x;
    uint y;
};

uint2 make_uint2(uint x, uint y) {
    uint2 v;
    v.x = x;
    v.y = y;
    return v;
}

uint2 repro(uint2 xy) {
    // A) Crash
    return (make_uint2(xy.x, xy.y)).y;
    
    // B) Error: syntax error, unexpected '.', expecting ',' or ';'
    // return make_uint2(xy.x, xy.y).y;
}
```

Compile with `ispc something.ispc -o something.obj`

A) outputs:
```
repro.ispc:16:12: Error: Assertion failed (D:\a\ispc\ispc\src\expr.cpp:5320): "m->errorCount > 0".
    return (make_uint2(xy.x, xy.y)).y;
           ^^^^^^^^^^^^^^^^^^^^^^^^^^

***
*** Please file a bug report at https://github.com/ispc/ispc/issues
*** (Including as much information as you can about how to reproduce this error).
*** You have apparently encountered a bug in the compiler that we'd like to fix!
***
```

B) outputs:
```
repro.ispc:19:34: Error: syntax error, unexpected '.', expecting ',' or ';'.
    return make_uint2(xy.x, xy.y).y;
                                 ^
```

Version:

```
c:\stuff>ispc --version
Intel(r) Implicit SPMD Program Compiler (Intel(r) ISPC), 1.26.0dev (build commit 190effe68a2d3a36 @ 20250127, LLVM 18.1.8)
Supported Visual Studio versions: Visual Studio 2015 and later.
```

System:
Windows 11 Pro
Version	10.0.26100 Build 26100
Processor	11th Gen Intel(R) Core(TM) i5-11300H @ 3.10GHz, 3110 Mhz, 4 Core(s), 8 Logical Processor(s)

nurmukhametov:
B is a duplicate of #2533, A looks like a unique crash to me.

At the moment, you can workaround it by something like this:

```ispc
uint repro(uint2 xy) {
    uint2 s = make_uint2(xy.x, xy.y);
    return s.y;
}
```

Compile explorer link: https://ispc.godbolt.org/z/1r31e4575

Issue #2533, opened 2023-06-01
Accessing struct member in the function call causes syntax error


aneshlya:
Accessing struct member right in the function call causes ISPC parser syntax error:
```
struct X {
    int x;
};

uniform X getx() {
    uniform X x;
    return x;
}

export uniform int test(){
    return getx().x;
}
```
Error: syntax error, unexpected '.', expecting ',' or  ';'. 
    return getx().x;

PR #3432, opened 2025-06-16
Support struct access on function call returns

aneshlya:
Support struct access on function call returns (which is especially important when dealing with operators).
Fixes https://github.com/ispc/ispc/issues/3220 and https://github.com/ispc/ispc/issues/2533

## Related Issue

## Checklist

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -5429,17 +5429,17 @@ const Type *StructMemberExpr::GetTypeImpl() const {
         return type;
     }
 
-    // It's a struct, and the result type is the element type, possibly
-    // promoted to varying if the struct type / lvalue is varying.
-    const Type *exprType = nullptr, *lvalueType = nullptr;
+    const Type *exprType = nullptr;
     const StructType *structType = nullptr;
-    if (expr == nullptr || ((exprType = expr->GetType()) == nullptr) || ((structType = getStructType()) == nullptr) ||
-        ((lvalueType = GetLValueType()) == nullptr)) {
+
+    // It's a struct, and the result type is the element type, possibly
+    // promoted to varying if the struct type is varying.
+    if (expr == nullptr || ((exprType = expr->GetType()) == nullptr) || ((structType = getStructType()) == nullptr)) {
         AssertPos(pos, m->errorCount > 0);
         return nullptr;
     }
 
-    if (exprType->IsDependent() || structType->IsDependent() || lvalueType->IsDependent()) {
+    if (exprType->IsDependent() || structType->IsDependent()) {
         return AtomicType::Dependent;
     }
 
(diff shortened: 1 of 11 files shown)


tests/func-tests/operators5.ispc
```diff
@@ -40,12 +40,10 @@ task void f_f(uniform float RET[], uniform float aFOO[]) {
...
 
     // Test prefix increment
-    ++a;
-    RET[programIndex] = a.a;
+    RET[programIndex] = ++a.a;
 
     // Test prefix decrement
-    --a;
-    RET[programIndex] += a.a;
+    RET[programIndex] += --a.a;
```
nurmukhametov: Am I correct that it is (--a).a ? If so, then it seems strange to me because it is not aligned with the precedence of C/C++ operators (compiler explorer [link](https://ispc.godbolt.org/z/1avrMvqez))

src/parse.yy
```diff
@@ -11,7 +11,18 @@
 /* We expect these conflicts and accept the default resolution (shift). It means
 that for the example 'if (a) if (b) c; else d;' the else will be associated
 with the nearest if. Same happens for 'if (a) cif (b) c; else d;'.  */
-%expect 2
+
+/* supress 5 shift-reduces conflict messages for postfix and funccall expressions */
```
nurmukhametov: Can you provide examples when these conflicts happen? Bison can usually generate them using the `-Wcounterexamples` flag.

src/parse.yy
```diff
@@ -11,7 +11,18 @@
...
+/* The dependency chain looks like this:
+
+funcall_expression -> postfix_expression (first rule)
+postfix_expression -> various expressions that can include function calls
```
nurmukhametov: I am not sure I follow this chain. I can't find where `postfix_expression` includes `function_call`.

src/expr.cpp
```diff
@@ -4337,13 +4365,18 @@ const Type *FunctionCallExpr::GetLValueType() const {
...
-    } else {
-        // Only be a valid LValue type if the function
-        // returns a pointer or reference.
-        return nullptr;
+    if (ftype) {
+        const Type *returnType = ftype->GetReturnType();
+        if (returnType->IsPointerType() || returnType->IsReferenceType()) {
+            return returnType;
+        }
+        const StructType *structType = CastType<StructType>(returnType);
```
nurmukhametov: I don't think it's correct to treat a returned struct value as an lvalue. Moreover, the original behavior of treating returned pointers as lvalues also seems wrong to me. See the following [link](https://ispc.godbolt.org/z/ofWx8dhs4).

tests/func-tests/operators5.ispc
aneshlya: I forgot the brackets 😆. Please see updated.

src/parse.yy
aneshlya: [parser.log.txt](https://github.com/user-attachments/files/20782006/parser.log.txt)

src/expr.cpp
aneshlya: I reworked this. Also I've added lvalue restrictions on function returns to `member_access_errors.ispc` test

tests/func-tests/operators5.ispc
nurmukhametov: Ok, it is good that is not precedence issue.

src/parse.yy
nurmukhametov: > [parser.log.txt](https://github.com/user-attachments/files/20782006/parser.log.txt)

This is the most suspicious area for me, especially the first example, but I tried some examples and they seem fine to me.

nurmukhametov (approved):
LGTM!

PR #3433, opened 2025-06-18
Improve shuffle avx512

### LINKED ISSUE #3437: Improve shuffle for AVX-512
Cover all types and widths

nurmukhametov:
## Description

Improve implementations of `shuffle(T, perm)` for avx512 targets. This relates to #2924 and continues target specific optimization of shuffles started there. This fixes #3437.

The overall shuffle implementation for AVX-512 is the following now:

• i8 shuffles: vpshufb for small vectors, vpermb for ICL x16+,
  vpermw/vpermi2w for SKX larger vectors
• i16/half shuffles: vpermw consistently, vpermi2w for x64
• i32/float shuffles: vpermilps → vpermps → vpermi2d/vpermi2ps → vpgatherdd
• i64/double shuffles: vpermq → vpermi2q/vpermi2pd → vpgatherqq/vpgatherdq

Key improvements:
- Eliminates inefficient extract/insert instruction sequences using AVX-512 instructions for each type/width combination.
- Uses vpermb (AVX-512 VBMI) for ICL for efficient i8 shuffles for 16+  width
- Uses gather instructions for very wide vectors (x32/x64) because there is no instructions to cover cross lanes shuffles wider than 1024 overall.

|                | avx512skx-x4 | avx512skx-x8 | avx512skx-x16 | avx512skx-x32 | avx512skx-x64 |
|----------------|--------------|--------------|---------------|---------------|---------------|
| shuffle1_int8  | -65          | -86          | -86           | -88           | -87           |
| shuffle1_int16 | -69          | N/A          | N/A           | -88           | -95           |
| shuffle1_int32 | N/A          | N/A          | N/A           | -89           | -73           |
| shuffle1_int64 | N/A          | -80          | -81           | -68           | -85           |
| shuffle1_float16* | -79       | -88          | -89           | -84           | -97           |
| shuffle1_float | N/A          | N/A          | N/A           | -89           | -73           |
| shuffle1_double| N/A          | -80          | -81           | -68           | -85           |

|                | avx512icl-x4 | avx512icl-x8 | avx512icl-x16 | avx512icl-x32 | avx512icl-x64 |
|----------------|--------------|--------------|---------------|---------------|---------------|
| shuffle1_int8  | -75          | -76          | N/A           | -90           | -94           |
| shuffle1_int16 | N/A          | N/A          | N/A           | -88           | -95           |
| shuffle1_int32 | N/A          | N/A          | N/A           | -89           | -73           |
| shuffle1_int64 | -79          | -83          | -81           | -68           | -85           |
| shuffle1_float16* | -79       | -88          | -89           | -89           | -96           |
| shuffle1_float | N/A          | N/A          | N/A           | -89           | -73           |
| shuffle1_double| -79          | -83          | -80           | -68           | -85           |

Measured on Intel(R) Xeon(R) Platinum 8360Y CPU except float16 tests that were measured on Zen4.

## Related Issue
- [X] Linked to relevant issue(s)

## Checklist


builtins/target-avx512skx-x8.ll
```diff
@@ -37,6 +57,48 @@ define <WIDTH x float> @__shuffle_float(<WIDTH x float>, <WIDTH x i32>) nounwind
...
+    ; Two VPERMI2PD using corresponding indices  (0-3 = low, 4-7 = high)
+    %result1 = call <4 x double> @llvm.x86.avx512.mask.vpermi2var.pd.256(<4 x double> %low, <4 x i64> %indices1, <4 x double> %high, i8 15)
+    %result2 = call <4 x double> @llvm.x86.avx512.mask.vpermi2var.pd.256(<4 x double> %low, <4 x i64> %indices2, <4 x double> %high, i8 15)
+
+    ; Concatenate results
+    %final = shufflevector <4 x double> %result1, <4 x double> %result2, <8 x i32> <i32 0, i32 1, i32 2, i32 3, i32 4, i32 5, i32 6, i32 7>
+    ret <8 x double> %final
+}
+
+declare <8 x i64> @llvm.x86.avx512.mask.vpermi2var.q.256(<4 x i64> %a, <4 x i64> %idx, <4 x i64> %b, i8 %mask)
```
aneshlya: should it be declare **<4 x i64>** @llvm.x86.avx512.mask.vpermi2var.q.256(<4 x i64> %a, <4 x i64> %idx, <4 x i64> %b, i8 %mask)?

builtins/target-avx512skx-x64.ll
```diff
@@ -55,7 +55,107 @@ define i1 @__none(<WIDTH x MASK> %mask) nounwind readnone alwaysinline {
...
+}
+
+define <WIDTH x half> @__shuffle_half(<WIDTH x half>, <WIDTH x i32>) nounwind readnone alwaysinline {
+  %vals = bitcast <WIDTH x half> %0 to <WIDTH x i16>
+  %res = call <WIDTH x i16> @__shuffle_i16(<WIDTH x i16> %vals, <WIDTH x i32> %1)
+  %res_half = bitcast <WIDTH x i16> %res to <WIDTH x half>
+  ret <WIDTH x half> %res_half
+}
+
+declare <64 x i32> @llvm.masked.gather.v64i32.v64p0i32(<64 x i32*>, i32, <64 x i1>, <64 x i32>)
```
aneshlya: It probably doesn't matter much but in `v64p0i32` do we really need i32?

builtins/target-avx512skx-x4.ll
```diff
@@ -12,9 +12,47 @@ include(`target-avx512-utils.ll')
...
+  ; Use vpshufb to perform the shuffle
+  %result16 = call <16 x i8> @llvm.x86.ssse3.pshuf.b.128(<16 x i8> %data16, <16 x i8> %mask16)
+
+  %result8 = shufflevector <16 x i8> %result16, <16 x i8> undef, <8 x i32> <i32 0, i32 1, i32 2, i32 3, i32 4, i32 5, i32 6, i32 7>
+  %result = shufflevector <8 x i8> %result8, <8 x i8> undef, <4 x i32> <i32 0, i32 1, i32 2, i32 3>
+  ret <4 x i8> %result
+}
+
+; Use vpermw to perform the shuffle of i16 vectors.
+declare <WIDTH x i16> @llvm.x86.avx512.mask.permvar.hi.128(<WIDTH x i16>, <WIDTH x i16>, <WIDTH x i16>, i8)
```
aneshlya: Here the WIDTH is 4 so it's declared as 
`declare <4x i16> @llvm.x86.avx512.mask.permvar.hi.128(<4x i16>, <4x i16>, <4x i16>, i8)`
and then it is used as <8 x i16>.

I'm wondering why incorrect declarations didn't cause any compilation problems.

builtins/target-avx512skx-x4.ll
```diff
@@ -12,9 +12,47 @@ include(`target-avx512-utils.ll')
...
+define <4 x i16> @__shuffle_i16(<4 x i16> %src, <4 x i32> %indices) nounwind readnone alwaysinline {
+  %src_ext = shufflevector <4 x i16> %src, <4 x i16> undef, <8 x i32> <i32 0, i32 1, i32 2, i32 3, i32 4, i32 5, i32 6, i32 7>
+  %indices_ext = shufflevector <4 x i32> %indices, <4 x i32> undef, <8 x i32> <i32 0, i32 1, i32 2, i32 3, i32 4, i32 5, i32 6, i32 7>
+  %ind = trunc <8 x i32> %indices_ext to <8 x i16>
+  %res_ext = call <8 x i16> @llvm.x86.avx512.mask.permvar.hi.128(<8 x i16> %src_ext, <8 x i16> %ind, <8 x i16> zeroinitializer, i8 15)
```
aneshlya: Seems to me that we should set all bits in mask here: i8 15 -> i8 -1

builtins/target-avx512skx-x32.ll
```diff
@@ -56,7 +56,100 @@ define i1 @__none(<WIDTH x MASK> %mask) nounwind readnone alwaysinline {
...
+; Use vpermw for 16-bit shuffles.
+declare <WIDTH x i16> @llvm.x86.avx512.mask.permvar.hi.512(<WIDTH x i16>, <WIDTH x i16>, <WIDTH x i16>, i16)
+define <WIDTH x i16> @__shuffle_i16(<WIDTH x i16>, <WIDTH x i32>) nounwind readnone alwaysinline {
+  %ind = trunc <WIDTH x i32> %1 to <WIDTH x i16>
+  %res = call <WIDTH x i16> @llvm.x86.avx512.mask.permvar.hi.512(<WIDTH x i16> %0, <WIDTH x i16> %ind, <WIDTH x i16> zeroinitializer, i16 -1)
+  ret <WIDTH x i16> %res
+}
+
+define <WIDTH x i8> @__shuffle_i8(<WIDTH x i8>, <WIDTH x i32>) nounwind readnone alwaysinline {
+  %vals = zext <WIDTH x i8> %0 to <WIDTH x i16>
```
aneshlya: I don't think we can safely use zext/trunc here since they don't preserve the exact bit patterns. Same for x64 target.

builtins/target-avx512skx-x32.ll
```diff
@@ -56,7 +56,100 @@ define i1 @__none(<WIDTH x MASK> %mask) nounwind readnone alwaysinline {
...
+
+    ; Concatenate results
+    %final = shufflevector <16 x float> %result1, <16 x float> %result2, <32 x i32> <i32 0, i32 1, i32 2, i32 3, i32 4, i32 5, i32 6, i32 7, i32 8, i32 9, i32 10, i32 11, i32 12, i32 13, i32 14, i32 15, i32 16, i32 17, i32 18, i32 19, i32 20, i32 21, i32 22, i32 23, i32 24, i32 25, i32 26, i32 27, i32 28, i32 29, i32 30, i32 31>
+    ret <32 x float> %final
+}
+
+declare <32 x i64> @llvm.masked.gather.v32i64.v32p0i64(<32 x i64*>, i32, <32 x i1>, <32 x i64>)
+define <32 x i64> @__shuffle_i64(<32 x i64> %input, <32 x i32> %perm) nounwind readnone alwaysinline {
+  ; Store input vector to memory so we can gather from it
+  %input_alloca = alloca [32 x i64], align 64
```
aneshlya, line 114: Big allocation with alignment can be slow. Have you compared this implementation with multiple 4x shufflevectors?

aneshlya (changes requested):
Just an observation: these target files use a mix of explicit numerical widths like <16 x i32> and macros like <WIDTH x i32>. This inconsistency makes the code more error-prone.

nurmukhametov:
> Just an observation: these target files use a mix of explicit numerical widths like <16 x i32> and macros like . This inconsistency makes the code more error-prone.

What should I use for consistency?

builtins/target-avx512skx-x32.ll
nurmukhametov: > Big allocation with alignment can be slow. 

This is stack memory. 

> Have you compared this implementation with multiple 4x shufflevectors?

I'm not quite sure which implementation with 4x shufflevectors you are referring to.

builtins/target-avx512skx-x4.ll
nurmukhametov: Yes, I need to fix these declarations. They are pure copy-paste errors. AFAIK, LLVM rules for declarations of intrinsics are quite relaxed now. If I am not mistaken, we can skip their declaration entirely since LLVM 18. They are likely derived unambiguously from the call site.

aneshlya:
> What should I use for consistency?

I like numerical values more. Using macro `WIDTH` makes sense only in the code shared across different targets and utils file.

builtins/target-avx512skx-x32.ll
aneshlya: I mean something like this:
```
 %input_0 = shufflevector <32 x i64> %input, <32 x i64> undef, <8 x i32> <i32 0, i32 1, i32 2, i32 3, i32 4, i32 5, i32 6, i32 7>
  %input_1 = ...
  %input_2 = ...
  %input_3 = ...

;   Same for permutation indexes
  %perm_0 = shufflevector <32 x i32> %perm, <32 x i32> undef, <8 x i32> <i32 0, i32 1, i32 2, i32 3, i32 4, i32 5, i32 6, i32 7>
  %perm_1 = ...
  %perm_2 = ...
  %perm_3 = ...

  ; Perform 4 separate 8-wide shuffles using vpermi2q 
  ; Combine results
```

builtins/target-avx512skx-x8.ll
nurmukhametov: Fixed

builtins/target-avx512skx-x64.ll
nurmukhametov: Not really, since pointer are opaque. Fixed.

builtins/target-avx512skx-x4.ll
nurmukhametov: Fixed

builtins/target-avx512skx-x4.ll
nurmukhametov: This is for 4-wide vectors, so it is actually the same because the upper part is discarded anyway. However, I have changed it to -1 for consistency with the others.

builtins/target-avx512skx-x32.ll
nurmukhametov: Why not? I don't see any issues with zext/trunc here. zext extends with zeros, preserving the lower 8 bits, and then trunc discards these zeros.

builtins/target-avx512skx-x32.ll
nurmukhametov: This implementation doesn't work for all permutations.

The `vpermi2q` instruction permutes values from two input ZMM registers (2 x 512 bits = 1024 bits, i.e., <16 x i64>), so we can only refer to up to 16-wide i64 vectors. Here, we have 32, so we can only split the initial vector into two chunks and perform intra-chunk permutations.

aneshlya (approved):
LGTM!

builtins/target-avx512skx-x32.ll
aneshlya: Right, I was mistaken, the original bit pattern is indeed preserved

PR #3444, opened 2025-06-24
Fix #3439: function calls returning a pointer type is treated as an lvalue

### LINKED ISSUE #3439: The pointer return type is treated as an lvalue.
ISPC treats the return value of a pointer type as an lvalue, which seems incorrect to me: https://ispc.godbolt.org/z/ofWx8dhs4.

zephyr111:
## Description

As the name suggest, this fixes the issue #3439. Details are provided in an answer of the issue.
I am going to add 2 lit-tests later. 

## Related Issue

## Checklist


aneshlya (approved):
LGTM!

zephyr111:
I added the 2 missing lit-tests (as expected) so it should be Ok now.

tests/lit-tests/3439-1.ispc
```diff
@@ -0,0 +1,39 @@
...
+
+static uniform int *uniform foo() {
+    return &val;
+}
+
+static void bar() {
+    foo() = &val2;
+}
+
+static void bar2() {
```
nurmukhametov, line 25: There is also an error report here that needs to be checked:

```bash
t.ispc:13:5: Error: Can't assign to left-hand side of expression.
    foo() += 1;
    ^^^^^^^^^^
```

zephyr111: Indeed. Fixed. Thank you.

PR #3445, opened 2025-06-24
Docs fix for bold sections

### LINKED ISSUE #3440: Doc parsing issue with the new template
The new doc template is significantly better, but there is a small parsing issue with it: the `mono-spaced code` does not work in bold sentences. Here is what can be currently seen (not the `ispc.rst` code): 

```
Like other types in ``ispc``, pointers are ``varying`` by default, if an explicit ``uniform`` qualifier isn't provided. However, the default variability of the pointed-to type is uniform. This rule will be illustrated and explained in examples below.
```

A trivial workaround it to replace the \`\` with " (except for `ispc`), but it would be better if it was supported (e.g. this avoid mistake when existing sentence are updated to be bold one).

aneshlya:
## Description
This PR fixes formatting of code inside bold sections. 
Fixes #3440.
Now it will look like this:
Screenshot 2025-06-23 203552

## Related Issue

## Checklist

### DIFF
--- a/docs/ispc.rst
+++ b/docs/ispc.rst
@@ -2460,9 +2460,9 @@ changing values in memory with pointers, and so forth is supported as in C.
 As with other basic types, pointers can be both ``uniform`` and
 ``varying``.
 
-**Like other types in ``ispc``, pointers are ``varying`` by default, if an
-explicit ``uniform`` qualifier isn't provided.  However, the default
-variability of the pointed-to type is uniform.** This rule will be
+**Like other types in ispc, pointers are** ``varying`` **by default, if an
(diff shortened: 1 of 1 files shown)


nurmukhametov (approved):
LGTM!

zephyr111:
@aneshlya Thank you for the fix. I just have an naive question: why do we still see the old content in https://ispc.github.io/ispc.html while the PR has been merged. Is this automatically updated in a deferred way?

aneshlya:
@zephyr111, it's not updated automatically. We usually update the website content after the release, as we prefer to keep new features (and other updates added to ispc.rst during development) hidden until the official binary is available.
I'll update it manually later this week.

PR #3448, opened 2025-06-25
Fix the list of reserved keywords in the doc

zephyr111:
## Description

Everything is in the name.
The list of keywords was not up to date and also incorrect. For example, `char` was in the list while it is not a reserved keyword (explicitly mentioned earlier in the doc so it was inconsistent and confusing). Moreover, `template` and `typename` were missing in the first list while they was inconsistently added in the second one. For `NULL`, it is the opposite. Additionally, `alloca` was missing in both. There are few other issues like this in these two lists.

Interestingly, while `assert` is mentioned [here](https://github.com/ispc/ispc/blob/31b620ac6a8220333880a1272ccd7e27a4526da6/src/lex.ll#L50) as a reserved token, it does not seems to actually be one : https://ispc.godbolt.org/z/Ebdd9cdMe . This looks like a bug to me.
Thus, I did not add `assert` yet to the list.

By the way, I wonder how relevant it is to keep the first list in the doc. Keeping simultaneously the two lists in the doc is prone to mistakes. What do you think about this?

## Related Issue

## Checklist

### DIFF
--- a/docs/ispc.rst
+++ b/docs/ispc.rst
@@ -2076,12 +2076,13 @@ The following reserved words from C89 are also reserved in ``ispc``:
 
 ``ispc`` additionally reserves the following words:
 
-``bool``, ``delete``, ``export``, ``cdo``, ``cfor``, ``cif``, ``cwhile``,
+``bool``, ``cdo``, ``cfor``, ``cif``, ``cwhile``, ``delete``, ``export``,
 ``false``, ``float16``, ``foreach``, ``foreach_active``, ``foreach_tiled``,
-``foreach_unique``, ``in``, ``inline``, ``noinline``, ``__regcall``,
-``__vectorcall``, ``int8``, ``int16``, ``int32``, ``int64``, ``launch``,
-``new``, ``print``, ``uint8``, ``uint16``, ``uint32``, ``uint64``, ``soa``,
-``__attribute__``, ``sync``, ``task``, ``true``, ``uniform``, and ``varying``.
+``foreach_unique``, ``in``, ``inline``, ``int8``, ``int16``, ``int32``,
+``int64``, ``invoke_sycl``, ``launch``, ``new``, ``noinline``, ``print``,
+``soa``, ``sync``, ``task``, ``template``, ``true``, ``typename``, ``uint8``,
+``uint16``, ``uint32``, ``uint64``, ``uint``, ``uniform``, ``unmasked``,
+``varying``, ``__attribute__``, ``__regcall``, ``__vectorcall``.
 
 
 Lexical Structure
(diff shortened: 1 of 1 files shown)


nurmukhametov (commented):
Why `alloca` is reserved keyword? I though it is just more or less the standard library function. Moreover, `assume` and `assert` are almost identical to it.

aneshlya:
> Why `alloca` is reserved keyword? I though it is just more or less the standard library function. Moreover, `assume` and `assert` are almost identical to it.

This list of keywords should be probably be aligned with `tokenToName` in lex.ll. All these expressions `alloca`, `assume` and `assert` can be used without standard library, they are compiler-supported expressions.

zephyr111:
Yes, [`alloca` is in `lex.ll`](https://github.com/ispc/ispc/blob/31b620ac6a8220333880a1272ccd7e27a4526da6/src/lex.ll#L102) like `assert` but both do not behave the same strangely. `assume` is not in the file and is not a reserved keyword like `assert` in practice : https://ispc.godbolt.org/z/d1Pbb1Wjh. Maybe this is unexpected?

Then the questions are:
 - Which of them should be considered as keywords? IMHO I expect the three to be in the same list.
 - Shouldn't the lexer/compiler be modified to match with the expectations? (i.e. should we open an issue for this)

aneshlya:
After looking into C, I think we should follow the same concept in ISPC, no matter how these builtins are implemented in ISPC.
In C they are not keywords and you can declare variables using all these names: https://ispc.godbolt.org/z/KhPcj4xxY

So in ISPC it's almost like this but we probably should not produce error for `alloca` here: https://ispc.godbolt.org/z/5ffvnrenq

zephyr111:
I tend to agree.

I found why `assert` behave differently and it looks like a sleight of hand trick (with side effects).
Indeed, the pre-processor definition [here](https://github.com/ispc/ispc/blob/841579e06a70da010888d214287a8b4b98bc171d/stdlib/include/core.isph#L36) transform the macro `assert(...)` to `__assert(...)` and `__assert` is the [actual reserved keyword](https://ispc.godbolt.org/z/3hG7j3Yv8) reported as an "assert".
We can do the same trick with `alloca` : define an `alloca(...)` macro substituted  to the reserved `__alloca` token reported as  "alloca". However, this is just a renaming trick : there is a reserved keyword for both.
Should we do that and if so, shouldn't we report `__assert` and `__alloca` as reserved keyword like we do for `__regcall`, `__attribute__`, etc. ?

zephyr111:
An alternative solution is not to have them in the lexer and treat them as intrinsics in the parsing step while having condition in this step to to substitute AST nodes to `AllocaExpr` and a possibly new `AssertExpr`. However, this seems a non-trivial bug-prone change for such a minor issue IMHO (I do not expect people to name their variable `assert` or `alloca`, and no one to complain about this in practice).

aneshlya:
I agree that it's very minor issue and we should not spend too much time on it. I think that implementing the same approach which is used for `assert` would be fine. As for this PR, `alloca` should be removed from the list of keywords.

zephyr111:
I fixed the list and opened a PR (#3454) as expected.

nurmukhametov (approved):

PR #3451, opened 2025-06-27
Add auxiliary scripts for benchmarking

nurmukhametov:
- build_benchmarks.sh: Automated build script for ISPC benchmarks for specified targets between two ISPC revisions
- create_benchmark_comparison_table.py: Python tool to parse benchmark JSON results and generate comparison tables in ASCII/CSV/Markdown
- run_all_benchmark_comparisons.sh: Batch runner that executes benchmark comparisons for all target pairs in the current directory
- run_and_compare_benchmarks.sh: Core comparison script that runs benchmarks and uses Google Benchmark's compare.py tool to analyze performance differences

## Checklist


aneshlya:
Could you add example usage/help to shell scripts?

aneshlya (approved):
LGTM!
Just a note that the scripts don't have copyrights but it seems that it's not a requirement for our `scripts` folder.

nurmukhametov:
> LGTM! Just a note that the scripts don't have copyrights but it seems that it's not a requirement for our `scripts` folder.

I have added copyrights to these scripts.

PR #3453, opened 2025-06-27
Extract compiler options parsing from main()

aneshlya:
- Extracted command-line argument parsing logic into a new function `ParseCommandLineArgs`, simplifying the `main` function 
- Replaced `exit()` calls with a new `ArgsParseResult` enum.

Based on https://github.com/ispc/ispc/pull/3452 and required for further refactoring for support of ISPC as a library.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -597,97 +597,35 @@ void lFreeArgv(std::vector<char *> &argv) {
 
 extern int yydebug;
 
-int main(int Argc, char *Argv[]) {
-    std::vector<char *> argv;
-    lGetAllArgs(Argc, Argv, argv);
-    int argc = argv.size();
-#ifdef ISPC_HOST_IS_WINDOWS
-    // While ispc doesn't load any libraries explicitly using LoadLibrary API (or alternatives), it uses vcruntime that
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -1367,23 +1301,118 @@ int main(int Argc, char *Argv[]) {
...
+    Arch arch = Arch::none;
+    std::vector<ISPCTarget> targets;
+    const char *cpu = nullptr;
+
+    std::string ISPCAbsPath = llvm::sys::fs::getMainExecutable(argv[0], (void *)(intptr_t)main);
+    initializeBinaryType(ISPCAbsPath.c_str());
+
+    // Parse command line optionsAdd commentMore actions
+    bool isLinkMode = false;
+    ArgsParseResult parseResult = ParseCommandLineArgs(
```
nurmukhametov: What if `ParseComandLineArgs` directly fills the `Module::Output` in the first place:

```c++
    ArgsParseResult parseResult = ParseCommandLineArgs(argc, argv.data(), file, arch, cpu, targets, output, linkFileNames, isLinkMode);
```

aneshlya: I planned to do it as part of another refactoring PR but it makes sense here.

PR #3456, opened 2025-06-29
[WIP] Fix #3446: implement shuffle, rotate and shift for short vectors

### LINKED ISSUE #3446: Implement shuffle, rotate and shift for short uniform vectors
These functions could be added to stdlib:

```ispc
template <typename T, uint N>
uniform T<N> shuffle(uniform T<N> v, uniform int<N> perm);

template <typename T, uint N>
uniform T<N> shuffle(uniform T<N> v1, uniform T<N> v2, uniform int<N> perm);

template <typename T, uint N>
uniform T<N> rotate(uniform T<N> v, uniform int r);

template <typename T, uint N>
uniform T<N> shift(uniform T<N> v, uniform int r);
```

See the discussion in https://github.com/ispc/ispc/issues/3418 for more context.

zephyr111:
## Description

This PR address the issue #3446 . It is based on #3485. This is a work-in-progress PR so to get some *feedback*. Indeed, I do not know if this is the best approach to implement these functions. Besides, the performance of some functions can be a concern in some pathological cases (maybe we can optimize them later). I wrote some NOTE comment about open questions.

Overall, the new functions are fast on "simple-pumping" targets when `N < programCount` (often not that bad with `N < 2*programCount`) and the provided function argument (i.e. `perm` or `offset`) is a compile-time constant especially.

Some thought that are not in the code:
 - Does using conditions like `if(N < programCount)` is the best solution as ? It might increase the size of the generated code before optimizations (so the compilation time). Using template for that does not seems possible though since partial specialization is AFAIK not yet implemented.
 - Should such functions be implemented in stdlib.isph or as builtins? I think it cannot be builtins due to templates. Maybe some builtins can be written for some frequent cases if needed later.

I am going to write functional tests later. 

By the way, I also added some previously-missing `#undef`.

## Related Issue

## Checklist

### DIFF
--- a/stdlib/include/short_vec.isph
+++ b/stdlib/include/short_vec.isph
@@ -159,3 +159,194 @@ SHORT_VEC_ARG2(fmod)
 #undef SHORT_VEC_ARG1_BOOLRET
 #undef SHORT_VEC_ARG2
 #undef SHORT_VEC_ARG3
+
+// Short vector functions with a custom implementation
+
+template <typename T, uint N>
+uniform T<N> shuffle(uniform T<N> v, uniform int<N> perm) {
+    if(N <= programCount) {
+        varying T var_v;
+        varying int var_perm;
+
+        foreach(i = 0 ... N) {
+            var_v = v[i];
+            var_perm = perm[i];
+        }
+
+        // NOTE:
+        // Using a mask here can make the resulting assembly code slighly less
+        // efficient. That being said, this is certainly needed for generic 
+        // targets or possibly with double pumping.
+        // We should be careful about possible out-of-bounds acesses.
+        // Technically, this is AFAIK undefined behavior independently of the target.
+        if(programIndex < N) {
+            var_v = shuffle(var_v, var_perm);
+        }
+
+        foreach(i = 0 ... N) {
+            v[i] = var_v;
+        }
+        
+        return v;
+    }
+    else if(N <= 2 * programCount) {
+        varying T var1_v;
+        varying T var2_v;
+        varying int var1_perm;
+        varying int var2_perm;
+
+        foreach(i = 0 ... programCount) {
+            var1_v = v[i];
+            var1_perm = perm[i];
+        }
+
+        foreach(i = programCount ... N) {
+            var2_v = v[i];
+            var2_perm = perm[i];
+        }
+
+        varying T res_var1_v;
+        varying T res_var2_v;
+
+        res_var1_v = shuffle(var1_v, var2_v, var1_perm);
+
+        // See the note above about out-of-bounds accesses
+        if(programCount + programIndex < N) {
+            res_var2_v = shuffle(var1_v, var2_v, var2_perm);
+        }
+
+        foreach(i = 0 ... programCount) {
+            v[i] = res_var1_v;
+        }
+
+        foreach(i = programCount ... N) {
+            v[i] = res_var2_v;
+        }
+
+        return v;
+    }
+    else
+    {
+        uniform T<N> result;
+
+        // NOTE: slow implementation
+        foreach (i = 0 ... N) {
+            result[i] = v[perm[i]];
+        }
+
+        return result;
+    }
+}
+
+// TODO: to be check further
+template <typename T, uint N>
+uniform T<N> shuffle(uniform T<N> v1, uniform T<N> v2, uniform int<N> perm) {
+    if(N <= programCount) {
+        uniform T<N> result;
+        varying T var_v1;
+        varying T var_v2;
+        varying T var_result;
+        varying int var_perm;
+
+        foreach(i = 0 ... N) {
+            var_v1 = v1[i];
+            var_v2 = v2[i];
+            var_perm = perm[i];
+        }
+
+        // NOTE: see the above note about out-of-bound.
+        if(programIndex < N) {
+            var_result = shuffle(var_v1, var_v2, var_perm);
+        }
+
+        foreach(i = 0 ... N) {
+            result[i] = var_result;
+        }
+        
+        return result;
+    }
+    else
+    {
+        uniform int<N> perm1;
+        uniform int<N> perm2;
+
+        foreach (i = 0 ... N) {
+            // NOTE: `select` is only meant to avoid out-of-bounds. See the above note about it.
+            perm1[i] = select(perm[i] < N, perm[i], 0);
+            perm2[i] = select(perm[i] >= N, perm[i], 0);
+        }
+
+        const uniform T<N> res1 = shuffle(v1, perm1);
+        const uniform T<N> res2 = shuffle(v2, perm2);
+        return select(perm < N, res1, res2);
+    }
+}
+
+template <typename T, uint N>
+uniform T<N> rotate(uniform T<N> v, uniform int offset)
+{
+    uniform T<N> result;
+    uniform int<N> perm;
+
+    // NOTE: 
+    // The modulus is meant to support an extended range (i.e. beyond the ]-N;N[ interval),
+    // but this is expensive when `offset` is not a compile-time constant.
+    // Thus, should we use it or tell users this is not supported?
+    // `assume` can be used to reduce the overhead of this modulus.
+    offset = offset % N;
+    offset += select(offset < 0, (uniform int)N, 0);
+
+    foreach (i = 0 ... N) {
+        varying int var_offset = offset + i;
+        var_offset -= select(var_offset >= (uniform int)N, (uniform int)N, 0);
+        perm[i] = var_offset;
+    }
+
+    return shuffle<T, N>(v, perm);
+}
+
+// TODO: to be check further
+template <typename T, uint N>
+uniform T<N> shift(uniform T<N> v, uniform int offset) {
+    uniform T<N> result;
+    uniform int<N> perm;
+
+    // NOTE: using a the shift built-in is surprisingly quite inefficient for constant offsets
+#if 0 // TODO
+    if(N <= programCount)
+    {
+        uniform T<N> result;
+
+        varying T var_v;
+
+        foreach(i = 0 ... N) {
+            var_v = v[i];
+        }
+
+        var_v = shift(var_v, offset);
+
+        foreach(i = 0 ... N) {
+            result[i] = var_v;
+        }
+
+        return result;
+    }
+#endif
+
+    foreach (i = 0 ... N) {
+        // NOTE: `clamp` is only meant to avoid out-of-bounds. See the above note about it.
+        perm[i] = clamp(offset + i, (int)0, (int)(N-1));
+    }
+
+    // NOTE: this approach is fast for constant offsets on small vectors, but slow for variables offsets
+    const uniform T<N> shifted = shuffle<T, N>(v, perm);
+
+    // NOTE: we can specialize the loop so to consider either a positive offset, 
+    // or negative one, in order to avoid a cmp+and SIMD instructions.
+    // However, this does not seems much more efficient with this at first glance.
+    foreach (i = 0 ... N) {
+        result[i] = select(i >= -offset && i < N-offset, shifted[i], (varying T)0);
+    }
+
+    return result;
+}


stdlib/include/stdlib.isph
```diff
@@ -1430,3 +1432,234 @@ SHORT_VEC_ARG1(cbrt)
...
+#undef SHORT_VEC_UNIFORM_ARG1_BOOLRET
+#undef SHORT_VEC_VARYING_ARG1_BOOLRET
+#undef SHORT_VEC_UNIFORM_ARG2
+#undef SHORT_VEC_VARYING_ARG2
+#undef SHORT_VEC_UNIFORM_ARG3
+#undef SHORT_VEC_VARYING_ARG3
+#undef SHORT_VEC_ARG1
+#undef SHORT_VEC_ARG1_BOOLRET
+#undef SHORT_VEC_ARG2
+#undef SHORT_VEC_ARG3
```
nurmukhametov: Correct, they need to be undefined to avoid clobbering user namespace. Could you make this change in a separate commit?

stdlib/include/stdlib.isph
```diff
@@ -1430,3 +1432,234 @@ SHORT_VEC_ARG1(cbrt)
...
+
+    foreach (i = 0 ... N) {
+        result[i] = select(cond, t[i], f[i]);
+    }
+
+    return result;
+}
+
+template <typename T, uint N>
+inline uniform T<N> select(uniform bool<N> cond, uniform T<N> t, uniform T<N> f) {
```
nurmukhametov: I wanted to inspect the generated asm, but it fails to compile. See here: https://godbolt.org/z/bandq3avn. It would be good if you can share the link to the compiler explorer with implementations to look into.

stdlib/include/stdlib.isph
```diff
@@ -1430,3 +1432,234 @@ SHORT_VEC_ARG1(cbrt)
...
+    foreach (i = 0 ... N) {
+        result[i] = select(cond[i], t[i], f[i]);
+    }
+
+    return result;
+}
+
+template <typename T, uint N>
+uniform T<N> shuffle(uniform T<N> v, uniform int<N> perm) {
+    if(N < programCount) {
```
nurmukhametov: <= ?

stdlib/include/stdlib.isph
```diff
@@ -1430,3 +1432,234 @@ SHORT_VEC_ARG1(cbrt)
...
+        varying T var_v;
+        varying int var_perm;
+
+        foreach(i = 0 ... N) {
+            var_v = v[i];
+            var_perm = perm[i];
+        }
```
nurmukhametov: It is quite an interesting way to cast short vectors to varying values!

stdlib/include/stdlib.isph
```diff
@@ -1430,3 +1432,234 @@ SHORT_VEC_ARG1(cbrt)
...
+
+        varying T res_var1_v;
+        varying T res_var2_v;
+
+        res_var1_v = shuffle(var1_v, var2_v, var1_perm);
+
+        // See the note above about out-of-bounds accesses
+        if(programCount + programIndex < N) {
+            res_var2_v = shuffle(var1_v, var2_v, var2_perm);
```
nurmukhametov: This split into two shuffles doesn't seem correct to me. You split into two buckets, and an element inside the first can't be moved to the second, which should be possible according to the original function's semantics.

stdlib/include/stdlib.isph
```diff
@@ -1430,3 +1432,234 @@ SHORT_VEC_ARG1(cbrt)
...
+    uniform T<N> result;
+    uniform int<N> perm;
+
+    // NOTE: 
+    // The modulus is meant to support an extended range (i.e. beyond the ]-N;N[ interval),
+    // but this is expensive when `offset` is not a compile-time constant.
+    // Thus, should we use it or tell users this is not supported?
+    // `assume` can be used to reduce the overhead of this modulus.
+    offset = offset % N;
+    offset += select(offset < 0, (uniform int)N, 0);
```
nurmukhametov: I remember that same was done without modulus in builtins, [here](https://github.com/ispc/ispc/blob/768fc1242522d945017f6a0077f73b6ddd407e5f/builtins/generic.ispc#L990) and [here](https://github.com/ispc/ispc/blob/768fc1242522d945017f6a0077f73b6ddd407e5f/builtins/util.m4#L2267).

stdlib/include/stdlib.isph
```diff
@@ -1430,3 +1432,234 @@ SHORT_VEC_ARG1(cbrt)
...
+        }
+
+        return result;
+    }
+}
+
+// TODO: to be check further
+template <typename T, uint N>
+uniform T<N> shuffle(uniform T<N> v1, uniform T<N> v2, uniform int<N> perm) {
+    if(N < programCount) {
```
nurmukhametov: > Does using conditions like if(N < programCount) is the best solution as ? ...

At the moment, I don't know of any better solution. Later, we can probably support constexpr to deal with such situations.

nurmukhametov (commented):
> Should such functions be implemented in stdlib.isph or as builtins?

They definitely should be in the header. It looks like, as the number of functions supporting short vectors is growing, we need to create a separate one that is not supposed to be included by default. Let's say `short_vecs.isph`, and users need to write explicitly `#include <short_vecs.isph>`.

stdlib/include/stdlib.isph
zephyr111: Ok. Here is the separate PR for this:  #3467 .

stdlib/include/stdlib.isph
zephyr111: The thing is short vectors may not have a *power of two size*.
For the first link, a mask can be performed because `programCount` and `TARGET_WIDTH` are AFAIK guaranteed to be a power of 2.
For the second link, I am not sure `WIDTH` is always a power of two but it looks like it is.
In practice, LLVM is always able to optimize the modulus since `N` is guaranteed to be a compile-time constant. I do not think there is a better way to do that than what LLVM currently does (instruction sets are for power of 2 SIMD vectors so I do not expect a trick to be possible in x86 CPUs) unless we explicitly not support offset beyond a given limit (this is what I wanted to emphasize in the above comment). LLVM apparently removes the modulus-related instructions when `assume` precise the range of a runtime-defined `offset` (the modulus is optimized out for compile-time constants).
For short vectors having a size being a power of 2, I expect LLVM to replace the modulus by a mask (I did not check so far).

stdlib/include/stdlib.isph
zephyr111: Ok. 
You mean `if constexpr`? AFAIK C++ compiler do not prune the AST based on this or optimize the code early because it can have side effects. In D, they proposed `static if` for such a case which AFAIK enable compilers to do this. AFAIR, `static if` was initially proposed in C++, but members of the C++ committee (including the creator of the C++ language) claimed this was *"fundamentally flawed"* and they wrote a paper providing details the issues. I wonder if this would also applies in ISPC (it looks like it does at first glance). [Here](https://www.open-std.org/Jtc1/sc22/wg21/docs/papers/2013/n3613.pdf) is the research paper about that and [here](https://forum.dlang.org/post/gmewktoocqnwwvnpbrco@forum.dlang.org) is the reaction of D users. 
In C++, partial template specialization, concepts and things like `enable_if` can be used to avoid the need for `static_if` in most cases (so `if constexpr` and `constexpr` functions/containers seem enough once combined with other C++ features), but ISPC do not implement that yet. I expect them to be challenging to implement in ISPC though.
Anyway, this is certainly better to open an issue on this to discuss about that further.

stdlib/include/stdlib.isph
zephyr111: Unless I am missing something, I think element can move from one bucket to another because I use extended `shuffle` (i.e. one with 2 varying inputs). Let us consider the example, with `perm` set to `{3, 1, 2, 0}` (so `N = 4`) and a `programCount` of 2 (e.g. SSE/Neon with 64-bit elements):
 - `var1_v = {v[0], v[1]}`
 - `var2_v = {v[2], v[3]}`
 - `var1_perm = {3, 1}`
 - `var2_perm = {2, 0}`
 - `res_var1_v = shuffle({v[0], v[1]}, {v[2], v[3]}, {3, 1}) = {v[3], v[1]}`
 - `res_var2_v = shuffle({v[0], v[1]}, {v[2], v[3]}, {2, 0}) = {v[2], v[0]}`
 - the first bucket of `v` is then assigned to `res_var1_v` and the second to `res_var2_v`, hence `{v[3], v[1], v[2], v[0]}`
As we can see, the first bucket contains element that was initially stored in the second bucket and the other way around.

stdlib/include/stdlib.isph
zephyr111: I got the inspiration from your first try with vectors having a size smaller than `programCount` ;)  (I did not wrote a function for that though because it cannot be generalized for bigger vectors and I would like to reduce the number of loops and calls to keep the compilation time reasonable).

stdlib/include/stdlib.isph
zephyr111: `__is_compile_time_constant_uniform_int32` from the second code is pretty interesting though: it can be used to use a better implementation of some of these functions. That being said, I guess this intrinsic is for compile-time constants detected as such during the AST optimization and not after LLVM optimization steps, isn't it? I am wondering what happens in such a case:
```ispc
float test1(float v, uniform int offset) {
    // `offset` is not a compile-time constant in the ISPC AST
    // but it is a constant once test1 is inlined in test2 (i.e. after LLVM optimizations).
    // is __is_compile_time_constant_uniform_int32 true of false for `offset` in the IR of `test2` ?
    return rotate(v, offset); // or return shift(v, offset);
}

float test2(float v) {
    return test1(v, 1);
}
```

stdlib/include/stdlib.isph
zephyr111: You can fix this by specifying the template arguments (AFAIK template argument deduction is not yet implemented in ISPC), and by fixing the type of `cond` (from `int<4>` to `bool<4>`. Interestingly, having a bad type for `cond` results in an ISPC crash (so I reported this in #3474).
On my side, I used [this Godbolt code](https://ispc.godbolt.org/z/38MdP63Mc) to do some tests. The code of the rotate/shift/shuffle is certainly not up to date anymore, but the test functions are still relevant to analyze the generated code in many cases. I used the prefix `tmp_` during the development so to avoid conflicts with existing non-short-vectors functions when the calling code is incorrect. I can update the code if you wish.

stdlib/include/stdlib.isph
nurmukhametov: > The thing is short vectors may not have a _power of two size_.

Oh, indeed, right you are..

nurmukhametov: > `__is_compile_time_constant_uniform_int32` from the second code is pretty interesting though: it can be used to use a better implementation of some of these functions. That being said, I guess this intrinsic is for compile-time constants detected as such during the AST optimization and not after LLVM optimization steps, isn't it? I am wondering what happens in such a case:
> 
> ```ispc
> float test1(float v, uniform int offset) {
>     // `offset` is not a compile-time constant in the ISPC AST
>     // but it is a constant once test1 is inlined in test2 (i.e. after LLVM optimizations).
>     // is __is_compile_time_constant_uniform_int32 true of false for `offset` in the IR of `test2` ?
>     return rotate(v, offset); // or return shift(v, offset);
> }
> 
> float test2(float v) {
>     return test1(v, 1);
> }
> ```

No, it is not. This function call is resolved to `true` or `false` depending on the value of the argument at the point of running LLVM IR pass, `IsCompileTimeConstantPass`, called somewhere in the middle of the optimization pipeline ([here](https://github.com/ispc/ispc/blob/b54c1b72c14cd78086acab3dfeaeb1c6f090b9df/src/opt.cpp#L335)). If the argument happens to be a constant at this point then this would resolve to `true`, and `false` otherwise. The argument may become constant due to the chain of optimizations like inlining, constant propagation, etc. Conversely, there are situations when the argument is obviously constant but inlining doesn't inline the function call and we have false as the result. I recently came across this situation in [this test](https://github.com/ispc/ispc/blob/main/tests/lit-tests/const_shuffle_i8.ispc).
You can check this with the compiler explorer: https://godbolt.org/z/qrzf7W6eP. In `test1`, it is general permutation. In `test2`, call was inlined and there is only shufflevector.

stdlib/include/stdlib.isph
nurmukhametov: Right, my bad. I misremembered that shuffle across two input vectors is also named shuffle in stdlib, not shuffle2, and didn't notice the two input vectors in your code. Yes, it is correct. I literally did the same in builtins recently.

stdlib/include/stdlib.isph
nurmukhametov: > Ok. You mean `if constexpr`? AFAIK C++ compiler do not prune the AST based on this or optimize the code early because it can have side effects.

My understanding was that it can be pruned if the compiler is sure that it has no side effects.

Thanks for the broader context and links! This is for sure better discussed in a separate issue.

aneshlya:
It would be very helpful to split this PR to several (one per function) + add tests. 
`select` implementation is trivial and can be merged right away, but others are trickier and may require more time for test/review.

Issue #1134, opened 2015-12-15
About libispc


thebusytypist:
Hello,
I saw many people asked about a library form of ispc(#791).
And I also find it useful while I am doing my course project.

Today I try to implement this feature in my fork https://github.com/thebusytypist/ispc (the `libispc` branch).

What I have finished:

* Refactor `ispc::Module::CompileFile`.
  I extract the compilation process(the pre-processing and the LLVM module construction)
  into a new function `ispc::Module::compile`. And let `ispc::Module::CompileFile` calls it.
* Add `ispc::Module::CompileAndJIT` and `ispc::Module::GetFunctionAddress`.
  The `CompileAndJIT` function uses the compiled LLVM module
  to construct a `llvm::ExecutionEngine`.
  And the `GetFunctionAddress` returns the JIT compiled function's address.
* Add a new project called `libispc`, which is a static library.

Now a libispc.lib has been successfully compiled.

But I stopped here because I find it hard to integrate libispc into an external project
without CMake
and realize maybe I need more refactoring on the current code base.

I am considering following ideas for my next step:

* Add CMake build system. So I can `target_link_libraries` to integrate libispc easily.
  This also makes it easy to add new projects(e.g. tools, analyzers, etc) into ispc.
* Refactor the compilation process into a single class.
  Put JIT and normal emission process(e.g. output to `.o`) into separate classes.
  The reason is that in JIT the ownership of compiled `llvm::Module` is *transferred*
  to `ExecutionEngine`, while it is not in normal emission process.
  Mixing `CompileAndOutput` and `CompileAndJIT` in a single `Module` class
  may confuse the user and lead to memory access violation.
* Store global states(static variables, and LLVM context) in classes.
  libispc may be called from different threads.
* Implement the ispc executable based on libispc(Add libispc as a dependency of ispc executable).

Could you give me some advice?

By the way, I do not understand what is the "dispatch module" in `module.cpp`.
Could you explain it to me?
Thank you.

dbabokin:
Refactoring existing code sounds ok.

Question about CMake - I'm not sure I understand why you need it. Do you want to integrate building ISPC library into another project, which supposed to use the library? I think ISPC library should be supplied as part of binary distribution and integrated into other projects as third-party component.

Dispatch module is the module, which is responsible for multi-target compilation - i.e. when you have more than one target and generated code detects in runtime the CPU it runs on.

thebusytypist:
CMake build system is not an important part in the refactoring.
[...]

PR #3461, opened 2025-07-02
ISPC as a library

aneshlya:
## Description
This pull request enables ISPC to function as a library. The most important changes include adding support for building ISPC as a shared library, creating a new API for programmatic compilation using the `ispc::ISPCEngine` class, and introducing an example (`simple_lib`) that showcases how to use ISPC in library mode.

* Added new header file `ispc.h` defining the API for initializing, shutting down, and compiling ISPC code via library functions 
* Implemented the ISPC library functionality in `ispc_impl.cpp`, including methods for compilation and linking 
* Updated `main.cpp` to include the new library header and removed redundant methods, delegating functionality to the library implementation.
* Added new build target to build shared and static library
* Added `ispcConfig.cmake.in` to provide a package configuration file for consumers of the ISPC library, enabling easy integration via `find_package`
* Added a new example `simple_lib` demonstrating how to use the ISPC library.

Documentation will be updated after the first round of review.

## Related Issue
- [x] Linked to relevant issue(s). Related to https://github.com/ispc/ispc/issues/791 and https://github.com/ispc/ispc/issues/1134

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -478,7 +478,9 @@ list(APPEND FRONTEND_SOURCES
     src/expr.h
     src/func.cpp
     src/func.h
+    src/include/ispc/ispc.h
     src/intrinsics.cpp
+    src/ispc_impl.cpp
     src/module.cpp
     src/module.h
     src/preprocessor.cpp
@@ -642,6 +644,7 @@ list(APPEND INCLUDE_DIRECTORIES
     ${LLVM_INCLUDE_DIRS}
     ${XE_DEPS_DIR}/include
     ${CMAKE_CURRENT_SOURCE_DIR}/src
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/include
     ${CMAKE_CURRENT_SOURCE_DIR}/common
     ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_CFG_INTDIR}
     ${CMAKE_BINARY_DIR}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -307,6 +307,7 @@ find_package(Python3 REQUIRED)
     endif()
 
 include(GNUInstallDirs)
+include(CMakePackageConfigHelpers)
 
 find_package(BISON 3.0 REQUIRED)
     if (BISON_FOUND)
@@ -880,7 +881,7 @@ function(configure_ispc_lib TARGET)
         SOVERSION ${ISPC_VERSION_MAJOR})
 endfunction()
 
-# This function creates and configures an ISPC library target (static or shared)
+# This function creates and configures an ISPC library target (shared for now)
 function(add_ispc_library TARGET_NAME LIBRARY_TYPE OUTPUT_NAME_OVERRIDE)
     # Define the common source files and object libraries
     set(ISPC_LIB_SOURCES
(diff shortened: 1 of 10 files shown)


src/include/ispc/ispc.h
```diff
@@ -0,0 +1,66 @@
...
+ * This function parses command-line arguments and executes the appropriate action
+ * (compile, link, or help). Initialize() must be called successfully before using this function.
+ *
+ * @param argc Argument count.
+ * @param argv Argument vector.
+ * @return 0 on success, non-zero on failure.
+ */
+int CompileFromArgs(int argc, char *argv[]);
+
+class ISPCEngine {
```
aneshlya, line 53: The future JIT methods will reside in this class

examples/cpu/simple_lib/simple.cpp
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    std::cout << "Compiling simple.ispc using library mode with different options...\n";
+
+    // Set up a second compilation with different options
+    std::vector<const char *> args2 = {"ispc", "simple.ispc", "--target=host",  "-O0", "--emit-asm", "-o",
+                                       "simple_debug.s", "-h", "simple_debug.h", "-g"};
+
+    // Execute second compilation
+    int result2 = ispc::CompileFromArgs(static_cast<int>(args2.size()), const_cast<char **>(args2.data()));
```
nurmukhametov: If we provide only C++ API, maybe it makes sense to pass just vector to `CompileFromArgs`?

CMakeLists.txt
```diff
@@ -917,6 +973,61 @@ else()
     message(STATUS "ISPC will be built as a composite binary")
 endif()
 
+if (ISPC_LIBRARY AND NOT ISPC_SLIM_BINARY)
+    # Add libispc static library target with custom output name to avoid conflict
+    add_ispc_library(libispc STATIC "ispc_static")
```
nurmukhametov: What about placing libraries under `lib` in build directory?

src/include/ispc/ispc.h
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * @brief Compiles ISPC code from command-line arguments.
+ * This function parses command-line arguments and executes the appropriate action
+ * (compile, link, or help). Initialize() must be called successfully before using this function.
+ *
+ * @param argc Argument count.
+ * @param argv Argument vector.
+ * @return 0 on success, non-zero on failure.
+ */
+int CompileFromArgs(int argc, char *argv[]);
```
nurmukhametov: There is a quite unintuitive thing about argv. It contains the program name as the first element.

CMakeLists.txt
nurmukhametov: Do we really need to provide the static library? It requires LLVM installation and segfaults in my setup at the moment.

src/include/ispc/ispc.h
aneshlya: This method follows the same logic as `clang::CompilerInvocation::CreateFromArgs` which also requires first element as program name. Basically it can be any dummy string, it will be just ignored.

CMakeLists.txt
aneshlya: Not really. We can only provide dynamic library and C++ interface for the first release and see how it'll go.

src/include/ispc/ispc.h
nurmukhametov: Let's clearly state this point in the comments.

nurmukhametov (approved):
LGTM!

PR #3469, opened 2025-07-05
Fix #2207: make integer suffix definition stricter

### LINKED ISSUE #2207: Make integer suffix definition stricter
ISPC has too frivolous definition for integer suffixes in parser:
`INT_NUMBER (([0-9]+)|(0[xX][0-9a-fA-F]+)|(0b[01]+))[uUlL]*[kMG]?[uUlL]*` ([link](https://github.com/ispc/ispc/blob/fd297fa67ce52ce7d46da137ab3d2281a26d79c7/src/lex.ll#L370))

It has the following problems:
1. it doesn't limit the number of occurrences of [uUlL] symbols, for example this appears legal: `1uUuUlllLLllLLuUu` ([godbolt link](https://godbolt.org/z/Psb8hMbcd))
2. value modification suffix (i.e. [kMG]) may go after type modification suffix. This seems illogical to me and gives too much freedom where it's not needed.
3. Unlike C/C++ it allows `lL` and `Ll` suffixes (i.e. the mix of lower and upper case "L" to form a 'LL' suffix). Only `ll` and `LL` should be allowed.

zephyr111:
## Description

This PR fixes #2207 (i.e. it makes the integer suffix definition stricter).

## Related Issue

## Checklist

### DIFF
--- a/docs/ispc.rst
+++ b/docs/ispc.rst
@@ -799,6 +799,15 @@ existing code from intrinsics to ISPC, you can use these headers as a reference.
 The standard library ``select`` functions now support unsigned integer types
 ``uint8``, ``uint16``, ``uint32``, and ``uint64``.
 
+Integer literals are now stricter:
+
+* It now limits the number of occurrences of [uUlL] symbols (i.e. "ulll", 
+  "uul" and "lulu" are not valid anymore).
+* The value modification suffix (i.e. [kMG]) must preceed the type 
+  modification suffix (i.e. [uUlL] symbols).
+* Like C/C++, "lL" and "Ll" suffixes is not allowed anymore (i.e. the mix of 
+  lower and upper case "L" to form a "LL" suffix).
+
 
 Getting Started with ISPC
 =========================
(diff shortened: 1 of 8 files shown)


zephyr111:
I totally agree for the documentation.

Regarding the warning, maybe we can keep the old tokens and set a lower priority to them so the new one take precedence (and mark them as obsolete one. We then need to update the grammar consequently so to merge the obsolete tokens with the non obsolete ones. It will certainly decrease a bit the performance of the lexer and make the parser/lexer slightly more complicated. I fine with this as long as this is temporary and removed later in a future version (i.e. deprecated before removal). Then the question is what to put in the grammar in the documentation: should this trick be visible?

That being said, in my experience, please note that such things tend to stay for a very long or even sometime forever (if so, then the PR is pointless). It would be good not to forget this later. What about opening an issue so to remove the old format in a specified version and add this issue in the planning? We could even report the version where the old format will be removed to the user.

nurmukhametov:
> Regarding the warning, maybe we can keep the old tokens and set a lower priority to them so the new one take precedence (and mark them as obsolete one. We then need to update the grammar consequently so to merge the obsolete tokens with the non obsolete ones. It will certainly decrease a bit the performance of the lexer and make the parser/lexer slightly more complicated. I fine with this as long as this is temporary and removed later in a future version (i.e. deprecated before removal). Then the question is what to put in the grammar in the documentation: should this trick be visible?

At the same time, I don't want this simple change to be overengineered because of the fear of breaking user code. Anyway, even if it happens, it would require a quite simple fix from the users. Overall, I am fine with the current change if it is documented. Maybe @aneshlya has a stronger opinion here.

aneshlya:
I'm OK with this change if it's documented and highlighted in the "Updating ISPC Programs For Changes In ISPC 1.28.0" section.

aneshlya (approved):

zephyr111:
Ok. Let's keep it simple. I just fixed the documentation, including the "Updating ISPC Programs For Changes In ISPC 1.28.0"  section. It should be fine now.

zephyr111:
I just rebased the PR due to a conflicting merge issue on the doc.

docs/ispc.rst
```diff
@@ -2122,15 +2131,22 @@ Here are three ways of specifying the integer value "15":
...
+
+::
+
+   uint three = 3ul;
+   int two_kb = 2k;         // 2048
+   int two_megs = 2M;       // 2 * 1024 * 1024
+   int one_gig = 1G;        // 1024 * 1024 * 1024
+   uint three_gig = 3Gu;    // 3 * 1024 * 1024 * 1024
+   uint64 six_gig = 6GuLL;  // 6 * 1024 * 1024 * 1024
+   int64 ten_gig = 12Gll;   // 10 * 1024 * 1024 * 1024
```
nurmukhametov: I am confused ten or twelve?

nurmukhametov (approved):
LGTM!

docs/ispc.rst
zephyr111: Right! I fixed this too quickly (twelve was too big for the line so I replaced it with ten but forgot to fix the numbers)... Thank you.

Issue #1272, opened 2017-05-09
Cannot allocate unsigned types with operator 'new'


peterhuba:
Cannot allocate memory the following way:
```
void test()
{
  uniform unsigned int* uniform v = uniform new uniform unsigned int[16];
}
```

Compiler output:

> test.ispc:3:57: Error: syntax error, unexpected 'unsigned'. 
>   uniform unsigned int* uniform v = uniform new uniform unsigned int[16];
>                                                         ^^^^^^^^

Need to use `typedef` as a workaround.

zephyr111:
I am closing the issue since I confirm this is a duplicate issue (and there is a PR for that now).

PR #3473, opened 2025-07-06
Fix #707: support signed/unsigned/const in 'new' expressions and template arguments

### LINKED ISSUE #707: The parser doesn't properly parse 'ptr = new unsigned int'
// this produces Error: syntax error, unexpected 'unsigned'.
unsigned int \* ptr = new unsigned int;

// Ok using this workaround:
typedef unsigned int uint;
unsigned int \* ptr = new uint;

zephyr111:
## Description

This PR fixes #707 (and the duplicate issue #1272).
The modification also impacts template arguments (but not template parameters).

Short vectors can now be syntactically supported but they do not seems to be fully supported by the rest of the ISPC code. Thus, I explicitly disabled them so far. I put some examples in the lit-test code showing what could be easily supported and what breaks (or is ambiguous) if we enable them. I think supporting short vectors in `new` expression can be a good idea though it is certainly a low-priority feature since I do not expect much people to use it. I am not sure it make sense in template parameters (nor if this is already supported in the rest of the code, and otherwise if this is simple to do). What do you think about this?
Anyway, note that this can be implemented in another independent  PR if needed (and discussed in a separate issue).

Note that this PR is dependent of the PR #3472 for the lit-tests to pass when they will be complete. I will correct the lit-test once #3472 will be merged.

Note that `new const soa<4> Struct` is not supported yet, but `new soa<4> const Struct` is. I am not just we want the first to be supported though.

## Related Issue

## Checklist

### DIFF
--- a/tests/lit-tests/707.ispc
+++ b/tests/lit-tests/707.ispc
@@ -1,0 +1,37 @@
+// This test checks that the compiler supports `signed`, `unsigned` and `const`
+// quialifiers in `new` expressions and template arguments. 
+// It also check the compiler does not crash (fatal error).
+
+// RUN: %{ispc} --target=host --emit-llvm-text --nostdlib --nowrap -g %s -o - 2>&1 | FileCheck %s
+
+// CHECK-NOT: FATAL ERROR:
+
+uniform uint* uniform test_uint() {
+    return uniform new uniform unsigned int;
+}
+
+uniform int* uniform test_sint() {
+    return uniform new uniform signed int; 
+}
+
+uniform const int* uniform test_const() {
+    return uniform new uniform const int(815);
+}
+
+const unsigned uniform int* uniform test_mixed_unordered() {
+    return uniform new unsigned const uniform int(815);
+}
+
+template <typename T> void test_template() {}
+
+void test_template_instantiation()
+{
+    test_template<int>();
+    test_template<uniform int>();
+    test_template<varying int>();
+    test_template<const int>();
+    test_template<const signed int>();
+    test_template<const unsigned int>();
+    test_template<const unsigned uniform int>();
+    test_template<uniform const unsigned int>();
+}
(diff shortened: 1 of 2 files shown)


tests/lit-tests/707.ispc
```diff
@@ -0,0 +1,58 @@
...
+    uniform int<6> v = {4, 8, 15, 16, 23, 42};
+    return uniform new uniform int<6>(v); // Should be legit (does not work yet) if short-vectors would be supported.
+}
+
+uniform int<6>* uniform test_vec_with_init2() {
+    return uniform new uniform int<6>(4, 8, 15, 16, 23, 42); // Is is legit? (could build without apparent issue so far)
+}
+
+uniform int<6>* uniform test_vec_with_init3() {
+    return uniform new uniform int<6>({4, 8, 15, 16, 23, 42}); // Is is legit? (does not build yet)
```
aneshlya: This seems to me more like advanced feature of C++. It can be supported in ISPC, but it's optional.
Other cases looks legit to me.

zephyr111: Does your comment reference the last example/function or the whole (commented) block about short vectors?

zephyr111:
Please note I rebased this PR on main and enabled back the lines in lit-test related to the fixed `signed` bug. It should not be fine (besides the block about short vectors in the lit-tests).

tests/lit-tests/707.ispc
aneshlya: About the last example. Everything else looks reasonable.

zephyr111: Ok. Thank you for the feedback. I will see if I can come up with another PR for supporting simple cases with short vectors (and test this more carefully). 
Then the question is whether we want to support that in templates arguments. 
I removed the comment so this PR should be fine now.

aneshlya (approved):
LGTM!

PR #3477, opened 2025-07-08
Improve the doc and update the changelog

zephyr111:
## Description

This PR is only about documentation.
It add missing news in the changelog (addressed by recent past PR) and also updates the table mentioning all targets (add missing targets, bold table header, add information about AMD CPUs, add more precision about Intel one, cleaner/better-structured information).

## Related Issue

## Checklist

### DIFF
--- a/docs/ispc.rst
+++ b/docs/ispc.rst
@@ -382,13 +382,13 @@ Updating ISPC Programs For Changes In ISPC 1.9.0
 ------------------------------------------------
 
 The release doesn't contains language changes, which may affect compatibility with
-older versions. It introduces new AVX512 target: avx512knl-i32x16.
+older versions. It introduces new AVX-512 target: avx512knl-i32x16.
 
 Updating ISPC Programs For Changes In ISPC 1.9.1
 ------------------------------------------------
 
 The release doesn't contains language changes, which may affect compatibility with
-older versions. It introduces new AVX512 target: avx512skx-i32x16.
+older versions. It introduces new AVX-512 target: avx512skx-i32x16.
 
 Updating ISPC Programs For Changes In ISPC 1.9.2
 ------------------------------------------------
@@ -417,7 +417,7 @@ Starting 1.11.0 version auto-generated headers use ``#pragma once``. In the unli
 case when your C/C++ compiler is not supporting that, please use ``--no-pragma-once``
 ``ispc`` switch.
 
-This release also introduces new AVX512 target avx512skx-i32x8. It produces code,
+This release also introduces new AVX-512 target avx512skx-i32x8. It produces code,
 which doesn't use ZMM registers.
 
 Updating ISPC Programs For Changes In ISPC 1.12.0
@@ -516,13 +516,13 @@ Standard library library got new ``prefetchw_{l1,l2,l3}()`` intrinsics for
 prefetching in anticipation of write.
 
 The algorithms used for implementation of ``rsqrt(double)`` and ``rcp(double)``
-standard library functions have changed on AVX512 and may affect the existing
+standard library functions have changed on AVX-512 and may affect the existing
 code.
 
 Updating ISPC Programs For Changes In ISPC 1.18.0
 -------------------------------------------------
 
-AVX512 targets were renamed to drop "base type" (or "mask size"), old naming is accepted for
+AVX-512 targets were renamed to drop "base type" (or "mask size"), old naming is accepted for
 compatibility. New names are avx512skx-x4, avx512skx-x8, avx512skx-x16,
 avx512skx-x32, avx512skx-x64, and avx512knl-x16.
 
@@ -547,7 +547,7 @@ New targets were added:
 
 * avx512spr-x4, avx512spr-x8, avx512spr-x16, avx512spr-x32, avx512spr-x64 for
   4th generation Intel® Xeon® Scalable (codename Sapphire Rapids) CPUs. A macro
-  ISPC_TARGET_AVX512SPR was added.
+  ``ISPC_TARGET_AVX512SPR`` was added.
 * xehpc-x16 and xehpc-x32 for Intel® Data Center GPU Max (codename Ponte Vecchio).
 
 Function templates were introduces to the language, please refer to `Function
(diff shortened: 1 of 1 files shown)


nurmukhametov (approved):
LGTM!

docs/ispc.rst
```diff
@@ -1101,22 +1109,28 @@ of target, arch and target OS.
...
-neon         ARM NEON
-sse2         SSE2 (early 2000s era x86 CPUs)
-sse4.1       SSE4.1 (2007 Intel codename Penryn CPUs)
-sse4.2       SSE4.2 (2008-2010 Intel codename Nehalem CPUs)
-gen9         Intel Gen9 GPU
-xelp         Intel XeLP GPU
-xehpg        Intel Arc GPU
-xehpc        Intel Ponte Vecchio GPU
-============ =========================================================
+============= ========================= ===========================================================
```
aneshlya, line 1121: I really like this updated table, it makes things much clearer. But I think we should fill the gaps here and add some missing targets:
```
avx, avx1     AVX                       Intel Sandy Bridge (2011), AMD Bulldozer/Jaguar (2011/2013)
avx2          AVX 2                     Intel Haswell (2013) [#]_, AMD Excavator (2015)
avx2vnni      AVX 2 + VNNI              Intel Alder Lake (2021)
avx512skx     AVX 512                   Intel Skylake Xeon (2017), AMD Zen4 (2022)
avx512icl     AVX 512 + VNNI            Intel Ice Lake (2019), AMD Zen4 (2022)
avx512spr     AVX 512 + FP16            Intel Sapphire Rapids (2023)
avx10.2       AVX 10.2                  Intel Diamond Rapids (2026+)
neon          ARM NEON                  ARM Cortex-A35/A53/A57 (2012-2015)
sse2          SSE2                      Intel Pentium 4 (2001), AMD Athlon 64 (2003)
sse4.1        SSE4.1                    Intel Penryn (2007)
sse4.2        SSE4.2                    Intel Nehalem (2008)
gen9          Intel Gen9 GPU            Intel Skylake GPU (2015)
xelp          Intel XeLP GPU            Intel Tiger Lake LP GPU (2020)
xehpg         Intel XeHPG GPU           Intel Arc Alchemist GPU (2022)
xehpc         Intel XeHPC GPU           Intel Ponte Vecchio GPU (2022)
xelpg         Intel XeLPG GPU           Intel Meteor Lake GPU (2023)
xe2hpg        Intel Xe2HPG GPU          Intel Battlemage GPU (2024)
xe2lpg        Intel Xe2LPG GPU          Intel Lunar Lake GPU (2024)
generic       Platform-agnostic         Any compatible CPU/GPU
```

zephyr111: I agree. 

I applied the changes and split this in three parts: the CPU targets, the GPU ones and the generic one (other similar targets may be added in the last table like WASM).

Here are some questions:
 - Regarding avx512icl, shouldn't the target also support the (many) AVX-512 VL/DQ/BW/IFMA/VBMI/VBMI2/BITALG/VPCLMULQDQ/GFNI/VAES extensions? AFAIK, there is no special focus on them yet on ISPC but the generated code can contain such instruction with this target, isn't it?
 - Regarding avx512spr, it AFAIK also support VNNI, isn't it? So I guess it is better to say it is "VNNI+FP16".
 - Regarding Ice Lake, are we 100% sure the target support both client and server CPU?

aneshlya: > Regarding avx512icl, shouldn't the target also support the (many) AVX-512 VL/DQ/BW/IFMA/VBMI/VBMI2/BITALG/VPCLMULQDQ/GFNI/VAES extensions? AFAIK, there is no special focus on them yet on ISPC but the generated code can contain such instruction with this target, isn't it?
Regarding avx512spr, it AFAIK also support VNNI, isn't it? So I guess it is better to say it is "VNNI+FP16".

From ISPC perspective for `avx512icl` only VNNI feature matters, but you're right, instructions from these extensions can be generated. To avoid confusion let's remove the features from `SIMD Instruction Set` column (for both AVX512 and AVX2), CPU name should be sufficient. 

> Regarding Ice Lake, are we 100% sure the target support both client and server CPU?

ISPC `avx512icl` target maps to Icelake client by default. So if we want to be very precise, you can add Intel Icelake Client but I'm OK with current wording.

nurmukhametov: > > Regarding avx512icl, shouldn't the target also support the (many) AVX-512 VL/DQ/BW/IFMA/VBMI/VBMI2/BITALG/VPCLMULQDQ/GFNI/VAES extensions? AFAIK, there is no special focus on them yet on ISPC but the generated code can contain such instruction with this target, isn't it?
> > Regarding avx512spr, it AFAIK also support VNNI, isn't it? So I guess it is better to say it is "VNNI+FP16".
> 
> From ISPC perspective for `avx512icl` only VNNI feature matters, but you're right, instructions from these extensions can be generated. To avoid confusion let's remove the features from `SIMD Instruction Set` column (for both AVX512 and AVX2), CPU name should be sufficient.

VBMI, at least as well, we use `vpermb` for shuffle of `int8` for `avx512icl/avx512spr/avx10.2` but not for `avx512skx`. 

The best place to understand what `avx512icl` is supposed to support is in the dispatcher logic [here](https://github.com/ispc/ispc/blob/eada0e050859100ff437fb926382267d1177882e/src/isa.h#L194)

zephyr111: Ok. I removed the extension to make this clearer. I also added an external link with a pretty-clear table about this. I change also the term like "AVX512" to "AVX-512" when it was referring to the (whole) instruction set (i.e. not specific extensions) so to match with Intel technical documents (like the [intrinsics guide](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html) and the [AVX10.2 specification](https://www.intel.com/content/www/us/en/content-details/836199/intel-advanced-vector-extensions-10-2-intel-avx10-2-architecture-specification.html) as well as Wikipedia articles). It should be fine now.

aneshlya (approved):
Looks good!

PR #3483, opened 2025-07-11
Fixed crash caused by type signature mismatch when calling functions with uniform bool<N> parameters.

### LINKED ISSUE #3474: ISPC crash when calling a template function with short vectors having an incorrect atomic type
The following code cause a crash of ISPC (fatal error + segmentation fault):
```ispc
template <typename T, uint N>
inline uniform T<N> select(uniform bool<N> cond, uniform T<N> t, uniform T<N> f) {
    uniform T<N> result;

    foreach (i = 0 ... N) {
        result[i] = select(cond[i], t[i], f[i]);
    }

    return result;
}

#define N 4

uniform int<N> foo(uniform int<N> cond, uniform int<N> a, uniform int<N> b) {
    return select<int, N>(cond, a, b);
}
```

This is certainly due to `cond` being of the bad type (`int<4>` rather than `bool<4>`) and so no function match found.

See on [Godbolt](https://godbolt.org/z/vaePzGfb9).

aneshlya:
## Description
Fixed the crash caused by type signature mismatch when calling functions with `uniform bool<N>` parameters. The issue was in
`TypeCastExpr::GetValue()` where `SwitchBoolToStorageType` was being called for uniform bool vectors, converting from logical types `(<N x i1>)` to storage types `(<N x i8>)`, causing a mismatch with function signatures that expect logical types.
Function signatures use `LLVMType()` while argument conversion was switching to `LLVMStorageType()`.

## Related Issue
- [ ] Fixed #3474 

## Checklist

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -7857,9 +7857,6 @@ llvm::Value *TypeCastExpr::GetValue(FunctionEmitContext *ctx) const {
             if (!conv) {
                 return nullptr;
             }
-            if ((toVector->GetElementType()->IsBoolType())) {
-                conv = ctx->SwitchBoolToStorageType(conv, toVector->LLVMStorageType(g->ctx));
-            }
             return conv;
         } else {
             // Emit instructions to do type conversion of each of the elements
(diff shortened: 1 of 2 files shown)


nurmukhametov (approved):
LGTM!

PR #3486, opened 2025-07-14
Use std::move() to avoid unnecessary copying

aneshlya:
## Description
This PR fixes minor Coverity issues related to using `std::move` instead of object copying.

## Related Issue

## Checklist

### DIFF
--- a/src/args.cpp
+++ b/src/args.cpp
@@ -252,16 +252,6 @@ static ArgsParseResult devUsage() {
  */
 class ArgFactory {
   private:
-    char *AllocateString(std::string string) {
-        int len = string.length();
-        // We use malloc here because of strdup in lAddSingleArg
-        char *ptr = (char *)malloc(len + 1);
-        memset(ptr, 0, len + 1);
-        strncpy(ptr, string.c_str(), len);
-        ptr[len] = '\0';
-        return ptr;
-    }
-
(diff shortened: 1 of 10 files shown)


src/args.cpp
```diff
@@ -302,7 +302,7 @@ class ArgFactory {
             c = GetNextChar();
         }
 
-        return AllocateString(arg);
+        return AllocateString(std::move(arg));
```
nurmukhametov: It is not clear for me why we need `AllocateString` function at all. It seems to me that it reimplements `strdup`, so it looks `return strdup(arg.c_str());` is enough here.

src/expr.cpp
```diff
@@ -6470,15 +6470,15 @@ template <typename To> static inline void lConvertElement(llvm::APFloat from, To
 template <typename From>
 static inline void lConvertElement(From from, std::vector<llvm::APFloat> &to, llvm::Type *type) {
     llvm::APFloat apf = lCreateAPFloat((double)from, type);
-    to.push_back(apf);
+    to.push_back(std::move(apf));
```
nurmukhametov: Let's do `emplace_back`:

```c++
    to.push_back(lCreateAPFloat((double)from, type));
```

nurmukhametov (approved):
LGTM!

Issue #2903, opened 2024-07-03
Missing avx2-i64x8 target (ie. double pumping for 64-bit codes)


zephyr111:
Hello,

There is no `avx2-i64x8` target in ISPC while there is a `avx2-i32x16` target to enable double-pumping on 32-bit codes.
While ISPC is mainly design to speed up 32-bit codes, it can be very helpful to speed up 64-bit ones. Providing this missing target can help to significantly reduce the execution time of some 64-bit code on machines supporting AVX-2 but not AVX-512 yet (i.e. ~80% machines), especially since the latency of 64-bit instructions tends to be higher than 32-bit ones. 

Note it is not a problem with AVX-512 anymore since we can use targets like avx512skx-x8 or avx512skx-x16, but only a tiny fraction of mainstream client (Intel) CPUs support it (i.e. <15%), and even the newest client Intel CPUs still don't support it (so generating AVX-2 code will still be useful for >5 years).

Here is a simple example:

```
uniform double sum(uniform double v[], uniform size_t size)
{
    varying double s = 0.0d;
    foreach(i=0...size)
        s += v[i];
    return reduce_add(s);
}
```

Here is the generated code for the `avx2-i64x4` target ([see on GodBolt](https://godbolt.org/z/64r7f6c1j)):

```
.LBB1_3:
        mov     r8d, edx
        and     r8d, -32
        vaddpd  ymm1, ymm1, ymmword ptr [rdi + r8]
        add     rax, 4
        add     rdx, 32
        cmp     rax, rcx
        jb      .LBB1_3
```

This takes 4 cycles/iteration on all Intel CPU from Skylake to Rocket-Lake. 3 cycles/iteration on (P-cores of) newer Intel CPUs and all AMD Zen ones.

With double-pumping (ie. a new `avx2-i64x8` target), the following generated code is expected:

```
.LBB1_3:                                # =>This Inner Loop Header: Depth=1
        mov     r8d, edx
        and     r8d, -64
        vaddpd  ymm2, ymm2, ymmword ptr [rdi + r8 + 32]
        vaddpd  ymm1, ymm1, ymmword ptr [rdi + r8]
        add     rax, 8
        add     rdx, 64
        cmp     rax, rcx
        jb      .LBB1_3
```

It should take exactly the same number of cycles/iteration while computing 8 items/iteration rather than 4 resulting in a **twice faster code**. The code is still not optimal but it should be enough in most case to mitigate the latency in more complex use-cases.

[...]

Issue #666, opened 2013-11-22
Support for different floating point models.


jbrodman:
Add support for different models (similar to ICC's -fp-model argument).

This can allow us to define a default, fast, and strict mode to enable different optimizations.

For example, it'd be awesome if things getting divided by a loop invariant value got replaced with multiplies by the reciprocal.

zephyr111:
I am closing this issue since the new fast-math modes (which is kind of similar to the fp-model argument) should be able to perform the requested kind of optimizations. The default mode in ISPC is AFAIK the strict one (though FMA are enabled by default and can be explicitly disabled with `--opt=disable-fma`).

dbabokin:
> I am closing this issue since the new fast-math modes (which is kind of similar to the fp-model argument) should be able to perform the requested kind of optimizations. The default mode in ISPC is AFAIK the strict one (though FMA are enabled by default and can be explicitly disabled with `--opt=disable-fma`).

It's not the same. Historically there are two approaches to managing fp model in the compilers. Chaotic gcc approach, which has poorly defined default (poorly wrt mathematical expectation, but it's "well defined" in terms of "it's what everyone expects"), whole bunch of knobs to manage individual optimizations and fast-math mode, which is combination of that knobs basically.

And the approach implemented in Microsoft compiler and then adopted in Intel compiler with `-fp-model=[strict|precise|fast]`, which has better definition of these levels of optimization. `precise` is value safe optimizations. `strict` is the same plus preserving exception semantics. `fast` is value unsafe optimizations.

There was a push to introduce `-fp-model` switch in clang. A lot of work was done to clear up optimizations, but I haven't checked what is the current status.

Note, that math library is kinda orthogonal to that. As well as fp-contract / disable-fma.

So I suggest reopening the issue, as it about introducing MS-style `-fp-model` flag.

zephyr111:
@dbabokin 

> Note, that math library is kinda orthogonal to that. As well as fp-contract / disable-fma.

I did not reference fast-math library (enabled with `--math-lib=`), but *"fast-math modes"* (enabled with `--opt=fast-math`). More specifically, I was thinking about #3489 which now introduce new mode for this flag (e.g. `--opt=fast-math:balanced` or `--opt=fast-math:aggressive`). I apologies for the confusion or the lack of details.

[...]

PR #3489, opened 2025-07-15
Make fast-math optimizations more aggressive

### LINKED ISSUE #3402: Can't force optimizer to use logarithmic exponentiation even with fast-math ON
Simple example with applying hardcoded value expanentiation with fixed degree on set of floats:

```ISPC
varying float square(const varying float value)
{
    return value * value;
}

template<int N>
static varying float power(const varying float value)
{
    varying float accum = value;
    #pragma unroll
    for (uniform int i = 1; i < N; i++) {
        accum *= value;
    }
    return accum;
}

export void power_9(uniform float values[], const uniform int count)
{
    assume(((uniform uint64)((void*)values) & (32 * TARGET_WIDTH)-1) == 0);

    foreach(index = 0...count)
    {
        values[index] = power<9>(values[index]);
        // values[index] = square(square(square(values[index]))) * values[index];
    }
}
```

There are two options: handmade exponentiation with squaring and just inlined loop of repeated multiplication.
Both result in the same mathematial result while floats makes them different. However fast math option should make them equal so compiler will produce 4 multiplication instruction instead of 8.

This is so in theory, but this is compiler output:

<table>
 <tr>
  <td> power<9>(values[index]); </td>
  <td> square(square(square(values[index]))) * values[index]; </td>
 </tr>
 <tr>
  <td>

```Asm
.LBB1_3:                                # =>This Inner Loop Header: Depth=1
        vmovaps ymm0, ymmword ptr [rdi + 4*rax]
        vmulps  ymm1, ymm0, ymm0
        vmulps  ymm1, ymm0, ymm1
        vmulps  ymm1, ymm0, ymm1
        vmulps  ymm1, ymm0, ymm1
        vmulps  ymm1, ymm0, ymm1
        vmulps  ymm1, ymm0, ymm1
        vmulps  ymm1, ymm0, ymm1
        vmulps  ymm0, ymm0, ymm1
        vmovaps ymmword ptr [rdi + 4*rax], ymm0
        add     rax, 8
        cmp     rax, rcx
        jb      .LBB1_3
        cmp     eax, esi
        jge     .LBB1_6
.LBB1_5:
```
  </td>
  <td>

```Asm
.LBB1_3:                                # =>This Inner Loop Header: Depth=1
        vmovaps ymm0, ymmword ptr [rdi + 4*rax]
        vmulps  ymm1, ymm0, ymm0
        vmulps  ymm1, ymm1, ymm1
        vmulps  ymm1, ymm1, ymm1
        vmulps  ymm0, ymm0, ymm1
        vmovaps ymmword ptr [rdi + 4*rax], ymm0
        add     rax, 8
        cmp     rax, rcx
        jb      .LBB1_3
        cmp     eax, esi
        jge     .LBB1_6
.LBB1_5:
```
  </td>
 </tr>
</table>

Godbolt example: https://godbolt.org/z/xrMEzhnzM

zephyr111:
## Description

This PR implements an LLVM pass so to optimize floating-point math operations. It currently also removes the frond-end AST transformation (regarding divisions to reciprocals) since it is meant to be done in the back-end.
It fixes the issue #3402 when fast-math is enabled (i.e. using compilation flags `--opt=fast-math -O2`).
However, there are many things to consider and discuss.

One major issue with the current LLVM IR `fast` flag on FP math instructions is that is can cause *crazy transformations* during optimizations if the fast-math assumptions are broken. This a known issue and it also happens in C and C++ code compiled with `--fast-math`. This is described further [here](https://github.com/iree-org/iree/issues/19743). See https://llvm.org/docs/LangRef.html#fast-math-flags for more information about LLVM fast-math/rewrite-based flags.

IMHO, this is like opening the Pandora's box. Indeed, it is very easy to generate Inf, NaN, negative zeros values (e.g. from subnormal which are also quite common in pathological cases) and so to break codes compiled with this optimization. While this happens in C/C++, the benefit in C/C++ is higher because this optimization generally enables loops to be auto-vectorized (strongly improving performance in many cases), while in ISPC there is no need of this for loops to be efficiently vectorized (which is great since `--fast-math` is pretty dangerous and its effect are often underestimated). That being said, it would be nice to benefit from `contract` (i.e. `mul`+`add` to `fma`), and `reassoc` (i.e. associative FP math) to improve a bit further ISPC code without the dangerous impact of a full fast-math. #3402 is a good example for that (NaN and Inf could still be supported with twice less `mul` -- so a possibly twice faster code). In some cases, enabling reciprocal optimization can be also beneficial for performance though it is generally less precise and more dangerous. For users wanting performance at all costs, it would be good to provide a full fast-math optimization. The later is particularly dangerous because it should impact all our math functions and I have no idea of whether they still work well in this case (I just run some functional-tests to quickly evaluate this and it was surprisingly not that bad but I think some important test on math functions are disabled).

Based on this, I started to support one experimental version fully enabling fast-math optimization (enabled with the define `AGGRESSIVE_FAST_MATH_OPT`. It works well regarding optimizations: LLVM performs FP math operation re-association, reciprocal optimization (certainly better than we do so far), etc. Then I tried to support a less dangerous fast-math as aforementioned. On one hand, It turns out that re-association requires the `nsz` (i.e. "no signed zeros") flag to be actually effective. Fortunately, it *"does not imply that -0.0 is poison and/or guaranteed to not exist in the operation"* so this is a rather-"safe" assumption. Please note that *"the reassociation does lose both some infinites and some NaNs"* (aforementioned [here](https://github.com/iree-org/iree/issues/19743)), but IMHO I consider this is rather normal considering one plays with fire in such a case. Still, there is no poison values in this case. On the other hand, reciprocal optimizations requires the `ninf` flag (or apparently `nnan` for some non-trivial optimizations) to be enabled to be actually effective and this one is dangerous due to the aforementioned possibly-broken assumption during optimizations. It (kind of) makes sense since reciprocal operation (like the optimization we currently do in ISPC when fast-math is enabled) can sometimes generate infinities from subnormal values or even `-nan` values from infinities ([see on Godbolt](https://ispc.godbolt.org/z/qvdbn5b8v)).

Currently, in this PR, the fast-math optimizations are rather-safe and the reciprocal optimizations are not done anymore in the AST. The later should either be done by LLVM if we accept to perform a full fast-math optimization, or done manually in this new fast-math optimization pass (and consider the fact that it can wrongly produce NaN/Inf values in pathological cases). Indeed, the early transformation of the divisions in the AST apparently prevent [some checks](https://ispc.godbolt.org/z/z1jqWx1d8) to be reported so moving this to the back-end should solve this issue. I personally prefer the work to be done by LLVM since the optimization do not need the stdlib (as opposed to our current solution) and the code is better in some cases (e.g. more accurate for double-precision in AVX2 targets while being at least as fast). Our ISPC solution does not generate poison values when assumptions are broken but it can generate Inf/NaN values that will break assumption if the `nnan`, `ninf` or `fast` math-flags are used anyway...

At this point, here is the **list of questions we should answer**:
 - Should `--opt=fast-math` be equivalent to `--fast-math` in Clang?
 - Should we provide fine-grained fast-math optimizations like in Clang? Or alternatively, should we provide multiple level of fast-math optimizations? (IMHO the later help a LOT easily track issues in our math functions since there is far less combination to tests and it makes things simpler from the user PoV).
 - Should we use our own reciprocal function or let LLVM do the optimizations?

Last but not least, some other points needs to be considered like the fact fast-math will certainly have a different behavior, the documentation this and ways to avoid users to encounter issues from such an update (i.e. reduce surprising results after updating ISPC). One way to do that is by removing the current `--opt=fast-math` flag so to force users to switch to the new one. We could also just still support the old one and warn users temporary (i.e. deprecation warning) and remove this later (but this certainly make things a bit more complicated).

## Related Issue

## Checklist

### DIFF
--- a/docs/ispc.rst
+++ b/docs/ispc.rst
@@ -268,6 +268,12 @@ New Architecture Support:
   included in official ISPC binaries. To use it, build ISPC from source with
   the `PPC64_ENABLED=ON` CMake option. Usage: `ispc --arch=ppc64le
   --target=generic-i32x4 foo.ispc -o foo.o`.
+* Implementation of two new fast-math modes: ``balanced`` and ``aggressive``.
+  These modes bring more aggressive optimizations compared to the default
+  ``legacy`` mode. The latter was the mode provided so far in previous ISPC
+  versions. It is still available and left unchanged. The fast-math mode can be
+  specified with the option ``--opt=fast-math:<mode>``. See the optimization
+  settings section for more information about the new modes.
 
 Updating ISPC Programs For Changes In ISPC 1.30.0
 -------------------------------------------------
@@ -1674,11 +1680,55 @@ Available options:
   memory accesses beyond the end of an array, which could cause undefined
   behavior if not handled carefully.
 
-- ``fast-math``
+- ``fast-math[:<mode>]``
 
   Perform non-IEEE-compliant optimizations of numeric expressions. These
   optimizations may improve performance but can result in less precise results
-  or different behavior compared to IEEE-compliant math.
+  or different behavior compared to IEEE-compliant math. Various fast-math
+  modes are available. The default one is the ``legacy`` mode.
+
+  Available modes:
+
+  - ``legacy``
+
+    Replace divisions by constants (i.e., ``x / const``) to multiplications by
+    the inverse (i.e., ``x * (1/const)``), and perform reciprocal approximations
+    (i.e., ``x / y`` is replaced with ``x * rcp(y)``). Please read the
+    documentation about ``rcp()`` for more information about its precision.
+    This mode was used in version 1.30 and earlier.
+
+  - ``balanced``
+
+    Enable *imprecise* optimizations. This mode may dramatically change results
+    in floating-point in pathological cases (e.g., up to a point it can produce
+    unexpected infinities or even "not a number" values).
+    It can:
+
+    - Perform algebraically equivalent transformations such as re-association;
+    - Perform floating-point contractions (e.g., fusing a multiply followed
+      by an addition into a fused multiply-and-add);
+    - Treat divisions as multiplications by reciprocals
+      (i.e., consider ``a / b`` to be equivalent to ``a * (1.0 / b)``, and
+      ``a / (b / c)`` to ``a * (c / b)``, and vice versa for both);
+    - Treat the sign of a zero argument or zero result as insignificant.
+
+    As opposed to the ``aggressive`` mode, this one does not produce any
+    *undefined behavior* as such.
+    Technically, this mode currently enables the LLVM's following fast-math
+    flags: ``reassoc``, ``contract``, ``arcp``, ``nsz``. This may change in a
+    future release.
+
+  - ``aggressive``
+
+    Enable *unsafe* optimizations and even more imprecise ones. 
+    In addition to the optimizations enabled by the ``balanced`` mode, this one
+    also assumes there are neither +/- infinities nor "not a number" values in
+    order to perform further optimizations. Codes breaking these assumptions
+    exhibit an *undefined behavior*.
+    This mode can also perform approximations of math functions (e.g., ``sqrt()``)
+    and reciprocals (similar to what ``rcp()`` does).
+    Technically, this mode currently enables all the LLVM's fast-math flags and
+    unexpected values currently produce poison values.
 
 - ``force-aligned-memory``
 
@@ -5452,8 +5502,11 @@ above.
     template <typename T, uint N> T<N> ceil(T<N> a)
     template <typename T, uint N> T<N> trunc(T<N> a)
 
-``rcp()`` computes an approximation to ``1/v``.  The amount of error is
-different on different architectures.
+``rcp()`` computes an approximation to ``1/v``.  The precision is dependent of
+the target CPU architecture. Input/output denormalized numbers may be flushed to
+zeros. For regular numbers, the relative error is usually a few ULPs (unit of
+least precision). Newton-Raphson iterations can be performed to reach such a
+precision.
 
 ::
 
--- a/docs/ispc.rst
+++ b/docs/ispc.rst
@@ -1685,17 +1685,18 @@ Available options:
   Perform non-IEEE-compliant optimizations of numeric expressions. These
   optimizations may improve performance but can result in less precise results
   or different behavior compared to IEEE-compliant math. Various fast-math
-  modes are available. The default one is the ``legacy`` mode.
+  modes are available.
 
   Available modes:
 
-  - ``legacy``
+  - ``legacy`` (default)
 
     Replace divisions by constants (i.e., ``x / const``) to multiplications by
     the inverse (i.e., ``x * (1/const)``), and perform reciprocal approximations
     (i.e., ``x / y`` is replaced with ``x * rcp(y)``). Please read the
     documentation about ``rcp()`` for more information about its precision.
-    This mode was used in version 1.30 and earlier.
+    This was the unique implicit mode available in version 1.30 and earlier. 
+    Its behavior is the same in newer versions.
 
   - ``balanced``
 
--- a/src/opt.cpp
+++ b/src/opt.cpp
@@ -403,7 +403,7 @@ void ispc::Optimize(llvm::Module *module, int optLevel) {
 
         // This should be early (before the InstCombinePass and 
         // InstructionSimplifyPass to actually be effective).
-        if (g->opt.fastMath) {
+        if (g->opt.fastMath != Opt::FastMathMode::None) {
             optPM.addFunctionPass(FastMathPass());
         }
 
--- a/src/opt.cpp
+++ b/src/opt.cpp
@@ -401,7 +401,7 @@ void ispc::Optimize(llvm::Module *module, int optLevel) {
         }
         optPM.addFunctionPass(llvm::DCEPass(), 220);
 
-        // This should be early (before the InstCombinePass and 
+        // This should be early (before the InstCombinePass and
         // InstructionSimplifyPass to actually be effective).
         if (g->opt.fastMath != Opt::FastMathMode::None) {
             optPM.addFunctionPass(FastMathPass());
--- a/src/opt.cpp
+++ b/src/opt.cpp
@@ -401,6 +401,12 @@ void ispc::Optimize(llvm::Module *module, int optLevel) {
         }
         optPM.addFunctionPass(llvm::DCEPass(), 220);
 
+        // This should be early (before the InstCombinePass and 
+        // InstructionSimplifyPass to actually be effective).
+        if (g->opt.fastMath) {
+            optPM.addFunctionPass(FastMathPass());
+        }
+
         // On to more serious optimizations
         optPM.addFunctionPass(llvm::SROAPass(llvm::SROAOptions::ModifyCFG));
         optPM.addFunctionPass(llvm::InferAlignmentPass());
--- a/src/opt.cpp
+++ b/src/opt.cpp
@@ -403,7 +403,7 @@ void ispc::Optimize(llvm::Module *module, int optLevel) {
 
         // This should be early (before the InstCombinePass and
         // InstructionSimplifyPass to actually be effective).
-        if (g->opt.fastMath != Opt::FastMathMode::None) {
+        if (g->opt.fastMath != Opt::FastMathMode::None && g->opt.fastMath != Opt::FastMathMode::Legacy) {
             optPM.addFunctionPass(FastMathPass());
         }
 
--- a/src/opt/FastMath.cpp
+++ b/src/opt/FastMath.cpp
@@ -1,13 +1,7 @@
 #include "FastMath.h"
 #include "builtins-decl.h"
 
-// Enable all fast math flags. This improves the performance of generated
-// codes in some cases but it is much more dangerous. Indeed, it assumes there
-// are no NaN, Inf, negative zeros or subnormal values computed by the target 
-// code. If this assumption is broken, then LLVM store poison values in the 
-// output so the target code basically has an undefined behaviour and may do
-// completely crazy things (very hard to debug for ISPC users).
-// See https://github.com/iree-org/iree/issues/19743 for more information.
+
 //#define AGGRESSIVE_FAST_MATH_OPT
 
 namespace ispc {
@@ -19,15 +13,23 @@ bool FastMathPass::optimizeFpInstructions(llvm::BasicBlock &bb) {
 
     llvm::FastMathFlags fmFlags;
 
-#ifdef AGGRESSIVE_FAST_MATH_OPT
-    fmFlags.setFast(true);
-#else
-    // The three first are required to perform basic FP math reassociation.
-    fmFlags.setAllowContract(true);
-    fmFlags.setAllowReassoc(true);
-    fmFlags.setNoSignedZeros(true);
-    fmFlags.setAllowReciprocal(true);
-#endif
+    if (g->opt.fastMath == Opt::FastMathMode::Safe) {
+        fmFlags.setAllowContract(true);
+        fmFlags.setAllowReassoc(true);
+        fmFlags.setNoSignedZeros(true);
+    } else if (g->opt.fastMath == Opt::FastMathMode::Unsafe) {
+        // This improves the performance of generated codes in some cases but
+        // it is much more dangerous. Indeed, it assumes there are no NaN,
+        // infinities, negative zeros or subnormal values computed by the
+        // target code. If this assumption is broken, then LLVM stores poison
+        // values in the output so the target code basically has an undefined
+        // behaviour and may do completely crazy things (very hard to debug
+        // for ISPC users). For more information, please read:
+        // https://github.com/iree-org/iree/issues/19743
+        fmFlags.setFast(true);
+
+        // TODO: alternative solution: only enable SafeMode + ISPC reciprocals
+    }
 
     // Note: we do modify instruction list during the traversal, so the iterator
     // is moved forward before the instruction is processed.
--- a/src/opt/FastMath.cpp
+++ b/src/opt/FastMath.cpp
@@ -11,9 +11,13 @@ bool FastMathPass::optimizeFpInstructions(llvm::BasicBlock &bb) {
     llvm::FastMathFlags fmFlags;
 
     if (g->opt.fastMath == Opt::FastMathMode::Balanced) {
+        // Note that imprecise rcp instructions are apparently only generated
+        // when at least both math function approximations and reciprocal are
+        // allowed.
         fmFlags.setAllowContract(true);
         fmFlags.setAllowReassoc(true);
         fmFlags.setNoSignedZeros(true);
+        fmFlags.setAllowReciprocal(true);
     } else if (g->opt.fastMath == Opt::FastMathMode::Aggressive) {
         // This improves the performance of generated codes in some cases but
         // it is much more dangerous. Indeed, it assumes there are no NaN,
--- a/src/opt/FastMath.cpp
+++ b/src/opt/FastMath.cpp
@@ -13,11 +13,11 @@ bool FastMathPass::optimizeFpInstructions(llvm::BasicBlock &bb) {
 
     llvm::FastMathFlags fmFlags;
 
-    if (g->opt.fastMath == Opt::FastMathMode::Safe) {
+    if (g->opt.fastMath == Opt::FastMathMode::Balanced) {
         fmFlags.setAllowContract(true);
         fmFlags.setAllowReassoc(true);
         fmFlags.setNoSignedZeros(true);
-    } else if (g->opt.fastMath == Opt::FastMathMode::Unsafe) {
+    } else if (g->opt.fastMath == Opt::FastMathMode::Aggressive) {
         // This improves the performance of generated codes in some cases but
         // it is much more dangerous. Indeed, it assumes there are no NaN,
         // infinities, negative zeros or subnormal values computed by the
@@ -27,8 +27,6 @@ bool FastMathPass::optimizeFpInstructions(llvm::BasicBlock &bb) {
         // for ISPC users). For more information, please read:
         // https://github.com/iree-org/iree/issues/19743
         fmFlags.setFast(true);
-
-        // TODO: alternative solution: only enable SafeMode + ISPC reciprocals
     }
 
     // Note: we do modify instruction list during the traversal, so the iterator
--- a/src/opt/FastMath.cpp
+++ b/src/opt/FastMath.cpp
@@ -1,9 +1,6 @@
 #include "FastMath.h"
 #include "builtins-decl.h"
 
-
-//#define AGGRESSIVE_FAST_MATH_OPT
-
 namespace ispc {
 
 bool FastMathPass::optimizeFpInstructions(llvm::BasicBlock &bb) {
--- a/src/opt/FastMath.cpp
+++ b/src/opt/FastMath.cpp
@@ -1,0 +1,68 @@
+#include "FastMath.h"
+#include "builtins-decl.h"
+
+// Enable all fast math flags. This improves the performance of generated
+// codes in some cases but it is much more dangerous. Indeed, it assumes there
+// are no NaN, Inf, negative zeros or subnormal values computed by the target 
+// code. If this assumption is broken, then LLVM store poison values in the 
+// output so the target code basically has an undefined behaviour and may do
+// completely crazy things (very hard to debug for ISPC users).
+// See https://github.com/iree-org/iree/issues/19743 for more information.
+//#define AGGRESSIVE_FAST_MATH_OPT
+
+namespace ispc {
+
+bool FastMathPass::optimizeFpInstructions(llvm::BasicBlock &bb) {
+    DEBUG_START_BB("FastMath");
+
+    bool modifiedAny = false;
+
+    llvm::FastMathFlags fmFlags;
+
+#ifdef AGGRESSIVE_FAST_MATH_OPT
+    fmFlags.setFast(true);
+#else
+    // The three first are required to perform basic FP math reassociation.
+    fmFlags.setAllowContract(true);
+    fmFlags.setAllowReassoc(true);
+    fmFlags.setNoSignedZeros(true);
+    fmFlags.setAllowReciprocal(true);
+#endif
+
+    // Note: we do modify instruction list during the traversal, so the iterator
+    // is moved forward before the instruction is processed.
+    for (llvm::BasicBlock::iterator iter = bb.begin(), e = bb.end(); iter != e;) {
+        llvm::BasicBlock::iterator curIter = iter++;
+        llvm::Instruction *inst = &*curIter;
+        llvm::BinaryOperator *binOpInst = llvm::dyn_cast<llvm::BinaryOperator>(inst);
+        llvm::FPMathOperator *fpMathInst = llvm::dyn_cast<llvm::FPMathOperator>(inst);
+        if (binOpInst && fpMathInst) {
+            binOpInst->setFastMathFlags(fmFlags);
+            modifiedAny = true;
+        }
+    }
+
+    DEBUG_END_BB("FastMath");
+
+    return modifiedAny;
+}
+
+llvm::PreservedAnalyses FastMathPass::run(llvm::Function &F, llvm::FunctionAnalysisManager &FAM) {
+    llvm::TimeTraceScope FuncScope("FastMathPass::run", F.getName());
+    bool modifiedAny = false;
+
+    for (llvm::BasicBlock &BB : F) {
+        modifiedAny |= optimizeFpInstructions(BB);
+    }
+
+    if (!modifiedAny) {
+        // No changes, all analyses are preserved.
+        return llvm::PreservedAnalyses::all();
+    }
+
+    llvm::PreservedAnalyses PA;
+    PA.preserveSet<llvm::CFGAnalyses>();
+    return PA;
+}
+
+} // namespace ispc
--- a/src/opt/FastMath.cpp
+++ b/src/opt/FastMath.cpp
@@ -1,3 +1,9 @@
+/*
+  Copyright (c) 2026, Intel Corporation
+
+  SPDX-License-Identifier: BSD-3-Clause
+*/
+
 #include "FastMath.h"
 #include "builtins-decl.h"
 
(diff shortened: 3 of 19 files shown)


aneshlya:
> Should --opt=fast-math be equivalent to --fast-math in Clang?

Not necessary - our priority is maximizing safe, predictable speedups for ISPC’s already‑vectorized kernels, not reproducing every single Clang optimization.

> Should we provide fine-grained fast-math optimizations like in Clang? Or alternatively, should we provide multiple level of fast-math optimizations? (IMHO the later help a LOT easily track issues in our math functions since there is far less combination to tests and it makes things simpler from the user PoV).

Layered FastMath optimization makes more sense to me. Probably for us it will be sufficient to have 2 layers of fast-math optimization, like `fast-math:1` and `fast-math:2` (AFAIR ICC has something similar). For example, enabling contract and reassoc + no-signed-zero might yield most of the performance benefit in many cases without inviting as much trouble. These could be considered “fast-math level 1,” whereas “level 2” could be the full-blown mode including reciprocal (and maybe nnan/ninf) transformations. 

> Should we use our own reciprocal function or let LLVM do the optimizations?
LLVM can definitely do a better job but we should do a lot of testing before proceeding with such change.

We should preserve the current behavior of the option during the transition. One way of doing it is having different modes.
```
--opt=fast-math[=<mode>]
    legacy     = today’s reciprocal transforms  ← default
    moderate   = IR flags: contract, reassoc, no-signed-zero
    aggressive = IR flags: contract, reassoc, no-signed-zero, + (either LLVM-driven reciprocal +no-nan + no-inf or manual reciprocal)
```
For legacy mode, we can try to do reciprocal optimization in backend and, if we can achieve the same behavior as now, keep it there (it maybe beneficial if we decide to do reciprocal manually instead of utilizing LLVM).
Eventually we can remove `legacy` mode and switch default to `aggressive`.

zephyr111:
Hello,

I updated this pull request. Here is some notes about it:
 - Tests running on MacOS failed because of the non support of the avx512spr target (required for native half-precision). Is this normal? It is a bit surprising since I do not understand the reason why it would not be possible. I finally split the tests in two parts and disabled one part on MacOS like some other existing tests do. It now works.
 - I fixed a copyright issue with an Intel copyright header with the year like other file, but I am not sure of the policy about it. I am wondering if I am the first external contributor to add a new file. Can you shed some light on it?
 - I finally enabled back `arcp` because it is safer than I thought first. Indeed, it does not result in the generation of `rcp` instructions per se (it does only if the `afn` flag is also enabled), and it is particularly useful to optimize expressions with divisions in many cases. It does not seems really harmful so far (unless `afn` is enabled, especially when combined with `ninf` and `nnan` -- so `fast` for example). Please tell me if this is fine to you.

I am going to write some details about this in the documentation soon.

There is one important concern I found while testing this: the standard library is impacted by the flags. While I did not found critical issues so far, I wonder if the fast-math optimizations could generate numerical instabilities in some pathological cases. Indeed, in some case, numerical instabilities (especially catastrophic cancellation) can be fixed using double-double precision (or an equivalent in single precision) and changing the order of operations (mainly due to the `reassoc` flag) will basically disable numerical correction. For example, `sin(x-2*3.14159265)` is not as precise as `sin(x - 2*3.141592 - 2*6.278330e-07)` when `x` (having a `float` type) is very close to `2 * pi`. The `reassoc` flag enabled by the new fast-math mode should pre-compute `2*3.141592 - 2*6.278330e-07` so the small correction vanishes. That being said, this is not a very good example since even the corrected expression is not very precise... and I cannot find cases so far where such a method completely fix a numerical issue only due to the actual implementation. I think the legacy mode is also impacted but the optimizations are generally not sufficiently aggressive to be an issue in practice. On C compilers (including clang), it looks like functions are not inlined so the problem does not appear, although specific alternative implementations can be called when enabling the fast-math flags. That being said, inlining improves performance, especially on cheap math functions (and especially in single precision). Anyway, I wonder how surprising it can be for users to find out numerical issues on standard math functions when enabling the new fast-math modes. What do you think about this? Should we try to disable this on standard math functions (which means disabling some legit optimizations improving performance)? And if so, then the question is how we can do that in a clean way?

nurmukhametov:
>     * Tests running on MacOS failed because of the non support of the avx512spr target (required for native half-precision). Is this normal? It is a bit surprising since I do not understand the reason why it would not be possible. I finally split the tests in two parts and disabled one part on MacOS like some other existing tests do. It now works.

There is no MacOS with SPR, right @dbabokin?

>     * I fixed a copyright issue with an Intel copyright header with the year like other file, but I am not sure of the policy about it. I am wondering if I am the first external contributor to add a new file. Can you shed some light on it?

If I understand correctly, you need to increment year for existing files and add the standard copyright header to new files. @aneshlya 

>     * I finally enabled back `arcp` because it is safer than I thought first. Indeed, it does not result in the generation of `rcp` instructions per se (it does only if the `afn` flag is also enabled), and it is particularly useful to optimize expressions with divisions in many cases. It does not seems really harmful so far (unless `afn` is enabled, especially when combined with `ninf` and `nnan` -- so `fast` for example). Please tell me if this is fine to you.

I am not sure I follow which mode you are talking about here.

> I am going to write some details about this in the documentation soon.

Yes, please do. It would be much appreciated.

> There is one important concern I found while testing this: the standard library is impacted by the flags. While I did not found critical issues so far, I wonder if the fast-math optimizations could generate numerical instabilities in some pathological cases.

Did you find them in non-legacy mode? Minding the issues like this [one](https://github.com/iree-org/iree/issues/19743) I think that numerical instabilities in some pathological cases are the least of the problems with fast-math flags.

src/opt.cpp
```diff
@@ -401,6 +401,12 @@ void ispc::Optimize(llvm::Module *module, int optLevel) {
         }
         optPM.addFunctionPass(llvm::DCEPass(), 220);
 
+        // This should be early (before the InstCombinePass and
+        // InstructionSimplifyPass to actually be effective).
+        if (g->opt.fastMath != Opt::FastMathMode::None) {
+            optPM.addFunctionPass(FastMathPass());
```
nurmukhametov, line 407: Is this pass no-op in legacy mode?

src/opt.cpp
```diff
@@ -401,6 +401,12 @@ void ispc::Optimize(llvm::Module *module, int optLevel) {
...
+        }
+
```
nurmukhametov, line 409: I wonder if some other passes can be included or enabled under non-legacy mode? For example, will transformations like [these](https://github.com/llvm/llvm-project/blob/main/llvm/lib/Transforms/Utils/SimplifyLibCalls.cpp) be applicable to ISPC generated code with fast-math flags?

aneshlya:
> There is no MacOS with SPR, right @dbabokin?

Right. I think the last Intel HW Apple used was Icelake.

> If I understand correctly, you need to increment year for existing files and add the standard copyright header to new files. 

That's right.

src/opt.cpp
zephyr111: Indeed, It is probably better to disable the whole pass in legacy mode. I am going to fix that. Currently, It should not impact instructions because `fmFlags` is not modified in `FastMathPass::optimizeFpInstructions`, assuming instructions are not flagged yet (AFAIK, no optimization pass does currently that in ISPC). This explains why the tests are successful.
EDIT: fixed.

zephyr111:
Thank you for the review.

> I think the last Intel HW Apple used was Icelake.

That's right. There is an exhaustive list [here](https://en.wikipedia.org/wiki/List_of_Mac_models_grouped_by_CPU_type).

> I am not sure I follow which mode you are talking about here.

I was talking about the `balanced` mode. The `aggressive` mode enabled all optimization.

> Did you find them in non-legacy mode?

No. I only saw some rather-unsafe optimizations of the assembly code of the `exp` function during my tests and assumed it *could* be a problem. In practice, I tested the result of the `exp` function on the full `float` range and I did not see any major precision/numerical issue. The existing imprecision related to subnormal numbers (for x~=-87) appears independently of the fast-math flag mode. There is no issue introduced by the fast math mode `balanced`. I tested a bit the `aggressive` mode too and it looked fine overall (besides computations involving NaN/Inf, which is expected). I tested the `balanced` mode on the modified `log` of the PR #2937 (because the current implementation has too many numerical/precision issues to be actually testable) with `--opt=reset-ftz-daz` enabled (so to avoid subnormal issue existing in both the current implementation and the PR) and results looked fine on the full `float` range. I also partially check with `sin` (in `balanced` mode) and it was similar (i.e. no huge additional issue introduced). Thus, let's say this is not a real problem so far.

> Minding the issues like this https://github.com/iree-org/iree/issues/19743 I think that numerical instabilities in some pathological cases are the least of the problems with fast-math flags.

I fully agree with you that numerical instabilities are the least of the problems for the `aggressive` mode. I am more skeptical about the `balanced` mode.

Indeed, as a user, I expect the `balanced` mode to generate a faster program at the expense of a possibly less-accurate result, but ideally not garbage, e.g. due to any numerical instability issue. I also do not expect any undefined behavior. Moreover, while imprecision issues can indeed results in Inf/NaN values as mentioned in the link, I expect this to be pretty rare in practice (i.e. pathological use-cases). Meanwhile, I expect the `aggressive` mode to optimize the code further at all cost (i.e. no trade-off). This includes undefined behaviors if the rules are not fulfilled.

Regarding the `balanced` mode, there is no undefined behaviors resulting from the fast-math flags, so it can only be less precise (possibly so much it generates unexpected Inf/NaN). As a user, I wish for its precision not to be too much affected. That being said, I have to say that we unfortunately cannot provide any guarantee/bound on its precision in general.

src/opt.cpp
zephyr111: I am wondering whether this is worth it. AFAIK, math functions should be inlined quite early in practice. I am going to do some simple tests. Anyway, I think this can be done in a separate PR later if it worth it.

zephyr111: I did some basic test (on the current implementation) with a function `f` returning a constant (varying) `float`. Interestingly, `f() + f()` and `log(f())` are fully optimized (to a constant), but `exp(f())`, `pow(2.0f, f())` and `pow(f(), 1.0f)`, `sqrt(x*x)` are not, independently of the fast-math mode or even its presence. `sqrt(x)*sqrt(x)` is optimized only in aggressive mode (it generates a sub-optimal code, but better).
From what I can read, SimplifyLibCalls mainly optimizes system's libm calls and/or math intrinsics when it comes to math-related optimizations. I do not expect it to recognize ISPC math functions in stdlib.ispc (it would be really great in `aggressive` mode though). Do you know how it works and if we can adapt this?
Considering that the FastMathPass just add flags and does not actually optimize the code per se and the above notes about SimplifyLibCalls, I do not think SimplifyLibCalls will really help here. It might on double-precision (DP), in the `aggressive` mode and without the SVML though since DP math functions currently call the system's libm. This will likely change in the future though (because the system libm scalar functions are pretty slow). It also might help further for `sqrt` or similar intrinsics.

I think it clearly deserve a separate discussion considering the large number of optimizations of this pass (not just related to fast-math by the way), its impact, and the possible implication on ISPC so to benefit from it (as well as if it is worth it).

zephyr111: Oh, I found surprising things:
- functions are more optimized on ARM/Neon, even without fast-math! Do you know why?
- `1 / sqrt(x)` is optimized using a `vrsqrt` instruction + 1 Newton-Raphson iter, even in `balanced` mode (i.e. without math approx). It seems quite precise in practice, so it is OK.
See: https://godbolt.org/z/YPa76fEfY

elena901:
~~Why FMA is not enabled?~~ Ok, it's enabled. Speaking in general, my proposal is untypical and not unquestionable. I would like to have `fastmath{}` scoping similar to `unmasked{}`.

zephyr111:
@elena901 Yes it is enabled, [even without FastMath already](https://godbolt.org/z/1EsncTvoh). See also the tests. I like the `fastmath{}` proposal as an ISPC user, but I am afraid to introduce a possibly-flawed or ambiguous language construct as a contributor. I think we should discuss about this in a separate channel. Indeed, fast-math is already provided as a global flag so far (which is consistent with C/C++ compilers), and this PR only makes optimizations more aggressive (depending on the mode picked by the users). Besides, there is a lot to say about such a construct so it will be more clear in a dedicated channel than here. I am going to open a feature-request issue about it referencing your message.

src/opt.cpp
nurmukhametov: > From what I can read, SimplifyLibCalls mainly optimizes system's libm calls and/or math intrinsics when it comes to math-related optimizations. 

Right, so it can be useful for ISPC only for very limited use cases when math functions have only generic implementation via llvm.builtins. 

> I do not expect it to recognize ISPC math functions in stdlib.ispc (it would be really great in `aggressive` mode though). Do you know how it works and if we can adapt this?

They have a handful of patterns that they recognize and rewrite in LLVM IR level. ISPC probably can have similar patterns on AST level around the place where current x / y -> x * rcp(y) are.

nurmukhametov: > Oh, I found surprising things:
> 
>     * functions are more optimized on ARM/Neon, even without fast-math! Do you know why?

No, I don't.

nurmukhametov (changes requested):
The code change looks good to me, but you need to rebase and resolve the conflict. Please also update `docs/ispc.rst` with information about these new fast-math modes.

elena901:
My expectations (as an ISPC user) of the fast-math mode are simple: the default mode can handle "poisons", the fast-math doesn't and the precision of trig. functions is less (speaking roughly), +fma. But i'm not expecting reassoc at all because of the proposed SPMD model and the user agreement.

UPDATE:

Please don't get me wrong, speaking in general `reassoc` is a good thing. And to my surprise this PR wasn't merged with [the matmul example](https://github.com/ispc/ispc/pull/3307) last year.

I hope I'm not rude but i don't see too much sense to run neural networks while ISPC is not able to "pipeline" programs due to the latency of the main instructions (and `vnni` for example) and (increasing) number of execution ports and the unit width on modern cpus. On the other hand the reassoc approach [ruins](https://github.com/ispc/ispc/issues/2903#issuecomment-3968700158) the SPMD model.

zephyr111:
@elena901
I did not understand your expectation about fast-math. Can you clarify them? Please note that reassoc is the most impacting flag of fast-math regarding performance. Without it, a lot of transformations of complex expression are not possible preventing a significant performance boost. If you do not want it, it is certainly better to stay with the legacy mode or just do the rcp optimization yourself (this is what I do most of the time so far).

> i don't see too much sense to run neural networks while ISPC is not able to "pipeline" programs due to the latency of the main instructions [...]

First of all, ISPC is not only used for neural networks. There are a many users using it in games, 3D tools, scientific/HPC computing, image processing, etc. I am one of them. Moreover, ISPC can (nearly) saturate the CPU in many cases (IDK for VNNI specifically though). If you reference the avx2-i64x8-target discussion, then be aware that the main issue happens because of loop-carried dependencies (on the basic sum), and the AVX2 target (while AVX-512 has quad-pumping for FP32, see `avx512skx-x64`). You should neither generalize to all codes, nor all targets.

zephyr111:
I rebased the branch, added some documentation about the new modes and updated the changelogs. I also added some information about the precision of `rcp()` (used by the old `legacy` mode) and clarified/detailed a bit its implementation.

elena901:
@zephyr111 Speaking in general all is ok 👍 

@aneshlya Arina proposed `moderate` as the derived name for the first layer. Honestly i like it much more than `balanced`.

elena901:
I am not a great software architect, but i don't like `FastMath.cpp` as a separate module (at this moment of development process at least).

_remark_> Btw i think ISPC needs a module for prefetching. I hope to see an option `--disable-prefetching` one day.

elena901:
EDITED a bit.

@zephyr111 there is a separate discussion about ~~reciprocals~~ [the floating point model](https://github.com/ispc/ispc/issues/666).

docs/ispc.rst
```diff
@@ -1674,11 +1680,55 @@ Available options:
...
+
+  Available modes:
+
+  - ``legacy``
+
+    Replace divisions by constants (i.e., ``x / const``) to multiplications by
+    the inverse (i.e., ``x * (1/const)``), and perform reciprocal approximations
+    (i.e., ``x / y`` is replaced with ``x * rcp(y)``). Please read the
+    documentation about ``rcp()`` for more information about its precision.
+    This mode was used in version 1.30 and earlier.
```
nurmukhametov: It is slightly confusing. Will 1.31 change the default mode?

nurmukhametov (approved):
LGTM!

elena901:
@zephyr111 What about to leave `reassoc` for the aggressive mode only?

elena901:
@zephyr111 What about to leave `reassoc` for the aggressive mode only?

My state is clear: `reassoc` is not politically correct in terms of SPMD.

docs/ispc.rst
zephyr111: Is this better now?

nurmukhametov: Yes, thank you!

zephyr111:
> What about to leave reassoc for the aggressive mode only?

@elena901 As said previously, `reassoc` is the most impacting flag of fast-math regarding performance. Removing it makes the `balanced` not very interesting IMHO. More importantly, people who want it (like me as a user) without the scary undefined behavior issues or the aggressive mode would have no real alternative. I cannot really advise people (including the one of the initial issue meant to be solved by this PR) to use the aggressive mode. That being said, we could add another fast-math mode for that, although I prefer not to add it unless there is a good example showing its interest (just to keep fast-math modes rather simple). If you have found a practical use-case where it make sense, then please share it.

-----

> reassoc is not politically correct in terms of SPMD.

@elena901 I do not understand what you mean. This is rather independent notions to me. The fast-math applies on each program instance independently. Btw, the same thing applies to fast-math options in vectorized C code.

-----

> Arina proposed moderate as the derived name for the first layer. Honestly i like it much more than balanced.

@elena901 Regarding the name, I have to say I prefer a bit to keep the term `balanced` (especially now the code/doctests have been written/modified). I am open to changes on this if anyone else also prefer the term `moderate` too. @aneshlya @nurmukhametov What do you prefer?

nurmukhametov:
> > Arina proposed moderate as the derived name for the first layer. Honestly i like it much more than balanced.
> 
> @elena901 Regarding the name, I have to say I prefer a bit to keep the term `balanced` (especially now the code/doctests have been written/modified). I am open to changes on this if anyone else also prefer the term `moderate` too. @aneshlya @nurmukhametov What do you prefer?

No preference from me.

dbabokin:
@codex review

src/opt/FastMath.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+    // Note: we do modify instruction list during the traversal, so the iterator
+    // is moved forward before the instruction is processed.
+    for (llvm::BasicBlock::iterator iter = bb.begin(), e = bb.end(); iter != e;) {
+        llvm::BasicBlock::iterator curIter = iter++;
+        llvm::Instruction *inst = &*curIter;
+        llvm::BinaryOperator *binOpInst = llvm::dyn_cast<llvm::BinaryOperator>(inst);
+        llvm::FPMathOperator *fpMathInst = llvm::dyn_cast<llvm::FPMathOperator>(inst);
+        if (binOpInst && fpMathInst) {
+            binOpInst->setFastMathFlags(fmFlags);
```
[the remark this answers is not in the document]
zephyr111: Do not think adding the fast-math flags to a `Call` is actually useful. If the call is in the same TU, it should have no impact. If it is in another translation unit (TU) with the same fast-math mode, it should also have no impact. If the mode is different, then adding it with LTO might even be dangerous (I wonder if this could add the flags to the instructions coming from the function call once inlined). It could have been useful for libm functions, but ISPC has its own implementation in stdlib.ispc.

Regarding `Select`, I have no idea if this could actually help to add the flags on them. At first glance, it looks it is not a problem for optimizing code. Indeed, the following ISPC code generates no IR `sqrt` call (nor `vsqrtps` ASM instruction in AVX) and a single `fmul` in `aggressive` mode:

```ispc
float test(float x)
{
	float a = select(x < 1.0, 2*x, sqrt(x));
	return select(x < 1.0, 3*x, a * a);
}
```

LLVM IR result:

```
define <4 x float> @test___vyf(<4 x float> %x, <4 x i32> %__mask) local_unnamed_addr #0 {
allocas:
  %less_x_load_ = fcmp olt <4 x float> %x, splat (float 1.000000e+00)
  %mul__x_load10 = fmul fast <4 x float> %x, splat (float 3.000000e+00)
  %blend.i.i30 = select <4 x i1> %less_x_load_, <4 x float> %mul__x_load10, <4 x float> %x
  ret <4 x float> %blend.i.i30
}
```

The `balanced` mode do generate an `sqrt`.

Regarding `PHI`, I do not expect this to have any significant impact but I am not very familiar with it. I just tested on [this IR](https://godbolt.org/z/oTG88KMYn) (coming from [this code](https://godbolt.org/z/35aGM41Kb) in `-O0`) and adding `fast` to the `phi` did not impact the optimizations.

Regarding `fneg`, this is true. However, not having the flags does not prevent the optimizations in all the tests I did so far. For example, `-(1.0 / (1.0/-x))`, `1.0 - (-x)`, `-(-x)`, `(-x) * (-x)` and `-((-x) * x)` are properly optimized in the `balanced` mode. I do not see what kind of optimization could be done with the flags on `fneg`. Thus, I do not think this is really useful though it looks better to do it (and also more consistent). I can add this if you wish.

Regarding `sqrt`, this is currently not a problem either: `sqrt(-x)`, `1.0/sqrt(x)` and `sqrt(x) * sqrt(x)` are properly optimized in aggressive mode. Only `sqrt(x * x)` is strangely not.

elena901:
> Btw, the same thing applies to fast-math options in vectorized C code.

@zephyr111 This is exactly what i meant. ISPC is not C/C++. It's a SPMD compiler. So (it's just a formula for all users) wanna `reassoc` ? => goto C/C++. It's like how things seem to me.

> if anyone else also prefer

It's not about preferences. It's more about organization at the development process.

Sorry for the straightforward answer.

**UPDATE:**
>reassoc is not politically correct in terms of SPMD

In terms of ISPC's SPMD first of all.

src/opt/FastMath.cpp
```diff
@@ -0,0 +1,69 @@
+#include "FastMath.h"
```
aneshlya, line 7: Please add copyright here and in the header file:
```
/*
  Copyright (c) 2026, Intel Corporation

  SPDX-License-Identifier: BSD-3-Clause
*/
```

src/opt/FastMath.cpp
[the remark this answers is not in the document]
aneshlya: @zephyr111 Thanks for the detailed analysis. 
I think the Codex suggestion has some sense especially for aggressive mode.

Regarding `call`: You're right that for stdlib.ispc functions this doesn't matter since they get inlined. However, LLVM intrinsics like `llvm.sqrt.*`,  `llvm.fma.*`, `llvm.pow.*` etc. with FP types could benefit from the flags.
As for other operators, we may add them for consistency with LLVM: Clang is applying the flags to all created `FPMathOperator` instructions (including FNeg, Call, Select, PHI with FP types). 
I briefly looked into LLVM code and it looks like `select` with proper fast math flags can be transformed to `fadd` or `@llvm.maxnum.f32`.  PHI nodes with fast-math flags are less common and used in loop optimizations. The flags propagate through PHI nodes to enable vectorization and reduction pattern detection which is usually not applicable to ispc but again, I don't see why not to add it for consistency.

If you think adding the fast math flags to all `FPMathOperator` brings more harm than benefit, please add justification in the comments to this pass.

aneshlya (commented):
The changes look good to me and I don't see anything blocking. The `balanced` wording looks good. Please see my two comments in `FastMath.cpp`

src/opt/FastMath.cpp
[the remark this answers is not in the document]
zephyr111: Thank you for the information! Adding fast-math flags is on more instructions is OK to me overall and I agree with you on the consistency aspect. 
That being said, I am worried about what fast-math flags could do on *user function-call* instructions specifically. It is not clear what it does for me. I am scared it could cause instructions from inlined non-fast-math functions to be incorrectly tagged as fast-math one (i.e. transitivity) -- which should only happens with LTO AFAIK. I did not found documentation about it. Did you see how/whether user call instruction could be optimized when tagged?

> If you think adding the fast math flags to all FPMathOperator brings more harm than benefit, please add justification in the comments to this pass.

AFAIR, the main initial reason why I did not called `setFastMathFlags` directly on `fpMathInst` is that it is a private member function. Then, I found out that call instructions could also be flagged and I got worried about it, so I considered that only targeting binary-operations should be a safe subset and certainly enough to get good performance boost.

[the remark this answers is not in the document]
zephyr111: @aneshlya It looks like the inlining of `call fast` user-functions does not results in FPMath instructions being tagged with `fast`: the `fsub` is not optimized out nor tagged after optimization steps while the function call is properly inlined. See [this example](https://godbolt.org/z/e617WK5ja) coming from [this code](https://godbolt.org/z/W7YarbbxT) (but without the `noinline` on `base` so it can be optimized by `opt` later and no debugging information). This makes me less concerned about this potential issue, but I would be more relieved if you can confirm the fast-math flags are never transitively applied on `fast`-tagged inlined user function calls.

PS: I fixed the code so to include all kind of FP-math operators

zephyr111:
@aneshlya I applied the changes! Please read my answer about the `FPMathOperator` (and more specifically the `fast` tagging of user-function calls).

zephyr111:
@elena901 I still do not understand the link between SPMD and `reassoc`. CUDA and OpenCL (as well as GPU shaders AFAIK) also perform re-association with (full) fast-math flags. Both are very similar to ISPC and follow the same SPMD programming approach. To quote the ISPC User's Guide:

> The SPMD programming approach is widely known to graphics and GPGPU programmers; it is used for GPU shaders and CUDA* and OpenCL* kernels, for example.

Fast-math flags are not enabled by default and shall never be (since they break the IEEE-754 standard).

src/opt/FastMath.cpp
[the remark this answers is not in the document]
aneshlya: In LLVM when a call site has FMF and the callee gets inlined, it uses intersection semantics - a flag is preserved in the inlined instruction only if it was present on both the call site and the original instruction. So I think it should be fine!

aneshlya (approved):
Looks good! Thank you so much for the contribution!

elena901:
@zephyr111 as i said all is ok :+1: .

> It's not about preferences. It's more about organization in the development process.

:point_up:

elena901:
>CUDA and OpenCL

@zephyr111 
CUDA and OpenCL are not SPMD compilers (in terms of ISPC) first of all. But as i said, all is good :+1:

elena901:
> Speaking in general, my proposal is untypical and not unquestionable. I would like to have `fastmath{}` scoping similar to `unmasked{}`.

I slightly change my initial proposal. I would like to see an option `--enable-fma` or FMA macros at least. [Related.](https://github.com/ispc/ispc/issues/3757)

zephyr111:
@elena901 FMA are already enabled by default (because they are actually more precise than what the user wrote) if the target supports it, and they can be manually disabled with `--opt=disable-fma`.

elena901:
> @elena901 FMA are already enabled by default (because they are actually more precise than what the user wrote) if the target supports it

@zephyr111 I know and it [doesn't](https://github.com/ispc/ispc/issues/3757) work properly.

Issue #3495, opened 2025-07-16
cbrt-float.ispc test is failing on avx512*-x64 targets


aneshlya:
Example run: https://github.com/ispc/ispc/actions/runs/16325104505/job/46114369209

PR #3496, opened 2025-07-16
Compute relative error in cbrt-float.ispc

aneshlya:
## Description
Compute relative error in cbrt-float.ispc. Currently the test fails with big inputs (on 64-width targets).

## Related Issue
- [x] Fixes https://github.com/ispc/ispc/issues/3495

## Checklist

### DIFF
--- a/tests/func-tests/cbrt-float.ispc
+++ b/tests/func-tests/cbrt-float.ispc
@@ -3,13 +3,10 @@
 task void f_f(uniform float RET[], uniform float aFOO[]) {
     varying float f = aFOO[programIndex];
     varying float calc = cbrt(f) * cbrt(f) * cbrt(f);
-    if (calc < 0.) calc = -calc;
-    if (abs(calc - f) < 1e-5)
-    {
+    if (isnan(calc) == isnan(f) && (abs(f) >= 1e-6 && abs((calc - f) / f) < 2e-6 || abs(calc - f) < 2e-12 || isnan(f))) {
         RET[programIndex] = 1;
     }
-    else
-    {
+    else {
         RET[programIndex] = 0;
     }
 }


zephyr111:
It is a quite surprising for the result to be target-dependent. Indeed, AFAIK all ISPC code executed on CPU should be (mostly) IEEE-754 compliant unless fast-math optimizations are enabled.
Thus, I am curious of what kind of input value can cause this issue. Do you know which input values fail?
I think we should check if this is normal. It might be normal with subnormal numbers for example (due to the intrinsic lower precision). If the problem is coming from subnormal, then I think we should adapt the condition (i.e. add something specific for subnormal values) rather than increasing the threshold.

aneshlya:
The inputs are:
```
f: [1.000000,2.000000,3.000000,4.000000,5.000000,6.000000,7.000000,8.000000,9.000000,10.000000,11.000000,12.000000,13.000000,14.000000,15.000000,16.000000,17.000000,18.000000,19.000000,20.000000,21.000000,22.000000,23.000000,24.000000,25.000000,26.000000,27.000000,28.000000,29.000000,30.000000,31.000000,32.000000,33.000000,34.000000,35.000000,36.000000,37.000000,38.000000,39.000000,40.000000,41.000000,42.000000,43.000000,44.000000,45.000000,46.000000,47.000000,48.000000,49.000000,50.000000,51.000000,52.000000,53.000000,54.000000,55.000000,56.000000,57.000000,58.000000,59.000000,60.000000,61.000000,62.000000,63.000000,64.000000]
 cbrt(f): [1.000000,1.259921,1.442250,1.587401,1.709976,1.817121,1.912931,2.000000,2.080084,2.154435,2.223980,2.289428,2.351335,2.410142,2.466212,2.519842,2.571281,2.620741,2.668402,2.714418,2.758924,2.802039,2.843867,2.884499,2.924018,2.962496,3.000000,3.036589,3.072317,3.107233,3.141381,3.174802,3.207535,3.239612,3.271066,3.301927,3.332222,3.361975,3.391211,3.419952,3.448218,3.476027,3.503398,3.530349,3.556893,3.583048,3.608826,3.634241,3.659306,3.684031,3.708430,3.732511,3.756285,3.779763,3.802953,3.825862,3.848501,3.870877,3.892997,3.914868,3.936497,3.957891,3.979058,4.000000]
calc: [1.000000,2.000000,3.000000,4.000000,5.000000,6.000000,6.999998,8.000000,9.000000,10.000000,10.999998,12.000000,13.000002,13.999999,15.000000,16.000002,16.999998,18.000000,19.000000,20.000002,20.999996,21.999996,23.000002,23.999996,24.999998,26.000006,27.000000,27.999998,29.000000,30.000000,30.999998,32.000000,33.000008,33.999996,35.000004,35.999996,37.000004,37.999992,38.999992,40.000000,41.000011,41.999996,42.999996,44.000008,44.999992,46.000000,47.000004,47.999996,48.999996,50.000000,50.999992,52.000008,52.999985,54.000000,55.000015,55.999981,57.000004,57.999996,59.000008,60.000000,61.000000,61.999992,63.000015,64.000000]
 calc - f: [0.000000,0.000000,-0.000000,0.000000,0.000000,-0.000000,-0.000002,0.000000,0.000000,0.000000,-0.000002,0.000000,0.000002,-0.000001,0.000000,0.000002,-0.000002,0.000000,0.000000,0.000002,-0.000004,-0.000004,0.000002,-0.000004,-0.000002,0.000006,0.000000,-0.000002,0.000000,0.000000,-0.000002,0.000000,0.000008,-0.000004,0.000004,-0.000004,0.000004,-0.000008,-0.000008,0.000000,0.000011,-0.000004,-0.000004,0.000008,-0.000008,0.000000,0.000004,-0.000004,-0.000004,0.000000,-0.000008,0.000008,-0.000015,0.000000,0.000015,-0.000019,0.000004,-0.000004,0.000008,0.000000,0.000000,-0.000008,0.000015,0.000000]
RET: [1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,0.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,0.000000,1.000000,0.000000,0.000000,1.000000,1.000000,1.000000,1.000000,1.000000,1.000000,0.000000,1.000000]
```

zephyr111:
Ah, Ok. It makes sense. This happens because we compute an absolute difference so the bigger the input the bigger the error (and the input value is indeed dependent of the size). 

For floating-point numerical functions, it is generally better to compute the relative error but for 0 there is a singularity so in this case the absolute error can be computed instead. This is for example what Numpy and Scipy do (e.g. `np.allclose` and `np.isclose`). Here is a simple fix based on this approach:

```ispc
if (isnan(calc) == isnan(f) && (abs(f) >= 1e-6 && abs((calc - f) / f) < 2e-6 || abs(calc - f) < 2e-12 || isnan(f))) {
    // OK
}
```

This example performs an absolute difference with a threshold of 1e-12 (in `]-1e-6;+1e-6[`) and a relative difference with a threshold of 1e-6 (outside the pathological range). It assumes `f` is not +/- inf (like the current one), but checks for incorrect NaN generation (which is important) unlike the current test (which ignores them). You can tweak the constants (which are rather strict for single precision -- and much more than the current test).

A better (more accurate) alternative solution is to compute the error in units of least precision (ULP). Sleef does that for example. Computing the error in ULP is AFAIR a bit more complex though, but it would make the test far more precise than the current one to detect numerical issues.

EDIT: the formula was incorrect (the comment should have been *"// OK"* and the NaN check was incorrect), so I fixed it. Sorry.

aneshlya:
> For floating-point numerical functions, it is generally better to compute the relative error but for 0 there is a singularity so in this case the absolute error can be computed instead.

Thanks, it's a good idea!

Currently we don't have unified approach for computing error for float tests. It would be good to compute relative error everywhere and eventually switch to ULP when we have some understanding about the precision of our math functions.

tests/func-tests/cbrt-float.ispc
```diff
@@ -4,12 +4,10 @@ task void f_f(uniform float RET[], uniform float aFOO[]) {
     varying float f = aFOO[programIndex];
     varying float calc = cbrt(f) * cbrt(f) * cbrt(f);
     if (calc < 0.) calc = -calc;
-    if (abs(calc - f) < 1e-5)
-    {
+    if (abs(f) >= 1e-6 && abs((calc - f) / f) < 2e-6 || abs(calc - f) < 2e-12 || isnan(calc) != isnan(f)) {
```
zephyr111: The provided formula was not correct. Sorry. I fixed it. It should be:
```ispc
if (isnan(calc) == isnan(f) && (abs(f) >= 1e-6 && abs((calc - f) / f) < 2e-6 || abs(calc - f) < 2e-12 || isnan(f))) {
```
Overall, the difference is only the error handling of NaN: the new formula correctly tracks NaN issues. I verified it by testing NaN values for both `f` and `calc`.

tests/func-tests/cbrt-float.ispc
```diff
@@ -4,12 +4,10 @@ task void f_f(uniform float RET[], uniform float aFOO[]) {
     varying float f = aFOO[programIndex];
     varying float calc = cbrt(f) * cbrt(f) * cbrt(f);
     if (calc < 0.) calc = -calc;
```
zephyr111: I do not think we should keep this line. Indeed, the sign must be preserved by the `cbrt` function (except for pathological input if fast-math is enabled). If it does not, then this is a bug.

nurmukhametov (approved):

zephyr111 (changes requested):

tests/func-tests/cbrt-float.ispc
aneshlya: Updated

PR #3506, opened 2025-07-21
Fix global state issues for ispc usage as a library

aneshlya:
## Description
This PR fixes several issues with global state management found during testing of ispc as a library. It also adds support of ispc library to lit configuration and adds the lit test.

* Added a new lit test (`tests/lit-tests/ispc-library.ispc`) to validate the ISPC C++ library's API
* Updated error reporting in `src/preprocessor.cpp` by isolating diagnostic infrastructure and directly outputting errors to avoid corrupted diagnostic states.
* Transitioned from `char*` to `std::string` for `file` and `CPU` arguments in `src/args.cpp`, `src/args.h`, and `src/ispc_impl.cpp`, to avoid pointer corruption.
* Introduced a `GlobalStateGuard` class in `src/ispc_impl.cpp` to manage and restore global state during compilation
* Extended `simple_lib` example.

## Related Issue

## Checklist


tests/lit-tests/ispc-library.ispc
```diff
@@ -0,0 +1,110 @@
+// This test verifies that API for ISPC C++ library works correctly
+// and without state corruption.
+
+// RUN: %{cxx} -x c++ -std=c++17 -I%{ispc_include} %s -L%{ispc_lib} -lispc -o %t.bin
```
nurmukhametov: I think we need to allow having *.cpp files in lit-tests. It looks strange that we name files that are only C++ as .ispc.

src/ispc_impl.cpp
```diff
@@ -66,6 +83,8 @@ class ISPCEngine::Impl {
     }
 
     int Execute() {
+        GlobalStateGuard guard; // The guard to protect global state
```
nurmukhametov, line 87: I think we need to add here: TODO!: remove the global state.

tests/lit-tests/ispc-library.ispc
nurmukhametov: It seems to me that this `ispc-library.cpp` is not running until you add it to `config.suffixes` in `lit.cfg`.

aneshlya: Oh, it's a good catch! Thanks!

nurmukhametov (approved):
LGTM!

Issue #2376, opened 2022-09-22
ISPC should produce error when varying index is used with short vectors


aneshlya:
Currently ISPC crashes on such code: https://godbolt.org/z/zaoMh6ajv
```
typedef int<3> int3;

export void test(uniform float ret[], uniform float b) {
    int3 v0 = { b, 2*b, 3*b };
    int3 v1 = { b, b/2, b/3 };
    ret[programIndex] = (v0+v1)[programIndex];
}
```
Expected behavior: produce error and exit nicely.

zephyr111:
This issue might be related to #3220 and #2496 because of the `(v0+v1)`. 
Indeed, it correctly builds without segmentation fault if `(v0+v1)` is *stored in a temporary variable* (see https://godbolt.org/z/ba3veTTWE).

aneshlya:
> Expected behavior: produce error and exit nicely.

Actually, it's totally valid code. The compiler should compile it successfully.

PR #3507, opened 2025-07-22
Fix of several crashes related to short vector initialization/indexing

### LINKED ISSUE #3484: ISPC fails to handle case when function with varying return type is used to initialize uniform variable
This is the way to trigger ISPC crash:
```ISPC
static varying float load(const uniform float *values) { return 0.0f; }

export void func(uniform float values[][programCount])
{
    uniform float<3> col_xyz = {load(&values[0][0]),
                                load(&values[0][0]),
                                load(&values[0][0])};
}
```

Godbolt example: https://godbolt.org/z/37TE8jb5Y

aneshlya:
## Description
Fixed three separate compiler crashes that occurred when processing invalid or edge-case ISPC code related to short vectors.
1. Use ISPC's `symVectorType->LLVMType(g->ctx)` instead of manually constructing `llvm::FixedVectorType::get()`
2. Null pointer check (src/expr.cpp:4831)
3. Use PointerType::GetVarying() when indexType->IsVaryingType()
4. Early exit if we can't construct short vector from initializer list due to different variability.

## Related Issue
- [x] Fixes #3484, #2376 

## Checklist

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -4996,7 +4996,12 @@ llvm::Value *IndexExpr::GetValue(FunctionEmitContext *ctx) const {
             Assert(m->errorCount > 0);
             return nullptr;
         }
-        lvType = PointerType::GetUniform(st->GetElementType());
+        // Use varying pointer type if the index is varying, uniform otherwise
+        if (indexType->IsVaryingType()) {
+            lvType = PointerType::GetVarying(st->GetElementType());
+        } else {
+            lvType = PointerType::GetUniform(st->GetElementType());
+        }
 
         // And do the indexing calculation into the temporary array in memory
         ptr = ctx->GetElementPtrInst(tmpPtrInfo->getPointer(), LLVMInt32(0), index->GetValue(ctx),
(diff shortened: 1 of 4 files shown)


src/expr.cpp
```diff
@@ -941,6 +939,11 @@ void ispc::InitSymbol(AddressInfo *ptrInfo, const Type *symType, Expr *initExpr,
 
                             // Make type conversion
                             if (!Type::EqualIgnoringConst(symType, vResType)) {
+                                // Early return for unsupported conversions (e.g., varying to uniform)
+                                if (symVectorType->IsUniformType() && vResType->IsVaryingType()) {
```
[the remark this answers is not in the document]
aneshlya: This will be caught later by semantic analysis

src/expr.cpp
```diff
@@ -4987,7 +4996,12 @@ llvm::Value *IndexExpr::GetValue(FunctionEmitContext *ctx) const {
             Assert(m->errorCount > 0);
             return nullptr;
         }
-        lvType = PointerType::GetUniform(st->GetElementType());
+        // Use varying pointer type if the index is varying, uniform otherwise
+        if (indexType->IsVaryingType()) {
+            lvType = PointerType::GetVarying(st->GetElementType());
+        } else {
+            lvType = PointerType::GetUniform(st->GetElementType());
+        }
```
nurmukhametov, line 5004: This fixes #2376, but it suggests a different expected behavior (produce an error and exit nicely), so I am confused.

aneshlya: Right, the issue description was not correct, I've updated it. It's valid code and ispc should be able to compile it.

PR #3516, opened 2025-07-29
WIP: JIT functionality

aneshlya:
## Description
Brief description of changes

## Related Issue

## Checklist

### DIFF
--- a/src/include/ispc/ispc.h
+++ b/src/include/ispc/ispc.h
@@ -77,11 +77,58 @@ class ISPCEngine {
      */
     int Execute();
 
+    /**
+     * @brief Compiles ISPC code from a file using JIT compilation.
+     * @param filename Path to the ISPC source file.
+     * @return 0 on success, non-zero on failure.
+     */
+    int CompileFromFileToJit(const std::string &filename);
+
+    /**
+     * @brief Retrieves a function pointer from JIT-compiled code.
+     * @param functionName Name of the exported function.
+     * @return Function pointer or nullptr if not found.
+     */
+    void *GetJitFunction(const std::string &functionName);
+
+    /**
+     * @brief Sets a user-provided runtime function for JIT compilation.
+     * Runtime functions must be provided before calling CompileFromFileToJit().
+     * @param functionName Name of the runtime function (e.g., "ISPCLaunch", "ISPCSync", "ISPCAlloc")
+     * @param functionPtr Pointer to the user-provided function implementation
+     * @return true on success, false if functionName is invalid or functionPtr is null
+     */
+    bool SetJitRuntimeFunction(const std::string &functionName, void *functionPtr);
+
+    /**
+     * @brief Clears a specific runtime function.
+     * @param functionName Name of the function to clear
+     */
+    void ClearJitRuntimeFunction(const std::string &functionName);
+
+    /**
+     * @brief Clears all runtime functions.
+     */
+    void ClearJitRuntimeFunctions();
+
     ~ISPCEngine();
 
   private:
     ISPCEngine();
 
+    /**
+     * @brief Checks if the engine is in JIT mode.
+     * @return true if JIT mode is active, false otherwise.
+     */
+    bool IsJitMode() const;
+
+  public:
+    /**
+     * @brief Clears all JIT-compiled code.
+     */
+    void ClearJitCode();
+
+  private:
     class Impl;
     std::unique_ptr<Impl> pImpl;
 };
--- a/src/module.cpp
+++ b/src/module.cpp
@@ -2688,6 +2688,60 @@ int Module::CompileAndOutput(const char *srcFile, Arch arch, const char *cpu, st
     }
 }
 
+std::unique_ptr<llvm::Module> Module::CompileToLLVMModule(const char *srcFile, Arch arch, const char *cpu,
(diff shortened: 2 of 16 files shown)


aneshlya:
Ready for initial review.

tests/func-tests/memset-uniform.ispc
```diff
@@ -1,5 +1,6 @@
 #include "test_static.isph"
 // rule: skip on arch=xe64
+// rule: skip on jit=true
```
nurmukhametov, line 3: What is wrong with this test?

src/module.cpp
```diff
@@ -2688,6 +2688,57 @@ int Module::CompileAndOutput(const char *srcFile, Arch arch, const char *cpu, st
...
+        Module::OutputFlags defaultFlags;
+        ISPCTarget target = targets.empty() ? ISPCTarget::host : targets[0];
+        auto targetPtr =
+            Target::Create(arch, cpu, target, defaultFlags.getPICLevel(), defaultFlags.getMCModel(), false);
+        if (!targetPtr) {
+            Error(SourcePos(), "Failed to create target for JIT compilation");
+            return nullptr;
+        }
+        // Set global target (this will be cleaned up by Target destructor)
+        g->target = targetPtr.release();
```
nurmukhametov, line 2699: Manipulating this global state is not thread-safe, so we probably need to mention it somewhere.

src/include/ispc/ispc.h
```diff
@@ -77,11 +77,57 @@ class ISPCEngine {
...
+
+    /**
+     * @brief Clears all runtime functions.
+     */
+    void ClearJitRuntimeFunctions();
+
+    /**
+     * @brief Clears all JIT-compiled code.
+     */
+    void ClearJitCode();
```
nurmukhametov, line 113: Is it possible to destroy ISPCEngine but continue using JITed code?

src/module.cpp
aneshlya: It will go to the docs.

src/include/ispc/ispc.h
aneshlya: Technically yes, but I would not say that it's "officially supported".
The destructor in ispc_impl.cpp leaks the JIT engine:
```
  void ReleaseJitForShutdown() {
      // Store the released pointer to acknowledge we're intentionally leaking it
      [[maybe_unused]] auto *leaked_jit = m_jit.release();
  }
```
Since the JIT engine is leaked (not destroyed), the machine code remains in memory and function pointers continue to work.

tests/func-tests/memset-uniform.ispc
aneshlya: Don't know. It's failing in CI only on macos. Can't reproduce it locally.

aneshlya:
The `print` tests are disabled now because I can't find a way to export `__do_print` symbol into ispc.lib on Windows yet. On Linux/macOS there is no such problem, the symbol is present in libispc.so.

aneshlya:
I've added 2 more commits: one with docs and another removing dummy first argument in the APIs.

docs/ispc.rst
```diff
@@ -1616,6 +1619,130 @@ For more control over the compilation process, use the ``ISPCEngine`` class::
...
+  Multi-target compilation will result in an error.
+
+* **Thread Safety**: JIT compilation is not thread-safe. Use JIT functionality
+  from a single thread only.
+
+* **Function Lifetime**: JIT-compiled function pointers remain valid only as long
+  as the ``ISPCEngine`` instance exists. Do not use function pointers after the
+  engine is destroyed.
+
+* **Error Handling**: Always check return values and function pointers for null.
```
nurmukhametov, line 1743: Where do all warning and error messages reported by ISPC go?

aneshlya: They will be sent to stderr. Added to the docs.

nurmukhametov (commented):
I tried using the JIT mode for autotuning on [this](https://github.com/nurmukhametov/ispc/tree/jit-only2) branch and came across the following issue:
- ISPC reports that I redefined macros, even though I didn't.

aneshlya:
> * ISPC reports that I redefined macros, even though I didn't.

Thanks for checking! Indeed, there are some fields in `Globals` structure (like `cppArgs`) that accumulated data from previous `ISPCEngine` instances. It's fixed now. Now when ISPC is used as a library (no matter JIT or not) it will create new `Globals`  object for each compilation. When ISPC is used as executable a new Global object is created in `Initialize` function once as it was before.

Also I found that `ISPC_IS_LIBRARY` macros was set incorrectly before, it's also fixed.

nurmukhametov:
> > * ISPC reports that I redefined macros, even though I didn't.
> 
> Thanks for checking! Indeed, there are some fields in `Globals` structure (like `cppArgs`) that accumulated data from previous `ISPCEngine` instances. It's fixed now. Now when ISPC is used as a library (no matter JIT or not) it will create new `Globals` object for each compilation. When ISPC is used as executable a new Global object is created in `Initialize` function once as it was before.
> 
> Also I found that `ISPC_IS_LIBRARY` macros was set incorrectly before, it's also fixed.

I confirm that it works now.

nurmukhametov (approved):
LGTM!

Issue #3513, opened 2025-07-24
Bool representation for targets with i8 mask


nurmukhametov:
ISPC targets with i8 masks (e.g., `sse4.1-i8x16`) produce non-standard bool representations (`0xff` instead of `0x01`) that can cause comparison failures with standard C++ bool values, even when both values are logically equivalent.

**ISPC code (`bool_bug_minimal.ispc`):**
```ispc
export void test_bool_bug(uniform bool *uniform dst) {
    foreach (i = 0...1) {
        dst[i] = true;
    }
}
```

**C++ test (`bool_bug_minimal.cpp`):**
```cpp
#include <stdio.h>
#include <cstdint>
#include "bool_bug_minimal_ispc.h"

int main() {
    bool dst;

    // Call ISPC function that sets dst[0] = true
    ispc::test_bool_bug(&dst);

    // Check the bool representation
    printf("dst = %s, bit representation: 0x%02x\n",
           dst ? "true" : "false", (unsigned char)dst);

    return 0;
}
```

**Compilation and results:**
```bash
# Normal target (works correctly)
$ ispc bool_bug_minimal.ispc -o bool_bug_minimal.o -h bool_bug_minimal_ispc.h --target=avx512icl-x16
$ clang++ -O2 bool_bug_minimal.cpp bool_bug_minimal.o -o bool_bug_minimal
$ ./bool_bug_minimal
dst = true, bit representation: 0x01  # ✅ Correct                                                                                                                                                                                                          
# i8 mask target (shows the bug)
$ ispc bool_bug_minimal.ispc -o bool_bug_minimal_i8.o -h bool_bug_minimal_ispc.h --target=sse4.1-i8x16
$ clang++ -O3 bool_bug_minimal.cpp bool_bug_minimal_i8.o -o bool_bug_minimal_i8
$ ./bool_bug_minimal_i8
dst = true, bit representation: 0xff  # ❌ Bug: should be 0x01
```

In `benchmarks/01_trivial/15_reduce_equal`, the C++-side code is more complex, which results in comparison failures on the line with `            if (dst[i + j] != expected) {`, printing 
```bash
ERROR: ... mismatch at index ?: expected true, got ture
```
It seems to me that `0xff` representation of bool true value quite likely triggers some kind of UB in this case.

zephyr111:
[...]

PR #3518, opened 2025-07-30
Fix select-short-vec.ispc for targets with i8 mask

### LINKED ISSUE #3491: Test select-short-vec is failing accross platforms on neon-i8x16 with -O1
https://github.com/ispc/ispc/actions/runs/16286474415/job/45986562102#step:6:3066
```
Test ./tests/func-tests/select-short-vec.ispc failed (return code 1)            
./tests/func-tests/select-short-vec.ispc.run: value 0 disagrees: returned 2.000000 [0x1p+1], expected 0.000000 [0x0p+0]
./tests/func-tests/select-short-vec.ispc.run: value 2 disagrees: returned 2.000000 [0x1p+1], expected 0.000000 [0x0p+0]
./tests/func-tests/select-short-vec.ispc.run: value 3 disagrees: returned 2.000000 [0x1p+1], expected 0.000000 [0x0p+0]
./tests/func-tests/select-short-vec.ispc.run: value 4 disagrees: returned 2.000000 [0x1p+1], expected 0.000000 [0x0p+0]
./tests/func-tests/select-short-vec.ispc.run: value 6 disagrees: returned 2.000000 [0x1p+1], expected 0.000000 [0x0p+0]
./tests/func-tests/select-short-vec.ispc.run: value 8 disagrees: returned 2.000000 [0x1p+1], expected 0.000000 [0x0p+0]
./tests/func-tests/select-short-vec.ispc.run: value 10 disagrees: returned 2.000000 [0x1p+1], expected 0.000000 [0x0p+0]
./tests/func-tests/select-short-vec.ispc.run: value 12 disagrees: returned 2.000000 [0x1p+1], expected 0.000000 [0x0p+0]
./tests/func-tests/select-short-vec.ispc.run: value 13 disagrees: returned 2.000000 [0x1p+1], expected 0.000000 [0x0p+0]
./tests/func-tests/select-short-vec.ispc.run: value 14 disagrees: returned 2.000000 [0x1p+1], expected 0.000000 [0x0p+0]
```

nurmukhametov:
## Description

This PR fixes #3491 properly. With this fix nightly tests should pass [link](https://github.com/ispc/ispc/actions/runs/16624470651). However in full testing, there are still failed jobs with, e.g, avx1-i32x16 for -O1 for LLVM <= 19 [link](https://github.com/ispc/ispc/actions/runs/16624474802).

This PR changes:

* Gather missed conversion of bool to storage type for targets with i8 mask.

* This also fixes FunctionEmitContext::lSwitchBoolSize_2 to be able to cast bool <-> mask of equal length to each other, i.e., performing either 0x01 -> 0xff or 0xff -> 0x01 conversion.

* maskedStore: convert only bool to bool storage representation. Plain int8 or uint8 do not to be represented or converted like bool types.

### Debugging utilities:
* Added debug wrapper functions (`dv` and `dt`) for dumping LLVM values and types in debug builds, aiding GDB/LLDB debugging.

### Reverted workaround:
* https://github.com/ispc/ispc/commit/d22999b8cc152af7edf685bc86acd748405add2e

## Related Issue
- [X] Linked to relevant issue: #3491 and #3513

## Checklist

### DIFF
--- a/src/ctx.cpp
+++ b/src/ctx.cpp
@@ -2548,6 +2548,8 @@ llvm::Value *FunctionEmitContext::lSwitchBoolSize_2(llvm::Value *value, llvm::Ty
     // To support all cases with less code do bool casting in two stages:
     // 1) truncate to LLVM IR native bool types i1 or <N x i1>
     // 2) zero or sign extend from native bool types to storage or internal correspondingly.
+    // Varying bool in ispc is '-1'. The most significant bit being set to 1
+    // is important for blendv operations to work as expected.
     llvm::Value *i1Bool = value;
     llvm::Twine newName = name.isTriviallyEmpty() ? (llvm::Twine(value->getName()) + "_toi1") : name;
     if (llvm::dyn_cast<llvm::FixedVectorType>(fromType)) {
@@ -2906,28 +2908,8 @@ llvm::Value *FunctionEmitContext::gather(llvm::Value *ptr, const PointerType *pt
 
     // bool type is stored as i8. So, it requires some processing.
     if (returnType->IsBoolType()) {
-        // TODO!: should we just call SwitchBoolToStorageType here unconditionally?
-        if (g->target->getDataLayout()->getTypeSizeInBits(returnType->LLVMStorageType(g->ctx)) <
-            g->target->getDataLayout()->getTypeSizeInBits(llvmReturnType)) {
-            // This is needed when array of bool is passed in from cpp side
-            // TRUE in clang is '1'. This is zero extended to i8.
-            // In ispc, this is uniform * varying which after gather becomes
-            // varying bool. Varying bool in ispc is '-1'. The most
-            // significant bit being set to 1 is important for blendv
-            // operations to work as expected.
-            if (ptrType->GetBaseType()->IsUniformType()) {
-                gatherCall = TruncInst(gatherCall, LLVMTypes::Int1VectorType);
-                gatherCall = SExtInst(gatherCall, llvmReturnType);
-            } else {
-                gatherCall = SExtInst(gatherCall, llvmReturnType);
-            }
-        } else if (g->target->getDataLayout()->getTypeSizeInBits(returnType->LLVMStorageType(g->ctx)) >
-                   g->target->getDataLayout()->getTypeSizeInBits(llvmReturnType)) {
-            gatherCall = TruncInst(gatherCall, llvmReturnType);
-        } else {
-            gatherCall =
-                SwitchBoolToMaskType(gatherCall, llvmReturnType, llvm::Twine(gatherCall->getName()) + "_bool_to_mask");
-        }
+        gatherCall =
+            SwitchBoolToMaskType(gatherCall, llvmReturnType, llvm::Twine(gatherCall->getName()) + "_switch_bool");
     }
     return gatherCall;
 }
--- a/src/ctx.cpp
+++ b/src/ctx.cpp
@@ -3178,7 +3178,10 @@ void FunctionEmitContext::maskedStore(llvm::Value *value, llvm::Value *ptr, cons
         maskedStoreFunc = m->module->getFunction(builtin::__pseudo_masked_store_i16);
     } else if (llvmValueStorageType == LLVMTypes::Int8VectorType) {
         maskedStoreFunc = m->module->getFunction(builtin::__pseudo_masked_store_i8);
-        value = SwitchBoolToStorageType(value, llvmValueStorageType);
+        // Do not mix i8 types with bool type, only bool type requires the special handling.
+        if (valueType->IsBoolType()) {
+            value = SwitchBoolToStorageType(value, llvmValueStorageType);
+        }
     }
     AssertPos(currentPos, maskedStoreFunc != nullptr);
 
--- a/src/ctx.cpp
+++ b/src/ctx.cpp
@@ -2525,8 +2525,11 @@ llvm::Value *FunctionEmitContext::lSwitchBoolSize_2(llvm::Value *value, llvm::Ty
     llvm::Type *fromType = value->getType();
 
     if (fromType == toType) {
-        // No cast is needed
-        return value;
+        // if mask size and bool size is equal then cast is needed (for the proper sign extension)
+        if (g->target->getMaskBitCount() != 8) {
+            // No cast is needed
+            return value;
+        }
     }
 
     // Internal representation of bool matches mask behaviour that requires
@@ -2903,6 +2906,7 @@ llvm::Value *FunctionEmitContext::gather(llvm::Value *ptr, const PointerType *pt
 
     // bool type is stored as i8. So, it requires some processing.
     if (returnType->IsBoolType()) {
+        // TODO!: should we just call SwitchBoolToStorageType here unconditionally?
         if (g->target->getDataLayout()->getTypeSizeInBits(returnType->LLVMStorageType(g->ctx)) <
             g->target->getDataLayout()->getTypeSizeInBits(llvmReturnType)) {
             // This is needed when array of bool is passed in from cpp side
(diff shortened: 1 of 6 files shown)


src/ctx.cpp
```diff
@@ -2903,6 +2906,7 @@ llvm::Value *FunctionEmitContext::gather(llvm::Value *ptr, const PointerType *pt
 
     // bool type is stored as i8. So, it requires some processing.
     if (returnType->IsBoolType()) {
+        // TODO!: should we just call SwitchBoolToStorageType here unconditionally?
```
aneshlya: Yeah, I think this logic should be consolidated in `SwitchBoolToMaskType`. If you can spend a little more time on it, it would be good to 
1. resolve this TODO
2. describe all the bool/mask conversion behavior in design document

src/ctx.cpp
```diff
@@ -3178,7 +3178,10 @@ void FunctionEmitContext::maskedStore(llvm::Value *value, llvm::Value *ptr, cons
         maskedStoreFunc = m->module->getFunction(builtin::__pseudo_masked_store_i16);
     } else if (llvmValueStorageType == LLVMTypes::Int8VectorType) {
         maskedStoreFunc = m->module->getFunction(builtin::__pseudo_masked_store_i8);
-        value = SwitchBoolToStorageType(value, llvmValueStorageType);
+        // Do not mix i8 types with bool type, only bool type requires the special handling.
+        if (valueType->IsBoolType()) {
```
aneshlya, line 3164: Consider adding a `valueType` parameter to `SwitchBoolToStorageType`. This would allow us to perform an early return if the function receives a type other than bool, which could help prevent potential issues with future usage

tests/lit-tests/3513.ispc
```diff
@@ -0,0 +1,41 @@
+// This test checks whether ISPC produces correct representation of bool type
+// on ABI boundary for targets with i8 masks.
+
+// RUN: %{ispc} --pic --target=sse4.1-i8x16 -h %t.h %s -o %t.o
```
zephyr111: I think it would be good to test more targets. For example one that is not i8, and maybe an AVX-512 like avx512skl-x16 and/or avx512skl-x64 (I wonder what is the value of `g->target->getMaskBitCount()` in this case).

stdlib/include/short_vec.isph
```diff
@@ -166,7 +166,7 @@ template <typename T, uint N> inline uniform T<N> select(uniform bool cond, unif
     uniform T<N> result;
 
     foreach (i = 0 ... N) {
-        result[i] = select((UIntMaskType)cond, t[i], f[i]);
+        result[i] = select(cond, t[i], f[i]);
```
zephyr111, line 169: This is great!

src/ctx.cpp
```diff
@@ -2920,6 +2924,9 @@ llvm::Value *FunctionEmitContext::gather(llvm::Value *ptr, const PointerType *pt
         } else if (g->target->getDataLayout()->getTypeSizeInBits(returnType->LLVMStorageType(g->ctx)) >
                    g->target->getDataLayout()->getTypeSizeInBits(llvmReturnType)) {
             gatherCall = TruncInst(gatherCall, llvmReturnType);
+        } else {
+            gatherCall =
```
zephyr111: I think it should be good to have a lit-test with gathers since they are impacted here.

By the way, is there a scatter operation somewhere in ISPC? If yes, it might be impacted too.

src/ctx.cpp
zephyr111: I am not sure to fully understand the logic of the whole code, but won't this cause an issue if we call a function and use the result as a mask (like with `select(test(), a, b)`) ? Maybe it is safe because the bool (with a "bool storage type") will then be converted back to a mask?

I think it would be good to add a lit-test with many booleans operations (and function call returning boolean), just to be on the safe side. Ideally, something tested with both i8 and non-i8 targets.

zephyr111 (changes requested):
Thank you. This is great. 
I added some (certainly naive) remarks and questions. 
Overall, I think more lit-tests are needed.

By the way, does this PR impacts the behaviour of casts from/to boolean?

tests/lit-tests/3513.ispc
nurmukhametov: Indeed, it makes sense. To test more targets, I either need to create a lit configuration command (to avoid a lot of copy-paste), or somehow test in func-tests.

For all avx512 targets, the mask bit count is 1.

src/ctx.cpp
nurmukhametov: gathers and scatters are generated in this test. You can inspect the initial LLVM IR generated for the user code in `pre*GenerateIR*.ll` file that can be obtained via this command inside the directory `dbg`:
```bash
$ ispc --target=host input.ispc --no-discard-value-names --debug-phase=pre:last --dump-file=dbg
```

Actually, all three fixes from this PR are required to fix this test.

src/ctx.cpp
nurmukhametov: I have added test like that in `tests/func-tests/select-short-vec-func.ispc`. Apparently, it triggers a crash in macos linker e.g., [here](https://github.com/ispc/ispc/actions/runs/16750557498/job/47421333173) :
```bash
Compilation of test ./tests/func-tests/select-short-vec-func.ispc failed            
0  0x103360c3b  __assert_rtn + 64
1  0x10332c3c6  ld::AtomFileConsolidator::addAtomFile(ld::AtomFile const*, ld::AtomFile const*, bool) + 9446
2  0x10333a198  ld::AtomFileConsolidator::addAtomFile(ld::AtomFile const*) + 136
3  0x10335efb9  ld::pass::stubs(ld::Options const&, ld::AtomFileConsolidator&) + 1801
4  0x103343d94  ld::AtomFileConsolidator::resolve() + 19828
5  0x1032ac5e8  main + 12552
ld: Assertion failed: (slot < _sideTableBuffer.size()), function addAtom, file AtomFileConsolidator.cpp, line 2278.
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

tests/lit-tests/3513.ispc
nurmukhametov: I tested more targets but not all, see the comment in `tests/lit-tests/3513.ispc` under TODO.

src/ctx.cpp
nurmukhametov: I didn't find in some cases the proper `valueType` in the caller context.

src/ctx.cpp
nurmukhametov: TODO is resolved here: https://github.com/ispc/ispc/pull/3518/commits/39395099e7f204bd4dabbadb24e364549789fbe4

tests/lit-tests/3513.ispc
```diff
@@ -0,0 +1,117 @@
...
+// RUN: %{cc} -O2 -x c -c %s -o %t-sse4.1-i8x16.c.o --include %t-sse4.1-i8x16.h
+// RUN: %{cc} %t-sse4.1-i8x16.o %t-sse4.1-i8x16.c.o -o %t-sse4.1-i8x16.c.bin
+// RUN: %t-sse4.1-i8x16.c.bin | FileCheck %s
+// RUN: %{cc} -O2 -x c++ -c %s -o %t-sse4.1-i8x16.cpp.o --include %t-sse4.1-i8x16.h
+// RUN: %{cc} %t-sse4.1-i8x16.o %t-sse4.1-i8x16.cpp.o -o %t-sse4.1-i8x16.cpp.bin
+// RUN: %t-sse4.1-i8x16.cpp.bin | FileCheck %s
+
+// TODO!: ideally we need to introduce AVX2_HOST and similar features to
+// reliably test tests like this but for now assume that AVX2 is always
+// available practically on any x86_64 hosts
```
aneshlya, line 86: I hope this is true for all package managers builds of ISPC :)

aneshlya (approved):
LGTM!

PR #3519, opened 2025-07-31
WIP: opt: enable AggressiveInstCombine

nurmukhametov:
## Description

Enable AggressiveInstCombine in the ISPC optimization pipeline.

## Related Issue

## Checklist

### DIFF
--- a/src/opt.cpp
+++ b/src/opt.cpp
@@ -70,6 +70,7 @@
 #if ISPC_LLVM_VERSION >= ISPC_LLVM_18_1
 #include <llvm/Transforms/Scalar/InferAlignment.h>
 #endif
+#include <llvm/Transforms/AggressiveInstCombine/AggressiveInstCombine.h>
 #include <llvm/Transforms/Scalar/InstSimplifyPass.h>
 #include <llvm/Transforms/Scalar/JumpThreading.h>
 #include <llvm/Transforms/Scalar/LICM.h>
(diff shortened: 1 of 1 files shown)


zephyr111:
I am just curious about this. Do you have a practical example where this is useful?
I found [this discussion](https://reviews.llvm.org/D38313) but it is still quite nebulous to me.

By the way, it would be interesting to measure the execution time of compilation with/without it once just to be sure it stay reasonable (e.g. on script of `ispc/examples/cpu` directory).

nurmukhametov:
> I am just curious about this. Do you have a practical example where this is useful? 

I came across this while creating more lit tests for PR #3518. The compiler explorer [link](https://ispc.godbolt.org/z/rd94E4xv7). I can't say that this particular example has a big value because the generated ASM is the same but the IR is simpler.

> I found [this discussion](https://reviews.llvm.org/D38313) but it is still quite nebulous to me.

It can optimize extra patterns, but I am not sure if it has much value for us. I am going to measure the performance of the generated code as this can reveal some practical examples.

> By the way, it would be interesting to measure the execution time of compilation with/without it once just to be sure it stay reasonable (e.g. on script of `ispc/examples/cpu` directory).

Yes, sure. I will measure how the compile time has changed. Note that this pass is enabled in the default LLVM -O2 pipeline ([link](https://github.com/llvm/llvm-project/blob/7d815c7642ef202741ea27c3aa6a3ffd4113ecf7/llvm/lib/Passes/PassBuilderPipelines.cpp#L1991)) since LLVM 17, so I think this pass is generally OK.

nurmukhametov:
> > I found [this discussion](https://reviews.llvm.org/D38313) but it is still quite nebulous to me.
> 
> It can optimize extra patterns, but I am not sure if it has much value for us. I am going to measure the performance of the generated code as this can reveal some practical examples.

I measured performance on GameDev benchmarks. They didn't show any performance gains.

PR #3522, opened 2025-08-01
Fix couple of issues found in internal testing

aneshlya:
## Description
Brief description of changes

## Related Issue

## Checklist

### DIFF
--- a/benchmarks/04_intrinsics/02_sse2_intrinsics.cpp
+++ b/benchmarks/04_intrinsics/02_sse2_intrinsics.cpp
@@ -3843,8 +3843,8 @@ class mm_cvtepi32_ps : public TestBase {
             break;
 #ifdef IS_X86_ARCH
         case 1:
-            for (int k = 0; k < Iterations; k += 2) {
-                const __m128i S1 = _mm_load_si128((__m128i *)&Source1[k]);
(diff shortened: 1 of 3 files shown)


benchmarks/04_intrinsics/02_sse2_intrinsics.cpp
```diff
@@ -3843,7 +3843,7 @@ class mm_cvtepi32_ps : public TestBase {
             break;
 #ifdef IS_X86_ARCH
         case 1:
-            for (int k = 0; k < Iterations; k += 2) {
+            for (int k = 0; k < Iterations; k += 4) {
                 const __m128i S1 = _mm_load_si128((__m128i *)&Source1[k]);
```
zephyr111: It should be `_mm_loadu_si128` here since AFAIK `Source1` is not guaranteed to be aligned. It should not impact performance on all recent CPUs (<6 years-old and possibly more). In practice, the default `new` tends to allocate data with an alignment of 16-byte so it should be fine on all machine but AFAIK this is not guaranteed (only an alignment of `sizeof(int)` is). And even if it would, then `_mm_storeu_ps` should have been `_mm_store_ps` instead.

aneshlya: Agree.

Issue #3504, opened 2025-07-21
Linker errors for float16 conversion functions with MSVC


jspohr:
When targeting non-FP16 capable ISAs, and trying to link with the MSVC linker, I get the following linker errors:

> error LNK2019: unresolved external symbol __truncsfhf2
> error LNK2019: unresolved external symbol __extendhfsf2

I found these as part of LLVMs compiler-rt. I'm aware that fp16 support isn't finished on all platforms/targets (#2290).
But maybe there's a workaround for this particular scenario? I tried to implement the two functions, but there's obviously an ABI mismatch, because I see that the values that are returned don't end up being correct on the ISPC side. Should this call use the Windows x64 ABI or is there something special about it?
Am I just missing a dependency? I tried to link against ispcrt_static.lib but that gave me a different linker error.

If fp16 is just not ready to use with MSVC yet, that would be fair enough as an answer as well.

Thanks!

aneshlya:
Fp16 is not ready to use with MSVC yet on x86.

jspohr:
Thanks for the reply!

nurmukhametov:
> When targeting non-FP16 capable ISAs, and trying to link with the MSVC linker, I get the following linker errors:
> 
> > error LNK2019: unresolved external symbol __truncsfhf2
> > error LNK2019: unresolved external symbol __extendhfsf2
> 
> I found these as part of LLVMs compiler-rt. I'm aware that fp16 support isn't finished on all platforms/targets ([#2290](https://github.com/ispc/ispc/issues/2290)). But maybe there's a workaround for this particular scenario?

@jspohr, after the pull request https://github.com/ispc/ispc/pull/3524, you can use --include-float16-conversions, which provides these functions from the ISPC standard library for the compiled user module.

PR #3524, opened 2025-08-05
Implement --include-float16-conversions

nurmukhametov:
## Description

A new command-line option ``--include-float16-conversions`` was added. This option makes the compiler include float16 conversion functions in the compiled module. This is useful for targets that do not have native instructions for float16 conversions, which are x86 targets prior to AVX2. This option is disabled by default.

For targets without native instructions for half-precision conversions, LLVM generates calls to these functions. They are supposed to be provided by the compiler runtimes: libgcc for GCC and Clang, and compiler-rt for LLVM. libgcc provides them starting from GCC 12 version, LLVM provides them with float16 types starting from LLVM 15. MSVC does not provide these functions. LLVM backend generates calls to these functions during ISEL, so we need to preserve them until it by adding them to llvm.used if --include-float16-conversions is provided. To remove them if they are not used we need to modify machine optimization pipeline, that is not implemented yet.

## Related Issue
- [X] Linked to relevant issue: #3504 

## Checklist
- [X] Language/stdlib changes include new [functional tests](https://github.com/ispc/ispc/tree/main/tests/func-tests) for runtime behavior
- [X] [Documentation](https://github.com/ispc/ispc/tree/main/docs/ispc.rst) updated if needed

### DIFF
--- a/tests/func-tests/round-float16-uniform.ispc
+++ b/tests/func-tests/round-float16-uniform.ispc
@@ -1,4 +1,5 @@
 #include "test_static.isph"
+// rule: skip on target=generic.*
 task void f_f(uniform float RET[], uniform float aFOO[]) {
     uniform float error = 0;
     RET[programIndex] = 0;
--- a/tests/func-tests/round-float16-varying.ispc
+++ b/tests/func-tests/round-float16-varying.ispc
@@ -1,6 +1,8 @@
 #include "test_static.isph"
 // rule: skip on arch=wasm32
 // rule: skip on arch=wasm64
+// rule: skip on target=generic.*
+
 task void f_f(uniform float RET[], uniform float aFOO[]) {
     float error = 0;
     RET[programIndex] = 0;


src/builtins.cpp
```diff
@@ -286,6 +286,38 @@ void lCreateLLVMUsed(llvm::Module &M, std::vector<llvm::Constant *> &ConstPtrs)
...
+            ConstPtrs.push_back(llvm::ConstantExpr::getBitCast(F, type));
+        }
+    }
+
+    if (ConstPtrs.empty()) {
+        return;
+    }
+
+    // Create the array of i8* that llvm.used will hold
+    llvm::ArrayType *ATy = llvm::ArrayType::get(type, ConstPtrs.size());
```
aneshlya: Most of the code in this function duplicates `lCreateLLVMUsed`. Could you extract common part to separate function?

src/builtins.cpp
```diff
@@ -286,6 +286,38 @@ void lCreateLLVMUsed(llvm::Module &M, std::vector<llvm::Constant *> &ConstPtrs)
...
+    llvm::Constant *ArrayInit = llvm::ConstantArray::get(ATy, ConstPtrs);
+
+    // Create llvm.used and initialize it with the float16 functions
+    llvm::GlobalVariable *llvmUsed = new llvm::GlobalVariable(
+        M, ArrayInit->getType(), false, llvm::GlobalValue::AppendingLinkage, ArrayInit, "llvm.used");
```
aneshlya: Could you check that global variable "llvm.used" wasn't created yet (just for safe future usage)?
```
llvm::GlobalVariable *existingUsed = M.getGlobalVariable("llvm.used");
if (existingUsed) { ...
```

src/builtins.cpp
aneshlya: Also `lCreateLLVMUsed` has comments using `llvm.used` instead of `llvm.compiler.used` which is misleading.

src/builtins.cpp
KseniyaTikhomirova: @aneshlya  I am curious. lCreateLLVMUsed seems to imply unconditional creation & initialization of variable and don't have any check of variable existence. Why lCreateLLVMUsedForFloat16 is different? Do we expect this variable creation somewhere else?

src/builtins.cpp
```diff
@@ -286,6 +286,38 @@ void lCreateLLVMUsed(llvm::Module &M, std::vector<llvm::Constant *> &ConstPtrs)
...
+    llvmUsed->setSection("llvm.metadata");
```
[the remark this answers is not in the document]
KseniyaTikhomirova: I like this suggestion, variable name probably can be passed as parameter with default value.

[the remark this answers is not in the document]
nurmukhametov: Right, that makes sense. I have reused common functionality.

src/builtins.cpp
nurmukhametov: Yes, it is a duplicate. I have fixed it.

src/builtins.cpp
nurmukhametov: They should not be any different. I have added this check in both cases.

src/builtins.cpp
```diff
@@ -272,24 +272,51 @@ void lCollectPreservedFunctions(llvm::Module *M, std::vector<llvm::Constant *> &
     }
 }
 
-void lCreateLLVMUsed(llvm::Module &M, std::vector<llvm::Constant *> &ConstPtrs) {
+void lCreateLLVMUsed(llvm::Module &M, std::vector<llvm::Constant *> &ConstPtrs, const std::string &VarName) {
```
KseniyaTikhomirova, line 275: Please check if it makes sense to change the name then

Issue #3540, opened 2025-08-14
Target::checkIntrinsticSupport doesn't consider amx intrinsics


nurmukhametov:
`Target::checkIntrinsticSupport` in `src/ispc.cpp` doesn't consider AMX intrinsics as supported, even for CPUs that do support them (e.g., spr, gnr).

PR #3542, opened 2025-08-15
Amx features

nurmukhametov:
## Description

Implement CPU feature validation for AMX intrinsics to prevent usage on
unsupported processors with clear error messages and allow usage on
supported processors.

Changes:
- Add X86IntrinsicToFeature mapping for special intrinsic validation
- Update CPUFeatures with AMX support by CPU generation:
  * SPR: amxtile, amxint8, amxbf16
  * GNR: adds amxfp16
  * DMR: adds amxfp8
- Enhance Target::checkIntrinsticSupport() for composite feature requirements

This PR also adds the missied DMR entry to CPUFeatures. This allows to actually support x86 specific intrinsics (sse, avx and etc.)

## Related Issue
- [X] Linked to relevant issue: #3540

## Checklist


src/ispc.cpp
```diff
@@ -424,8 +452,17 @@ std::map<DeviceType, std::set<std::string>> CPUFeatures = {
...
+    {CPU_SPR,
+     {"mmx", "sse", "sse2", "ssse3", "sse41", "sse42", "avx", "avx2", "avx512", "avx_vnni", "avx512_vnni", "amxtile",
+      "amxint8", "amxbf16"}},
+    {CPU_GNR,
+     {"mmx", "sse", "sse2", "ssse3", "sse41", "sse42", "avx", "avx2", "avx512", "avx_vnni", "avx512_vnni", "amxtile",
+      "amxint8", "amxbf16", "amxfp16"}},
+#if ISPC_LLVM_VERSION >= ISPC_LLVM_20_0
+    {CPU_DMR,
+     {"mmx", "sse", "sse2", "ssse3", "sse41", "sse42", "avx", "avx2", "avx512", "avx_vnni", "avx512_vnni", "amxtile",
+      "amxint8", "amxbf16", "amxfp16", "amxfp8"}},
```
aneshlya: There are more AMX features on DMR: `FeatureAMX_MOVRS | FeatureAMX_AVX512 | FeatureAMX_FP8 | FeatureAMX_TF32 | FeatureAMX_TRANSPOSE | FeatureAMX_COMPLEX`.
Why only `amxfp8` was used?

tests/lit-tests/intrinsic_target_specific_llvm20.ispc
```diff
@@ -0,0 +1,14 @@
+//; RUN: %{ispc} %s --target=avx2-i32x8 --enable-llvm-intrinsics
+//; RUN: %{ispc} %s --cpu=ps4 --target=avx1-i32x8 --enable-llvm-intrinsics
+//; RUN: %{ispc} %s --nowrap --cpu=spr --target=avx512spr-x8 --enable-llvm-intrinsics
+//; RUN: %{ispc} %s --nowrap --cpu=gnr --target=avx512spr-x8 --enable-llvm-intrinsics
+//; RUN: %{ispc} %s --nowrap --cpu=dmr --target=avx512spr-x8 --enable-llvm-intrinsics
```
aneshlya: It's not a big deal but I would use here and in other tests `--target=avx10.2-x8` for `dmr` cpu.

src/ispc.cpp
aneshlya: As we discussed it would make sense to keep only features that can be represented in ISPC type system so fp8, bf16 can be removed.

YixingZhang007:
LGTM! Just a small suggestion, but not a big deal. We might consider changing `CPUFeatures` from a `map` to an `unordered_map` in `ispc.cpp`, since the keys aren’t arranged in order.

YixingZhang007 (approved):

src/ispc.cpp
nurmukhametov: I have removed fp8 and bf16.

tests/lit-tests/intrinsic_target_specific_llvm20.ispc
```diff
@@ -0,0 +1,14 @@
+//; RUN: %{ispc} %s --target=avx2-i32x8 --enable-llvm-intrinsics
+//; RUN: %{ispc} %s --cpu=ps4 --target=avx1-i32x8 --enable-llvm-intrinsics
+//; RUN: %{ispc} %s --nowrap --cpu=spr --target=avx512spr-x8 --enable-llvm-intrinsics
+//; RUN: %{ispc} %s --nowrap --cpu=gnr --target=avx512spr-x8 --enable-llvm-intrinsics
+//; RUN: %{ispc} %s --nowrap --cpu=dmr --target=avx10.2-x8 --enable-llvm-intrinsics
+//; RUN: not %{ispc} %s --target=sse2-i32x8 --enable-llvm-intrinsics --nowrap 2>&1 | FileCheck %s -check-prefix=CHECK_SSE2
+
+//; REQUIRES: X86_ENABLED && LLVM_20_0+ && !MACOS_HOST
```
KseniyaTikhomirova, line 8: Could you please clarify why we have LLVM 20+ requirement here for all platforms? from the code I see that  only DMR has any difference.

KseniyaTikhomirova: same comment for the test below. does it make sense to disable these tests for < LLVM 20.0 because of 1 platform?

KseniyaTikhomirova (approved):
LGTM, just a question about tests

tests/lit-tests/intrinsic_target_specific_llvm20.ispc
nurmukhametov: This test is not disabled. It is tested with LLVM < 20 in a different file named `intrinsic_target_specific.ispc`. We need LLVM 20+ to test ISPC with DMR targets because ISPC supports DMR only when it is built with LLVM 20+.

Issue #3545, opened 2025-08-16
Openmoonray fails to compile with version 1.28.0


gonsolo:
Trying to compile [Openmoonray](https://github.com/dreamworksanimation/openmoonray)
with the patch from https://github.com/dreamworksanimation/openmoonray/discussions/192, I have this error:

> /usr/src/debug/ispc/ispc/src/ast.cpp:367: Assertion failed: "node->IsTypeChecked() && "Node must be type-checked before optimization"".
> ***
> *** Please file a bug report at https://github.com/ispc/ispc/issues
> *** (Including as much information as you can about how to reproduce this error).
> *** You have apparently encountered a bug in the compiler that we'd like to fix!
> ***
> /usr/src/debug/ispc/ispc/src/main.cpp(40): FATAL ERROR: Unhandled signal sent to process; terminating.
> /usr/src/debug/ispc/ispc/src/ast.cpp:367: Assertion failed: "node->IsTypeChecked() && "Node must be type-checked before optimization"".
> ***
> *** Please file a bug report at https://github.com/ispc/ispc/issues
> *** (Including as much information as you can about how to reproduce this error).
> *** You have apparently encountered a bug in the compiler that we'd like to fix!
> ***
> /usr/src/debug/ispc/ispc/src/main.cpp(40): FATAL ERROR: Unhandled signal sent to process; terminating.

which I can reproduce with

[...]

PR #3546, opened 2025-08-18
CI: add job to build openmoonray

nurmukhametov:
## Description

This PR adds the openmoonray build job to CI. It can help to prevent regression like this #3545 in future.

At the moment, this job is failing (https://github.com/ispc/ispc/actions/runs/17044829107?pr=3546) and this is expected.

## Related Issue
- [X] Linked to relevant issue: #3545

## Checklist

### DIFF
--- a/.github/workflows/openmoonray.yml
+++ b/.github/workflows/openmoonray.yml
@@ -57,8 +57,6 @@ jobs:
       run: |
         pacman -Syu --noconfirm
         pacman -S --needed --noconfirm base-devel git sudo
-        # AUR PKBUILD misses these dependencies for OpenMoonRay
-        pacman -S --noconfirm libmicrohttpd cppunit qt5-base qt5-script openimagedenoise usd
 
     - name: Create build user
       run: |


YixingZhang007:
LGTM! Thanks!

.github/workflows/openmoonray.yml
```diff
@@ -0,0 +1,93 @@
+# Copyright 2025, Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause
+
+name: Build OpenMoonRay
+
+permissions: read-all
+
+on:
+  pull_request:
```
KseniyaTikhomirova, line 9: this job will be triggered for any changes in repo, even in 'docs'. I think it is better to exclude such folders

KseniyaTikhomirova (approved):
LGTM

.github/workflows/openmoonray.yml
nurmukhametov: Yes, sure. I have ignored some paths in the repo.

KseniyaTikhomirova: May I ask you why do you update PR with amending initial commit? If you push more commits it will be squashed for merge anyway but can significantly simplify work for reviewers since they will be able to see what exactly is changed since last review.

nurmukhametov: I don't think we have a policy to enforce squashing, so it is up to the person who merges to decide whether to squash or not. Squashing can indeed simplify the review process, but GitHub also has the compare button alongside the message 'someone force-pushed the branch from X to Y,' which likely shows the same information as the new commit above the feature branch.

KseniyaTikhomirova (approved):
LGTM

Issue #3549, opened 2025-08-19
Assertion failed with IndexExpr for FunctionCallExpr returning pointer type


nurmukhametov:
The following code 

```ispc
struct S;
inline varying float *uniform func(const varying struct S &t) {
    return (varying float *uniform) & t;
};
void foo(const varying struct S &x, uniform int c) {
    (func(x))[c];
}
```

fails with:

```bash
/root/ispc/src/expr.cpp:4873: Assertion failed: "ptrType != nullptr".
***
*** Please file a bug report at https://github.com/ispc/ispc/issues
*** (Including as much information as you can about how to reproduce this error).
*** You have apparently encountered a bug in the compiler that we'd like to fix!
***
/root/ispc/src/main.cpp(40): FATAL ERROR: Unhandled signal sent to process; terminating.
Program terminated with signal: SIGSEGV
Compiler returned: 139
```

Compiler explorer link: https://ispc.godbolt.org/z/7jT8Ge995

The assert happens after https://github.com/ispc/ispc/commit/536c26d7ba3d607ebaca427a69231dd1af76d796

PR #3550, opened 2025-08-19
Fix IndexExpr::GetLValueType() for function calls returning pointers

nurmukhametov:
## Description

Fixes assertion failure in lAddVaryingOffsetsIfNeeded() when compiling expressions like `(func(x))[c]` where func returns a pointer type.

When indexing into the result of a function call that returns a pointer (e.g., `func()[index]`), IndexExpr::GetLValueType() would incorrectly require baseExpr->GetLValueType() to be non-null. However, function calls returning pointers are rvalues, so FunctionCallExpr::GetLValueType() correctly returns null for non-reference return types.

The fix allows IndexExpr to fall back to using the pointer type directly when GetLValueType() returns null but GetType() returns a pointer type. This enables indexing operations on function call results that return pointers, which should create valid lvalues.

## Related Issue
- [X] Linked to relevant issue: #3549 

## Checklist

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -5223,10 +5223,21 @@ const Type *IndexExpr::GetLValueType() const {
         return lvalueType;
     }
 
-    const Type *baseExprType = nullptr, *baseExprLValueType = nullptr, *indexType = nullptr;
+    const Type *baseExprType = nullptr, *indexType = nullptr;
     if (baseExpr == nullptr || index == nullptr || ((baseExprType = baseExpr->GetType()) == nullptr) ||
-        ((baseExprLValueType = baseExpr->GetLValueType()) == nullptr) || ((indexType = index->GetType()) == nullptr)) {
+        ((indexType = index->GetType()) == nullptr)) {
         return nullptr;
+    }
+
+    // If baseExpr->GetLValueType() is null but baseExpr->GetType() is a pointer,
+    // use the pointer type directly (e.g., function calls returning pointers)
+    const Type *baseExprLValueType = baseExpr->GetLValueType();
+    if (baseExprLValueType == nullptr) {
+        if (baseExprType->IsPointerType()) {
+            baseExprLValueType = baseExprType;
+        } else {
+            return nullptr;
+        }
     }
 
     // regularize to a PointerType
--- a/tests/lit-tests/3549.ispc
+++ b/tests/lit-tests/3549.ispc
@@ -1,0 +1,26 @@
+// This test checks that the ISPC compiler does not crash when compiling
+// IndexExpr with FunctionCallExpr returning pointer type.
+
+// RUN: %{ispc} --target=host --emit-llvm-text --nostdlib --nowrap %s -o - 2>&1 | FileCheck %s
+// RUN: %{ispc} --target=avx512icl-x16 -O2 --emit-asm --x86-asm-syntax=intel --nostdlib --nowrap %s -o - 2>&1 | FileCheck %s --check-prefix=CHECK-AVX512
+
+// REQUIRES: X86_ENABLED
+
+// CHECK-NOT: .*expr.cpp:.*: Assertion failed: "ptrType != nullptr".
+
+// CHECK-AVX512-LABEL: foo___{{.*}}:
+// CHECK-AVX512-NEXT: # %bb.0:
+// CHECK-AVX512-NEXT:  movsxd  [[REG:r[a-z]+]], {{.*}}
+// CHECK-AVX512-NEXT:  shl     [[REG]], 6
+// CHECK-AVX512-NEXT:  vmovups zmm0, zmmword ptr [r{{.*}} + [[REG]]]
+// CHECK-AVX512-NEXT:  ret
+
+struct S {
+    float b[100];
+};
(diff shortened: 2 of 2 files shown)


src/expr.cpp
```diff
@@ -5225,7 +5225,19 @@ const Type *IndexExpr::GetLValueType() const {
 
     const Type *baseExprType = nullptr, *baseExprLValueType = nullptr, *indexType = nullptr;
     if (baseExpr == nullptr || index == nullptr || ((baseExprType = baseExpr->GetType()) == nullptr) ||
-        ((baseExprLValueType = baseExpr->GetLValueType()) == nullptr) || ((indexType = index->GetType()) == nullptr)) {
+        ((indexType = index->GetType()) == nullptr)) {
+        return nullptr;
+    }
+
+    baseExprLValueType = baseExpr->GetLValueType();
```
YixingZhang007: We might consider declaring `baseExprLValueType` here instead of earlier in the code. For example, we can change to this `const Type *baseExprLValueType = baseExpr->GetLValueType();`.

YixingZhang007 (approved):
LGTM. Just a small suggestion. Thanks!

tests/lit-tests/3549.ispc
```diff
@@ -0,0 +1,26 @@
...
+// IndexExpr with FunctionCallExpr returning pointer type.
+
+// RUN: %{ispc} --target=host --emit-llvm-text --nostdlib --nowrap %s -o - 2>&1 | FileCheck %s
+// RUN: %{ispc} --target=avx512icl-x16 -O2 --emit-asm --x86-asm-syntax=intel --nostdlib --nowrap %s -o - 2>&1 | FileCheck %s --check-prefix=CHECK-AVX512
+
+// REQUIRES: X86_ENABLED
+
+// CHECK-NOT: .*expr.cpp:.*: Assertion failed: "ptrType != nullptr".
+
+// CHECK-AVX512-LABEL: foo___{{.*}}:
```
zephyr111, line 11: Nitpicking: this part of the test seems a bit too strict to me.
Wouldn't this be better to analyse the (optimized) LLVM IR code assuming it is small enough? I am a bit worried about this test to fail due to future LLVM updates or future new optimisation passes (or modification of the existing ones).

src/expr.cpp
```diff
@@ -5223,9 +5223,21 @@ const Type *IndexExpr::GetLValueType() const {
...
-        ((baseExprLValueType = baseExpr->GetLValueType()) == nullptr) || ((indexType = index->GetType()) == nullptr)) {
+        ((indexType = index->GetType()) == nullptr)) {
+        return nullptr;
+    }
+
+    const Type *baseExprLValueType = baseExpr->GetLValueType();
+
+    // If baseExpr->GetLValueType() is null but baseExpr->GetType() is a pointer,
+    // use the pointer type directly (e.g., function calls returning pointers)
+    if (baseExprLValueType == nullptr && baseExprType->IsPointerType()) {
```
zephyr111: Nitpicking: the following code seems a bit more clear IMHO (than the two conditions):
```c++
if (baseExprLValueType == nullptr) {
    if (baseExprType->IsPointerType()) {
        baseExprLValueType = baseExprType;
    } else {
        return nullptr;
    }
}
```
It may be also a bit faster.

zephyr111 (approved):
LGTM (though I do not master this part of the code yet). I just added 2 minor comments.

tests/lit-tests/3549.ispc
nurmukhametov: My opinion actually is quite the opposite. The x86 ISA seems to me more stable than LLVM IR. During LLVM updates, we quite often had to change LLVM IR in tests. The reasons for these changes were mostly not the optimization of code but the evolution of IR itself. For example, here we have this:
```llvm-ir
define <8 x float> @foo___REFs_5B__c_vyS_5D_uni(ptr noalias readonly %x, i32 %c, <8 x i32> %__mask) local_unnamed_addr #1 !dbg !59 {
allocas:
  %c_load_to_int64 = sext i32 %c to i64, !dbg !73
  %calltmp_offset = getelementptr <8 x float>, ptr %x, i64 %c_load_to_int64, !dbg !75
  %calltmp_offset_load = load <8 x float>, ptr %calltmp_offset, align 1, !dbg !75
  ret <8 x float> %calltmp_offset_load, !dbg !75
}
```
We have `getelementptr` here that will eventually be changed to `ptradd` (links: [one](https://discourse.llvm.org/t/rfc-replacing-getelementptr-with-ptradd/68699), [two](https://www.npopov.com/2025/01/05/This-year-in-LLVM-2024.html)), requiring us to change the test. Moreover, we support several LLVM versions at the same time, so we will need to have something here that checks for both `getelementptr` and `ptradd`.

At the same time, if this assembly code changes in the future (for better or worse), it will be good to know it.

src/expr.cpp
nurmukhametov: I agree

tests/lit-tests/3549.ispc
zephyr111: Ok. I did not think about that, so I agree.

Issue #3552, opened 2025-08-20
Extend CI with more open-source user code builds


nurmukhametov:
Following the discussion in #3548, we can consider adding more tasks to the CI to catch ISPC miscompilation bugs earlier.

Some list of possible candidates to add:
- https://github.com/RenderKit/oidn
- https://github.com/isl-org/Open3D/
- https://github.com/LeelaChessZero/lc0
- https://github.com/GameTechDev/ISPCTextureCompressor
- https://github.com/GPSnoopy/RayTracingInOneWeekend
- https://github.com/proog128/smallpt-ispc
- https://github.com/jakubtomsu/ispc-3d-renderer
- https://gitlab.com/rossinelli/ispc-upwind/
- https://github.com/DensoITLab/Demitasse
- https://github.com/ZipXuan/FlashAttentionGPTonCPU
- https://github.com/chinmaygarde/merle
- https://github.com/re-esper/CPUShadertoy
- https://hg.sr.ht/~hfisher/nbody

PR #3554, opened 2025-08-22
CI: add a job to build more user code

nurmukhametov:
## Description

This pull request introduces a new GitHub Actions workflow, `.github/workflows/opensource-projects.yml`, to automate the building and testing of various open source projects that depend on ISPC from the list mentioned in #3552. This is the follow-up to the discussion here: https://github.com/ispc/ispc/pull/3548#issuecomment-3201471302

## Related Issue
- [X] Linked to relevant issue: #3552

### DIFF
--- a/.github/workflows/opensource-projects.yml
+++ b/.github/workflows/opensource-projects.yml
@@ -1,0 +1,192 @@
+# Copyright 2025, Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause
+
+name: Build opensource projects
+
+permissions: read-all
+
+on:
+  pull_request:
+    paths-ignore:
+      - 'docs/**'
+      - '*.md'
+      - 'snap/**'
+      - 'benchmarks/**'
+      - 'tests/**'
+      - 'examples/**'
+      - 'tools/**'
+      - 'utils/**'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build-ispc:
+    name: Build ISPC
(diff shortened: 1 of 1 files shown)


.github/workflows/opensource-projects.yml
```diff
@@ -0,0 +1,192 @@
...
+      - 'utils/**'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build-ispc:
+    name: Build ISPC
```
KseniyaTikhomirova, line 27: does it make sense to build ispc once and reuse within workflows? I believe it is possible with GH workflows.
you have the same build for https://github.com/ispc/ispc/blob/main/.github/workflows/openmoonray.yml, for example

.github/workflows/opensource-projects.yml
```diff
@@ -0,0 +1,192 @@
+# Copyright 2025, Intel Corporation
```
YixingZhang007, line 1: I saw that this project build workflow file is sharing a lot of workflow commands with `.github/workflows/openmoonray.yml`. Are we considering moving the shared code (such as the `build-ispc` part) into a separate reusable workflow file? 
Also, we might want to consider separating the `build-projects` list from the general workflow logic to keep things more modular and easier for adding new tests in the future.

dbabokin:
@codex review

.github/workflows/opensource-projects.yml
aneshlya: It's a good idea but I think it's out of the scope of this PR. It can be done separately.

PR #3557, opened 2025-09-03
Add the ability to enable SSP (stack smash protection) for ISPC functions.

wolfy1961:
## Introduction
We at Sony PlayStation are interested in ISPC, and we have been doing some work on ISPC targeting our CPU. Security is a high priority for us, and as a consequence, we have done some work on adding SSP (Stack Smash Protection) to ISPC. We'd like to contribute SSP to the community.

## Description
Make the LLVM SSP functionality available in ISPC by introducing the command line option --stack-protector. By itself it enables SSP for some functions the LLVM backend deems vulnerable. When present with the specifier 'strong' (as in --stack-protector=strong), it enables SSP for functions with any arrays. With --stack-protector=all SSP is enabled for all functions. Details are available in the [LLVM documentation](https://llvm.org/docs/LangRef.html).

## Checklist
- [x] Compiler changes are covered by [ssp.ispc](https://github.com/ispc/ispc/compare/main...wolfy1961:ispc:main#diff-7d6b3933fa83af609c3f97e1d1e5dfb07911cebfcc9b42df9dfc081e4092bb98)
- [x] [Documentation](https://github.com/ispc/ispc/compare/main...wolfy1961:ispc:main#diff-f014868f1ccd5256b5fa04823a766153c82b0eb98c0b3d9a2b4d154e2ee7275a) updated if needed

### DIFF
--- a/docs/ispc.rst
+++ b/docs/ispc.rst
@@ -1303,6 +1303,20 @@ compiled with the default ``--addressing=32`` and others were compiled with
 ``--addressing=64``.
 
 
+Stack smash protection (SSP)
+----------------------------
+
+``ispc`` can generate code to guard against stack smashing by enabling
+the LLVM functionality. The ``--stack-protector`` command-line argument
+enables stack protectors for some functions vulnerable to stack smashing.
+This option is equivalent to ``--stack-protector=on``.
+``--stack-protector=strong`` enables stack protectors for functions that
+contain arrays of any size, as well as for functions that take the arguments
+of local variables. ``--stack-protector=all`` enables stack protectors for
+all functions. ``--stack-protector=none`` will not generate any stack protectors,
+which is the default.
+
+
 The Preprocessor
 ----------------
 
--- a/src/args.cpp
+++ b/src/args.cpp
@@ -147,6 +147,13 @@ static ArgsParseResult usage() {
     printf("    [--PIC]\t\t\t\tGenerate position-independent code avoiding any limit on the size of the global offset "
            "table. Ignored for Windows target\n");
     printf("    [--quiet]\t\t\t\tSuppress all output\n");
+    printf("    [--stack-protector]\t\t\tEnable stack protectors for functions with larger stack variables.\n");
+    printf("    [--stack-protector=<option>]\tEnable stack protectors\n");
+    printf("        all\t\t\t\tfor all functions.\n");
+    printf("        none\t\t\t\tfor no functions (default).\n");
+    printf("        on\t\t\t\tfor functions with larger stack variables (same as --stack-protector).\n");
+    printf("        strong\t\t\t\tfor functions with stack variables of any size or taking addresses of local "
+           "variables.\n");
     printf("    [--support-matrix]\t\t\tPrint full matrix of supported targets, architectures and OSes\n");
     printf("    ");
     char targetHelp[2048];
@@ -908,6 +915,21 @@ ArgsParseResult ispc::ParseCommandLineArgs(int argc, char *argv[], std::string &
             g->disableLineWrap = true;
         } else if (!strcmp(argv[i], "--wno-perf") || !strcmp(argv[i], "-wno-perf")) {
             g->emitPerfWarnings = false;
+        } else if (!strcmp(argv[i], "--stack-protector")) {
+            g->SSPLevel = SSPKind::SSPOn;
+        } else if (!strncmp(argv[i], "--stack-protector=", 18)) {
+            const char *level = argv[i] + strlen("--stack-protector=");
+            if (!strcmp(level, "all")) {
+                g->SSPLevel = SSPKind::SSPReq;
+            } else if (!strcmp(level, "on")) {
+                g->SSPLevel = SSPKind::SSPOn;
+            } else if (!strcmp(level, "none")) {
+                g->SSPLevel = SSPKind::SSPNone;
+            } else if (!strcmp(level, "strong")) {
+                g->SSPLevel = SSPKind::SSPStrong;
+            } else {
+                errorHandler.AddError("Unknown --stack-protector= option \"%s\".", level);
+            }
         } else if (!strcmp(argv[i], "-o")) {
             if (++i != argc) {
                 output.out = argv[i];
--- a/src/ispc.cpp
+++ b/src/ispc.cpp
@@ -2358,6 +2358,7 @@ Target::Target(Arch arch, const char *cpu, ISPCTarget ispc_target, PICLevel picL
             fattrBuilder->addAttribute("target-cpu", this->m_cpu);
         }
 #endif
+
         for (auto const &f_attr : m_funcAttributes) {
             fattrBuilder->addAttribute(f_attr.first, f_attr.second);
         }
@@ -3162,6 +3163,7 @@ Globals::Globals() {
     timeTraceGranularity = 500;
     target = nullptr;
     ctx = new llvm::LLVMContext;
+    SSPLevel = SSPKind::SSPNone;
 
 #ifdef ISPC_XE_ENABLED
     stackMemSize = 0;
--- a/src/ispc.h
+++ b/src/ispc.h
@@ -137,6 +137,9 @@ class StorageClass {
     Kind m_kind = Kind::NONE;
 };
 
+// Stack smash protection
+enum class SSPKind { SSPNone, SSPOn, SSPStrong, SSPReq };
+
 // Enumerant for address spaces.
 enum class AddressSpace {
     ispc_default,  // 0 = ispc_private
@@ -989,6 +992,9 @@ struct Globals {
 
     /** When true, include float16 conversion functions permanently to the compiled module */
     bool includeFloat16Conversions;
+
+    /* Stores the stack smash protection setting. */
+    SSPKind SSPLevel;
 };
 
 // This is used when empty string is used for "--darwin-version-min"
--- a/tests/lit-tests/ssp-link.ispc
+++ b/tests/lit-tests/ssp-link.ispc
@@ -1,0 +1,29 @@
+// Check that linking the executable succeeds in the presence of a reference to the function
+// that is called in the event of a stack violation. The symbol is guaranteed to be generated
+// by --stack-protector=all.
+
+// RUN: %{ispc} --target=host --stack-protector=all -h %t.h -o %t.o %s
+// RUN: %{cc} -x c++ -c -o %t-cxx.o --include %t.h %s
+// RUN: %{cc} -o %t.bin %t.o %t-cxx.o
+// On x86 mac hosts clang seems to produce an arm executable, causing this
+// test to fail. Until this is fixed, disable the test on all macs.
+// REQUIRES: !MACOS_HOST
+
+struct S {
+    int arr[20];
+};
+
+#ifdef ISPC
+export uniform int test()
+{
+    uniform S s;
+    return s.arr[14];
+}
+#else
+using namespace ispc;
+
+int main()
+{
+    (void) test();
+}
+#endif /* ISPC */
--- a/tests/lit-tests/ssp.ispc
+++ b/tests/lit-tests/ssp.ispc
@@ -1,0 +1,18 @@
+// Make sure the appropriate ssp function attribute is set.
+
+// RUN: %{ispc} %s --emit-llvm-text -o - | FileCheck %s -check-prefixes=CHECK_ALL,DEFAULT
+// RUN: %{ispc} %s --emit-llvm-text --stack-protector -o - | FileCheck %s -check-prefixes=CHECK_ALL,SSPON
+// RUN: %{ispc} %s --emit-llvm-text --stack-protector=on -o - | FileCheck %s -check-prefixes=CHECK_ALL,SSPON
+// RUN: %{ispc} %s --emit-llvm-text --stack-protector=none -o - | FileCheck %s -check-prefixes=CHECK_ALL,DEFAULT
+// RUN: %{ispc} %s --emit-llvm-text --stack-protector=strong -o - | FileCheck %s -check-prefixes=CHECK_ALL,SSPSTRONG
+// RUN: %{ispc} %s --emit-llvm-text --stack-protector=all -o - | FileCheck %s -check-prefixes=CHECK_ALL,SSPALL
+
+// CHECK_ALL:   define {{.*}}foo{{.*}}#[[ATTRNO:[0-9]+]]
+// DEFAULT-NOT: attributes #[[ATTRNO]] = { {{.* ssp}} }
+// SSPON:       attributes #[[ATTRNO]] = { {{.* ssp .*}} }
+// SSPSTRONG:   attributes #[[ATTRNO]] = { {{.* sspstrong .*}} }
+// SSPALL:      attributes #[[ATTRNO]] = { {{.* sspreq .*}} }
+
+int foo( int val ) {
+    return val;
+}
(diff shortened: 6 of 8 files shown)


dbabokin:
@codex review

src/ispc.cpp
```diff
@@ -2327,6 +2327,15 @@ Target::Target(Arch arch, const char *cpu, ISPCTarget ispc_target, PICLevel picL
             fattrBuilder->addAttribute("target-cpu", this->m_cpu);
         }
 #endif
+
+        // Add SSP attributes if requested.
+        if (g->SSPLevel == SSPOn)
+            fattrBuilder->addAttribute(llvm::Attribute::StackProtect);
+        else if (g->SSPLevel == SSPStrong)
+            fattrBuilder->addAttribute(llvm::Attribute::StackProtectStrong);
+        else if (g->SSPLevel == SSPReq)
```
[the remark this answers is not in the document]
wolfy1961: This review seems incorrect. Setting one of the ssp* attributes on a function does not automatically cause the function to get a canary. The StackProtector pass in the LLVM backend makes that decision based on the characteristics of the function and the attribute (ssp, sspstrong, ssprreq).

The Clang frontend unconditionally attaches the appropriate attribute to all functions in the TU when requested without going through any kind of evaluation of the functions themselves. The suggested implementation does the same thing for ISPC.

src/args.cpp
```diff
@@ -908,6 +913,16 @@ ArgsParseResult ispc::ParseCommandLineArgs(int argc, char *argv[], std::string &
             g->disableLineWrap = true;
         } else if (!strcmp(argv[i], "--wno-perf") || !strcmp(argv[i], "-wno-perf")) {
             g->emitPerfWarnings = false;
+        } else if (!strcmp(argv[i], "--stack-protector")) {
+            g->SSPLevel = SSPOn;
+        } else if (!strncmp(argv[i], "--stack-protector=", 18)) {
+            const char *level = argv[i] + strlen("--stack-protector=");
+            if (!strcmp(level, "strong"))
```
nurmukhametov: The ISPC style guide requires putting curly braces around any amount of if-else statements. The clang-tidy job checks this, see https://github.com/ispc/ispc/actions/runs/17443066415/job/49579489365?pr=3557

src/args.cpp
```diff
@@ -147,6 +147,11 @@ static ArgsParseResult usage() {
     printf("    [--PIC]\t\t\t\tGenerate position-independent code avoiding any limit on the size of the global offset "
            "table. Ignored for Windows target\n");
     printf("    [--quiet]\t\t\t\tSuppress all output\n");
+    printf("    [--stack-protector]\t\tEnable stack protectors for some functions vulnerable to stack smashing.\n");
+    printf("    [--stack-protector=<option>]\t\tEnable stack protectors\n");
+    printf("        all\t\t\tfor all functions.\n");
+    printf(
+        "        strong\t\t\tfor functions containing arrays of any size and taking addresses of local variables.\n");
```
nurmukhametov: I think that it is worth covering all possible option values here; none and default (corresponds to ssp attr) are missing at the moment. Maybe there is better naming for them. It would just be more clear if users are not sure about the default value or if we change it in the future.

src/ispc.cpp
```diff
@@ -2327,6 +2327,15 @@ Target::Target(Arch arch, const char *cpu, ISPCTarget ispc_target, PICLevel picL
...
+            fattrBuilder->addAttribute(llvm::Attribute::StackProtectReq);
```
nurmukhametov: `fattrBuilder` is for target-specific function attributes. SSP* attributes do not look like target-specific ones to me. However, function creation is a little bit messy in the codebase and different function attributes created here and there with no clear guidance. Anyway, I think it is better to follow `NoOmitFramePointer` flag implementation here.

tests/lit-tests/ssp.ispc
```diff
@@ -0,0 +1,16 @@
+// Make sure the appropriate ssp function attribute is set.
```
nurmukhametov, line 1: It would be good to also add a test to check the presence of stack smash protection in the generated asm.

nurmukhametov (changes requested):
Thank you for submitting this PR! This is definitely a valuable contribution. The overall implementation looks good. After addressing the feedback and resolving the discussion points mentioned above, this should be ready to merge.

src/ispc.h
```diff
@@ -137,6 +137,9 @@ class StorageClass {
     Kind m_kind = Kind::NONE;
 };
 
+// Stack smash protection
+enum SSPKind { SSPNone, SSPOn, SSPStrong, SSPReq };
```
dbabokin: I would prefer to have `enum class` instead of `enum`. We are trying to walk away from untypes enums wherever possible.

tests/lit-tests/ssp.ispc
dbabokin: +1 here. LLVM may silently break something and do not generate stack protection in the binary. We need to verify that it happens.

tests/lit-tests/ssp.ispc
```diff
@@ -0,0 +1,16 @@
...
+
+// RUN: %{ispc} %s --emit-llvm-text -o - | FileCheck %s -check-prefixes=CHECK_ALL,DEFAULT
+// RUN: %{ispc} %s --emit-llvm-text --stack-protector -o - | FileCheck %s -check-prefixes=CHECK_ALL,SSPON
+// RUN: %{ispc} %s --emit-llvm-text --stack-protector=strong -o - | FileCheck %s -check-prefixes=CHECK_ALL,SSPSTRONG
+// RUN: %{ispc} %s --emit-llvm-text --stack-protector=all -o - | FileCheck %s -check-prefixes=CHECK_ALL,SSPALL
+
+// CHECK_ALL:   define {{.*}}foo{{.*}}#[[ATTRNO:[0-9]+]]
```
dbabokin, line 10: I also noticed that clang passed `"stack-protector-buffer-size"="8"` in the attributes, I assume we don't need that as we just stick with the default, right?

tests/lit-tests/ssp.ispc
```diff
@@ -0,0 +1,16 @@
+// Make sure the appropriate ssp function attribute is set.
+
+// RUN: %{ispc} %s --emit-llvm-text -o - | FileCheck %s -check-prefixes=CHECK_ALL,DEFAULT
```
dbabokin, line 3: It could make sense also verify that with different opt levels, including `-O0`.

docs/ispc.rst
```diff
@@ -1303,6 +1303,18 @@ compiled with the default ``--addressing=32`` and others were compiled with
...
+Stack smash protection (SSP)
+----------------------------
+
+``ispc`` can generate code to guard against stack smashing by enabling
+the LLVM functionality. The ``--stack-protector`` command-line argument
+enables stack protectors for some functions vulnerable to stack smashing,
+whereas ``stack-protector=strong`` enables it for functions that contain
+arrays of any size, as well as for functions that take the arguments of
+local variables. ``stack-protector=all`` enables stack protectors for
+all functions.
```
dbabokin: The feature implementation in LLVM CodeGen calls `__stack_chk_fail` in case of failing the check, right? The implementation comes from libC (glibc on Linux, libSystem on macOS, not sure about Windows). Does it make sense to explicitly mention that failure is handled by `__stack_chk_fail`?

Also, this mean that we probably need a test that after using this option, we can link to C++ file and produce working binary, i.e. the linker will be able to resolve the `__stack_chk_fail` symbol. This will highlight OSes / environments where this symbol is not avaliable for some reason.

dbabokin (commented):
Thanks for the feature contribution! It's definitely useful one! Please address the remarks in the individual comment.

src/args.cpp
wolfy1961: Added the braces.

src/args.cpp
wolfy1961: I added the options 'on' and 'none' (I didn't like "default' as that may cause confusion. The default is to not generate ssp at all).

src/ispc.cpp
wolfy1961: I modeled the attribution after what you suggested. Hope I got it right.

tests/lit-tests/ssp.ispc
wolfy1961: I added a couple of tests to check that the appropriate fail function is called, one for a typical linux target and one for windows. It may be that we need more? I'm unfortunately not aware of all possible target OSes.

src/ispc.h
wolfy1961: Done.

tests/lit-tests/ssp.ispc
wolfy1961: The LLVM default is 8, so perhaps we're OK there. Maybe we'll need to spell it out explicitly, but I don't immediately see a reason to pick a different value.

tests/lit-tests/ssp.ispc
wolfy1961: I test all optimization levels in the test that checks the assembly level.

docs/ispc.rst
wolfy1961: I'm not sure about the documentation requirement either. We could just refer the user to the LLVM documentation to avoid having to update anything in case it changes. I can certainly add something if you think it's necessary.

I added a test to see if an executable can be generated. The test should work on linux and windows at least. 
I would assume that a failure function would be supported on all platforms that LLVM supports, so I think we may be OK there.

src/ispc.cpp
nurmukhametov: It is fine with me.

tests/lit-tests/ssp.ispc
nurmukhametov: I am fine with the added tests. Ideally, it would be good to add a test for macOS as well or enable the Linux one (it might be identical).

nurmukhametov: Linux test also fails on aarch64 linux runner: https://github.com/ispc/ispc/actions/runs/17570063375/job/49929675765?pr=3557

tests/lit-tests/ssp.ispc
dbabokin: Ok, let's leave it as is.

tests/lit-tests/ssp-asm-linux.ispc
```diff
@@ -0,0 +1,23 @@
+// Make sure the backend honors the request to generate code to detect stack smashing.
+// We check for a call to __stack_chk_fail.
+// RUN: %{ispc} -O0 --emit-asm %s --target-os=linux --stack-protector=strong -o - | FileCheck %s
+// RUN: %{ispc} -O1 --emit-asm %s --target-os=linux --stack-protector=strong -o - | FileCheck %s
+// RUN: %{ispc} -O2 --emit-asm %s --target-os=linux --stack-protector=strong -o - | FileCheck %s
+// RUN: %{ispc} -O3 --emit-asm %s --target-os=linux --stack-protector=strong -o - | FileCheck %s
+// REQUIRES: LINUX_ENABLED
```
dbabokin: I suggest having just one test for "host" os that universally works on ARM/x86 and Linux/Windows/macOS. Regexp syntax for matching expression is flexible enough to allow matching both `call` and `bl` instruction, as well as all variations of security functions that we aware of. If we see fail on other platforms, we'll just add more variations later.

If you specifically care about cross-compilation scenario (I suspect you do), it's ok to have one more test for cross compilation scenario (i.e. `--target-os=<your os>`) that you care about. But it's up to you. I think if `host` is works, we should be good.

docs/ispc.rst
dbabokin: Ok, these functions are platform specific and then probably it doesn't worth it to go much into the details in the docs.

tests/lit-tests/ssp-link.ispc
```diff
@@ -0,0 +1,27 @@
+// Check that linking the executable succeeds in the presence of a reference to the function
+// that is called in the event of a stack violation. The symbol is guaranteed to be generated
+// by --stack-protector=all.
+
+// RUN: %{ispc} --target=host --stack-protector=all -h %t.h -o %t.o %s
+// RUN: %{cc} -x c++ -c -o %t-cxx.o --include %t.h %s
+// RUN: %{cc} -o %t.bin %t.o %t-cxx.o
+// REQUIRES: !MACOS_HOST || MACOS_ARM_ENABLED
```
nurmukhametov: I think we need just `!MACOS_HOST` to skip running this test on macOS. Apart from this test, everything else looks good to me.

dbabokin: It should pass on macOs as well. I'm not sure what the problem is, I'm trying to build and test.

wolfy1961: Seems like ispc is generating x86 code, but clang is producing arm code. Maybe a problem in the test SDK?

nurmukhametov:
```bash
$ ":" "RUN: at line 5"
$ "ispc" "--target=host" "--stack-protector=all" "-h" "/Users/runner/work/ispc/ispc/build/bin/tests/Output/ssp-link.ispc.tmp.h" "-o" "/Users/runner/work/ispc/ispc/build/bin/tests/Output/ssp-link.ispc.tmp.o" "/Users/runner/work/ispc/ispc/tests/lit-tests/ssp-link.ispc"
$ ":" "RUN: at line 6"
$ "clang" "-x" "c++" "-c" "-o" "/Users/runner/work/ispc/ispc/build/bin/tests/Output/ssp-link.ispc.tmp-cxx.o" "--include" "/Users/runner/work/ispc/ispc/build/bin/tests/Output/ssp-link.ispc.tmp.h" "/Users/runner/work/ispc/ispc/tests/lit-tests/ssp-link.ispc"
$ ":" "RUN: at line 7"
$ "clang" "-o" "/Users/runner/work/ispc/ispc/build/bin/tests/Output/ssp-link.ispc.tmp.bin" "/Users/runner/work/ispc/ispc/build/bin/tests/Output/ssp-link.ispc.tmp.o" "/Users/runner/work/ispc/ispc/build/bin/tests/Output/ssp-link.ispc.tmp-cxx.o"
# command stderr:
-macosx_version_min has been renamed to -macos_version_min
ld: warning: ignoring file '/Users/runner/work/ispc/ispc/build/bin/tests/Output/ssp-link.ispc.tmp.o': found architecture 'x86_64', required architecture 'arm64'
ld: Undefined symbols:
  _test, referenced from:
      _main in ssp-link.ispc.tmp-cxx.o
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

IIRC, this problem is typical for similar tests (all of them just skip MACOS_HOST) and doesn't relate to the current PR.

dbabokin: @nurmukhametov We should fix that (not as part of this PR). Something is off here (should we update lit package? should we make sure we call "host" target for clang? should we make sure we are not running in emulation mode?)

wolfy1961: Normally clang should pick the host cpu as target if you don't give it a triple or -mcpu option, unless it's configured to cross compile by default. I don't know of an equivalent to -target=host in clang.

I'll just disable the test for MACOS_HOST for now.

nurmukhametov: > @nurmukhametov We should fix that (not as part of this PR). Something is off here (should we update lit package? should we make sure we call "host" target for clang? should we make sure we are not running in emulation mode?)

Sure, I have created https://github.com/ispc/ispc/issues/3559 to follow this.

wolfy1961:
> LGTM! Could you please squash and rebase on top of the current main branch?

Should be done now.

nurmukhametov:
@wolfy1961, thank you for your contribution!

PR #3561, opened 2025-09-12
Add amx example

nurmukhametov:
## Description

This pull request adds a new Intel AMX-FP16 matrix multiplication example to the CPU examples, including hardware or emulator detection.

* Added `amx` subdirectory to the CPU examples.
* `examples/cpu/amx/CMakeLists.txt`: Implements detection for AMX hardware support via CPUID and falls back to Intel SDE emulation if hardware is not available. Adds a build/run target only when AMX support is detected, and sets up all necessary compile and run configurations.
* `examples/cpu/amx/amx_matmul.ispc`: an ISPC implementation of AMX-FP16 matrix multiplication, including AMX tile configuration, matrix initialization, AMX and naive multiplication kernels, result verification.
* `examples/cpu/amx/amx_matmul.cpp`: C++ code that requests AMX permissions, runs both the AMX and naive implementations, times and verifies results, and prints performance metrics.

## Related Issue

## Checklist

### DIFF
--- a/.github/workflows/ispc-ci.yml
+++ b/.github/workflows/ispc-ci.yml
@@ -669,7 +669,7 @@ jobs:
     - name: Run examples
       shell: cmd
       run: |
-        ctest --test-dir examples-build -C Release
+        ctest --test-dir examples-build -C Release --output-on-failure
 
   linux-examples:
     needs: [linux-package-examples, linux-build-ispc]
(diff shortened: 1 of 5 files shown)


dbabokin:
@codex review

aneshlya:
So eventually it would be good to compare this implementation with whatever we have in `attention` example. To do this we'll need to 
1. update matmul used in `attention` to use fp16
2. use jit to configure tiles there
3. anything else?

@nurmukhametov, do you recommend any changes/optimizations in amx example before such comparison?

PR #3567, opened 2025-09-17
Add script to compare function assembly between ISPC compilers

nurmukhametov:
Added `scripts/compare_func.sh` for detailed assembly comparison between old and new ISPC compiler versions. The script supports:
- Function-specific assembly extraction and comparison
- Side-by-side diff output with `--side-by-side` flag
- LLVM-MCA performance analysis integration
- Support for both mangled and unmangled function names
- Flexible compiler path configuration via environment variables

This tool enables regression testing and performance analysis when developing compiler optimizations.

Generated with Claude Code


dbabokin:
@codex review

Issue #3563, opened 2025-09-16
Enum is empty in C/C++ header


nurmukhametov:
Consider the following example:

```ispc
enum E { A };
export void foo(const uniform E x) {}
```

when a C/C++ header is generated it contains an empty enum declaration:

```c++
#ifndef __ISPC_ENUM_E__
#define __ISPC_ENUM_E__
enum E {
    A = 0
};
#endif
```

nurmukhametov:
This was introduced somewhere between v1.26.0 and v1.27.0

PR #3568, opened 2025-09-17
Fix enum fields missing from generated headers

nurmukhametov:
This commit fixes a bug introduced by the Type system refactoring in commit fdb994e64 where enum fields were not included in generated C/C++ header files.

The issue was caused by a change in caching behavior during type transformations:
- Old system: EnumType had its own GetAsConstType()/GetAsUniformType() methods that created new copies without caching
- New system: EnumType uses base Type class methods that cache the results

The parsing order in lCreateEnumType() called lFinalizeEnumeratorSymbols() before SetEnumerators(), causing cached const/uniform type copies to be created with empty enumerator lists. When SetEnumerators() was later called, it only updated the original EnumType, not the cached copies used for header generation.

## Related Issue
- [X] Linked to relevant issue: #3563

## Checklist

### DIFF
--- a/src/parse.yy
+++ b/src/parse.yy
@@ -3761,10 +3761,10 @@ lCreateEnumType(const char *name, std::vector<Symbol *> *enums, SourcePos pos) {
     if (name != nullptr)
         m->symbolTable->AddType(name, enumType, pos);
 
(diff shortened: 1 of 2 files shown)


dbabokin:
@codex review

PR #3572, opened 2025-09-18
Add script for debugging ISPC optimization pipeline

nurmukhametov:
Combines function extraction and HTML diff generation to help debug middle-end optimization passes. Extracts specified functions from LLVM IR debug files and creates side-by-side comparisons showing how the function transforms through each optimization phase.

Generated with Claude Code


aneshlya:
I agree with Copilot review

Issue #3571, opened 2025-09-18
Bool return value regression


nurmukhametov:
Consider the following example.

C++:

```c++
#include <iostream>
#include "ispc_exact_repro_ispc_stubs.h"

inline bool asCppBool(bool ispcBool) {
    int i = ispcBool ? 1 : 0;
    return (bool)i;
}

int main() {
    float scl = 0.001f;
    uint32_t isectsEqual = 0;
    
    bool rawResult = ispc::foo(scl, isectsEqual);
    
    bool convertedResult = asCppBool(rawResult);
    
    std::cout << "Raw ISPC result: " << (int)rawResult << " (as int)" << std::endl;
    std::cout << "Raw ISPC result: " << rawResult << " (as bool)" << std::endl;
    std::cout << "asCppBool result: " << (int)convertedResult << " (as int)" << std::endl;
    std::cout << "asCppBool result: " << convertedResult << " (as bool)" << std::endl;
    
    return 0;
}
```

ISPC:
```ispc
export uniform bool foo(uniform float scl, uniform uint32 &isectsEqual) {
    uniform bool success = true;
    uint32 localIsectsEqual = 0;

    for (uniform uint32 y = 0; y < 2; y++) {
        float testValue = (float)y * scl;

        bool passed = true;
        bool asserted = false;

        if (!asserted) {
            if (testValue > 1.0f) {
                passed = false;
                if (testValue > 0.1f) {
                    asserted = true;
                }
            }
        }

        if (asserted) {
            success = false;
        }

        localIsectsEqual += passed ? 1 : 0;
    }

    isectsEqual = reduce_add(localIsectsEqual);

    return success;
}
```

Run like this:

```bash
#!/bin/bash
# Test the boolean bug with different ISPC versions

set -e
cd "$(dirname "$0")"

echo "=== Testing ISPC Boolean Bug Across Versions ==="

ISPC_V126="ispc-v1.26.0-linux/bin/ispc"
ISPC_V127="ispc-v1.27.0-linux/bin/ispc"

echo "Found both ISPC versions:"
echo "v1.26.0: $ISPC_V126"
echo "v1.27.0: $ISPC_V127"
echo ""

# Function to test with a specific ISPC version
test_with_ispc() {
    local ispc_bin="$1"
    local version="$2"
    
    echo "=== Testing with ISPC $version ==="
    
    # Clean previous builds
[...]

PR #3573, opened 2025-09-18
Replace trunc i8 to i1 with icmp ne i8 in bool conversion

nurmukhametov:
## Description

Replace trunc i8 to i1 with icmp ne i8, 0 for proper bool conversion. LLVM 19+ removed the InstCombine canonicalization that converted trunc i8 to i1 to icmp ne i8, 0. This caused uniform bool return values to potentially be 255 instead of the expected 1.

The fix updates lSwitchBoolSize_2() in ctx.cpp to generate proper comparison instructions instead of relying on removed LLVM InstCombine canonicalization. This canonicalization for trunc to `i1` in LLVM was removed in https://github.com/llvm/llvm-project/pull/84628 that caused regression in #3571. 

## Related Issue
- [X] Linked to relevant issue: #3571

## Checklist

### DIFF
--- a/src/ctx.cpp
+++ b/src/ctx.cpp
@@ -2555,9 +2555,10 @@ llvm::Value *FunctionEmitContext::lSwitchBoolSize_2(llvm::Value *value, llvm::Ty
     if (llvm::dyn_cast<llvm::FixedVectorType>(fromType)) {
         llvm::VectorType *i1VecType = LLVMVECTOR::get(
             llvm::Type::getInt1Ty(*g->ctx), llvm::dyn_cast<llvm::FixedVectorType>(fromType)->getNumElements());
-        // trunc only if needed
+        // Use icmp ne instead of trunc for proper bool conversion
         if (fromType != i1VecType) {
-            i1Bool = TruncInst(value, i1VecType, newName);
+            llvm::Value *zeroVec = llvm::Constant::getNullValue(fromType);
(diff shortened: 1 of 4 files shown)


nurmukhametov:
At the same time, we can fix this regression on our side.

The first approach is to emit icmp instead of trunc in `lSwitchBoolSize_2()` https://github.com/ispc/ispc/pull/3573/commits/af09fef5fa5251f1ec5849c7929baf35e2b1b535. When we emitted trunc to i1 there we implicitly relied on canonicalization of that to icmp.

The second one is to change the boolean storage constants from 0xFF to 0x01 (https://github.com/ispc/ispc/pull/3573/commits/99c5aabfe85011d9fa25db96b3209f155a844d06).

I am not sure which approach is better.

aneshlya:
> 
> The first approach is to emit icmp instead of trunc in `lSwitchBoolSize_2()` [af09fef](https://github.com/ispc/ispc/commit/af09fef5fa5251f1ec5849c7929baf35e2b1b535). When we emitted trunc to i1 there we implicitly relied on canonicalization of that to icmp.
> 
> The second one is to change the boolean storage constants from 0xFF to 0x01 ([99c5aab](https://github.com/ispc/ispc/commit/99c5aabfe85011d9fa25db96b3209f155a844d06)).
> 
> I am not sure which approach is better.

I think the first approach is safer from user's perspective. Someone may rely on boolean storage constants values.

aneshlya (approved):
Looks good!

PR #3575, opened 2025-09-22
Claude Code development configuration

aneshlya:
This PR introduces the initial `CLAUDE.md` file based on best practices from the team.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,76 @@
+# ISPC Project
+
+## Project Overview
+
+ISPC (Intel SPMD Program Compiler) is a performance-oriented compiler for
+vectorized parallel programming.
+
+## Build System
+
+**CMake-based** (Unix Makefiles/Ninja)
+- Build directory: `build/` (ask user permission before creating if missing)
+- Generated compiler: `build/bin/ispc`
+
+**Build commands:**
+```bash
+cmake -B build
+cmake --build build -j $(nproc)
+```
+
+**Test commands:**
+Lit tests:
+```bash
+cmake --build build --target check-all
+```
+
+To test the specific test, run:
+```bash
+TEST=/full/path/test.ispc cmake --build build --target check-one
+```
+
+Functional tests:
+```bash
+PATH=`pwd`/build/bin:$PATH ./scripts/run_tests.py --target=avx2-i32x8
+```
+
+**Important**: Both build and tests take several minutes - use longer timeouts.
+Some tests may show UNSUPPORTED/XFAIL status normally.
+
+**Testing strategy**: Lit tests for compilation/parsing changes,
+functional tests for runtime behavior changes.
+
+## Source Code Architecture
+### Key directories:
+
+- **src/**: Core compiler source (C++ frontend, AST, codegen, optimizations)
+- **stdlib/**: ISPC standard library (built-in functions, target-specific implementations)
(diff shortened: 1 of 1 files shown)


.claude/hooks/code_format.py
```diff
@@ -0,0 +1,150 @@
...
+    if target_path in clang_targets and has_required_clang_format:
+        apply_clang_format(target_path)
+    else:
+        strip_trailing_whitespace(target_path)
+
+    print("Done!")
+    sys.exit(0)
+
+
+if __name__ == "__main__":
```
dbabokin: Should we have this script as separate script and call it from here?

dbabokin (commented):
I suggest adding `.claude/settings.local.json` to git ignore.

.claude/hooks/code_format.py
aneshlya: What for? It seems to me like a small self-contained hook. What are the reasons to separate it?

CLAUDE.md
```diff
@@ -0,0 +1,55 @@
...
+
+**Lit tests** (`tests/lit-tests/`, 650+ tests): Compilation correctness
+```bash
+cd build/ && make check-all  # or: cmake --build . --target check-all or ninja check-all
+```
+
+**Functional tests** (`tests/func-tests/`, 1500+ tests): Runtime behavior
+```bash
+python3 scripts/run_tests.py
+```
```
nurmukhametov, line 34: I think it is worth mentioning the `check-one` target:

To test the specific test, run TEST=`full_path_to/tests/lit-tests` ninja check-one

CLAUDE.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Source Code Architecture
+
```
nurmukhametov, line 54: What about adding here at least stdlib, examples and benchmarks?

CLAUDE.md
```diff
@@ -0,0 +1,55 @@
...
+## Build System
+
+**CMake-based** (Unix Makefiles/Ninja)
+- Build directory: `build/` (ask user permission before creating if missing)
+- Generated compiler: `build/bin/ispc`
+
+**Build commands:**
+```bash
+# Manual setup
+mkdir build && cd build && cmake ../
```
nurmukhametov: What about avoiding directories changes as much as possible, i.e.:

```bash

cmake -B build
cmake --build build -j $(nproc)
cmake --build build --target check-all
```

.claude/hooks/code_format.py
dbabokin: This should be standard way to format the code - both with manual development and through the tools.

CLAUDE.md
aneshlya: Do you by any chance have it described in your CLAUDE.md? Just want to use something that was used in practice.

nurmukhametov: My CLAUDE.md is [here](https://github.com/nurmukhametov/ispc/blob/claude-code/CLAUDE.md)

.claude/hooks/code_format.py
aneshlya: I'm not adding formatting as part of this PR, there are many ways to implement auto-formatting better then using Claude hooks (e.g. git precommit hook). The example of how to do it using hooks can be found here: https://github.com/aneshlya/ispc/tree/claude-dev-hook

PR #3577, opened 2025-09-24
Sample profiling debug info and its usage for PGO

nurmukhametov:
## Description

Add `--sample-profiling-debug-info flag` to enable the generation of debug information optimized for sample-based profiling. When this flag is used:
- Variable and function parameter debug information is skipped to reduce overhead.
- The `DICompileUnit` is configured with `LineTablesOnly` emission kind and `DebugInfoForProfiling` set to true, which is suitable for profiling tools.

Add `--profile-sample-use=<file>` to enable profile-guided optimization (PGO) using sample profile data. When this flag is provided, the ISPC compiler will load the specified sample profile data file and use it to guide optimization decisions during compilation. This is achieved by adding an `llvm::SampleProfileLoaderPass` to the optimization pipeline.

## Related Issue

## Checklist


aneshlya:
What is missing now in this PR is a documentation update. It would be good to have a whole section related to HWPGO, not just simple description of options but the whole workflow - how to get the profile, how to use it, compile, link with C++ (do we need to pass any additional options to linker?)... If it's easier to add an example, it can be in the form of example.

I'm also wondering if you had a chance to measure any perf impact from this change and what HW did you use?

PR #3579, opened 2025-09-25
Fix 3358: Support structures in lIsSafeToBlend

### LINKED ISSUE #3358: Assigning to struct prevents blended stores
[Godbolt](https://godbolt.org/z/6jraa7xEo)
```ispc
struct Compound
{
    float<16> value;
};

unmasked void test1(Compound & data, float test)
{
    float<16> copy = data.value;
    if (test <= 0) {
        copy += 0.01;
    }
    data.value = copy;
}

unmasked void test2(Compound & data, float test)
{
    Compound copy = data;
    if (test <= 0) {
        copy.value += 0.01;
    }
    data = copy;
}
```

In this example, blended stores will be used in `test1` while masked stores will be used in `test2`. On targets without masked stores like NEON, this leads to worse codegen for `test2` which has to use many branches to emulate the masked stores.

Maybe this is because [`lIsSafeToBlend`](https://github.com/ispc/ispc/blob/ea5022b5db507839f7c004c65cca77ed0f040a15/src/opt/ReplacePseudoMemoryOps.cpp#L23) doesn't work with struct members?

There seems to be many ways to work around this, but thought I would create an issue in case this wasn't known.

aneshlya:
## Description
This PR supports structures in analysis allowing to promote masked store to blend store. It's especially beneficial on the targets without HW masked stores (like neon, sse4) where emulating of masked store results in multiple branches.
No regressions were observed. One improvement of 53% is present on one GameDev benchmark.

## Related Issue
- [x] Fixes #3358 

## Checklist

### DIFF
--- a/src/opt/ReplacePseudoMemoryOps.cpp
+++ b/src/opt/ReplacePseudoMemoryOps.cpp
@@ -1,5 +1,5 @@
 /*
-  Copyright (c) 2022-2024, Intel Corporation
+  Copyright (c) 2022-2025, Intel Corporation
 
   SPDX-License-Identifier: BSD-3-Clause
 */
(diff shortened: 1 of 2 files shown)


dbabokin:
@codex review

Issue #3238, opened 2025-02-18
For loop counter codegen


nurmukhametov:
Consider the following loop:

```ispc
export void Fill(uint32* uniform buffer, uniform int count) {
  for (uniform uint i = 0; i < count; i += programCount) {
    buffer[i + programIndex] = 7;
  }
}
```

Generated asm is not ideal

```asm
.LBB0_2:                                # =>This Inner Loop Header: Depth=1
        cdqe
        vmaskmovps      ymmword ptr [rdi + rax], ymm0, ymm1
        add     ecx, 8
        add     eax, 32
        cmp     ecx, esi
        jb      .LBB0_2
```

because of `cdqe` and extra `add`.

Compiler explorer link https://ispc.godbolt.org/z/9GdGME6sE

It reminds me of the similar issue with `foreach` loop counters #2870

nurmukhametov:
This issue is related to #3242. ISPC generates `sext` instead of `zext` during address calculation (https://ispc.godbolt.org/z/vxcKoq5P7).

With proper sign extension and `!"llvm.loop.mustprogress"` being attached, `IndVarSimplify` would have been able to promote the loop counter to i64 (https://godbolt.org/z/jEsTjE7Ya)

aneshlya:
@karthik-senthil, could you please into this?
ISPC should be built against LLVM 20.
You can use `--debug-phase=pre:last --dump-file=<out_folder>` ispc flags to dump IR after each pass.

karthik-senthil:
I believe `IndVarSimplify` is unable to promote `i32` type of loop IV phi to `i64` due to missing wrap flags (`nuw` and `nsw`) in arithmetic operations that involve the IV. Check https://godbolt.org/z/zeqj61M1q where adding both flags to the `shl` and IV increment `add` operation enables `IndVarSimplify` to promote the data type. I have some local fixes to update `ispc` to emit both these flags for binary operations when applicable. With these changes in my local WS, following is the output for the test case -

``` bash
$ ../build-ispc/bin/ispc -O1 gh3238.ispc --emit-llvm-text -o fixed_gh3238.ll
$ cat fixed_gh3238.ll
```
``` llvm
for_loop:                                         ; preds = %for_loop.lr.ph, %for_loop
  %indvars.iv = phi i64 [ 0, %for_loop.lr.ph ], [ %indvars.iv.next, %for_loop ]
  %1 = shl nuw nsw i64 %indvars.iv, 2
[...]

PR #3026, opened 2024-09-25
ImproveMemOps: consider zext in lOffsets32BitSafe

nurmukhametov:
After LLVM change [d77067d08](https://github.com/llvm/llvm-project/commit/d77067d08), InstCombine substitutes 
```llvm-ir
%v2 = sext <4 x i32> %v1 to <4 x i64>
```
with
```llvm-ir
%v2 = zext nneg <4 x i32> %v1 to <4 x i64>
```

`lOffsets32BitSafe` should consider `zext` too to be able to replace `__pseudo_scatter64_i32` with `__pseudo_scatter_factored_base_offsets32_i32`. Otherwise, it replaces it with `__pseudo_scatter_factored_base_offsets32_i32` that somehow results in generation of `vunpckhps`.

This fixes https://github.com/ispc/ispc/blob/main/tests/lit-tests/interleaved.ispc (issue #2813) with LLVM 18.1.

Re-apply this after #3009 as it should not change codegen for loop counters now.

PR #3581, opened 2025-10-01
[ImproveMemOps] Do not truncate offset computation for 64-bit targets.

karthik-senthil:
Currently ImproveMemOps pass tries to truncate all offset computations to 32-bits
even for 64-bit targets. This is not conducive for downstream LLVM passes like
InstCombine when incoming ISPC source has unsigned loop induction variable. It ends
up generating a sext instead of zext while computing addresses using the unsigned
induction variable. This PR updates ImproveMemOps to not truncate offsets when
target is 64-bit, to keep inline with ISPC frontend. Additionally it also updates
ISPC frontend to not emit nsw flags on binops using unsigned loop induction variable.

Github issue: #3238

## Description
Brief description of changes

## Related Issue

## Checklist

### DIFF
--- a/tests/lit-tests/for_unsigned_loop_counter.ispc
+++ b/tests/lit-tests/for_unsigned_loop_counter.ispc
@@ -1,0 +1,22 @@
+// This test ensures that ispc compiler generates a zext for an unsigned
+// loop induction variable. We also check that nsw flag is not incorrectly
+// set on binop operating on the induction variable.
+
+// RUN: %{ispc} %s --emit-llvm-text --target=host --no-discard-value-names -o - -O1 --nowrap | FileCheck %s
+
+// CHECK-LABEL: define void @fill(
+// CHECK:      for_loop:
+// CHECK-NEXT:   %i.019 = phi i32 [ %add_i_load10_, %for_loop ], [ 0, %allocas ]
+// CHECK-NEXT:   %shl_i_load4_broadcast5_.elt0 = shl i32 %i.019, 2
+// CHECK-NEXT:   %offset_cast.elt0 = zext i32 %shl_i_load4_broadcast5_.elt0 to i64
+// CHECK-NEXT:   %ptr = getelementptr i8, ptr %buffer, i64 %offset_cast.elt0
+// CHECK-NEXT:   store <8 x i32> splat (i32 7), ptr %ptr, align 4
+// CHECK-NEXT:   %add_i_load10_ = add i32 %i.019, 8
+// CHECK-NEXT:   %less_i_load_count_load = icmp ult i32 %add_i_load10_, %count
+// CHECK-NEXT:   br i1 %less_i_load_count_load, label %for_loop, label %for_exit
+
+export void fill(uint32* uniform buffer, uniform int count) {
+    for (uniform uint i = 0; i < count; i += programCount) {
+        buffer[i + programIndex] = 7;
+    }
+}
--- a/src/ctx.cpp
+++ b/src/ctx.cpp
@@ -2145,6 +2145,8 @@ llvm::Value *FunctionEmitContext::applyVaryingGEP(llvm::Value *basePtr, llvm::Va
 
     bool indexIsVarying = llvm::isa<llvm::VectorType>(index->getType());
     llvm::Value *offset = nullptr;
+    // Determine wrap semantics based on index type.
+    WrapSemantics ws = (indexType && indexType->IsSignedType()) ? WrapSemantics::NSW : WrapSemantics::None;
     if (indexIsVarying == false) {
         // Truncate or sign extend the index as appropriate to a 32 or
         // 64-bit type.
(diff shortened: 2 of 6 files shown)


tests/lit-tests/for_unsigned_loop_counter.ispc
```diff
@@ -0,0 +1,22 @@
...
+
+// RUN: %{ispc} %s --emit-llvm-text --target=host --no-discard-value-names -o - -O1 --nowrap | FileCheck %s
+
+// CHECK-LABEL: define void @fill(
+// CHECK:      for_loop:
+// CHECK-NEXT:   %i.019 = phi i32 [ %add_i_load10_, %for_loop ], [ 0, %allocas ]
+// CHECK-NEXT:   %shl_i_load4_broadcast5_.elt0 = shl i32 %i.019, 2
+// CHECK-NEXT:   %offset_cast.elt0 = zext i32 %shl_i_load4_broadcast5_.elt0 to i64
+// CHECK-NEXT:   %ptr = getelementptr i8, ptr %buffer, i64 %offset_cast.elt0
+// CHECK-NEXT:   store <8 x i32> splat (i32 7), ptr %ptr, align 4
```
aneshlya: `--target=host` will use different target width on different HW. Please adjust `store` instruction to be width-independent.

tests/lit-tests/3238.ispc
```diff
@@ -0,0 +1,22 @@
+// This test ensures that ispc compiler generates a zext for an unsigned
```
aneshlya: Please rename this test to 3238.ispc. It's easier to track in the future.

dbabokin:
@codex review

aneshlya:
I think this fix is too aggressive and it prevents lots of optimizations coming from 32-bit addressing. I'd like to come back to it tomorrow but one idea is to modify `lOffsets32BitSafe()` to return `false` when encountering `zext`. `sext` can still be optimized (truncated and re-extended safely). Note that there are the single-parameter and two-parameter versions of this functions and probably both should be updated.
I found related fix mentioning llvm commit (just for the reference): https://github.com/ispc/ispc/pull/3026

Claude's analysis of the issue:
In ctx.cpp, we explicitly generate different extension instructions based on type signedness:
  - Unsigned types _ ZExtInst
  - Signed types _ SExtInst

  In ImproveMemoryOps.cpp::lOffsets32BitSafe(), we need to recognize and preserve this distinction:
  - When we encounter ZExtInst from i32_i64, it signals an unsigned type from ctx.cpp
  - When we encounter SExtInst from i32_i64, it signals a signed type from ctx.cpp

The extension instruction type (zext vs sext) is the only remaining marker of the original type's signedness once we're at the LLVM IR level. If we truncate the zext back to i32, we lose this crucial information. When LLVM later needs to re-extend that i32 to i64, it has no way to know it should use zext instead of sext.

tests/lit-tests/3238.ispc
karthik-senthil: Done.

tests/lit-tests/for_unsigned_loop_counter.ispc
karthik-senthil: Done.

karthik-senthil (commented):
Thanks for the review Arina! I've addressed the comments below.


> I think this fix is too aggressive and it prevents lots of optimizations coming from 32-bit addressing. I'd like to come back to it tomorrow but one idea is to modify `lOffsets32BitSafe()` to return `false` when encountering `zext`. `sext` can still be optimized (truncated and re-extended safely). Note that there are the single-parameter and two-parameter versions of this functions and probably both should be updated. I found related fix mentioning llvm commit (just for the reference): #3026
> 
> Claude's analysis of the issue: In ctx.cpp, we explicitly generate different extension instructions based on type signedness:
> 
> * Unsigned types _ ZExtInst
> * Signed types _ SExtInst
> 
> In ImproveMemoryOps.cpp::lOffsets32BitSafe(), we need to recognize and preserve this distinction:
> 
> * When we encounter ZExtInst from i32_i64, it signals an unsigned type from ctx.cpp
> * When we encounter SExtInst from i32_i64, it signals a signed type from ctx.cpp
> 
> The extension instruction type (zext vs sext) is the only remaining marker of the original type's signedness once we're at the LLVM IR level. If we truncate the zext back to i32, we lose this crucial information. When LLVM later needs to re-extend that i32 to i64, it has no way to know it should use zext instead of sext.

Done. I've updated `ImproveMemoryOps` to use `sext` vs. `zext` instead of target.

aneshlya:
Since `loopmustprogress` is already merged I think it would be beneficial to add one more test here checking the assembly as was initially described in https://github.com/ispc/ispc/issues/3238.

karthik-senthil:
> Since `loopmustprogress` is already merged I think it would be beneficial to add one more test here checking the assembly as was initially described in #3238.

Done. I've updated the LIT test `3238.ispc` to also check for ASM.

aneshlya:
Looks good but the test is still failing on some platforms. Let me know if you nee help with fixing/reproducing.

PR #3582, opened 2025-10-06
Use generic max/min and masked_blend_stores for int32/int64 for sse2

aneshlya:
## Description
There is a regression after https://github.com/ispc/ispc/pull/3579 on `struct-test-127.ispc` test in nightly testing on `sse2-i32x8` target: https://github.com/ispc/ispc/actions/runs/18272766892/job/52081357993, https://github.com/ispc/ispc/actions/runs/18272766892/job/52081357968.

This PR fixes a regression by switching to the generic implementation for blend stores and max/min operations instead of using the SSE2-specific one. The previous implementation didn’t rely on any hardware intrinsics (since SSE2 doesn’t support blend instructions) but was overly specialized and not reused elsewhere.

From a code generation perspective, both the previous and new implementations produce very similar code. However, the new generic version is leaner and cleaner.

The main issue with the old implementation was its interaction with InstCombine and SROA, which caused unintended transformations for this test:
For x8 (256-bit, wider than SSE2’s native 128-bit):
- ISPC/LLVM emits a `llvm.masked.load` with an undef passthrough.
- InstCombine converts this into a shufflevector with undef values.
- When a blend store is used, inlined, and then SROA’d, InstCombine incorrectly eliminates the -150 initialization.

Lit test tests/lit-tests/mask_extract_first_el.ispc was removed since the original issue covered by it was also caused by implementation of blend stores for sse2 targets. 

## Related Issue
- [x] Linked to relevant issue(s). Finishes implementation of #3358.

## Checklist


dbabokin:
@codex review

dbabokin (approved):
LGTM.

Since we unified the implementation, should we move the definitions to `util.m4`?

aneshlya:
> Since we unified the implementation, should we move the definitions to `util.m4`?

I've moved common parts to `builtins/target-sse2-common.ll`.

PR #3578, opened 2025-09-25
Support RISC-V

nurmukhametov:
## Description

Implements foundation for RISC-V 64-bit (riscv64) target support with RISC-V Vector Extension (RVV) ISA, introducing the rvv-x4 target for 4-wide vectorization.

Key additions:
- RISCV_ENABLED CMake configuration option
- RISC-V target enumeration and ISA definitions
- LLVM backend initialization for RISC-V
- RISC-V target with 4-wide vectorization
- Architecture detection and triple configuration

This PR also adds GitHub Actions workflow for testing RISC-V Vector Extension (rvv-x4) target using Arch Linux container, riscv64-linux-gnu cross-compiler, and QEMU user-mode emulation to test compiled binaries.

## Related Issue
- [X] Linked to relevant issue: #1582

## Checklist
- [X] Code has been formatted with `clang-format` (e.g., `clang-format -i src/ispc.cpp`)
- [X] Git history has been squashed to meaningful commits (one commit per logical change)
- [X] Compiler changes are covered by [lit tests](https://github.com/ispc/ispc/tree/main/tests/lit-tests)
- [X] [Documentation](https://github.com/ispc/ispc/tree/main/docs/ispc.rst) updated if needed

PR #3586, opened 2025-10-07
LLVM update

aneshlya:
## Description
1. Use 21.1.3 version of LLVM when building LLVM 21.1
2. Add RISCV target to LLVM build targets required by https://github.com/ispc/ispc/pull/3578

## Related Issue

## Checklist


dbabokin:
21.1.3 👻

dbabokin (approved):
LGTM.

We should probably keep an eye on how we cook LLVM for the release, so we don't ship RISC-V backend if we don't ship RISC-V targets.

And if we do, then we need to keep an eye on the binary size.

But it's outside the scope of this PR.

Issue #3574, opened 2025-09-19
Make dispatcher more efficient


aneshlya:
Example of current ISPC dispatch
```
@__system_best_isa = internal unnamed_addr global i32 -1, align 4
...
declare i1 @foo_sse4(float, ptr)

declare i1 @foo_avx2(float, ptr)

define i1 @foo(float %0, ptr %1) {
entry:
  call void @__set_system_isa()
  %system_isa = load i32, ptr @__system_best_isa, align 4
  %isa_ok = icmp sge i32 %system_isa, 5
  br i1 %isa_ok, label %do_call, label %next_try

do_call:                                          ; preds = %entry
  %ret_value = call i1 @foo_avx2(float %0, ptr %1)
  ret i1 %ret_value

next_try:                                         ; preds = %entry
  %isa_ok1 = icmp sge i32 %system_isa, 2
  br i1 %isa_ok1, label %do_call2, label %next_try3

do_call2:                                         ; preds = %next_try
  %ret_value4 = call i1 @foo_sse4(float %0, ptr %1)
  ret i1 %ret_value4

next_try3:                                        ; preds = %next_try
  call void @__terminate_now()
  ret i1 undef
}
```
The more targets a given source is compiled for, the more bloated the logic for function selection. It would be good for ISPC to perform the `cpuid` check once, cache the target function address, and perform an indirect branch which should predict perfectly.

ivankelarev:
My proposal here is to create a global array of function pointers for each target ISA and another global pointer that will point to one of these global arrays depending on the hardware the program is executed on (initially uninitialized):

```
@__system_func_ptrs = internal global ptr null
@__system_func_ptrs_5 = internal constant [3 x ptr] [ptr @bar_avx2, ptr @foo_avx2, ptr @noise_ispc_avx2]
@__system_func_ptrs_3 = internal constant [3 x ptr] [ptr @bar_avx, ptr @foo_avx, ptr @noise_ispc_avx]
@__system_func_ptrs_2 = internal constant [3 x ptr] [ptr @bar_sse4, ptr @foo_sse4, ptr @noise_ispc_sse4]
```

In the dispatch function, we check this global pointer value: if it's already initialized, it points to the proper function pointers array and we can perform an indirect call using a unique offset in this array for each function. Otherwise, we initialize the pointer by selecting the best available ISA variant and then perform the indirect call:

```
declare void @foo_sse4(ptr, ptr)

declare void @foo_avx(ptr, ptr)

[...]

PR #3587, opened 2025-10-08
Make dispatcher more efficient

ivankelarev:
This patch introduces a caching mechanism in the generated dispatcher code. It also refactors the `__set_system_isa` function into `__get_system_best_isa` and enables LLVM optimization passes for the dispatcher module to eliminate unnecessary load instructions. The observed performance improvement for dispatcher code after applying this patch is approximately 50%.

Related issue: #3574


aneshlya:
Please squash the commits to one. Next time please create the branch in your fork and make PR from there.

dbabokin:
@codex review

ivankelarev:
> Please squash the commits to one. 

Done.

src/module.cpp
```diff
@@ -2005,12 +2003,45 @@ static void lCreateDispatchFunction(llvm::Module *module, llvm::Function *setISA
     }
     llvm::BasicBlock *bblock = llvm::BasicBlock::Create(*g->ctx, "entry", dispatchFunc);
 
-    // Start by calling out to the function that determines the system's
-    // ISA and sets __system_best_isa, if it hasn't been set yet.
-    llvm::CallInst::Create(setISAFunc, "", bblock);
+    // Create a global variable to cache the chosen function pointer.
+    auto *ptrTy = llvm::PointerType::getUnqual(*g->ctx);
+    auto *nullPtr = llvm::Constant::getNullValue(ptrTy);
+    auto *funcPtrCache = new llvm::GlobalVariable(*module, ptrTy, false, llvm::GlobalValue::InternalLinkage, nullPtr,
```
aneshlya, line 2013: If there are several functions in the module, I'd prefer to have "__system_func_ptr_cache_" + name.
Also please add second function to the lit test.

src/module.cpp
```diff
@@ -2005,12 +2003,45 @@ static void lCreateDispatchFunction(llvm::Module *module, llvm::Function *setISA
...
+                                                  "__system_func_ptr_cache");
+
+    // At the beginning, check whether we already have a callee address in cache.
+    llvm::IRBuilder builder(bblock);
+    auto *ptrF = builder.CreateLoad(ptrTy, funcPtrCache);
+    auto *cond = builder.CreateCmp(llvm::CmpInst::ICMP_NE, ptrF, nullPtr);
+
+    // Two successor basic blocks: make indirect call on cached value or determine the desired variant.
```
aneshlya, line 2021: Hmm, actually on my machine (12th Gen Intel(R) Core(TM) i9-12900) the indirect approach shows slower results using microbenchmark from here: https://github.com/ispc/ispc/issues/3574#issuecomment-3392220232.
Do you see the same?

aneshlya:
Overall, looks like in practice (at least on my machine) we're not getting expected better result. Let's approach the problem gradually. 
1. Caching of __set_system_isa()
2. Measure perf impact
3. Enabled LLVM optimization passes for the dispatcher module
4. Measure perf impact
5. Added function pointer caching mechanism in dispatcher code generation
6. Measure perf impact

if indirect calls brings regression, let's brainstorm what else can be done. May be switch jump table would perform better, e.g.
```
    switch i32 %system_isa, label %default [
      i32 0, label %call_sse2
      i32 1, label %call_sse2
      i32 2, label %call_sse4
      i32 3, label %call_avx
      i32 4, label %call_avx
      i32 5, label %call_avx2
      i32 6, label %call_avx2
      i32 7, label %call_avx2
      i32 8, label %call_avx512
      i32 9, label %call_avx512
      i32 10, label %call_avx512
    ]
```

src/module.cpp
ivankelarev: Done.

src/module.cpp
ivankelarev: No, I don't see the same results. My machine is quite different (Intel(R) Xeon(R) Platinum 8260M CPU @ 2.40GHz), but I consistently see dispatcher performance improvements of around 50% with my patch. It's surprising if you don't see any improvement at all. I've added a benchmark to this patch that, when built with `-DBENCHMARKS_ISPC_TARGETS=sse2-i32x4,sse4-i32x8,avx1-i32x8,avx2-i32x8,avx512skx-x16`, also shows performance gains on my machine.

src/module.cpp
```diff
@@ -2033,32 +2064,10 @@ static void lCreateDispatchFunction(llvm::Module *module, llvm::Function *setISA
...
-            llvm::ReturnInst::Create(*g->ctx, callBBlock);
-        } else {
-            llvm::CallInst *callInst = llvm::CallInst::Create(targetFuncs[i], args, "ret_value", callBBlock);
-            callInst->setCallingConv(targetFuncs[i]->getCallingConv());
-            llvm::ReturnInst::Create(*g->ctx, callInst, callBBlock);
-        }
+        // Store the chosen variant in the cache.
+        builder.SetInsertPoint(callBBlock);
+        builder.CreateStore(targetFuncs[i], funcPtrCache);
+        builder.CreateBr(bbCall);
```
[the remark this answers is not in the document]
aneshlya: Could you provide some comments why it is resolved? ISPC can be invoked in multithreaded environment, we don't have control over it.

aneshlya:
There are some linter issues that should be resolved: https://github.com/ispc/ispc/actions/runs/18539641836/job/52843585694?pr=3587

src/module.cpp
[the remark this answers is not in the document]
ivankelarev: Sure. The comment is correct but irrelevant. Yes, the proposed dispatcher implementation is not thread-safe, but the current implementation is not thread-safe either. So we are not making things worse here. Of course we can address this issue as well, but it should be a separate task.

aneshlya (approved):
LGTM. Maybe @dbabokin or @karthik-senthil can review as well.

ivankelarev:
Thanks @dbabokin for the feedback! I completely agree with your comments. I'm planning to address the thread safety issues in a follow-up patch.

Issue #389, opened 2012-09-30
soa<> can't be used inside structs or typedefs


Reedbeta:
Trying to use soa<> on a member of a struct doesn't work:

```
struct point { float x, y, z; };
struct pointbatch
{
    soa<8> point p[64];
};
```

gives the error `syntax error, unexpected 'soa'.       soa<8> point p[64];`  That prevents doing things like making an array struct with a length and a pointer to items in soa layout.

Also, the compiler allows defining a typedef for a soa<>, but then it behaves weirdly:

```
struct point { float x, y, z; };
typedef soa<8> point pointbatch[64];
export void test ()
{
    soa<8> point batch[64];
    foreach (xi = 0 ... 64)
    {
        batch[xi].x = (float)xi;
    }

    pointbatch batch2;
    foreach (xi = 0 ... 64)
    {
        batch2[xi].x = (float)xi;    // Generates performance warning: scatter
    }
}
```

In the first loop there is no performance warning, as expected; the soa layout makes all the writes to 'x' adjacent.  In the second loop, I'm not sure what it thinks the data layout is, but apparently it's not the same as in the first loop.  I found the struct problem when trying to work around this.

PR #3592, opened 2025-10-12
Legalize soa<> members in structs

thirtythreeforty:
## Description
This change legalizes `soa<>`-qualified members in structs.

The only change needed is to the grammar - the produced type is handled correctly by the rest of the compiler.

## Related Issue
- https://github.com/ispc/ispc/issues/389

## Checklist

### DIFF
--- a/.github/workflows/codex-pr-review.yml
+++ b/.github/workflows/codex-pr-review.yml
@@ -13,15 +13,20 @@ name: Codex PR Review
 
 # zizmor is not wrong, but we need to access token in the target repo for the PR to make it work.
 # The token is created with minimal scope possible.
+
+# Temporary disable pull_requst_target due to unclear security policies.
 on:
-  pull_request_target:  # zizmor: ignore[dangerous-triggers]
-    types: [opened, reopened]
+  #pull_requst_target:  # zizmor: ignore[dangerous-triggers]
+  #  types: [opened, reopened]
   workflow_dispatch:  # Manual trigger
     inputs:
       pr_number:
         description: 'PR number to comment on'
         required: true
         type: number
+
+permissions:
+  contents: read
 
 jobs:
   comment:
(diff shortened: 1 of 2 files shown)


thirtythreeforty:
I'm not exactly sure where to add a lit test for this - happy to take advice.

aneshlya:
Hi @thirtythreeforty, thanks for the reporting the issue and the proposed PR! However the CI is failing with the fix. Would you like to take care of it?
The lit tests reside in tests/lit-tests folder.

thirtythreeforty:
Sorry, I wasn't clear - I meant exactly what should the lit test look for. But I added a test that should require the type to get generated; if that's not what's needed, let me know.

I'm not sure how I missed the CI failures locally, but that problem does reproduce in a simple copy-paste test. Fixed by moving the soa<> parsing clause to `specifier_qualifier_list`. It makes me worried that I'm apparently not running the full test suite locally, but the func and lit tests do pass.

thirtythreeforty:
Getting close on CI - the lit test used `--target=host` which didn't work when ISPC chose a different width.

tests/lit-tests/389.ispc
```diff
@@ -0,0 +1,15 @@
...
+
+struct Point { float x, y, z; };
+struct SoaPoints {
+    soa<8> Point pts[10];
+};
+
+extern bar(SoaPoints&);
+
+// CHECK: %v16_varying_SoaPoints = type { [10 x %v16_soa8_Point] }
+// CHECK: %v16_soa8_Point = type { [8 x float], [8 x float], [8 x float] }
```
thirtythreeforty: @aneshlya, question: this fails on aarch64 targets (`Error: Invalid architecture "x86-64"`); what's the preferred way to ask for "a 16-wide target on host architecture"? `--target=generic-i32x16` isn't it.

Or, should I do separate `RUN`s for aarch64 and x86_64?

aneshlya: You can still use `--target=host` but make the test target independent like use `[{{[0-9]*}} x float]` and `%v{{[0-9]*}}_varying_SoaPoints`.

thirtythreeforty: Ah, thank you! Done.

aneshlya:
@codex review

PR #3602, opened 2025-10-17
Make dispatcher thread-safe

ivankelarev:
This is a follow-up change addressing feedback from the #3587 code review. The patch marks loads and stores to variables that may be shared across threads as atomic operations. While multiple threads may still execute code for selecting the optimal ISA or function variant, this change eliminates data races when accessing the corresponding shared variables. The patch is not expected to impact dispatcher performance.


aneshlya:
@codex review

ivankelarev:
> The changes look good to me, the codex remark would be good to address.

Done.

PR #3606, opened 2025-10-20
Fix internal error when casting SOA pointer to opaque pointer (#3591)

### LINKED ISSUE #3591: Internal error casting `soa<>` pointer to opaque pointer
The following reproducer:
```ispc
struct point {
    int x, y;
};

struct foo;

__attribute__((external_only))
export foo* uniform foo_init() {
    soa<8> point* uniform points = uniform new soa<8> point[123];
    return (foo* uniform)points;
}
```
causes the following crash:
```text
$ ispc --version
Intel(r) Implicit SPMD Program Compiler (Intel(r) ISPC), 1.28.2 (build commit 5bf8fb90b34122b5 @ 20250925, LLVM 20.1.8)
$ ispc test.ispc
Warning: No --target specified on command-line. Using default system target "avx2-i32x8". 
Invalid bitcast
  %17 = bitcast { ptr, i32 } %16 to ptr
Invalid bitcast
  %27 = bitcast { ptr, i32 } %26 to ptr
/usr/src/debug/ispc/ispc/src/main.cpp(40): FATAL ERROR: Unhandled signal sent to process; terminating.
***
***
Aborted                    (core dumped) ispc test.ispc
```

aneshlya:
When casting a slice/SOA pointer (e.g., soa<8> point* uniform) to an opaque pointer type (e.g., foo* uniform), the compiler was attempting to bitcast the entire slice pointer structure ({ptr, i32}) directly to a simple pointer (ptr), which is invalid in LLVM.

The fix extracts the pointer component from the slice structure before performing the bitcast, allowing the cast to succeed.

Fixes #3591

## Checklist

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -8025,10 +8025,21 @@ const Type *TypeCastExpr::GetLValueType() const {
     }
 }
 
+// Special helper to remove const from both pointer and pointee type recursively.
+// We can't directly use PointerType::GetAsNonConstType() here since
+// it does not recursively deconstify the base type - It only removes const from
+// the pointer itself, not from nested pointer types.
+// For example:
+//   Input: const int * const *
+//   GetAsNonConstType() gives: const int * * (only outer const removed)
+//   lDeconstifyType() should give: int * * (all const removed)
 static const Type *lDeconstifyType(const Type *t) {
     const PointerType *pt = CastType<PointerType>(t);
     if (pt != nullptr) {
-        return new PointerType(lDeconstifyType(pt->GetBaseType()), pt->GetVariability(), false);
+        // Preserve slice property when removing constness
+        unsigned int prop = pt->IsSlice() ? PointerType::SLICE : PointerType::NONE;
+        return new PointerType(lDeconstifyType(pt->GetBaseType()), pt->GetVariability(), false, prop,
+                               pt->GetAddressSpace());
     } else {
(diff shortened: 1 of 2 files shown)


aneshlya:
@codex review

aneshlya:
@codex review

src/expr.cpp
```diff
@@ -8028,7 +8028,13 @@ const Type *TypeCastExpr::GetLValueType() const {
 static const Type *lDeconstifyType(const Type *t) {
     const PointerType *pt = CastType<PointerType>(t);
     if (pt != nullptr) {
-        return new PointerType(lDeconstifyType(pt->GetBaseType()), pt->GetVariability(), false);
+        // Preserve slice property when removing constness
```
nurmukhametov, line 8039: IIRC, types system has `AsNonConstType` implemented for every type. Why do we need this local helper here instead of properly implemented method for PointetType (if the current one is not suitable)?

aneshlya: It's a good question. Looks like `PointerType, GetAsNonConstType()` removes the const qualifier from the pointer itself, not from what it points to.
`lDeconstifyType` removes `const` from what the pointer points to.

nurmukhametov: Right, so this function removes const from both pointer and pointee type for pointer types.

nurmukhametov (approved):
LGTM!

PR #3607, opened 2025-10-20
Fix missing IRBuilder.h include

ivankelarev:
Add missing IRBuilder.h include to fix compilation error with LLVM 21:
```
/home/runner/work/ispc/ispc/src/module.cpp:2017:11: error: no member named 'IRBuilder' in namespace 'llvm'; did you mean 'DIBuilder'?
 2017 |     llvm::IRBuilder builder(bblock);
      |           ^~~~~~~~~
```

### DIFF
--- a/src/module.cpp
+++ b/src/module.cpp
@@ -35,6 +35,7 @@
 #include <llvm/Bitcode/BitcodeWriter.h>
 #include <llvm/IR/DataLayout.h>
 #include <llvm/IR/DerivedTypes.h>
+#include <llvm/IR/IRBuilder.h>
 #include <llvm/IR/LegacyPassManager.h>
 #include <llvm/IR/Module.h>
 #include <llvm/IR/Type.h>


src/module.cpp
```diff
@@ -35,6 +35,7 @@
 #include <llvm/Bitcode/BitcodeWriter.h>
 #include <llvm/IR/DataLayout.h>
 #include <llvm/IR/DerivedTypes.h>
+#include <llvm/IR/IRBuilder.h>
```
aneshlya, line 38: I see that it's still a draft but just a reminder - if the fix is related to some particular version of LLVM it's better to surround it with macros to avoid problems with previous version. For example:
```
#if ISPC_LLVM_VERSION >= ISPC_LLVM_21_0
```

ivankelarev: Thanks @aneshlya for the feedback! I don't think we need version checks here - this is just fixing a missing include that should have been there anyway. Including `IRBuilder.h` when using `IRBuilder` is the right thing to do regardless of LLVM version.

Issue #3608, opened 2025-10-20
Refactor stdlib linkage to be target-independent


aneshlya:
The stdlib is linked once per target and is not hierarchical. In fact, the stdlib files for different targets are almost identical (the only differences are the mask and a few minor details. As a result adding a new target results in ~2mb size increase of final ispc executable.

We need to properly decouple the target from stdlib.
Start by analyzing the differences between stdlib implementations that share the same width and mask but belong to different targets.

PR #3617, opened 2025-11-04
Refactor target capabilities

aneshlya:
## Description
This PR refactors capability handling to eliminate duplication and establish a single source of truth for target capabilities.
It also moves target-specific capabilities from `target.isph` to `core.isph` delaying capability resolution from stdlib compilation to  user's code compilation. It makes stdlib independent from target capabilities and allows to easier reduce binary size.

## Related Issue
Related to #3608 

## Checklist

### DIFF
--- a/stdlib/include/core.isph
+++ b/stdlib/include/core.isph
@@ -215,56 +215,55 @@ static const int32 programIndex = {ISPC_PROGRAM_INDEX_INITIALIZER};
 
 // X-Macro for defining target capability constants
 // These correspond to hardware/ISA capabilities defined in src/target_capabilities.h
-#define ISPC_CAPABILITY_CONSTANTS                                                                                      \
-    ISPC_CAPABILITY_CONSTANT(__have_arm_dot_product, ISPC_TARGET_HAS_ARM_DOT_PRODUCT_VAL)                              \
-    ISPC_CAPABILITY_CONSTANT(__have_arm_i8mm, ISPC_TARGET_HAS_ARM_I8MM_VAL)                                            \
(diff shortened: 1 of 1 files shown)


aneshlya:
@codex review

aneshlya:
@codex review

stdlib/stdlib.ispc
```diff
@@ -520,128 +506,114 @@ template <typename T> unmasked varying T __shift_via_shuffle(varying T v, unifor
...
-
 __declspec(safe) static inline float shift(float v, uniform int i) {
-#ifdef USE_SHIFT_VIA_SHUFFLE_32
-    return __shift_via_shuffle(v, i);
-#else
-    varying float result;
-    unmasked { result = __shift_float(v, i); }
-    return result;
-#endif
+    if (__have_shift_via_shuffle_32) {
```
nurmukhametov, line 510: It doesn't feel right to me to have `__have_shift_via_shuffle_32` treated as something similar to a capability. It seems to me it would be better to move these implementations, along with the conditional logic, into the generic builtins implementation.

aneshlya: I agree that `__have_shift_via_shuffle_32` is not a capability.

Do you mean do something like this? https://github.com/aneshlya/ispc/commit/ab765afec4d09031421717542e4404dbd53c17bd
I'm not sure how to make it work because when we compile `generic.ispc` ISPC_TARGET_ macro is not set.

nurmukhametov: > I agree that `__have_shift_via_shuffle_32` is not a capability.
> 
> Do you mean do something like this? [aneshlya@ab765af](https://github.com/aneshlya/ispc/commit/ab765afec4d09031421717542e4404dbd53c17bd) I'm not sure how to make it work because when we compile `generic.ispc` ISPC_TARGET_ macro is not set.

Right, I completely missed that.

PR #3531, opened 2025-08-08
Fix or ignore some nightly CI issues

nurmukhametov:
## Description

- CI: ignore for now --opt=fast-math for macos-test-flags
- func-tests: skip round-float16-*.ispc for all targets on Windows
- ldexp-float16 func-test: reduce input range to avoid float16 overflow (happened for x64 targets).

Full test run: https://github.com/ispc/ispc/actions/runs/16835313082

## Related Issue
- [X] Linked to relevant issue: #3530 

## Checklist
- [X] Code has been formatted with `clang-format` (e.g., `clang-format -i src/ispc.cpp`)
- [X] Git history has been squashed to meaningful commits (one commit per logical change)

Issue #3529, opened 2025-08-08
round-float16 func-tests compilation error on pre-avx2 targets on Windows


nurmukhametov:
The call to roundevenf is generated that is not resolved.

```bash
Compilation of test .\tests\func-tests\round-float16-uniform.ispc failed            
test_static.cpp
round-float16-uniform.ispc.obj : error LNK2019: unresolved external symbol roundevenf referenced in function f_f___un_3C_unf_3E_un_3C_unf_3E_
round-float16-uniform.ispc.exe : fatal error LNK1120: 1 unresolved externals
Compilation of test .\tests\func-tests\round-float16-varying.ispc failed            
test_static.cpp
round-float16-varying.ispc.obj : error LNK2019: unresolved external symbol roundevenf referenced in function f_f___un_3C_unf_3E_un_3C_unf_3E_
round-float16-varying.ispc.exe : fatal error LNK1120: 1 unresolved externals
.....
+.\tests\func-tests\round-float16-uniform.ispc compfail  x86-64     sse2-i32x4 unspec Windows LLVM 18.1         cl O2 obj *
+.\tests\func-tests\round-float16-varying.ispc compfail  x86-64     sse2-i32x4 unspec Windows LLVM 18.1         cl O2 obj *
+.\tests\func-tests\round-float16-uniform.ispc compfail  x86-64     sse2-i32x4 unspec Windows LLVM 18.1         cl O1 obj *
+.\tests\func-tests\round-float16-varying.ispc compfail  x86-64     sse2-i32x4 unspec Windows LLVM 18.1         cl O1 obj *
+.\tests\func-tests\round-float16-uniform.ispc compfail  x86-64     sse2-i32x4 unspec Windows LLVM 18.1         cl O0 obj *
+.\tests\func-tests\round-float16-varying.ispc compfail  x86-64     sse2-i32x4 unspec Windows LLVM 18.1         cl O0 obj *
```

https://github.com/ispc/ispc/actions/runs/16824259671

alexfmpe:
Seems like this also happens on x86-64 mac.

PR #3623, opened 2025-11-06
func-tests: skip round-float16-*.ispc for all targets on x86_64 mac

alexfmpe:
## Description

I attempted to replicate https://github.com/ispc/ispc/pull/3531/commits/3d052870b3dc3c7553329cd6b9ccc4db2228e030 for x86_64-darwin to get around https://github.com/ispc/ispc/issues/3529, but couldn't find a way to match against arch/os in one go. Tried De Morgan's laws via multiple `run on` but that also doesn't seem to work?

## Related Issue

## Checklist

### DIFF
--- a/tests/func-tests/round-float16-uniform.ispc
+++ b/tests/func-tests/round-float16-uniform.ispc
@@ -1,6 +1,8 @@
 #include "test_static.isph"
 
 // See issue #3529
+// rule: run on OS=!darwin
+// rule: run on arch!=x86_64
 // rule: skip on OS=windows
 
 task void f_f(uniform float RET[], uniform float aFOO[]) {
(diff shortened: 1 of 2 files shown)


aneshlya:
I'm OK if you skip this test entirely on macOS, you can use `rule: skip on OS=mac`.

alexfmpe:
Considering Apple will soon drop x86_64 support, I'm reluctant to disable aarch64 for its sake.
Is there no way to disable on x86_64-mac specifically? 
For now we're applying the patch to skip the test only on x86_64-mac on nixpkgs.

nurmukhametov:
We can probably add the specific failing entry to fail_db.txt.

aneshlya:
We should emulate the rounding functionality on sse2 targets without using llvm intrinsics. I'm working on the fix.

aneshlya:
This PR should fix the issue: https://github.com/ispc/ispc/pull/3629

aneshlya:
Should be fixed by https://github.com/ispc/ispc/pull/3629. I'm closing this PR now.

PR #3625, opened 2025-11-07
Enable 2334_unroll.ispc test and fix RISC-V testing in CI

aneshlya:
## Description
LLVM 20.1 and 21.1 were rebuilt with required patch and updated in CI. The lit test 2334_unroll.ispc can be enabled now.
The freshly rebuilt LLVM also includes RISC-V target now so this PR includes some fixes for RISC-V testing as well.

## Related Issue

## Checklist

### DIFF
--- a/.github/workflows/riscv.yml
+++ b/.github/workflows/riscv.yml
@@ -18,6 +18,7 @@ jobs:
     name: RISC-V rvv-x4 test
     runs-on: ubuntu-22.04
     container:
+      # archlinux allows to use qemu 10.1.2 which has support of float16 in RISC-V
       image: archlinux@sha256:2fcdd55448255f87dd337ef5c0fdbd5e4fec432345db1aa5faea4bf2764b4ca8
       options: --privileged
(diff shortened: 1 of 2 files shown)


ivankelarev (approved):
Thanks for the patch Arina! LGTM.

aneshlya:
@codex review

.github/workflows/riscv.yml
```diff
@@ -36,29 +30,25 @@ jobs:
         pacman -S --needed --noconfirm llvm clang
```
nurmukhametov: I think you don't need to install clang and llvm anymore, and use the arch container for this job.

aneshlya: Updated

aneshlya: There is no `float16` support in qemu version (v8.2) for ubuntu 24.04. I'd like to keep testing inside arch linux since it allows usage of the newer qemu with `float16`.

PR #3631, opened 2025-11-12
Replace undef with poison in @__shuffle2_const_* builtins to eliminate redundant shufflevector instructions

ivankelarev:
A simple `shuffle` in the source code can unexpectedly blow up into a mess of redundant `shufflevector` instructions, killing performance. The culprit is InstCombine's inability to properly optimize `undef` values in certain scenarios, where it prefers `poison` values instead (see this change: https://github.com/llvm/llvm-project/commit/8f1c984325bd679b2634a6173db69548da87ac71). To mitigate this, `undef` value in `@_shuffle2_const*` builtin functions has been replaced with `poison` value. Eventually, other `undef`s should be considered for replacement with `poison`s to prevent similar performance pitfalls.

### DIFF
--- a/builtins/util.m4
+++ b/builtins/util.m4
@@ -2213,7 +2213,7 @@ forloop(i, 0, eval(WIDTH-1), `
 forloop(i, 0, eval(WIDTH-1), `
   %v_`'i = extractelement <eval(2*WIDTH) x $1> %v2, i32 %index_`'i')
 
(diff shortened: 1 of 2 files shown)


nurmukhametov (approved):
LGTM!

ivankelarev:
> Looks good! Did you have a chance to make any perf measurements on GameDev tests?

Yes, I tested one of the most severely regressed cases. The results are very close to pre-regression performance, actually slightly better. This fix seems to resolve the issue.

PR #3632, opened 2025-11-12
Optimize stdlib compilation

### LINKED ISSUE #3608: Refactor stdlib linkage to be target-independent

aneshlya:
## Description
This PR optimizes stdlib compilation by implementing a width family system that reduces bitcode duplication. Targets with the same vector width that produce identical stdlib bitcode are grouped into families and share a single stdlib, reducing binary size by ~40% (334MB → 250MB) and build times.
Currently the optimization is only performed for x86 targets but can be extended for other targets as well.

## Related Issue
Fixes #3608

## Checklist

### DIFF
--- a/cmake/StdlibFamilies.cmake
+++ b/cmake/StdlibFamilies.cmake
@@ -28,15 +28,30 @@ function(define_stdlib_families)
     set(all_families "")
     set(all_members "")
 
+    # Define AVX10 target suffixes if LLVM supports them (>= 20.1.2)
+    if (${LLVM_VERSION_NUMBER} VERSION_GREATER_EQUAL "20.1.2")
+        set(AVX10_X4 ",avx10_2-x4")
+        set(AVX10_X8 ",avx10_2-x8")
+        set(AVX10_X16 ",avx10_2-x16")
+        set(AVX10_X32 ",avx10_2-x32")
+        set(AVX10_X64 ",avx10_2-x64")
+    else()
+        set(AVX10_X4 "")
+        set(AVX10_X8 "")
+        set(AVX10_X16 "")
+        set(AVX10_X32 "")
+        set(AVX10_X64 "")
+    endif()
+
     # Single declarative list of all families
     # Note: nozmm targets are excluded - they produce different stdlib that avoids ZMM instructions
     set(FAMILY_DEFINITIONS
         # AVX512 families - use generic targets
-        "i1x4:generic-i1x4:avx512skx-x4,avx512icl-x4,avx512spr-x4"
-        "i1x8:generic-i1x8:avx512skx-x8,avx512icl-x8,avx512spr-x8"
-        "i1x16:generic-i1x16:avx512skx-x16,avx512icl-x16,avx512spr-x16"
-        "i1x32:generic-i1x32:avx512skx-x32,avx512icl-x32,avx512spr-x32"
-        "i1x64:generic-i1x64:avx512skx-x64,avx512icl-x64,avx512spr-x64"
+        "i1x4:generic-i1x4:avx512skx-x4,avx512icl-x4,avx512spr-x4${AVX10_X4}"
+        "i1x8:generic-i1x8:avx512skx-x8,avx512icl-x8,avx512spr-x8${AVX10_X8}"
+        "i1x16:generic-i1x16:avx512skx-x16,avx512icl-x16,avx512spr-x16${AVX10_X16}"
+        "i1x32:generic-i1x32:avx512skx-x32,avx512icl-x32,avx512spr-x32${AVX10_X32}"
+        "i1x64:generic-i1x64:avx512skx-x64,avx512icl-x64,avx512spr-x64${AVX10_X64}"
 
         # SSE/AVX families that match generic implementation
         "i32x4:generic-i32x4:sse2-i32x4,sse4-i32x4"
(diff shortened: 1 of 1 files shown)


aneshlya:
@codex review

aneshlya:
@claude review

aneshlya:
@codex review

cmake/StdlibFamilies.cmake
```diff
@@ -0,0 +1,178 @@
...
+        # These produce stdlib different from generic but identical to parent ISA
+        "avx_i32x8:avx1-i32x8:avx2-i32x8,avx2vnni-i32x8"
+        "avx_i32x4:avx2-i32x4:avx2vnni-i32x4"
+    )
+
+    # Add AVX10 targets to families if LLVM supports them (>= 20.1.2)
+    if (${LLVM_VERSION_NUMBER} VERSION_GREATER_EQUAL "20.1.2")
+        # Append AVX10 targets to existing AVX512 families
+        list(APPEND FAMILY_DEFINITIONS
+            "i1x4_avx10:generic-i1x4:avx10_2-x4"
```
nurmukhametov: Should we just append it to the previously declared `"i1x4:generic-i1x4:avx512skx-x4,avx512icl-x4,avx512spr-x4"`? It seems to be from the same family for me.

cmake/StdlibFamilies.cmake
```diff
@@ -0,0 +1,178 @@
...
+            "i1x8_avx10:generic-i1x8:avx10_2-x8"
+            "i1x16_avx10:generic-i1x16:avx10_2-x16"
+            "i1x32_avx10:generic-i1x32:avx10_2-x32"
+            "i1x64_avx10:generic-i1x64:avx10_2-x64"
+        )
+    endif()
+
+    # Parsing lop
```
nurmukhametov: Type: `loop`?

docs/design/TargetRedesign.md
```diff
@@ -83,6 +83,52 @@ The compilation process for ISPC programs, as outlined, necessitates adjustments
...
+### Benefits
+
+- **Reduced Binary Size**: ~40% reduction in stdlib bitcode files (334mb -> 250mb)
+- **Faster Build Times**: Fewer stdlib compilations during ISPC build
+- **Maintainability**: Single source of truth for family definitions
+- **Correctness**: Automated validation ensures family members are consistent
+
+### Key Design Principle
+
+Stdlib must be capability-neutral. All target-specific capability constants (`__have_intel_vnni`, `__have_native_half_converts`, etc.) are declared without initialization during stdlib compilation via `ISPC_INTERNAL_STDLIB_COMPILATION`, causing them to default to 0. This ensures all targets in the same width family produce identical stdlib bitcode, regardless of their ISA-specific capabilities. The capabilities are resolved later during user code compilation.
```
nurmukhametov: > `causing them to default to 0`

I don't think they are `0`. They should be just symbols without the real value at this point (externals).

src/target_registry.cpp
```diff
@@ -217,8 +242,48 @@ const BitcodeLib *TargetLibRegistry::getISPCTargetLib(ISPCTarget target, TargetO
     return lGetTargetLib(m_targets, target, os, arch);
 }
 
+// Include auto-generated stdlib target map from CMake
+// This map is generated by cmake/StdlibFamilies.cmake based on X86_TARGETS
+// to ensure consistency between build system and runtime lookup
+#include "stdlib_target_map_generated.h"
+
+// Initialize the stdlib target map (generated from CMake)
+static std::unordered_map<ISPCTarget, ISPCTarget> stdlibTargetMap = createStdlibTargetMap();
```
nurmukhametov: Do we really need to define a function to create the map? IMO, it is better to have in the included header the direct declaration of the map.

nurmukhametov (approved):
In my setup, it is 122s -> 96s (21% faster, build), 239M -> 153M (36% smaller). Looks great!

cmake/StdlibFamilies.cmake
aneshlya: You're right, it is the same family but It's made like this for convenience. List entries in CMake are atomic strings - you can’t directly “edit” one entry in place. So to avoid extracting, modifying, and replacing it, it's just easier to create separate family. I'll add a comment explaining it.

nurmukhametov: Can we just define, e.g., AVX_TARGET_X4 =",avx10_2-x4" if LLVM >= 20.1.2 "" (sorry for pseudo python code instead of proper CMake) and use it like `"i1x4:generic-i1x4:avx512skx-x4,avx512icl-x4,avx512spr-x4${AVX_TARGET_X4}"?

In the current state, this extra family exists only in CMake files, right?

cmake/CommonStdlibBuiltins.cmake
```diff
@@ -7,6 +7,25 @@
 # ispc CommonStdlibBuiltins.cmake
 #
 
+# Check if target should be skipped for the given OS/bit combination
+# Sets out_skip to TRUE if target should be skipped, FALSE otherwise
+function(should_skip_target_for_os target os bit out_skip)
+    set(skip FALSE)
+
+    if ("${os}" STREQUAL "unix" AND APPLE AND NOT ISPC_LINUX_TARGET)
```
dbabokin, line 15: The condition is built in the way that it's difficult to interpret the intent.

dbabokin (approved):
Looks good!

cmake/StdlibFamilies.cmake
aneshlya: Oh, it's a good idea!

aneshlya: Updated

PR #3633, opened 2025-11-13
Create explicit target to ensure parser generation completes before compilation

aneshlya:
This PR attempts to fix rarely happening issues while building the parser in CI: https://github.com/ispc/ispc/actions/runs/19319277766/job/55257126462?pr=3632
The possible reason is that there is implicit dependency race condition: while CMake should automatically create dependencies from `add_custom_command` OUTPUT to consuming targets, both frontend and frontend-lib can be built in parallel without explicit serialization
    - Bison may still be writing the file
    - The file handle hasn't been closed/flushed
    - The compiler reads a partial file with an unterminated comment
This PR creates an explicit custom target for parser generation. This ensures a proper happens-before relationship and eliminates the race condition.

## Related Issue

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -349,6 +349,8 @@ find_package(BISON 3.0 REQUIRED)
             COMMENT "Generating parse.cc"
             WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
     )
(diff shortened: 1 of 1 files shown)


nurmukhametov (commented):
I have seen same error on Linux arm runner. Thank you for the fix! LGTM!

Issue #3081, opened 2024-10-24
Sub-par code generation when using extract


emoon:
Hi,

I have the following code

```C
unmasked void draw_rects(
    uniform float output[], 
    uniform float rect_min_x[],
    uniform float rect_min_y[],
    uniform float rect_max_x[],
    uniform float rect_max_y[],
    uniform int count) 
{
    for (uniform int i = 0; i < count; i += programCount) {
        float min_x = rect_min_x[i + programIndex];
        float min_y = rect_min_y[i + programIndex];
        float max_x = rect_max_x[i + programIndex];
        float max_y = rect_max_y[i + programIndex];

        int min_xi = floor(min_x);
        int min_yi = floor(min_y);
        int max_xi = floor(max_x);
        int max_yi = floor(max_y);
            
        for (uniform int pi = 0; pi < programCount; ++pi)
        {
            uniform int y0 = extract(min_yi, pi);
            uniform int y1 = extract(max_yi, pi);
            uniform int x0 = extract(min_xi, pi);
            uniform int x1 = extract(max_xi, pi);

            for (uniform int y = y0; y < y1; y++) {
                for (uniform int x = x0; x < x1; x += programCount) {
                    uniform int32 index = (y * 128) + x;
                    output[index + programIndex] = 10.0; 
                }
            }
        }
    }
}
```
https://godbolt.org/z/8cTdnz4j9 this makes the compiler generate some stores/loads to the stack
```
        movaps  xmmword ptr [rsp - 24], xmm2
        mov     eax, r10d
        and     eax, 3
        cvttss2si       r12d, dword ptr [rsp + 4*rax - 24]
        movaps  xmmword ptr [rsp - 40], xmm4
        cvttss2si       r11d, dword ptr [rsp + 4*rax - 40]
        movaps  xmmword ptr [rsp - 56], xmm1
        cvttss2si       r15d, dword ptr [rsp + 4*rax - 56]
        movaps  xmmword ptr [rsp - 72], xmm3
```
If I change the code to look like this
```C
nmasked void rasterize_rect(uniform float output[],  uniform int x0,  uniform int y0,  uniform int x1,  uniform int y1)  {
    for (uniform int y = y0; y < y1; y++) {
        for (uniform int x = x0; x < x1; x += programCount) {
            uniform int32 index = (y * 128) + x;
            output[index + programIndex] = 10.0; 
        }
    }
 }

unmasked void draw_rects(
    uniform float output[], 
[...]

PR #3637, opened 2025-11-16
Restrict the canonicalization of extractelement(cast)

azwolski:
## Description
Add an LLVM patch to restrict the canonicalization of extractelement(cast) to cases where the index is constant or both instructions are in the same basic block. 

Lit test to the issue: 

There is also a opensource LLVM PR under the review: https://github.com/llvm/llvm-project/pull/166227

## Related Issue
This is fix for [Sub-par code generation when using extract issue](https://github.com/ispc/ispc/issues/3081)

## Checklist


llvm_patches/20_1_21_1_limit-canonicalization.patch
```diff
@@ -0,0 +1,21 @@
+diff --git a/llvm/lib/Transforms/InstCombine/InstCombineVectorOps.cpp b/llvm/lib/Transforms/InstCombine/InstCombineVectorOps.cpp
```
aneshlya, line 2: Could you add a comment at the beginning of the patch like
```
# This patch is backport of the https://github.com/llvm/llvm-project/pull/166227
```

aneshlya:
This patch will be only applied to LLVM 20, not LLVM 21. I don't remember if we discussed it. is it intentional?

aneshlya:
Thanks, looks good!

PR #3652, opened 2025-12-02
Rename avx10.2 targets to avx10.2dmr

### LINKED ISSUE #3647: Update avx10.2 targets
Recently `novalake` cpu was added to LLVM trunk: https://github.com/llvm/llvm-project/blob/778e104dee8186c8a8c8ed781692825fb3a8bec8/llvm/lib/TargetParser/X86TargetParser.cpp#L178

Even though the `novalake` (client) and `diamondrapids` (server, already exposed in ispc) are both `avx10.2`, they differ in some features, at least AMX and `avx512_vp2intersect` so we need to differentiate them on ISPC target level.

What needs to be done:
1. Rename current `avx10.2` targets to `avx10.2dmr` targets similar to `avx512skl`/`avx512spr`
2. Add `avx10.2nvl` targets starting LLVM version 22.

aneshlya:
## Description
DMR and NVL targets differs due to AMX (DMR has it and NVL hasn't). We should differentiate and properly direct novalake HW in ISPC. 
This PR makes step 1: renames existing avx10.2 targets to avx10.2dmr and makes necessary preparations for easy addition of nvl targets.

Partially fixes #3647 

## Checklist


aneshlya:
@codex review

ivankelarev:
I have a general question regarding backward compatibility. If any current ISPC users are already utilizing the `avx10.2` target in their code, won't this change break their builds?
Should we consider maintaining the existing `avx10.2` target as a deprecated alias (perhaps mapping it to `avx10.2dmr` for now) and only remove it in a future major release? This would give users time to migrate their code and avoid immediate breaking changes.

aneshlya:
> I have a general question regarding backward compatibility. If any current ISPC users are already utilizing the `avx10.2` target in their code, won't this change break their builds?

Since the HW is not available yet it's probably the last release where we can do such change without breaking the user's code. Otherwise we would fully support backward compatibility.

ivankelarev (approved):
@aneshlya That makes sense, thank you for the clarification! LGTM.

dbabokin (approved):
LGTM

PR #3660, opened 2025-12-04
Add avx10.2nvl family of targets

### LINKED ISSUE #3685: Expose the family of avx10.2nvl targets corresponding to Novalake HW
`avx10.2nvl` is almost identical to `avx10.2dmr` except AMX support.

aneshlya:
## Description
Adds the avx10.2nvl-x{4,8,16,32,64} x86 target family for Intel Nova Lake (NVL).

New NVL target family
  - NVL implements AVX10.2 but, unlike Diamond Rapids (DMR), has no AMX. It's modeled as "DMR without AMX": the target hierarchy inherits from ICL (avx512icl) rather than DMR/GNR, and no AMX capabilities (AmxTile/Int8/Bf16/Fp16) are advertised.
  - Half-precision rcp/rsqrt builtins are reused from the shared target-avx512fp16-x*-common.ll files (extracted out of the SPR builtins as part of this work).
  - Docs (docs/ispc.rst), CPU/ISA plumbing, check_isa, and alloy.py updated for NVL.

Dispatcher AMX-capability fix
  - The runtime dispatcher picked a variant via a linear ICMP_SGE walk over Target::ISA, assuming each ISA is a superset of all lower-numbered ones. NVL breaks that: it sorts above the AMX-bearing SPR/GNR tiers (it has AVX10.2) but lacks AMX, so an NVL system would mis-select an SPR/GNR variant and fault.
  - Replaces the prior NVL-specific blocklist with a capability requirement: the AMX server tiers (SPR/GNR/DMR) now require an orthogonal "system has AMX" check via a new __system_has_amx() query. Any future non-AMX AVX10.2 client is handled with no further dispatcher changes.

## Related Issue

## Checklist


dbabokin:
@codex review

dbabokin:
@codex review

dbabokin:
@codex review

dbabokin:
@codex review

src/ispc.cpp
```diff
@@ -257,6 +257,13 @@ static ISPCTarget lGetSystemISA() {
         return ISPCTarget::avx512spr_x16;
     case Target::ISA::GNR_AVX512:
         return ISPCTarget::avx512gnr_x16;
+    case Target::ISA::NVL_AVX10_2:
+        // Return ICL target for LLVM versions < 22.0
```
azwolski: i dont understand why

aneshlya: On LLVM < 22 we can't emit an NVL target (the target isn't available in the backend yet), but lGetSystemISA() can still detect NVL hardware. Since NVL is a superset of ICL — full AVX-512, just without AMX — falling back to the ICL target produces code that runs correctly on NVL. I've expanded the comment to say this.

dbabokin:
@codex review

dbabokin:
I found one dispatch-mechanism issue worth fixing before merge.

The new AMX guard is intended to answer "can this process safely run an AMX-bearing variant?", but `__system_has_amx()` currently only checks CPUID AMX feature bits via `get_x86_has_amx()`:

https://github.com/ispc/ispc/blob/5d5e8ec5ead4d9ec12b04c4c5550f8750484fbf7/src/isa.h#L284-L291

That is not quite in line with existing dispatcher practice. The dispatcher already uses `get_x86_isa()` from `builtins/dispatch.c`, and that path gates AVX/AVX-512 on OS XSAVE state, not just CPUID:

https://github.com/ispc/ispc/blob/5d5e8ec5ead4d9ec12b04c4c5550f8750484fbf7/src/isa.h#L85-L91
https://github.com/ispc/ispc/blob/5d5e8ec5ead4d9ec12b04c4c5550f8750484fbf7/src/isa.h#L153
https://github.com/ispc/ispc/blob/5d5e8ec5ead4d9ec12b04c4c5550f8750484fbf7/src/isa.h#L248

There is also repo-local precedent for AMX specifically: `tools/check_isa.cpp` reports AMX on/off using XCR0 tile-state bits (`xgetbv() & 0x60000`).

https://github.com/ispc/ispc/blob/5d5e8ec5ead4d9ec12b04c4c5550f8750484fbf7/tools/check_isa.cpp#L22-L33

So I think the dispatcher-side AMX predicate should include the OS AMX state as well, at least on non-macOS. Otherwise a process can dispatch to an SPR/GNR/DMR variant on a machine that advertises AMX in CPUID but has not enabled AMX tile state, and then fault when the selected variant executes AMX instructions.

One caveat: the dispatch cache means AMX permission needs to be requested before the first dispatched call. That matches the current test/example pattern, which requests `ARCH_REQ_XCOMP_PERM` before running AMX code.

aneshlya:
The review comment has been addressed. 
The job [Quick Start Build / windows (pull_request)](https://github.com/ispc/ispc/actions/runs/27480300344/job/81474490401?pr=3660) fail is not related to the changes in this branch, it happens on main as well, and is probably caused by switching to newer version of VS2026 in github runners.

dbabokin (approved):
Looks good!

PR #3702, opened 2026-01-16
Reapply struct support in lIsSafeToBlend with blend target size validation

ivankelarev:
This reapplies commit 11fc9cd6f7a5daaf3f602e4d41ea6eeb3a94a8ac which was reverted in 559b397216d79e001d769e5fced71ef007bc475b due to stability regressions. 

The original change added support for structures in `lIsSafeToBlend` but caused issues when blend operations were applied to data larger than the blend target type. Such oversized stores can be treated as undefined behavior and optimized away by the compiler. This reapplication adds blend target size validation to prevent blend operations when the stored data exceeds the target vector type capacity. 

This patch resolves issue #3358.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1024,7 +1024,7 @@ configure_ispc_obj(builtin)
 add_executable(ispc-opt tools/ispc-opt.cpp)
 configure_ispc_exe(ispc-opt)
 # TODO! should we link against all the same LLVM/Clang libraries? Clang libs is definetely not needed
-target_link_libraries(ispc-opt ${LINK_LIBRARIES} builtin optimization common)
+target_link_libraries(ispc-opt ${LINK_LIBRARIES} optimization common)
 
 # The final target has to be ispc due to two use-cases we do not want to
 # interfere with: make ispc and $<TARGET_FILE:ispc>.
--- a/src/ispc.cpp
+++ b/src/ispc.cpp
@@ -2177,7 +2177,7 @@ Target::Target(Arch arch, const char *cpu, ISPCTarget ispc_target, PICLevel picL
         llvm::Triple triple = GetTriple();
 
         // The last validity check to ensure that supported for this target was enabled in the build.
-        if (!g->target_registry->isSupported(m_ispc_target, g->target_os, arch)) {
+        if (!g->disableTargetValidation && !g->target_registry->isSupported(m_ispc_target, g->target_os, arch)) {
             std::string target_string = ISPCTargetToString(m_ispc_target);
             std::string arch_str = ArchToString(arch);
             std::string os_str = OSToString(g->target_os);
@@ -3175,6 +3175,7 @@ Globals::Globals() {
     includeFloat16Conversions = false;
 
     enableTimeTrace = false;
+    disableTargetValidation = false;
     // set default granularity to 500.
     timeTraceGranularity = 500;
     target = nullptr;
--- a/src/ispc.h
+++ b/src/ispc.h
@@ -1,5 +1,5 @@
 /*
-  Copyright (c) 2010-2025, Intel Corporation
+  Copyright (c) 2010-2026, Intel Corporation
 
   SPDX-License-Identifier: BSD-3-Clause
 */
@@ -917,6 +917,9 @@ struct Globals {
 
     /* When true, enable compile time tracing. */
     bool enableTimeTrace;
+
+    /* If true, disable validation of target support in Target constructor. */
+    bool disableTargetValidation;
 
     /* When compile time tracing is enabled, set time granularity. */
     int timeTraceGranularity;
--- a/src/target_registry.cpp
+++ b/src/target_registry.cpp
@@ -1,5 +1,5 @@
 /*
-  Copyright (c) 2019-2025, Intel Corporation
+  Copyright (c) 2019-2026, Intel Corporation
 
   SPDX-License-Identifier: BSD-3-Clause
 */
@@ -68,6 +68,9 @@ TargetLibRegistry::TargetLibRegistry() {
     // TODO: check for conflicts / duplicates.
     m_dispatch = nullptr;
     m_dispatch_macos = nullptr;
+    if (libs == nullptr) {
+        return;
+    }
     for (auto lib : *libs) {
         switch (lib->getType()) {
         case BitcodeLib::BitcodeLibType::Dispatch:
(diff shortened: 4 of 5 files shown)


src/opt/ReplacePseudoMemoryOps.cpp
```diff
@@ -13,36 +13,112 @@ namespace ispc {
...
-    false for memory that actually is stack allocated.  The basic strategy
-    is to traverse through the operands and see if the pointer originally
-    comes from an AllocaInst.
+/** Helper function to return a FixedVectorType if the type (after unwrapping arrays)
+    is a vector, or nullptr otherwise.
 */
-static bool lIsSafeToBlend(llvm::Value *lvalue) {
-    llvm::BitCastInst *bc = llvm::dyn_cast<llvm::BitCastInst>(lvalue);
-    if (bc != nullptr) {
+static llvm::FixedVectorType *lIsVectorMatchingTargetWidth(llvm::Type *type) {
```
aneshlya: Since the function is not returning `bool` anymore, I suggest renaming it to `lGetUnderlyingVectorType` or something similar.

src/opt/ReplacePseudoMemoryOps.cpp
```diff
@@ -13,36 +13,112 @@ namespace ispc {
...
+    llvm::Type *fieldType = st->getElementType(fieldIndex);
+
+    // If this field is a nested struct, recurse
+    if (llvm::StructType *nestedSt = llvm::dyn_cast<llvm::StructType>(fieldType)) {
+        uint64_t fieldOffset = SL->getElementOffset(fieldIndex).getFixedValue();
+        uint64_t offsetWithinField = targetOffset - fieldOffset;
+        return lCheckStructFieldAtOffset(nestedSt, offsetWithinField, DL);
+    }
+
+    // Leaf field - check if it contains matching vectors
```
aneshlya: Comment update needed

src/opt/ReplacePseudoMemoryOps.cpp
```diff
@@ -79,10 +155,16 @@ static bool lReplacePseudoMaskedStore(llvm::CallInst *callInst) {
...
-    // or serializing the masked store.  Even on targets with a native
+    // or serializing the masked store. Even on targets with a native
     // masked store instruction, this is preferable since it lets us
     // keep values in registers rather than going out to the stack.
-    bool doBlend = (!g->opt.disableBlendedMaskedStores && lIsSafeToBlend(lvalue));
+    bool doBlend = false;
+    if (!g->opt.disableBlendedMaskedStores) {
+        if (llvm::FixedVectorType *vt = lIsSafeToBlend(lvalue)) {
+            auto &DL = callInst->getModule()->getDataLayout();
+            doBlend = DL.getTypeSizeInBits(vt) >= DL.getTypeSizeInBits(rvalue->getType());
```
aneshlya, line 165: Is `>=` here intentional?
The blend operation loads/stores the vector type size, not the rvalue size. For example if vt is larger than rvalue:
  - Load: reads 64 bytes
  - Blend: only 32 bytes of meaningful data
  - Store: writes 64 bytes (potentially with garbage in upper half)

src/opt/ReplacePseudoMemoryOps.cpp
```diff
@@ -13,36 +13,112 @@ namespace ispc {
...
+            // For non-struct types, use the existing logic
+            return lIsVectorMatchingTargetWidth(allocatedType);
+        }
+    }
+
+    // Handle GEP instruction
+    if (llvm::GetElementPtrInst *gep = llvm::dyn_cast<llvm::GetElementPtrInst>(lvalue)) {
+        // Must have constant indices for analysis
+        if (!gep->hasAllConstantIndices()) {
+            return lIsSafeToBlend(gep->getOperand(0));
```
aneshlya: Looking into this code with fresh eyes, probably we should return `nullptr` here to prevent blending in such cases with non constant index access:
```
unmasked void test_non_constant_index(ArrayVectorStruct & data, uniform int idx, float test) {
    ArrayVectorStruct copy = data;
    if (test > 0) {
        copy.vectors[idx] *= 2.0;
    }
    data = copy;
}
```

src/opt/ReplacePseudoMemoryOps.cpp
ivankelarev: Done.

src/opt/ReplacePseudoMemoryOps.cpp
ivankelarev: Done.

src/opt/ReplacePseudoMemoryOps.cpp
ivankelarev: Yes, `>=` is intentional. We want to prevent blend operations when the target vector type (`vt`) is smaller than the data being stored (`rvalue`), as this creates oversized stores that can be treated as undefined behavior and optimized away. I added a test case to illustrate this scenario.

src/opt/ReplacePseudoMemoryOps.cpp
ivankelarev: Done.

aneshlya: Could you add the test case with non constant index to the lit test as well?

aneshlya:
The changes looks good! I'd like to spend some time and check it manually on different platforms.

src/opt/ReplacePseudoMemoryOps.cpp
ivankelarev: > Could you add the test case with non constant index to the lit test as well?

I added the non-constant index test case as requested. However, it seems to work correctly with the original code (without returning nullptr), so I reverted back to the original implementation. The code snippet you provided above also works fine for me. 

Since this appears unrelated to the oversized store UB issue I'm fixing here, I suggest to address it as a separate issue. I would also appreciate if you can provide a code example to reproduce the issue with non-constant indices.

ivankelarev:
@aneshlya Thank you for the review! It's great to hear about the significant performance gain on Apple HW. Regarding avx512skx-i32x16 performance, while there is one regression in the GameDev benchmarks (as we previously discussed), the overall geomean across all tests remains stable, so I suggest to proceed with this change.

If we later decide to disable this optimization for, say, AVX512 and newer targets, I think it would be better to handle that as a separate patch to keep this one focused. (I actually have a draft for such a change, but it's not entirely straightforward, so I'd prefer not to mix it in here.)

ivankelarev:
Since the change is approved, I'll go ahead with the merge. If any issues arise, we can address them in follow-up patches.

PR #3705, opened 2026-01-20
Add constexpr support (v1)

### LINKED ISSUE #3697: `constexpr` design proposal
## Overview
ISPC currently has `const` and constant folding, but no first-class `constexpr`.
This proposal adds `constexpr` variables and functions, enabling compile-time
evaluation in constant-expression contexts. V1 supports varying constexpr values,
requires constexpr-suitable validation at definition time, and does not support
constexpr aggregates (arrays/structs).

## Goals
- Provide explicit compile-time constants with `constexpr`.
- Allow constexpr function calls in constant-expression contexts.
- Support varying constexpr values (lane-wise compile-time constants).
- Enforce "constexpr-suitable" validation at definition time.

## Non-goals (v1)
- No constexpr aggregates (arrays, structs).
- No constexpr evaluation of memory dereference or dynamic allocation.
- No constexpr for `task`, `export`, or external ABI functions.

## Proposed syntax
- Variables: `constexpr` is a declaration specifier and implies `const`.
- Functions: `constexpr` marks functions eligible for constant evaluation.

Examples:
```c++
constexpr uniform int Tile = 16;
uniform float buf[Tile];

constexpr varying int lane = programIndex;
constexpr varying int off = lane * 4;

constexpr uniform int gcd(uniform int a, uniform int b) {
    while (b != 0) {
        uniform int t = a % b;
        a = b;
        b = t;
    }
    return a;
}

uniform int buf2[gcd(48, 18)];
```

## Semantics (v1)
- `constexpr` variables must have a constant-evaluable initializer.
- Allowed v1 types: atomic, enum, pointer, short vector of allowed types.
- Arrays and structs are rejected for `constexpr` variables in v1.
- `constexpr` functions are compiled normally and also eligible for constant
  evaluation when called in constant-expression contexts.
- Only `constexpr` overloads are eligible for compile-time evaluation.

## Constexpr-suitable validation (definition time)
Definition-time validation ensures the body is safe for constant evaluation.
Allowed constructs: local variables, arithmetic/bitwise/comparison, control flow
(`if`, `switch`, `for`, `while`, `do`), `break`/`continue`, `return`, casts,
`sizeof`/`alignof`, calls to other `constexpr` functions.
Disallowed in v1: `foreach*`, `goto`, `new`/`delete`, `launch`, `invoke_sycl`,
`sync`, `print`, atomics, dereference writes or writes to non-local storage,
calls to non-`constexpr` functions, `void` return.

## Constant-expression contexts
`constexpr` values and constexpr function calls are permitted in:
- Array sizes
- Short vector lengths (`float<N>`)
- `switch` case labels
- Enumerator values
- Non-type template arguments
- Default parameter values

## Varying constexpr values
`constexpr varying` values are lane-wise compile-time constants. Constant
evaluation produces a compile-time vector using target width. Contexts requiring
uniform constants (array sizes, vector lengths, template args) require the
constexpr result to be uniform.

## Target dependence
If constexpr evaluation depends on target width or `sizeof(varying T)`, the value
may differ per target. In multi-target compilation, reject such values where a
single target-independent constant is required (global initializers, template
args, etc.), matching existing behavior for `sizeof`.

## Proposed implementation plan
1. Parser/lexer: add `constexpr` token in `src/lex.ll` and keyword tables in
   `src/parse.yy`; add `TYPEQUAL_CONSTEXPR` in `src/decl.h`; allow it in
   `type_qualifier`.
2. Type/symbol flags: add `FunctionType::FUNC_CONSTEXPR` in `src/type.h` and set
   it in `src/decl.cpp`; add a `Symbol` flag for constexpr variables to
   distinguish `const` vs `constexpr`; reject illegal combinations (`task`,
   `export`, `extern "C"/"SYCL"`).
3. Constexpr-suitable validation pass: implement `ConstexprValidator` that walks
   the function AST and enforces allowed/disallowed constructs. Run it when a
   `constexpr` function definition is parsed, before the function is used.
4. Constant-evaluation engine: add `ConstValue` in new `src/constexpr.*` holding
   ISPC type, `llvm::Constant`, and a target-dependence flag. Evaluate `Expr`
   and `Stmt` with a local environment map, supporting uniform and varying
   results. Reuse `Expr::GetConstant` where possible and provide lane-wise
   evaluation for varying values.
5. Constant contexts integration: replace `lGetConstantIntOrSymbol` in
   `src/parse.yy` with const-eval for integer constants; enforce uniform-only
   where required (array sizes, short vector lengths, template args).
6. Template args: allow constexpr expressions as non-type arguments; update
   `TemplateArg` evaluation and mangling in `src/func.cpp` to use evaluated
   constant values rather than only `ConstExpr`.
7. Optimizer hook: in `FunctionCallExpr::Optimize`, if the call is in a
   constant-expression context, evaluate via const-eval; otherwise emit normal
   code.
8. Diagnostics: report clear reasons for constexpr validation failures and
   const-eval failures (side effect, non-constexpr call, non-uniform result,
   target-dependent value).
9. Tests: add positive/negative cases in `tests/lit-tests` and `tests/func-tests`
   covering constexpr vars/functions, varying constexpr, template args, array
   sizes, and error conditions.
10. Docs: add a `Constexpr` section in `docs/ispc.rst`, update the template
   non-type argument rules to allow constexpr expressions, and add a brief note
   in `docs/ReleaseNotes.txt`.

## Open questions
- Should v1 allow constexpr pointer arithmetic beyond globals, or restrict to
  constant addresses only?
- Should constexpr functions allow `switch` over varying conditions if they
  fold to constants?

dbabokin:
## Summary

This PR adds first-class `constexpr` support to ISPC (v1), including constexpr variables/functions, a definition-time validator, a constexpr evaluator, and full front-end plumbing (lexer/parser/AST/type system). It also migrates stdlib/builtins hacks to constexpr where possible, adds comprehensive tests, and documents the design and implementation notes.

## User-facing behavior

- New `constexpr` keyword for variables and functions.
- `constexpr` variables may be `uniform` or `varying` and can be used in constant-expression contexts.
- `constexpr` functions are validated at definition time (C++-style “constexpr-suitable”).
- V1 supports aggregates (arrays/structs), short vectors, and limited pointer constants (null or addresses of globals/functions).
- Uniform control flow required inside constexpr functions (if/switch/loops).
- Deferred constexpr evaluation for global initializers and default parameter values.
- LLVM immarg requirements are enforced in the front end and lowered as true immediates.

## Implementation overview

- Lexer/parser: new `constexpr` token and grammar support.
- AST/symbols/types: constexpr flags on symbols/function types, constexpr parameter defaults.
- Validator: definition-time checks for constexpr-suitable statements/expressions.
- Evaluator: interpreter-style constexpr evaluation with locals, control flow, and aggregate folding.
- Deferral: post-parse resolution for global init + default params that reference later constexpr definitions.
- Intrinsics: immarg metadata and argument rewriting to literal constants.

## stdlib/builtins changes

- Replaced immarg literal hacks in `builtins/generic.ispc` with constexpr constants.
- Migrated `programIndex`/`programCount` and target selectors to `static constexpr`.
- Kept `__is_compile_time_constant_*` probes as backend optimizations (no semantic change).

## Docs

- `docs/ispc.rst`: user documentation for constexpr variables/functions and constraints.
- `docs/design/constexpr.rst`: design and semantics.
- `docs/design/constexpr_impl_notes.md`: implementation notes + remaining work.
- `docs/blog/constexpr_implementation.md`: overview of the work, removed hacks, tests.

## Tests

Added lit tests covering:

- constexpr variables/functions, aggregates, short vectors
- varying constexpr evaluation
- control flow validation (uniform-only)
- template non-type args using constexpr expressions
- pointer constants (null + non-null + function pointers)
- immarg enforcement for LLVM intrinsics
- deferred evaluation for globals/default params

Tests run:

- `cmake --build superbuild/build --target ispc-stage2-check-all`

## Challenges and lessons learned

- **Forward references**: constexpr evaluation needs function bodies; only globals/default params can be deferred in v1. Array sizes, vector lengths, and template args still require definitions to be visible.
- **Grammar ambiguity**: template args use a `constant_expression_no_gt` workaround to avoid `<`/`>` parsing ambiguity; a dedicated template-arg parser is the long-term fix.
- **Immarg requirements**: LLVM requires literal immediates; it wasn’t enough to evaluate to a const value—we had to rewrite immarg arguments to `ConstExpr` during type checking.
- **Pointer constants**: enabling address-of globals/functions exposed assumptions in `ConstExpr::GetValues`; added pointer→bool conversion to support `if (p)` in constexpr.
- **Aggregates/const folding**: aggregate constexpr support required element-wise folding and adjustments to const symbol/address-of handling.
- **Validation timing**: immarg checks needed const locals to be visible earlier during type checking.
- **Target-dependent constants**: preserving target-dependence metadata across constexpr evaluation required additional plumbing to match existing constant folding behavior.

Follow-ups are tracked in `docs/design/constexpr_impl_notes.md`.

### DIFF
--- a/src/constexpr.cpp
+++ b/src/constexpr.cpp
@@ -1,0 +1,2102 @@
+/*
+  Copyright (c) 2026, Intel Corporation
+
+  SPDX-License-Identifier: BSD-3-Clause
+*/
+
+/** @file constexpr.cpp
+    @brief Helpers for constexpr validation and evaluation.
+*/
+
+#include "constexpr.h"
+
+#include "ast.h"
+#include "expr.h"
+#include "func.h"
+#include "llvmutil.h"
+#include "stmt.h"
+#include "sym.h"
+#include "type.h"
+#include "util.h"
+
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
+#include <llvm/IR/Constants.h>
+
+using namespace ispc;
+
+namespace {
+
+constexpr int kMaxConstexprSteps = 100000;
+constexpr int kMaxConstexprCallDepth = 64;
+
+struct EvalResult {
+    enum class Kind { Ok, Return, Break, Continue, Error } kind;
+    ConstExpr *value;
+};
+
+static ConstExpr *lMakeZeroConstExpr(const Type *type, SourcePos pos);
+static bool lIsNullPointerConstexpr(ConstExpr *value);
+static llvm::Constant *lConvertConstexprPointer(llvm::Constant *c, const Type *toType);
+
+static ConstExpr *lConvertConstExpr(ConstExpr *value, const Type *toType, SourcePos pos) {
+    if (value == nullptr || toType == nullptr) {
+        return nullptr;
+    }
+    if (Type::Equal(value->GetType(), toType)) {
+        return value;
+    }
+    if (Type::EqualIgnoringConst(value->GetType(), toType)) {
+        return value;
+    }
+    const Type *fromType = value->GetType();
+    const Type *toBase = toType->GetAsNonConstType();
+    const Type *fromBase = fromType ? fromType->GetAsNonConstType() : nullptr;
+    const VectorType *toVector = CastType<VectorType>(toBase);
+    const VectorType *fromVector = CastType<VectorType>(fromBase);
+    const PointerType *toPointer = CastType<PointerType>(toBase);
+    const PointerType *fromPointer = CastType<PointerType>(fromBase);
+
+    if (value->IsAggregate() || toVector != nullptr || fromVector != nullptr) {
+        if (toVector != nullptr && fromVector != nullptr) {
+            if (toVector->GetElementCount() != fromVector->GetElementCount()) {
+                return nullptr;
+            }
+            if (!value->IsAggregate()) {
+                return nullptr;
+            }
+            std::vector<ConstExpr *> converted;
+            const std::vector<ConstExpr *> &elements = value->GetAggregateValues();
+            if ((int)elements.size() != toVector->GetElementCount()) {
+                return nullptr;
+            }
+            converted.reserve(elements.size());
+            for (ConstExpr *elem : elements) {
+                ConstExpr *conv = lConvertConstExpr(elem, toVector->GetElementType(), pos);
+                if (conv == nullptr) {
+                    return nullptr;
+                }
+                converted.push_back(conv);
+            }
+            return new ConstExpr(toType, converted, pos);
+        }
+        if (toVector != nullptr && fromVector == nullptr) {
+            if (value->IsAggregate()) {
+                return nullptr;
+            }
+            ConstExpr *elem = lConvertConstExpr(value, toVector->GetElementType(), pos);
+            if (elem == nullptr) {
+                return nullptr;
+            }
+            std::vector<ConstExpr *> elements(toVector->GetElementCount(), elem);
+            return new ConstExpr(toType, elements, pos);
+        }
+        if (toVector == nullptr && fromVector != nullptr) {
+            if (!value->IsAggregate() || fromVector->GetElementCount() != 1) {
+                return nullptr;
+            }
+            const std::vector<ConstExpr *> &elements = value->GetAggregateValues();
+            if (elements.empty()) {
+                return nullptr;
+            }
+            return lConvertConstExpr(elements[0], toType, pos);
+        }
+        return nullptr;
+    }
+    auto isZero = [](ConstExpr *candidate) -> bool {
+        if (candidate == nullptr || candidate->IsAggregate()) {
+            return false;
+        }
+        if (candidate->IsPointerConst()) {
+            llvm::Constant *c = candidate->GetPointerConstant();
+            return c != nullptr && c->isNullValue();
+        }
+        const Type *base = candidate->GetType() ? candidate->GetType()->GetAsNonConstType() : nullptr;
+        const AtomicType *at = CastType<AtomicType>(base);
+        const EnumType *et = CastType<EnumType>(base);
+        const PointerType *pt = CastType<PointerType>(base);
+        if (pt == nullptr && et == nullptr && (at == nullptr || (!at->IsIntType() && !at->IsBoolType()))) {
+            return false;
+        }
+        int64_t vals[ISPC_MAX_NVEC] = {};
+        candidate->GetValues(vals, candidate->GetType()->IsVaryingType());
+        for (int i = 0; i < candidate->Count(); ++i) {
+            if (vals[i] != 0) {
+                return false;
+            }
+        }
+        return true;
+    };
+
+    if (toPointer != nullptr || fromPointer != nullptr) {
+        if (value->GetType() && value->GetType()->IsVaryingType() && toType->IsUniformType()) {
+            return nullptr;
+        }
+        if (toPointer != nullptr && fromPointer != nullptr) {
+            if (value->IsPointerConst()) {
+                llvm::Constant *c = lConvertConstexprPointer(value->GetPointerConstant(), toType);
+                if (c == nullptr) {
+                    return nullptr;
+                }
+                return new ConstExpr(toType, c, pos);
+            }
+            if (isZero(value)) {
+                return lMakeZeroConstExpr(toType, pos);
+            }
+            return nullptr;
+        }
+        if (fromPointer != nullptr && toPointer == nullptr) {
+            const AtomicType *toAtomic = CastType<AtomicType>(toBase);
+            if (toAtomic != nullptr && toAtomic->IsBoolType()) {
+                bool isNull = lIsNullPointerConstexpr(value) || isZero(value);
+                if (toType->IsVaryingType()) {
+                    bool vals[ISPC_MAX_NVEC];
+                    for (int i = 0; i < g->target->getVectorWidth(); ++i) {
+                        vals[i] = !isNull;
+                    }
+                    return new ConstExpr(toType->GetAsConstType(), vals, pos);
+                }
+                return new ConstExpr(toType->GetAsConstType(), !isNull, pos);
+            }
+            return nullptr;
+        }
+        if (toPointer != nullptr && fromPointer == nullptr) {
+            if (isZero(value)) {
+                return lMakeZeroConstExpr(toType, pos);
+            }
+            return nullptr;
+        }
+    }
+
+    Expr *converted = TypeConvertExpr(value, toType, "constexpr evaluation");
+    if (converted == nullptr) {
+        return nullptr;
+    }
+    converted = TypeCheckAndOptimize(converted);
+    return llvm::dyn_cast<ConstExpr>(converted);
+}
+
+static bool lIsNullPointerConstexpr(ConstExpr *value) {
+    if (value == nullptr || value->IsAggregate()) {
+        return false;
+    }
+    if (value->IsPointerConst()) {
+        llvm::Constant *c = value->GetPointerConstant();
+        return c != nullptr && c->isNullValue();
+    }
+    const Type *base = value->GetType() ? value->GetType()->GetAsNonConstType() : nullptr;
+    if (CastType<PointerType>(base) == nullptr) {
+        return false;
+    }
+    int64_t vals[ISPC_MAX_NVEC] = {};
+    value->GetValues(vals, value->GetType()->IsVaryingType());
+    for (int i = 0; i < value->Count(); ++i) {
+        if (vals[i] != 0) {
+            return false;
+        }
+    }
+    return true;
+}
+
+static llvm::Constant *lConvertConstexprPointer(llvm::Constant *c, const Type *toType) {
+    if (c == nullptr || toType == nullptr) {
+        return nullptr;
+    }
+    llvm::Type *llvmType = toType->LLVMType(g->ctx);
+    if (llvmType == nullptr) {
+        return nullptr;
+    }
+    if (c->getType() == llvmType) {
+        return c;
+    }
+    if (llvmType->isPointerTy()) {
+        if (c->getType()->isPointerTy()) {
+            return llvm::ConstantExpr::getBitCast(c, llvmType);
+        }
+        return llvm::ConstantExpr::getIntToPtr(c, llvmType);
+    }
+    llvm::Constant *intPtr =
+        c->getType()->isPointerTy() ? llvm::ConstantExpr::getPtrToInt(c, LLVMTypes::PointerIntType) : c;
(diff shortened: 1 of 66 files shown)


nurmukhametov:
I tried this draft with a few examples. I didn't encounter any obvious issues. The only thing that is unclear to me is the linkage of constexpr functions. At the moment, they have internal linkage by default, so they must be declared as static constexpr in headers to avoid linking conflicts.

Would it be possible, after this gets merged, to rewrite the fast_math stdlib functions as constexpr ones?

aneshlya:
> Would it be possible, after this gets merged, to rewrite the fast_math stdlib functions as constexpr ones?

Definitely. The best way to test this PR is to use `constexpr` to rewrite
1. fast_math
2. amx support
3. use it within templates

dbabokin:
Implemented the requested constexpr validation pass and rewrites, including fast-math, AMX, template usage, semantic fixes, expanded negative tests, and docs updates.

What was implemented

- Rewrote/validated feature usage in requested areas:
  - fast-math + templates (`cf088fa11`)
  - AMX support + templates (`56c1e1776`)
- Fixed semantic issues found during review:
  - Enforce missing initializer error for `extern constexpr` globals (`bf6001aeb`)
  - Keep `constexpr` out of generic declarator qualifier contexts (pointer qualifier path) (`bf6001aeb`)
  - Reject `typedef constexpr` (`73844e27b`)
  - Reject `constexpr` on function parameters (`73844e27b`)
  - Fix constexpr validator to reject address-of locals/params robustly (`73844e27b`)
- Aligned constexpr function linkage with C++ semantics:
  - non-`static constexpr` functions emit ODR linkage (`linkonce_odr`)
  - `static constexpr` functions remain internal (`c311c0b2f`)
- Updated docs for the above semantics/restrictions (`dca32184e`)

Test coverage added

- New positive coverage:
  - `tests/lit-tests/constexpr-contexts.ispc`
    - enum values from constexpr calls
    - short-vector lengths from constexpr calls
    - `switch case` labels from constexpr calls
- New negative coverage:
  - `tests/lit-tests/constexpr-extern-noinit-error.ispc`
  - `tests/lit-tests/constexpr-pointer-qualifier-error.ispc`
  - `tests/lit-tests/constexpr-typedef-error.ispc`
  - `tests/lit-tests/constexpr-parameter-qualifier-error.ispc`
  - `tests/lit-tests/constexpr-address-of-local-error.ispc`
  - `tests/lit-tests/constexpr-struct-member-error.ispc`
- New linkage regression coverage:
  - `tests/lit-tests/constexpr-linkage.ispc`

Validation

- Rebuilt compiler (`build-constexpr/bin/ispc`).
- Ran targeted lit for constexpr + immarg constexpr + AMX constexpr files on this host.
- Results: all applicable tests passed; AMX/X86-gated tests remained `UNSUPPORTED` on this host as expected.

Branch state

- Pushed branch `constexpr-v1-complete` to `origin`.

dbabokin:
Pushed fix commit: `76c78390` (`constexpr: propagate deferred failures through comma expr`).

What was fixed:
- `constexpr` evaluator incorrectly ignored the left side of comma expressions during deferred evaluation.
- In `src/constexpr.cpp`, `BinaryExpr::Comma` previously did:
  - evaluate `arg0` and ignore the result
  - return evaluation of `arg1`
- This allowed unresolved deferred calls to be dropped in forms like `(f(), 1)` and could incorrectly accept invalid initializers/default args.

Behavioral fix:
- `BinaryExpr::Comma` now returns failure/deferred immediately if `arg0` evaluation returns `nullptr`.
- This preserves the deferred/error state and prevents false acceptance.

Regression coverage added:
- `tests/lit-tests/constexpr-deferred-comma-global-error.ispc`
- `tests/lit-tests/constexpr-deferred-comma-param-error.ispc`

Both tests ensure unresolved `constexpr` calls in comma-left position are rejected with the expected diagnostics.

Validation:
- Full `check-all` after patch: `722 total / 626 passed / 90 unsupported / 6 xfail / 0 fail`.

dbabokin:
@nurmukhametov @aneshlya
- Linkage of constexpr function was made compliant with C++.
- Suggested stdlib/AMX/fast_math changes were implemented to prove that the implementation is robust.
- Codex found and fixed a few more edge cases that didn't work as expected.

The whole changeset was implemented by Codex and I haven't at the code myself yet.

THe key thing that I'd like to verify is how well it matches the current architecture and if it brings any new complexity that is not ok.

But it **seems** that the feature is solid now.

nurmukhametov:
>     * Suggested stdlib/AMX/fast_math changes were implemented to prove that the implementation is robust.

I am not quite sure that it was implemented.

> But it **seems** that the feature is solid now.

Yes, it definitely seemed to work when I tried it a while ago. The biggest problem here is how to review this huge patch. It would be good to somehow split this change into separate PRs of smaller size.

src/constexpr.cpp
```diff
@@ -0,0 +1,2168 @@
...
+    void DisallowedExpr(const Expr *expr, const char *what) {
+        if (!expr || !func) {
+            return;
+        }
+        Error(expr->pos, "constexpr function \"%s\" contains disallowed expression \"%s\".", funcName.c_str(), what);
+        ok = false;
+    }
+};
+
+void ConstexprValidator::GatherLocals(const Stmt *stmt) {
```
aneshlya: Validator re-implements AST traversal instead of reusing WalkAST .
`ConstexprValidator::VisitStmt/VisitExpr/GatherLocals` hand-roll a recursive walk with dyn_cast chains. ISPC has the canonical WalkAST utility (ast.h:253, ast.cpp:187) used by exactly this kind of pass (stmt.cpp:927, stmt.cpp:3063, ctx.cpp:1260). The hand-rolled walk silently skips unenumerated node types (e.g. no
`ForeachUniqueStmt` body recursion in GatherLocals, no default safety net), so future AST additions won't be validated. WalkAST has a `FATAL("Unhandled statement type")` backstop.

src/constexpr.cpp
```diff
@@ -0,0 +1,2168 @@
...
+            stmts.assign(sl->stmts.begin(), sl->stmts.end());
+        } else if (ss->stmts != nullptr) {
+            stmts.push_back(ss->stmts);
+        }
+
+        int startIndex = -1;
+        int defaultIndex = -1;
+        for (int i = 0; i < (int)stmts.size(); ++i) {
+            if (auto *cs = llvm::dyn_cast<CaseStmt>(stmts[i])) {
+                if (cs->value == (int)iv[0]) {
```
aneshlya: The switch condition is evaluated to UniformInt64 and stored in int64_t iv[], then truncated to int for case matching. This diverges from runtime: FunctionEmitContext::SwitchInst (src/ctx.cpp:1163) switches on the full-width 64-bit condition. A 64-bit condition whose low 32 bits collide with a case label (e.g.
condition 0x1_0000_0001 vs case 1:) matches at compile time but not at runtime.

Fix: compare in 64-bit — if ((int64_t)cs->value == iv[0]).

src/expr.cpp
```diff
@@ -4554,6 +4570,39 @@ Expr *FunctionCallExpr::TypeCheck() {
...
+        for (int i = 0; i < checkCount; ++i) {
+            if (!funcType->IsParameterImmArg(i)) {
+                continue;
+            }
+            Expr *argExpr = args->exprs[i];
+            if (argExpr == nullptr) {
+                continue;
+            }
+            ConstExpr *ce = ConstexprEvaluate(argExpr, funcType->GetParameterType(i));
+            if (ce == nullptr) {
```
aneshlya, line 4596: The immarg block calls `ConstexprEvaluate` and only errors when the result is nullptr. A varying compile-time constant evaluates successfully, passes the check, and is installed as the arg. LLVM immargs must be scalar, so this fails late (IR verifier / lowering) instead of producing a clean front-end diagnostic.
This is also inconsistent with `lGetConstantIntOrSymbol`, which explicitly rejects varying via requireUniform (src/parse.yy, new path).

Fix: reject a varying result in the immarg block with a proper Error(...), mirroring the uniform check used in parse.yy.

aneshlya (commented):
Also user docs are required.

src/constexpr.cpp
dbabokin: Addressed in the rewritten stack:

- compiler change: `b4a61b9f9` (`constexpr: add compiler support`)
- regression coverage: `3b7f80e97` (`tests: add constexpr coverage`)

`ConstexprValidator` no longer hand-recurses through `VisitStmt` / `VisitExpr` / `GatherLocals`. Local collection and validation now run through `WalkAST` callbacks, so traversal coverage is tied to the canonical AST walker and future unhandled AST node types hit the existing `WalkAST` safety backstop.

One subtlety I preserved from the old validator: `sizeof(expr)` remains unevaluated for validation purposes. I added coverage in `constexpr-basic.ispc` with `sizeof(non_constexpr_call())` inside a constexpr function to make sure the `WalkAST` conversion does not accidentally validate unevaluated operands.

Validated with focused lit plus full `check-all`: `741 tests`, `572 passed`, `163 unsupported`, `6 xfailed`.

src/constexpr.cpp
dbabokin: Addressed in the rewritten stack:

- compiler change: `b4a61b9f9` (`constexpr: add compiler support`)
- regression coverage: `3b7f80e97` (`tests: add constexpr coverage`)

The constexpr switch evaluator now compares the case label against the full-width condition value:

```c++
if ((int64_t)cs->value == iv[0])
```

I added a regression to `constexpr-switch.ispc`: `pick64(4294967297)` must choose `default` and produce `[22 x i32]`, rather than matching `case 1` through low-32-bit truncation.

I also checked the direct repro locally; it now emits/sizes as `22`. Focused lit and full `check-all` pass (`741 tests`, `572 passed`, `163 unsupported`, `6 xfailed`).

src/expr.cpp
dbabokin: Addressed in the rewritten stack in `b4a61b9f9` (`constexpr: add compiler support`).

After `ConstexprEvaluate`, the immarg block now explicitly rejects anything except a single uniform scalar constexpr result:

```c++
const Type *ceType = ce->GetType();
if (ceType == nullptr || !ceType->IsUniformType() || ce->Count() != 1) {
    emitImmArgError(argExpr, i, "must be a uniform scalar compile-time constant");
    immArgError = true;
    continue;
}
```

I could not make the simple source-level `constexpr varying int` cases reach this exact block because overload resolution already rejects them before immarg validation, but the immarg-specific invariant you requested is now enforced if such a folded value reaches this path.

Validated with `llvm-intrinsics-immarg-constexpr.ispc`, `llvm-intrinsics-immarg-error.ispc`, and full `check-all`: `741 tests`, `572 passed`, `163 unsupported`, `6 xfailed`.

PR #3724, opened 2026-01-27
Replace dots with underscores in avx10.2dmr target function suffix

### LINKED ISSUE #3723: Replace dots with underscore for avx10.2 dmr target in function postfix
Rename function suffix in dispatcher and generated object files for avx10.2 dmr target using underscore, for example "function_avx10_2dmr" (currently it's "function_avx10.2dmr").

ivankelarev:
Fixes #3723.

### DIFF
--- a/src/ispc.cpp
+++ b/src/ispc.cpp
@@ -2704,7 +2704,7 @@ const char *Target::ISAToString(ISA isa) {
         return "avx512gnr";
 #if ISPC_LLVM_VERSION >= ISPC_LLVM_20_0
     case Target::DMR_AVX10_2:
-        return "avx10.2dmr";
+        return "avx10_2dmr";
 #endif
 #ifdef ISPC_XE_ENABLED
     case Target::XELP:
--- a/tests/lit-tests/3723.ispc
+++ b/tests/lit-tests/3723.ispc
@@ -1,0 +1,10 @@
+// RUN: %{ispc} %s -O0 --target=sse4,avx10.2dmr-x16 --emit-llvm-text -o %t.ll
+// RUN: FileCheck --input-file=%t.ll %s
+// RUN: FileCheck --input-file=%t_avx10_2dmr.ll %s
+
+// REQUIRES: X86_ENABLED && !MACOS_HOST && LLVM_20_0+
+
+// CHECK-NOT: @function_avx10.2dmr
+// CHECK: @function_avx10_2dmr
+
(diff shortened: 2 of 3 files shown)


src/module.cpp
```diff
@@ -1845,6 +1845,7 @@ bool Module::writeObjectFileOrAssembly(llvm::Module *M, Output &CO) {
 // filename's suffix, like "foo_avx.obj".
 std::string lGetMangledFileName(std::string filename, Target *target) {
     std::string isaString = target->GetISAString();
+    std::replace(isaString.begin(), isaString.end(), '.', '_');
```
aneshlya: If I'm not missing anything I think it's safe to make the dot to underscore replacement as part of `GetISAString` function.
It will bring consistency between user-visible and internal naming and is good from maintainability point of view.
```
std::string Target::GetISAString() const {
    std::string isaString = ISAToString(m_isa);
    std::replace(isaString.begin(), isaString.end(), '.', '_');
    return isaString;
}

```
You can also covert it by lit test. For example:
```
// RUN: rm -rf %t_dbg && mkdir -p %t_dbg
// RUN: %{ispc} %s -O0 --target=sse4,avx10.2dmr-x8 -o %t.o --debug-phase=3:3 --dump-file=%t_dbg 2>&1
// RUN: FileCheck --input-file=%t_dbg/pre_3_LinkStdlib.ll_avx10_2dmr %s

// REQUIRES: X86_ENABLED && !MACOS_HOST && LLVM_20_0+

// Verify stdlib function is DEFINED with underscore suffix after linking
// CHECK: define {{.*}} @floatbits___vyu_avx10_2dmr(
// CHECK-NOT: define {{.*}} @floatbits___vyu_avx10.2dmr(

// Also verify debug dump file uses underscore in filename (tested by RUN line above)

export uniform float test(uniform int x) {
    // floatbits is a stdlib function that will be linked and renamed
    return floatbits(x);
}
```

ivankelarev: Thanks @aneshlya for the feedback! I see your point. However, this approach would require changing `GetISAString()`'s return type and adding `.c_str()` calls everywhere it's used - quite invasive changes throughout the codebase. While doable, it might be overkill. I think if we are going this way a better approach could be modifying the string constant right in `ISAToString()` function itself. Please see the updated patch.

aneshlya: That's even better, thanks!

Issue #3517, opened 2025-07-30
[Feature request] Add support for PPC64BE and PPC64LE SIMD vector


runlevel5:
Just want to write this issue up to double-check to see if anyone has attempted to add such support for POWER-based SIMD vector yet.

aneshlya:
We haven't had anyone attempt to add this functionality yet, but we'd definitely welcome contributions that extend support to additional architectures. If you're interested in working on this, feel free to submit a pull request. We're happy to provide guidance during the development process.

PR #3735, opened 2026-02-07
Add Linux ppc64le (PowerPC64) architecture support using generic targets

runlevel5:
## Description

Add Linux ppc64le as a new architecture in ISPC, enabling compilation for POWER8+ (POWER ISA 2.7) processors using the existing generic target system (generic-i1x4 through generic-i64x4). This follows the same approach as RISC-V — no PPC-specific ISPCTarget enum values, no ISA enum entry, no `.ll` builtin files.

Closes https://github.com/ispc/ispc/issues/3517

### Core compiler support
- Add `Arch::ppc64le` enum and parsing/serialization support
- Add `CPU_PPC64LE_Generic` device type with POWER8 baseline and `+vsx` features
- Initialize LLVM PowerPC backend when `ISPC_PPC64_ENABLED` is defined
- Generate `powerpc64le-unknown-linux-gnu` triple (with `powerpc64le-linux-gnu` for Debian)
- Auto-detect ppc64le host architecture and default to `generic-i32x4`
- Support cross-compilation: `lGetArchFromTarget()` infers ppc64le arch from generic targets when CPU is `ppc64le`

### Build system
- Add `PPC64_ENABLED` CMake option (OFF by default) with `PowerPC` LLVM component
- Add ppc64le support to builtin generation (`GenerateBuiltins.cmake`, `GenericTargets.cmake`, `bitcode2cpp.py`)
- Fix `generic-target` build: add `GENERATED` property and `add_dependencies` in `GenericTargets.cmake`
- Move `configure_ispc_obj(generic-target)` into `add_ispc_slim()` so it applies to both slim and composite builds
- Add `generic-target` to `ispc-opt` link libraries (required for `BitcodeLib` static registration)
- Add PowerPC to `LLVM_TARGETS_TO_BUILD` in `alloy.py` and `superbuild`

### Examples and documentation
- Add ppc64le support to example CMake build system (`AddISPCExample.cmake`, `AddISPCExampleLegacy.cmake`, `AddISPCExampleModern.cmake`)
- Add ppc64le support to examples timing utility (`timing.h`)
- Add ppc64le to supported architectures in documentation (`ispc.rst`, `ReleaseNotes.txt`)

### Tests
- Add lit test infrastructure: `PPC64_ENABLED` feature flag, `PPC64LE_HOST` host arch detection
- Add `generic-target-ppc64le.ispc` — verifies target/arch/triple correctness for all generic targets
- Add `ppc64le-vector-asm.ispc` — verifies PowerPC VSX vector instructions (`xvaddsp`, `xvmulsp`) are generated
- Add `run_tests.py` support: target config, gcc_arch flags, default target/arch detection for ppc64le hosts

### XFAIL tests on ppc64le

| Test | Reason | Version-conditional? |
|------|--------|---------------------|
| `stdlib-short-vec-funcs-3` | LLVM 20 can't select `froundeven` for vector types ([llvm#110753](https://github.com/llvm/llvm-project/issues/110753), fixed in LLVM 21 by [llvm#150597](https://github.com/llvm/llvm-project/pull/150597)) | Yes — auto-enables on LLVM 21+ |
| `stdlib-short-vec-funcs-6` | LLVM 20 can't promote `atan2`/`pow`/`fmod` results for vectors (fixed in LLVM 21) | Yes — auto-enables on LLVM 21+ |
| `fast_math_opt` | Generic target stdlib's `rcp` uses `1.0/x` (fdiv) instead of hardware rcp | No — generic target limitation |
| `num-limits` | ABI mismatch for `int8`/`int16` return types between ISPC and clang on ppc64le | No — ABI issue |
| `2169` | ISPC assertion in `lAllDivBaseEqual` (integer divide optimization) for generic targets | No — ISPC bug for generic targets |

## Test Results

Tested on Fedora 43 ppc64le ([Raptor Computing Blackbird](https://www.raptorcs.com/content/BK1SD1/intro.html), IBM POWER9) with LLVM 20.1.8:

**Lit tests (check-all):** 426 passed, 267 unsupported, 3 expectedly failed, **0 failures**

**Functional tests (run_tests.py --target=generic-i32x8):** 1503/1582 passed (95%)
- 66 compilation failures, 0 execution failures
- ~44 float16-related (ppc64le lacks native float16 in LLVM 20)
- 3 AMX + 3 Xe tests (x86/GPU-only, expected)
- ~16 LLVM 20 PPC backend limitations (round, fmod, broadcast, shuffle, etc.)

## TODO

- [x] Add ppc64le target in CI — @aneshlya please apply for ppc64le CI runner at https://github.com/IBM/actionspz
- [ ] If accepted, advise Linux distributions to enable ppc64le support (e.g., [Fedora ispc.spec](https://src.fedoraproject.org/rpms/ispc/blob/main/f/ispc.spec))

### DIFF
--- a/docs/ispc.rst
+++ b/docs/ispc.rst
@@ -257,6 +257,18 @@ of recent changes to the compiler.
 .. _ReleaseNotes.txt: https://raw.github.com/ispc/ispc/main/docs/ReleaseNotes.txt
 
 
+Updating ISPC Programs For Changes In ISPC 1.31.0
+-------------------------------------------------
+
+New Architecture Support:
+
+* Initial support for the PowerPC 64-bit little-endian (ppc64le) architecture
+  has been added using generic targets (e.g., generic-i32x4). The baseline is
+  POWER ISA 2.7 (POWER8+) with VSX. This support is experimental and not
+  included in official ISPC binaries. To use it, build ISPC from source with
+  the `PPC64_ENABLED=ON` CMake option. Usage: `ispc --arch=ppc64le
+  --target=generic-i32x4 foo.ispc -o foo.o`.
+
 Updating ISPC Programs For Changes In ISPC 1.30.0
 -------------------------------------------------
 
@@ -269,15 +281,6 @@ New Features:
   for INT8, BF16, and FP16 data types. AMX is supported on ``avx512spr``,
   ``avx512gnr``, and ``avx10.2dmr`` targets. Please refer to
   `Intel AMX (Advanced Matrix Extensions)`_ for more details.
-
-New Architecture Support:
-
-* Initial support for the PowerPC 64-bit little-endian (ppc64le) architecture
-  has been added using generic targets (e.g., generic-i32x4). The baseline is
-  POWER ISA 2.7 (POWER8+) with VSX. This support is experimental and not
-  included in official ISPC binaries. To use it, build ISPC from source with
-  the `PPC64_ENABLED=ON` CMake option. Usage: `ispc --arch=ppc64le
-  --target=generic-i32x4 foo.ispc -o foo.o`.
 
 Language Changes:
 


dbabokin:
@codex review

cmake/GenerateBuiltins.cmake
```diff
@@ -200,6 +200,9 @@ function (get_target_flags os arch out)
         elseif (${arch} STREQUAL "riscv64")
             set(triple ${arch}-unknown-linux-gnu)
             set(debian_triple ${arch}-linux-gnu)
+        elseif (${arch} STREQUAL "ppc64le")
+            set(triple ${arch}-unknown-linux-gnu)
+            set(debian_triple ${arch}-linux-gnu)
```
nurmukhametov: Debian triple is `powerpc64le-linux-gnu`.

CMakeLists.txt
```diff
@@ -1024,7 +1037,7 @@ configure_ispc_obj(builtin)
 add_executable(ispc-opt tools/ispc-opt.cpp)
 configure_ispc_exe(ispc-opt)
 # TODO! should we link against all the same LLVM/Clang libraries? Clang libs is definetely not needed
-target_link_libraries(ispc-opt ${LINK_LIBRARIES} builtin optimization common)
+target_link_libraries(ispc-opt ${LINK_LIBRARIES} builtin generic-target optimization common)
```
nurmukhametov: I don't understand why it is needed here.

tools/check_isa.cpp
```diff
@@ -62,13 +62,14 @@ static const char *lGetSystemISA() {
     }
     const char *isa = isa_strings[isa_id];
     if (isa_id >= SPR_AVX512) {
-        snprintf(amx_isa_string, sizeof(amx_isa_string), "%s (AMX %s)", isa,
-                 __os_enabled_amx_support() ? "on" : "off");
+        snprintf(amx_isa_string, sizeof(amx_isa_string), "%s (AMX %s)", isa, __os_enabled_amx_support() ? "on" : "off");
```
nurmukhametov: What has changed here? If no changes applied, please remove it.

src/ispc.cpp
```diff
@@ -839,6 +857,13 @@ Arch lGetArchFromTarget(ISPCTarget target) {
         return Arch::riscv64;
     }
 #endif
+#ifdef ISPC_PPC64_ENABLED
+    if (ISPCTargetIsGeneric(target)) {
+#if defined(ISPC_HOST_IS_PPC64LE)
+        return Arch::ppc64le;
+#endif
```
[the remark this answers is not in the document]
runlevel5: Hmm valid point, let me have a thought about this

nurmukhametov (changes requested):
Thanks for the submission!

I don't have access to PowerPC hardware, so I tested this on x86 Ubuntu using QEMU for ppc64le (similar to how we currently test RISC-V in our CI). I managed to compile ISPC on x86 with ppc64le support, but encountered a couple of issues that need to be addressed:

- Build error due to a new dependency for ispc-opt (see [this CI run](https://github.com/ispc/ispc/actions/runs/21774712401/job/62861837740?pr=3735)).
- Incorrect Debian triple.

Additionally, please add ppc64le support in `scripts/run_tests.py` and run the test suite as well. Under QEMU, I got a pass rate of 1500/1569 (95%) with run_tests.py, which is quite good!

CMakeLists.txt
```diff
@@ -981,6 +994,7 @@ function(add_ispc_slim NAME)
 
     # Defines targets for generic builtin implementations
     generate_generic_builtins(${NAME})
+    configure_ispc_obj(generic-target)
```
nurmukhametov: Why do you add this target here?

tools/check_isa.cpp
runlevel5: It's my text editor's auto-linting, I will revert those changes

tests/lit-tests/stdlib-short-vec-funcs-3.ispc
```diff
@@ -3,6 +3,8 @@
 
 // RUN: %{ispc} --target=host -o %t.o %s 2>&1
 
+// UNSUPPORTED: PPC64LE_HOST
```
nurmukhametov: Why do we need to ignore this test for ppc64le? I don't recall that it tests any arch specific things.

tests/lit-tests/generic-target-ppc64le.ispc
```diff
@@ -0,0 +1,33 @@
+// Test generic targets for ppc64le architecture
+// Verifies that generic targets produce correct LLVM IR vector types for ppc64le.
+
+// RUN: %{ispc} --target=generic-i1x4 --arch=ppc64le --nostdlib --nowrap %s --emit-llvm-text -o - 2>&1 | FileCheck %s --check-prefix=CHECK-I1X4
```
nurmukhametov, line 4: It would be good to add at least one test that checks generated asm for ppc64le.

CMakeLists.txt
runlevel5: Because I got failed test:

```
cd /home/tle/Work/ispc && TEST=/home/tle/Work/ispc/tests/lit-tests/ispc-opt.ll cmake --build build --target check-one -j \$(nproc) 2>&1 | tail -30"
Input was:
<<<<<<
            1: Error: generic-i32x4 target for ppc64le on Linux is not supported in current  
check:19'0     X~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ error: no match found
            2:  build.  
check:19'0     ~~~~~~~~~
            3: Error: Unsupported target  
check:19'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~
check:19'1     ?                           possible intended match
            4:   
check:19'0     ~~
>>>>>>
error: command failed with exit status: 1
--
********************
********************
Failed Tests (1):
  ispc :: ispc-opt.ll
Testing Time: 0.20s
  Failed: 1
gmake[3]: *** [tests/CMakeFiles/check-one.dir/build.make:72: tests/CMakeFiles/check-one] Error 1
gmake[2]: *** [CMakeFiles/Makefile2:1190: tests/CMakeFiles/check-one.dir/all] Error 2
gmake[1]: *** [CMakeFiles/Makefile2:1197: tests/CMakeFiles/check-one.dir/rule] Error 2
gmake: *** [Makefile:572: check-one] Error 2
```

ispc-opt now fails because it doesn't have the generic target bitcode libraries linked in. Without generic-target, it can't find the generic-i32x4 target support. So generic-target IS needed for ispc-opt after all — the BitcodeLib static objects need to register for ispc-opt to support generic targets. Unless I misunderstood the codes.

tests/lit-tests/stdlib-short-vec-funcs-3.ispc
runlevel5: I was running into 

```
LLVM ERROR: Cannot select: 0x14a050a30: v4f32 = froundeven 0x14a3a2db0
In function: common_round___vyf_3C_1_3E_
```

I will double check to see if it is an issue of LLVM or ISPC.

tests/lit-tests/generic-target-ppc64le.ispc
runlevel5: Added `tests/lit-tests/ppc64le-vector-asm.ispc`

tests/lit-tests/stdlib-short-vec-funcs-3.ispc
runlevel5: Okay I know why, this is an issue of LLVM

ref: https://github.com/llvm/llvm-project/issues/110753

It has been addressed in LLVM 21. Is there any plan to move to LLVM 21 as minimal requirement anytime soon? If not, I simply can add more condition to omit the test if LLVM is older than 21

cmake/GenerateBuiltins.cmake
runlevel5: Fixed in commit 51c2a30 — changed the triple to use the canonical `powerpc64le` prefix (e.g., `powerpc64le-linux-gnu` for Debian).

CMakeLists.txt
runlevel5: `configure_ispc_obj(generic-target)` needs to be inside `add_ispc_slim()` (and not just at the top-level scope for the composite build) because the slim build path also needs to compile and link the generic-target bitcode library. Without it, the slim build of ispc would fail with `bitcode_lib.h` not found errors since the generic-target object files are never generated.

This is analogous to how `configure_ispc_obj(stdlib)` is already called inside `add_ispc_slim()` on line 1000.

tests/lit-tests/stdlib-short-vec-funcs-3.ispc
runlevel5: I have now added explanatory comments to all `UNSUPPORTED: PPC64LE_HOST` test skips in commit 155f0f5. Here is the summary:

**LLVM 20 backend bugs (fixed in LLVM 21):**
- `stdlib-short-vec-funcs-3`: LLVM 20 PPC backend cannot select `froundeven` for vector types (`v4f32 = froundeven`). Fixed in LLVM 21 by [llvm/llvm-project#150597](https://github.com/llvm/llvm-project/pull/150597).
- `stdlib-short-vec-funcs-6`: LLVM 20 PPC backend crashes with "Do not know how to promote this operator's result!" on vector `atan2`/`pow`/`fmod`. Also confirmed fixed in LLVM 21 (tested with `llc-21`).

**Generic target limitations:**
- `fast_math_opt`: Generic target stdlib's `rcp` implementation uses `1.0/x` (fdiv) rather than a hardware rcp instruction, so the `CHECK_NOFDIV` check fails.
- `num-limits`: ABI mismatch for `int8`/`int16` return types between ISPC and clang on ppc64le (test passes with gcc but `int8_min` and `int16_min` return wrong values when linked with clang).
- `2169`: ISPC assertion failure in `lAllDivBaseEqual` (`llvmutil.cpp:893`) — the integer divide optimization passes a non-power-of-2 `baseValue` for generic targets.

dbabokin:
Thanks for the contribution! That's exciting to see the interest for ISPC on PowerPC!

Before merging this PR, I'd like to understand (just to set right expectations upfront):
- are you interested in supporting this target beyond initial enabling?
- do you have any specific application in mind that you'd like to enable with ISPC on PowerPC? What is assumed use-case?
- Looks like you have access to the hardware, were you able run any benchmarks / performance tests?
- Do you know how good is QEMU ppc64le support? Can we rely on it?

Given the limited access to hardware it would be difficult for us to do the performance testing and tuning.

runlevel5:
> are you interested in supporting this target beyond initial enabling?

Yes, I am more than happy to.

> do you have any specific application in mind that you'd like to enable with ISPC on PowerPC? What is assumed use-case?

I do have some personal applications that could leverage ispc and will use them to see if we could fine-tune ispc even more. Additionally I will encourage fellow PowerPC developers to give ispc a go and hopefully they will too extend PPC64 support further.

> Looks like you have access to the hardware, were you able run any benchmarks / performance tests?

Yes I am running RaptorCS Blackbird workstation

> Do you know how good is QEMU ppc64le support? Can we rely on it?

With x86 as host? If so, the support is quite okay though no where near as performant as real hardware. And yes I do not see any issue why we could not rely on it, though I highly encourage team to use a native PPC64 CI runner (please see https://github.com/IBM/actionspz)

tests/lit-tests/stdlib-short-vec-funcs-3.ispc
nurmukhametov: > It has been addressed in LLVM 21. Is there any plan to move to LLVM 21 as minimal requirement anytime soon? If not, I simply can add more condition to omit the test if LLVM is older than 21

ISPC supports LLVM 21. You can simply rebuild with LLVM 21 and retest.

Also, it looks like you should use XFAIL instead of UNSUPPORTED for known issues. XFAIL is more appropriate for tests that are expected to fail rather than tests that aren't applicable.

scripts/run_tests.py
```diff
@@ -659,6 +661,10 @@ def run_test(testname, host, target, jit_lib_path=None):
                     gcc_arch = '-march=rv64gcv'
                     if options.wrapexe and options.wrapexe.startswith('qemu'):
                         gcc_arch += ' -static'
+                elif target.arch == 'ppc64le':
+                    gcc_arch = ''
```
nurmukhametov: I believe we need to pass the appropriate -march here.

runlevel5: I can add `-mcpu=power8` here

aneshlya:
Hi @runlevel5, thanks a lot for your contribution! I’ve applied for the runner so we can use it for functional testing and running examples.

I’d also like to ask if you could add a CI workflow that builds ISPC with PPC64_ENABLED=ON and runs the lit tests. You can use the original riscv.yml workflow as a reference:
https://github.com/nurmukhametov/ispc/blob/3cd0d3a106a2a7a4f701993860fa770b8ebce79a/.github/workflows/riscv.yml
It would be great if you could avoid Arch Linux and use ubuntu instead - this would make it easier for us to port later if we decide to include PowerPC as part of the ISPC binary.

docs/ispc.rst
```diff
@@ -1246,7 +1246,7 @@ To compile for Intel Xe LP platform:
    ispc foo.ispc -o foo.bin --target=xelp-x16 --device=tgllp --emit-zebin
 
 Currently-supported architectures are ``x86``, ``x86-64``, ``xe64``,
-``arm``, and ``aarch64``.
+``arm``, ``aarch64``, and ``ppc64le``.
```
aneshlya: Since it's experimental and is not in the official binaries, please remove `ppc64le` here.

nurmukhametov:
> I’d also like to ask if you could add a CI workflow that builds ISPC with PPC64_ENABLED=ON and runs the lit tests. You can use the original riscv.yml workflow as a reference: https://github.com/nurmukhametov/ispc/blob/3cd0d3a106a2a7a4f701993860fa770b8ebce79a/.github/workflows/riscv.yml It would be great if you could avoid Arch Linux and use ubuntu-22.04 instead - this would make it easier for us to port later if we decide to include PowerPC as part of the ISPC binary.

When I tested locally, I used Ubuntu 25.10:
```bash
$ apt-get install llvm llvm-dev clang libclang-dev llvm-tools
$ apt-get install gcc-powerpc64le-linux-gnu g++-powerpc64le-linux-gnu libc6-dev-ppc64el-cross qemu-user
$ PATH=PATH=/usr/lib/llvm-20/bin:$PATH cmake ../ -DPPC64_ENABLED=ON
$ PATH=`pwd`/build-ppc64le/bin:$PATH python3 scripts/run_tests.py --target=generic-i32x4 --arch=ppc64le --wrap-exe="qemu-ppc64le" --compiler=powerpc64le-linux-gnu-g++ -o O2
```

runlevel5:
> Looks like you have access to the hardware, were you able run any benchmarks / performance tests?

@dbabokin it's very easy to find POWER8 hardware on eBay. Lots of decommissioned POWER IBM workstations are available at steep discount. POWER9 is too getting cheaper because POWER10 line is slowly eating into the POWER9 customer base.

scripts/run_tests.py
```diff
@@ -1356,6 +1362,9 @@ def run_tests(options1, args, print_version):
 elif platform.machine() == "arm64":
     default_target = "neon-i32x4"
     default_arch = "aarch64"
+elif platform.machine() == "ppc64le":
+    default_target = "generic-i32x8"
```
aneshlya: Here the default target is `generic-i32x8`, while src/ispc.cpp_s host ISA default is `generic_i32x4`. Better to align them.

tools/check_isa.cpp
```diff
@@ -69,6 +69,8 @@ static const char *lGetSystemISA() {
     return isa;
 #elif defined(__riscv)
     return "RISC-V";
+#elif defined(__powerpc64__)
+    return "PPC64LE";
```
aneshlya, line 73: It will reports PPC64LE for any `__powerpc64__` build, regardless of endianness. On big-endian ppc64 this will misreport the host ISA. Use a `__BYTE_ORDER__` check (like in src/ispc.cpp) or explicitly guard for LE

CMakeLists.txt
nurmukhametov: OK, I see now that on a PPC host, `--target=host` maps to `--arch=ppc64le --target=generic-i32x4`, so `ispc-opt` checks whether the generic libraries are in place. Yes, in the current state this target is needed, but it introduces another unnecessary dependency, similar to [#3737](https://github.com/ispc/ispc/issues/3737).

I'm starting to think it would be better to add at least one proper target, e.g., `pwr-x4`. This would align the target detection logic for the `--target=host` case with how other platforms work, and would also avoid this issue. @aneshlya, what is your opinion about the proper PowerPC target?

.github/workflows/ppc64.yml
```diff
@@ -0,0 +1,67 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  ppc64-test:
+    name: PowerPC64LE generic-i32x4 test
+    runs-on: ubuntu-22.04
+    container:
+      image: ubuntu:24.04
```
runlevel5: @aneshlya alternatively we could run 25.10 and stick with Ubuntu's LLVM package, please let me know which option you prefer

runlevel5: I chose 22.04 deliberately because it is LTS

.github/workflows/ppc64.yml
```diff
@@ -0,0 +1,67 @@
...
+
+    - name: Build ISPC with PPC64 support
+      run: |
+        PATH=/usr/lib/llvm-20/bin:$PATH \
+          cmake -B build -DPPC64_ENABLED=ON
+        cmake --build build -j"$(nproc)"
+
+    - name: Run lit tests
+      run: |
+        cmake --build build --target check-all
```
nurmukhametov, line 99: Can you please run `run_tests.py` here too?

.github/workflows/ppc64.yml
nurmukhametov: I don't think that you chose 22.04 here

.github/workflows/ppc64.yml
runlevel5: Added

.github/workflows/ppc64.yml
runlevel5: Sorry typo, 24 instead of 22

aneshlya: 24 is OK

CMakeLists.txt
aneshlya: I agree that having a dedicated pwr-x4 target would be beneficial for consistency with support on other platforms. If it also helps address additional issues along the way, that provides even more motivation to introduce it.

runlevel5:
@aneshlya I am still trying to fix the CI on my own fork, please refrain from running the build on your side until I say it is ready. Many thanks

CMakeLists.txt
nurmukhametov: Please rebase your change on the current main branch. After https://github.com/ispc/ispc/pull/3739, you can omit this dependency and all other changes related to generic-target/ispc-opt dependency issues.

CMakeLists.txt
nurmukhametov: https://github.com/ispc/ispc/pull/3735#discussion_r2782678061

runlevel5:
@nurmukhametov I would like to seek your advice on the failed CI https://github.com/runlevel5/ispc/actions/runs/21974377177/job/63482625295?pr=1

I am considering of adding `REQUIRES: X86_ENABLED` to all tests that assume x86.

nurmukhametov:
> @nurmukhametov I would like to see your advice on the failed CI https://github.com/runlevel5/ispc/actions/runs/21974377177/job/63482625295?pr=1
> 
> I am considering of adding `REQUIRES: X86_ENABLED` to all tests that assume x86.

No, don't do that. You just need to remove `-DX86_ENABLED=OFF` from cmake command. In your setup, you're running lit tests on an x86 host, so `--target=host` correctly detects x86, but you've turned it off in your CMake configuration. 

Lit tests using `--target=host` don't hardcode a specific architecture. They rely on whatever the host provides. However, they still need to be set up and configured properly to work.

In this build configuration of the test job, we're mostly testing x86 (except for tests with an explicit `--target=ppc64le`, which require `REQUIRES: PPC64_ENABLED`), but that's fine unless we have ppc64 hardware to run `--target=host` tests against a real ppc64 target.

cmake/GenericTargets.cmake
```diff
@@ -115,5 +121,7 @@ function (generate_generic_builtins ispc_name)
     add_custom_target(generic-stdlib-bc DEPENDS ${GENERIC_STDLIB_BC_FILE})
     add_custom_target(generic-stdlib-cpp DEPENDS ${GENERIC_STDLIB_CPP_FILE})
 
+    set_source_files_properties(${GENERIC_TARGET_CPP_FILE} ${GENERIC_STDLIB_CPP_FILE} PROPERTIES GENERATED true)
     add_library(generic-target OBJECT EXCLUDE_FROM_ALL ${GENERIC_TARGET_CPP_FILE} ${GENERIC_STDLIB_CPP_FILE})
+    add_dependencies(generic-target generic-target-cpp generic-stdlib-cpp)
```
nurmukhametov: Let's remove this line. I think this is the artifact of the fix related to ispc-opt.

runlevel5:
@nurmukhametov thanks for your guidance. I'd say we should wait for @aneshlya to set up the native PPC64 CI runner so I could run the build on that host then the PR should be ready for reviewing

nurmukhametov:
> @nurmukhametov thanks for your guidance. I'd say we should wait for @aneshlya to set up the native PPC64 CI runner so I could run the build on that host then the PR should be ready for reviewing

We can, but the CI job testing build and run_tests under QEMU is anyway useful as it represents the development environment easily available to anyone. I'd prefer to have them both.

I have a fix for the job you created [here](https://github.com/nurmukhametov/ispc/commit/6c40ece84c226969b91a81e664d2ec59f95f8582); a corresponding CI [run](https://github.com/nurmukhametov/ispc/actions/runs/22006610742).

It also has two lit-tests failing, because the default LLVM 20 was used, which lacks some patches from https://github.com/ispc/ispc/tree/main/llvm_patches. This is fine until the LLVM tarballs [here](https://github.com/ispc/ispc.dependencies/releases) are rebuilt with the ppc64 target enabled.

runlevel5:
It is ready for review

runlevel5:
@aneshlya https://github.com/ispc/ispc/actions/runs/22025689444/job/63688757869 is expected
I have added `sudo` fix to https://github.com/ispc/ispc/actions/runs/22025689458/job/63688758079?pr=3735

nurmukhametov:
> It also has two lit-tests failing, because the default LLVM 20 was used, which lacks some patches from https://github.com/ispc/ispc/tree/main/llvm_patches. This is fine until the LLVM tarballs [here](https://github.com/ispc/ispc.dependencies/releases) are rebuilt with the ppc64 target enabled.

> @aneshlya https://github.com/ispc/ispc/actions/runs/22025689444/job/63688757869 is expected

I have actually built the LLVM tarball with PowerPC and uploaded it as llvm-21.1.8-ubuntu22.04-Release+Asserts-x86.arm.ppc.wasm.tar.xz [here](https://github.com/ispc/ispc.dependencies/releases/llvm-21.1-ispc-dev). All lit-tests are passing with it; CI run [link](https://github.com/nurmukhametov/ispc/actions/runs/22044966435/job/63692074455), CI job [patch](https://github.com/nurmukhametov/ispc/commit/6498dbe4fade1e70a4869023b844698c4703b3a7).

Out of run_tests, I see only compilation fails related to fp16 support (undefined reference to `__extendhfsf2`/`__truncsfhf2`). As far as I understand, they will be resolved in this CI job setup at the moment when we pick GCC 16 that added fp16 support to PowerPC libgcc [here](https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;h=698a22ded27f3a1b85f496b2a777730186994605).

aneshlya:
@nurmukhametov, thanks for uploading LLVM with ppc and your patch!
@runlevel5, please add the CI patch to this PR. You can skip fp16 tests for now on powerpc using `run_tests.py` skip mechanism for now.
After that I think this PR is good to go.

aneshlya:
@runlevel5, in case we decide to do an "official" PowerPC ispc binary - do you think the most beneficial is to have it PowerPC native (so users build and run on PowerPC), cross x86->PowerPC (users build on x86 and run on PowerPC) or both?

runlevel5:
> in case we decide to do an "official" PowerPC ispc binary - do you think the most beneficial is to have it PowerPC native (so users build and run on PowerPC), cross x86->PowerPC (users build on x86 and run on PowerPC) or both?

@aneshlya that is a great question... I do not speak for the whole PowerPC community but personally I would go with the native route. Cross-compilation is the last resort that most PowerPC users would have to go with.

runlevel5:
@nurmukhametov I've adapted your ppc64 pipeline in https://github.com/ispc/ispc/pull/3735/changes/f461cd6738f3c10b0dec80e4cff19dc8eec660c3

The native runner one I am unsure if would get far as I do not have permission to run it.

nurmukhametov:
> @nurmukhametov I've adapted your ppc64 pipeline in [f461cd6](https://github.com/ispc/ispc/commit/f461cd6738f3c10b0dec80e4cff19dc8eec660c3)

Great! Could you please skip all miscompiled float16 tests with 
```ispc
// rule: skip on arch=ppc64le
```

> The native runner one I am unsure if would get far as I do not have permission to run it.

At the moment, it fails to install `g++-multilib`. I suppose you need to drop this package and probably add `-DX86_ENABLED=OFF` to cmake command.

src/ispc.cpp
```diff
@@ -2232,6 +2272,14 @@ Target::Target(Arch arch, const char *cpu, ISPCTarget ispc_target, PICLevel picL
                 m_cpu = "";
             }
         }
+#endif
+#ifdef ISPC_PPC64_ENABLED
+        if (arch == Arch::ppc64le) {
+            if (m_cpu == "ppc64le") {
+                featuresString = "+vsx";
+                m_cpu = "pwr8";
```
nurmukhametov: I am confused with this cpu redefinition. Is "ppc64le" legitimate CPU in LLVM?

src/ispc.cpp
```diff
@@ -1109,7 +1143,7 @@ Target::Target(Arch arch, const char *cpu, ISPCTarget ispc_target, PICLevel picL
     }
 
     if (arch == Arch::none) {
-        arch = lGetArchFromTarget(m_ispc_target);
+        arch = lGetArchFromTarget(m_ispc_target, CPUID);
```
nurmukhametov: I don't like that we now pass CPUID to the function that maps target to arch. How about reverting this change and adding a fix-up for the specific case you addressed, right here, like:

```cpp
  if (arch == Arch::none) {
        arch = lGetArchFromTarget(m_ispc_target);
  }
  #ifdef ISPC_PPC64_ENABLED
  // comment why we need to fix-up arch returned earlier for this case
  // TODO: add a proper PowerPC target
  if (CPUID == CPU_PPC64LE_Generic) {
      arch = Arch::ppc64le;
  }
  #endif
```

src/ispc.cpp
runlevel5: `ppc64le` is not really a legitimate LLVM name, it was registered at `src/ispc.cpp:652`. However the `pwr8` is a legit name. Basically POWER8 is the first CPU that supports native little endian.

src/ispc.cpp
runlevel5: Please advise me when PPC64 get a dedicated target

runlevel5:
@nurmukhametov could you please help review my last commit? Many thanks

.github/workflows/ppc64.yml
```diff
@@ -0,0 +1,113 @@
...
+    runs-on: ubuntu-24.04-ppc64le
+    env:
+      LLVM_TAR: llvm-21.1.8-ubuntu22.04-Release+Asserts-x86.arm.ppc.wasm.tar.xz
+
+    steps:
+    - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+      with:
+        submodules: true
+
+    - name: Install dependencies
```
nurmukhametov, line 32: You can't use x86 prebuilt LLVM from ispc.dependencies on PowerPC. It fails to run: 
```bash
/tmp/_actions-runner-working-dir/ispc/ispc/bin-21.1/bin/llvm-config: cannot execute binary file: Exec format error
```

src/ispc.cpp
nurmukhametov: Let's then use `pwr8` as a CPU entry instead of `ppc64le`.

src/ispc.cpp
nurmukhametov: > Please advise me when PPC64 get a dedicated target

You can add right now before we merge it or we have an option to land this first and then add the dedicated target.

.github/workflows/ppc64.yml
```diff
@@ -0,0 +1,113 @@
...
+
+    - name: Run PPC64LE tests
+      run: |
+        PATH="$PWD"/build/bin:"$PATH" \
+        python3 scripts/run_tests.py \
+          --target=generic-i32x4 \
+          --arch=ppc64le \
+          --wrap-exe="qemu-ppc64le" \
+          --compiler=powerpc64le-linux-gnu-g++ \
+          -o O2
```
nurmukhametov, line 109: There are few tests that still need to be skipped on `ppc64le` due to fp16 support. 

https://github.com/ispc/ispc/actions/runs/22122113686/job/64041118211?pr=3735

```bash
NEW COMPFAILS:
	./tests/func-tests/broadcast-5.ispc
	./tests/func-tests/exclusive-scan-add-11.ispc
	./tests/func-tests/reduce-equal-14.ispc
	./tests/func-tests/reduce-equal-15.ispc
	./tests/func-tests/rotate-7.ispc
	./tests/func-tests/shift-4.ispc
	./tests/func-tests/shuffle-6.ispc
	./tests/func-tests/shuffle2-12.ispc
```

.github/workflows/ppc64.yml
runlevel5: @nurmukhametov do we have ppc64le prebuilt LLVM? If not, I could go with the llvm19 that Ubuntu 24 ships with but I still prefer if we use the same LLVM package if possible

src/ispc.cpp
runlevel5: I prefer to add the dedicated target later

src/ispc.cpp
```diff
@@ -2281,7 +2281,7 @@ Target::Target(Arch arch, const char *cpu, ISPCTarget ispc_target, PICLevel picL
 #endif
 #ifdef ISPC_PPC64_ENABLED
         if (arch == Arch::ppc64le) {
-            if (m_cpu == "ppc64le") {
+            if (m_cpu == "pwr8") {
                 featuresString = "+vsx";
                 m_cpu = "pwr8";
```
nurmukhametov: Is this line needed?

.github/workflows/ppc64.yml
```diff
@@ -0,0 +1,109 @@
...
+    steps:
+    - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+      with:
+        submodules: true
+
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y git curl cmake xz-utils m4 flex bison python3 python3-dev libtbb-dev \
+          llvm-19-dev clang-19 lld-19
```
nurmukhametov: You also need to install `libclang-19-dev` to fix these configuration errors:

```bash
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
clangASTMatchersPath
    linked by target "ispc-slim" in directory /tmp/_actions-runner-working-dir/ispc/ispc
    linked by target "ispc" in directory /tmp/_actions-runner-working-dir/ispc/ispc
    linked by target "libispc_shared" in directory /tmp/_actions-runner-working-dir/ispc/ispc
    linked by target "ispc-opt" in directory /tmp/_actions-runner-working-dir/ispc/ispc
clangBasicPath
    linked by target "ispc-slim" in directory /tmp/_actions-runner-working-dir/ispc/ispc
    linked by target "ispc" in directory /tmp/_actions-runner-working-dir/ispc/ispc
    linked by target "libispc_shared" in directory /tmp/_actions-runner-working-dir/ispc/ispc
    linked by target "ispc-opt" in directory /tmp/_actions-runner-working-dir/ispc/ispc
clangEditPath
    linked by target "ispc-slim" in directory /tmp/_actions-runner-working-dir/ispc/ispc
    linked by target "ispc" in directory /tmp/_actions-runner-working-dir/ispc/ispc
    linked by target "libispc_shared" in directory /tmp/_actions-runner-working-dir/ispc/ispc
    linked by target "ispc-opt" in directory /tmp/_actions-runner-working-dir/ispc/ispc
clangFrontendPath
    linked by target "ispc-slim" in directory /tmp/_actions-runner-working-dir/ispc/ispc
    linked by target "ispc" in directory /tmp/_actions-runner-working-dir/ispc/ispc
    linked by target "libispc_shared" in directory /tmp/_actions-runner-working-dir/ispc/ispc
    linked by target "ispc-opt" in directory /tmp/_actions-runner-working-dir/ispc/ispc
clangLexPath
    linked by target "ispc-slim" in directory /tmp/_actions-runner-working-dir/ispc/ispc
    linked by target "ispc" in directory /tmp/_actions-runner-working-dir/ispc/ispc
    linked by target "libispc_shared" in directory /tmp/_actions-runner-working-dir/ispc/ispc
    linked by target "ispc-opt" in directory /tmp/_actions-runner-working-dir/ispc/ispc
clangSupportPath
    linked by target "ispc-slim" in directory /tmp/_actions-runner-working-dir/ispc/ispc
    linked by target "ispc" in directory /tmp/_actions-runner-working-dir/ispc/ispc
    linked by target "libispc_shared" in directory /tmp/_actions-runner-working-dir/ispc/ispc
    linked by target "ispc-opt" in directory /tmp/_actions-runner-working-dir/ispc/ispc
```

src/ispc.cpp
```diff
@@ -638,6 +648,10 @@ class AllCPUs {
 #endif // ISPC_LLVM_VERSION
 #endif // ISPC_RISCV_ENABLED
 
+#ifdef ISPC_PPC64_ENABLED
+        names[CPU_PPC64LE_Generic].push_back("pwr8");
```
nurmukhametov: Can you make clear the naming of the enum entry for pwr8 CPU? It was generic, but it is pwr8 now.

.github/workflows/ppc64.yml
nurmukhametov: > @nurmukhametov do we have ppc64le prebuilt LLVM? If not, I could go with the llvm19 that Ubuntu 24 ships with but I still prefer if we use the same LLVM package if possible

No, we don't. LLVM 19 from Ubuntu 24 is fine with me.

elena901:
> it's very easy to find POWER8 hardware on eBay.

Just a reminder: POWER8 is PowerISA 2.07 only (no PowerISA 3.0+ support).

CMakeLists.txt
elena901: pwr-x4 is more grammatically correct.
ppc-x4 is more canonical.

Anyway the vector instruction set is called VSX, so vsx-x4 is also possible (analogous to rvv-x4).

docs/ReleaseNotes.txt
```diff
@@ -1,3 +1,14 @@
+=== v1.31.0 === (TBD)
+
+Experimental PowerPC64 Support:
+
+* Initial support for the PowerPC 64-bit little-endian (ppc64le) architecture
```
nurmukhametov: ReleaseNotes is typically updated when a release is being prepared, but we update the changelog in `docs/ispc.rst` with every user-visible PR. Please create a new section there and move this paragraph above this line: https://github.com/ispc/ispc/blob/0ebf3899c6383eb159cb4dca69a96c38dc655f80/docs/ispc.rst?plain=1#L260

docs/ReleaseNotes.txt
```diff
@@ -1,3 +1,14 @@
...
+  has been added using generic targets (e.g., generic-i32x4). The baseline is
+  POWER ISA 3.0 (POWER8+) with VSX. This support is experimental and not
```
nurmukhametov: Fix the error mentioned by @ elena901 here https://github.com/ispc/ispc/pull/3735#issuecomment-3934749012

nurmukhametov (changes requested):
Please fix the remaining docs issues (see inline comments) and it's good to go.

docs/ispc.rst
```diff
@@ -270,6 +270,15 @@ New Features:
   ``avx512gnr``, and ``avx10.2dmr`` targets. Please refer to
   `Intel AMX (Advanced Matrix Extensions)`_ for more details.
 
+New Architecture Support:
```
aneshlya: Please create a new section 
`Updating ISPC Programs For Changes In ISPC 1.31.0` and move this PowerPC architecture support there. Now it's in the section for previous release (1.30)

Issue #3737, opened 2026-02-08
ispc-opt has unnecessary dependency


nurmukhametov:
The `ispc-opt` currently links against the `builtin` target [(CMakeLists.txt)](https://github.com/ispc/ispc/blob/c2d88669f286d0ada8e5e646d42089302aeb6593/CMakeLists.txt#L1027), which contains all the generated bitcode libraries for every supported target/OS/architecture combination. However, `ispc-opt` never actually uses this bitcode. It only runs optimization passes on user-provided LLVM IR.

If I am not mistaken, the only reason for this dependency is a single validation check in the Target constructor here: https://github.com/ispc/ispc/blob/c2d88669f286d0ada8e5e646d42089302aeb6593/src/ispc.cpp#L2180

The `target_registry` is populated by self-registering BitcodeLib objects that are only instantiated when the `builtin` object files are linked in. Without `builtin`, the registry is empty and all targets fail validation, even though `ispc-opt` doesn't need the actual bitcode libraries.

aneshlya:
Was fixed by #3739

PR #3739, opened 2026-02-09
Remove unnecessary dependency of ispc-opt

nurmukhametov:
## Description

`ispc-opt` runs optimization passes on LLVM IR and does not require the bitcode libraries provided by the builtin target. Previously, it was linked against builtin primarily to satisfy a target validation check in the Target constructor that relies on the target registry.

This change adds a `disableTargetValidation` flag to `ispc::Globals` that is used in `ispc-opt` to skip the target support check.

This reduces the binary size of ispc-opt from ~9MB to 0.9MB.

## Related Issue
- [x] Linked to relevant issue: https://github.com/ispc/ispc/issues/3737

## Checklist


dbabokin:
@codex review

PR #3749, opened 2026-02-22
Fix #3748: Align varying bool<N> LLVMType with mask representation

### LINKED ISSUE #3748: compiler bug: call parameter type does not match function signature
Hi, I periodically hit this fatal compiler error when using bool vectors, so thought I should report it. Repro case as follows:

```ispc
inline varying bool all_elements(bool<3> v)
{
    return v.x && v.y && v.z;
}

varying bool foo(float<3> b, float<3> c) 
{
    #if 0
        // compiles
        const bool<3> tmp = (b == c);
        return all_elements(tmp);
    #else
        // does not compile
        return all_elements(b == c);
    #endif
}
```

This should produce something like the following (from ispc 1.30.0 in godbolt):

```
Call parameter type does not match function signature!
  %36 = insertvalue [3 x <8 x i32>] %33, <8 x i32> %35, 2, !dbg !109
 [3 x <8 x i8>]  %39 = call <8 x i32> @all_elements___vyb_3C_3_3E_([3 x <8 x i32>] %36, <8 x i32> %38), !dbg !112
/usr/local/src/ispc/src/main.cpp(40): FATAL ERROR: Unhandled signal sent to process; terminating.
***
***
Compiler returned: 139
```

Changing the `#if 0` in the repro to `#if 1` avoids the error. Thanks in advance!

nurmukhametov:
## Description
The commit f80c6e0c changed I1VecToBoolVec to produce [N x <W x iMASK>] for bool short-vector comparison results, intending to make varying bool<N> consistent with scalar varying bool (which uses BoolVectorType = iMASK internally). However it did not update lGetVectorLLVMType, which still returned [N x <W x i8>] (storage type) for VectorType::LLVMType() on varying bool<N>. This mismatch caused "Call parameter type does not match function signature" when a comparison result was passed directly to a function expecting bool<N>.

Fix: complete the original intent by removing the !base->IsUniformType() exception in lGetVectorLLVMType so that LLVMType() for varying bool<N> returns [N x <W x iMASK>] (consistent with I1VecToBoolVec), while LLVMStorageType() continues to return [N x <W x i8>]. Update TypeCastExpr to match: build the output array with LLVMType and use SwitchBoolToMaskType for bool elements in both the vector-to-vector and scalar-to-vector paths.

This design is now fully symmetric: both scalar varying bool and varying bool<N> use the target-specific mask type internally (e.g. i32 on AVX2, i1 on AVX-512) and convert to i8 only at store/load/ABI boundaries.

## Related Issue
- [X] Linked to relevant issue: #3748

## Checklist

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -1,5 +1,5 @@
 /*
-  Copyright (c) 2010-2025, Intel Corporation
+  Copyright (c) 2010-2026, Intel Corporation
 
   SPDX-License-Identifier: BSD-3-Clause
 */
@@ -7945,7 +7945,7 @@ llvm::Value *TypeCastExpr::GetValue(FunctionEmitContext *ctx) const {
         } else {
             // Emit instructions to do type conversion of each of the elements
             // of the vector.
-            llvm::Value *cast = llvm::UndefValue::get(toType->LLVMStorageType(g->ctx));
+            llvm::Value *cast = llvm::UndefValue::get(toType->LLVMType(g->ctx));
             for (int i = 0; i < toVector->GetElementCount(); ++i) {
                 llvm::Value *ei = ctx->ExtractInst(exprVal, i);
                 llvm::Value *conv = lTypeConvAtomicOrUniformVector(ctx, ei, toVector->GetElementType(),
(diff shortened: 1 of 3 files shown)


dbabokin:
@codex review

aneshlya (approved):
Thanks for your fix! Looks good!

Issue #3743, opened 2026-02-17
Ninja build and Cmake 3.19+ not caching build files


ZilverBlade:
I have tried compiling 2 ISPC projects (one is ispc_texcomp, and another is a library i wrote myself), and both always seem to be triggering a rebuild for the ispc files every single time i build (even when no changes have been made!). Here is a cmake file i'm using, as far as i know, I followed the documentation on how to set up an ISPC project on cmake.

```cmake
cmake_minimum_required(VERSION 3.26)
project(ispc_texcomp LANGUAGES CXX ISPC) 

add_library(ispc_texcomp STATIC
    ISPCTextureCompressor/ispc_texcomp/ispc_texcomp.cpp
    ISPCTextureCompressor/ispc_texcomp/kernel.ispc      
)
set_target_properties(ispc_texcomp PROPERTIES 
    ISPC_INSTRUCTION_SETS "avx512skx-i32x16;avx2-i32x8;sse4-i32x4;sse2-i32x4"
    ISPC_HEADER_SUFFIX "_ispc.h"
)
target_include_directories(ispc_texcomp PUBLIC 
    ${CMAKE_CURRENT_SOURCE_DIR}/ISPCTextureCompressor
    $<TARGET_PROPERTY:ISPC_HEADER_DIRECTORY>
)
```

aneshlya:
@azwolski, were you able to reproduce it? I don't see rebuild of ispc files on my system.

azwolski:
I am unable to reproduce it as well, i am using the same cmake file, with `ISPCTextureCompressor` repo
```
ninja: no work to do.
```


@ZilverBlade could you please provide some logs `ninja -C build -d explain`?

ZilverBlade:
@aneshlya  @azwolski  I think I've might've found how to reproduce issue, there seems to be an issue when there are spaces in the path that the ISPC compiler is building
```
ninja explain: D:/Directories/Documents/Visual is dirty
ninja explain: Studio is dirty
ninja explain: 2019/Projects/Shard3D-Next/vendor/ISPCTextureCompressor/ispc_texcomp/kernel.ispc has no in-edge and is missing
ninja explain: 2019/Projects/Shard3D-Next/vendor/ISPCTextureCompressor/ispc_texcomp/kernel.ispc is dirty
```

azwolski:
[...]

PR #3753, opened 2026-02-25
Add path escaping for make/Ninja depfile compatibility

azwolski:
## Description
Fix depfile generation when source paths contain spaces.

ISPC emits make-style depfiles for `-M -MF`. Paths containing spaces were written without proper escaping (e.g. `/path/with space/file.ispc`), which violates make depfile syntax. Ninja then parsed them as multiple paths, causing targets to be always considered dirty and rebuilt on every invocation.

This patch escapes special characters in depfile output (e.g. space → `\ `, `\` → `\\`, `#` → `\#`, `$` → `$$`, `:` → `\:`).

After this change, CMake + Ninja builds are stable even when project paths contain spaces.

## Related Issue
- [ ] https://github.com/ispc/ispc/issues/3743

## Checklist


azwolski:
This would probably need some kind of regression test, but i am not sure how should it be implemented

dbabokin:
@codex review

dbabokin:
I think we should explicitly frame this around GCC/Clang behavior, because for depfile syntax/escaping they are the de-facto compatibility reference points used by CMake/Ninja/Make ecosystems.

Context for why `-MT` / `-MQ` came up:
- This PR fixes depfile parsing/caching when paths contain spaces (issue #3743), which is absolutely the right area.
- While reviewing, I noticed the same escaping function is currently applied to both depfile targets and dependency filenames.
- In GCC/Clang those are intentionally treated differently, and that is where `-MT` vs `-MQ` matters.

Terminology in depfile rule:
- LHS = target side (left-hand side), before `:`
  Example: `foo.o` in `foo.o: a.h b.h`
- RHS = dependency filenames (right-hand side), after `:`
  Example: `a.h b.h`

Why LHS and RHS need different rules:
- LHS may be user-authored expressions (`-MT '$(OBJDIR)/foo.o'`) that are intentionally interpreted by Make.
- RHS are filesystem paths discovered by preprocessing/scanning and should be escaped so parsers (Make/Ninja dep parser) don’t split/comment-expand tokens incorrectly.

Detailed behavior (4 scenarios):

1. Explicit raw target (`-MT <t>`)
- Reference behavior (GCC/Clang): write target as-is (raw).
- Example:
  - input: `-MT '$(OBJDIR)/foo.o'`
  - depfile LHS: `$(OBJDIR)/foo.o:`
- Effect: Make variable expansion is preserved.

2. Explicit quoted target (`-MQ <t>`)
- Reference behavior (GCC/Clang): write Make-quoted target.
- Example:
  - input: `-MQ '$(OBJDIR)/foo.o'`
  - depfile LHS: `$$(OBJDIR)/foo.o:`
- Effect: special chars are literalized (not expanded as Make syntax).

3. Default target (no `-MT`/`-MQ`)
- Reference behavior (GCC/Clang): auto-derived target is quoted (effectively like `-MQ`).
- Rationale: generated target path should be robust if it has special characters.

4. Dependency filenames on RHS
- Needed for #3743 regardless of `-MT/-MQ`.
- RHS paths should be depfile-escaped so Ninja/Make parse one path token per real path.
- Example: `D:/My Directory/project/file.ispc` must not be split into `D:/My` + `Directory/...`.

Do we need `-MQ` to fix #3743?
- Strictly no. #3743 is solved by correctly escaping RHS dependency filenames.
- But supporting `-MQ` is still recommended for GCC/Clang parity and to keep explicit semantics clear.

Recommended implementation in ispc:
- Keep RHS escaping fix (core of #3743).
- Do not force-escape explicit `-MT` target (keep it raw).
- Auto-quote only default (implicit) target.
- Add `-MQ` for explicit quoted target semantics.
- Keep LHS and RHS escaping logic separated in code (two helpers), not one shared function.

This keeps the Windows/Ninja cache fix, while avoiding behavioral drift for existing `-MT` users and aligning with GCC/Clang contracts.

dbabokin (changes requested):
I dived into this issues with Codex and the comment I has posted is explanatory comment that summarizes my findings and the general plan of what we need to fix that properly.

Please let me know if you need more info or have any question.

azwolski:
@dbabokin bump

azwolski:
@dbabokin is it good now?

azwolski:
@aneshlya

PR #3788, opened 2026-04-09
Massively speed-up 32-bit shuffles on the target avx2-i32x16

### LINKED ISSUE #3771: Inefficient shuffle implementation
In **single precision**, `shuffle` with a dynamic index is fast since [it generates a `vpermps` instruction](https://godbolt.org/z/57eYKKWrz) using the `avx2-i32x8` target. [Same thing](https://godbolt.org/z/b33Wqjv4z) with the target `avx2-i32x4` (or even the `avx2-i64x4` one). However, when it comes to the target `avx2-i32x16`, the [generated code](https://godbolt.org/z/bzad8avd7) is very inefficient: a lot of instructions (~90) are generated, instead of 2 `vpermps` with possibly few more instructions (to consider bigger IDs and wider gangs) like blending instructions.

-----

In addition, regarding **double-precision**, the generated code is inefficient:

On the target [`avx2-i64x4`](https://godbolt.org/z/9xoM3nTxd), there are 1 `vpaddq`, 1 `vpcmpgtq`, 2 `vpermpd`, 2 `vpermilpd`, a `vblendvpd` and few other instructions. The main reason (partially) justifying a bigger code is that there are AFAIK no 64-bit permute across lanes supporting dynamic indices in AVX-2 (i.e. `vpermpd` with an `ymm` register for the last operand instead of an `imm8`). That being said, there is a 32-bit instruction for that: `vpermps`. Thus, I wonder if we could use a single `vpermps` using an indexing trick so to move the upper/lower half of each 64-bit value. This should be faster since permutations are quite expensive (especially on Intel CPUs). Something like that:
 - `vpshufd` of `xmm1` (containing the 4 ids) so to get `[id0, id0, id1, id1]`, stored in `xmm2`
 - `vpshufd` of `xmm1` so to get `[id2, id2, id3, id3]`, stored in `xmm3`
 - `vinserti128 ymm2, xmm2, xmm3, 1` so to merge the two xmm registers
 - Load a ymm3 register containing `[0, 1, 0, 1, 0, 1, 0 1]`
 - `vpaddd ymm2, ymm2, ymm2`so ymm2 contains `[id0*2, id0*2+1, id1*2, id1*2, etc.]`
 - `vpaddd ymm2, ymm2, ymm3` so ymm2 contains `[id0*2, id0*2+1, id1*2, id1*2+1, etc.]`
 - Final `vpermps ymm2, ymm0, ymm2` so to actually do the shuffle

The 3 first operations can actually be replaced with a single permutation using a `vpermd` + load of the (constant) permutation indices. This looks like a lot and not significantly better, but the 2 loads are constants so they can be hoisted in hot loops and loaded only once in double-pumping targets. Regarding the `vpaddd`, they are rather cheap (i.e. 3 instruction/cycle on Skylake CPUs or more recent). In the end, there will be like 6 instructions for the new shuffle and often 4 in loops in practice (where the IDs are not constant). This is better than the current 9 instructions (certainly 8 in loops). Moreover, each of the 2 `vpermpd`, 2 `vpermilpd` and 1 `vblendvpd` should take ~1 cycle of throughput on most Intel CPUs (from Skylake to AlderLake). This means at least 5 cycles of throughput compared to only 3 for the proposed approach.

On the target [`avx2-i32x8`](https://godbolt.org/z/asfr73Wx8), the double-precision shuffles are *very inefficient*. I do not really see why the code is not just about twice bigger than the `avx2-i32x4` target, with certainly few more instruction just to consider bigger IDs and wider gangs... There are 49 instructions (with a lot of load and silly replicated rather-expensive stores). We could certainly perform a `vblendvpd` on the two `ymm` registers containing the values (i.e. `v` in the code) based on the indices and then duplicate the code of the `avx2-i32x8` target (similarly of what this target actually does). Even better, we could use the above method twice, certainly resulting in only about 11~15 instructions (instead of the current 49 instructions).

On [neon targets](https://godbolt.org/z/v9r6j7d1q), it does not look great either. I wonder whether [`TBX`/`TBL` instructions](https://developer.arm.com/documentation/102159/0400/Permutation---Neon-instructions) could be used instead. It looks like a good fit at first glance and much better than the currently-generated code.

Fortunately, on AVX-512 targets, there are no issues on usual targets (see [`avx512-x4`](https://godbolt.org/z/fqr6an57v),  [`avx512-x8`](https://godbolt.org/z/1a655jjfK),  [`avx512-x16`](https://godbolt.org/z/7164K7YP9)). Only the [`avx512skx-x32`](https://godbolt.org/z/bxqPnjG7j) target is not optimal since it uses 4 gather instructions (certainly about twice slower than an optimal code).

zephyr111:
This PR significantly improve the code generated by **32-bit** (i.e. `int` and `float`) **shuffle functions on the target avx2-i32x16** when the index is not a compile-time constant. It fixes the first issue stated in #3771. On my i5-9600KF CPU, calls to `shuffle(value, index)` are now about **5~10 times faster than the current code**, and calls to `shuffle(value1, value2, index)` are now about **1.9~2.6 times faster** than the current code. I used the benchmark `09_shuffle` to measure that. I added a simple lit-test and expect the benchmark to also check values. Just to be on the safe side, I manually checked the shuffles using the following code:

```
noinline unmasked float f32_test1(float v, int id) { return shuffle(v, id); }
noinline unmasked int i32_test1(int v, int id) { return shuffle(v, id); }
noinline unmasked float f32_test2(float v1, float v2, int id) { return shuffle(v1, v2, id); }
noinline unmasked int i32_test2(int v1, int v2, int id) { return shuffle(v1, v2, id); }

extern "C" uniform int main(uniform int argc, uniform int8 argv[])
{
    int i32_values = programIndex;
    float f32_values = programIndex + 0.5;
    int ids = {1, 4, 0, 9, 3, 6, 2, 7,  5, 0, 11, 17, 29, 1, 16, 15};
    print("%\n", i32_test1(i32_values, ids));
    print("%\n", f32_test1(f32_values, ids));
    print("%\n", i32_test2(i32_values, i32_values+100, ids));
    print("%\n", f32_test2(f32_values, f32_values+100.0, ids));
    return 0;
}
```

This program prints:

```
[1,4,0,9,3,6,2,7,5,0,11,9,13,1,8,15]
[1.500000,4.500000,0.500000,9.500000,3.500000,6.500000,2.500000,7.500000,5.500000,0.500000,11.500000,9.500000,13.500000,1.500000,8.500000,15.500000]
[1,4,0,9,3,6,2,7,5,0,11,101,113,1,100,15]
[1.500000,4.500000,0.500000,9.500000,3.500000,6.500000,2.500000,7.500000,5.500000,0.500000,11.500000,101.500000,113.500000,1.500000,100.500000,15.500000]
```

The results seem correct to me, at least as long as the index does not be bigger or equal to the gang size (i.e. `index < 16`) for the first shuffle, or twice the gang size for the second shuffle (i.e. `index < 32`). This matches with the constraint stated in the documentation for the first shuffle:

> The provided values for permutation must all be between 0 and the gang size. 

That being said, there is no explicit constraint stated in the doc for the second shuffle.

Here are raw benchmark results for 32-bit types (other types are left unchanged):

```
Benchmark function      |  Main branch   |    This PR     |     ~iter
------------------------|----------------|----------------|-----------
shuffle1_int/64         |    35.3 ns     |    5.01 ns     |   19859621
shuffle1_int/128        |    67.3 ns     |    8.50 ns     |    9486617
shuffle1_int/256        |     129 ns     |    15.4 ns     |    5261903
shuffle1_int/512        |     260 ns     |    34.0 ns     |    2684585
shuffle1_int/1024       |     554 ns     |    57.5 ns     |    1253011
shuffle1_int/2048       |    1161 ns     |     118 ns     |     614145
shuffle1_int/4096       |    2378 ns     |     231 ns     |     296468
shuffle1_int/8192       |    5011 ns     |     570 ns     |     139482
shuffle1_int/16384      |   10087 ns     |    1396 ns     |      69114
shuffle1_int/32768      |   20441 ns     |    3014 ns     |      34283
shuffle1_int/65536      |   41856 ns     |    8037 ns     |      16709
shuffle1_int_BigO       |    0.63 N      |    0.11 N      | 
shuffle1_int_RMS        |       3 %      |      28 %      | 
------------------------|----------------|----------------|-----------
shuffle1_float/64       |    35.3 ns     |    5.01 ns     |   19848737
shuffle1_float/128      |    67.0 ns     |    8.49 ns     |    9899539
shuffle1_float/256      |     130 ns     |    15.5 ns     |    5412408
shuffle1_float/512      |     260 ns     |    33.8 ns     |    2694908
shuffle1_float/1024     |     542 ns     |    57.4 ns     |    1295697
shuffle1_float/2048     |    1064 ns     |     117 ns     |     655429
shuffle1_float/4096     |    2241 ns     |     230 ns     |     313097
shuffle1_float/8192     |    4979 ns     |     582 ns     |     140410
shuffle1_float/16384    |   10058 ns     |    1412 ns     |      69525
shuffle1_float/32768    |   20310 ns     |    3018 ns     |      34479
shuffle1_float/65536    |   41771 ns     |    8039 ns     |      16743
shuffle1_float_BigO     |    0.63 N      |    0.11 N      | 
shuffle1_float_RMS      |       3 %      |      27 %      | 
------------------------|----------------|----------------|-----------
shuffle2_int/64         |    22.4 ns     |    10.0 ns     |   31227547
shuffle2_int/128        |    41.0 ns     |    17.9 ns     |   16683396
shuffle2_int/256        |    76.9 ns     |    33.5 ns     |    9336087
shuffle2_int/512        |     159 ns     |    69.2 ns     |    4386883
shuffle2_int/1024       |     321 ns     |     128 ns     |    2181363
shuffle2_int/2048       |     642 ns     |     257 ns     |    1091682
shuffle2_int/4096       |    1406 ns     |     544 ns     |     497426
shuffle2_int/8192       |    2842 ns     |    1125 ns     |     245872
shuffle2_int/16384      |    5674 ns     |    2236 ns     |     123367
shuffle2_int/32768      |   11771 ns     |    5572 ns     |      59473
shuffle2_int/65536      |   23402 ns     |   12398 ns     |      29897
shuffle2_int_BigO       |    0.36 N      |    0.18 N      | 
shuffle2_int_RMS        |       2 %      |      16 %      | 
------------------------|----------------|----------------|-----------
shuffle2_float/64       |    22.3 ns     |    10.0 ns     |   31317244
shuffle2_float/128      |    42.0 ns     |    17.8 ns     |   17153561
shuffle2_float/256      |    76.6 ns     |    33.5 ns     |    9344551
shuffle2_float/512      |     159 ns     |    69.1 ns     |    4409842
shuffle2_float/1024     |     325 ns     |     128 ns     |    2151894
shuffle2_float/2048     |     642 ns     |     257 ns     |    1089692
shuffle2_float/4096     |    1406 ns     |     544 ns     |     496659
shuffle2_float/8192     |    2846 ns     |    1126 ns     |     246214
shuffle2_float/16384    |    5679 ns     |    2230 ns     |     123158
shuffle2_float/32768    |   11787 ns     |    5576 ns     |      59584
shuffle2_float/65536    |   23396 ns     |   12398 ns     |      29926
shuffle2_float_BigO     |    0.36 N      |    0.18 N      | 
shuffle2_float_RMS      |       2 %      |      16 %      | 
```

Besides, I also modified a bit the code so to consistently use `WIDTH` on both the const and not-const version of the shuffle (which does not impact the generated code produced by this part). I also moved some macro code so to be able to use it so make the code smaller and more clear.

## Related Issue

## Checklist


dbabokin:
@codex review

dbabokin:
@codex review

nurmukhametov (commented):
Could you also test `13_rotate` and `14_shift`? They are likely affected by this change as well.

I  would also recommend using [this script](https://github.com/ispc/ispc/blob/main/scripts/compare_func.sh) to do a side-by-side comparison of the old and new versions across different targets. It is particularly useful for evaluating performance across different CPU generations and vendors (e.g. Intel Skylake, Intel Panther Lake, AMD Zen, and etc.). It uses `llvm-mca` and gives you a performance report with latency, throughput and etc..

zephyr111:
@nurmukhametov

It surprisingly looks like the PR does not impact the `rotate` and `shift` functions. Here are results:

```
-----------------------------------------------------------------------------
Benchmark function           |   Main branch   |    This PR     |     ~iter
-----------------------------------------------------------------------------
rotate_int32/256             |      78.3 ns    |      78.4 ns   |   1787247
rotate_int32/4096            |      1270 ns    |      1267 ns   |    110261
rotate_int32/32768           |     11284 ns    |     11078 ns   |     12338
rotate_int32/1048576         |    472386 ns    |    469980 ns   |       297
rotate_const_int32/256       |      13.9 ns    |      13.9 ns   |   9992535
rotate_const_int32/4096      |       226 ns    |       226 ns   |    611120
rotate_const_int32/32768     |      2971 ns    |      2965 ns   |     46976
rotate_const_int32/1048576   |    250646 ns    |    253211 ns   |       555
rotate_float/256             |      78.3 ns    |      78.4 ns   |   1785430
rotate_float/4096            |      1270 ns    |      1270 ns   |    110115
rotate_float/32768           |     11044 ns    |     11047 ns   |     12596
rotate_float/1048576         |    378700 ns    |    379489 ns   |       369
rotate_const_float/256       |      21.8 ns    |      20.9 ns   |   6691857
rotate_const_float/4096      |       336 ns    |       336 ns   |    416792
rotate_const_float/32768     |      3107 ns    |      3123 ns   |     45113
rotate_const_float/1048576   |    167276 ns    |    166911 ns   |       821
-----------------------------------------------------------------------------
shift_int32/256              |      74.6 ns    |     74.8 ns    |   1875657
shift_int32/4096             |      1206 ns    |     1202 ns    |    115729
shift_int32/32768            |     11279 ns    |    11229 ns    |     12317
shift_int32/1048576          |    446056 ns    |   444131 ns    |       314
shift_const_int32/256        |      10.7 ns    |     10.7 ns    |  12381463
shift_const_int32/4096       |       182 ns    |      182 ns    |    769312
shift_const_int32/32768      |      3135 ns    |     2644 ns    |     44544
shift_const_int32/1048576    |    252950 ns    |   250844 ns    |       550
shift_float/256              |      74.6 ns    |     74.6 ns    |   1872443
shift_float/4096             |      1203 ns    |     1202 ns    |    115816
shift_float/32768            |     11300 ns    |    11247 ns    |     12301
shift_float/1048576          |    383613 ns    |   382047 ns    |       365
shift_const_float/256        |      13.9 ns    |     13.9 ns    |  10015129
shift_const_float/4096       |       227 ns    |      227 ns    |    614904
shift_const_float/32768      |      3140 ns    |     2979 ns    |     44550
shift_const_float/1048576    |    169582 ns    |   168706 ns    |       808
```

I was surprised so I check the compilation flags and benchmark flags but this looks fine:
```
BENCHMARKS_ISPC_TARGETS: avx2-i32x16
BENCHMARKS_ISPC_FLAGS: -O3 --woff
CMAKE_BUILD_TYPE: Release
```

I also run this again multiple time with a shorter benchmark time. It is less precise, but enough (i.e. <~5%) to see there is not much difference between the main branch and this PR on the 2 other functions (while there is still a clear difference on shuffles):

```
# branch: main

$ (./build/bin/09_shuffle --benchmark_min_time=0.02s; ./build/bin/13_rotate --benchmark_min_time=0.02s; ./build/bin/14_shift --benchmark_min_time=0.02s) 2>/dev/null | grep -E "(rotate|shift|shuffle[12])_float/32768"
shuffle1_float/32768        20360 ns        20359 ns         1377
shuffle2_float/32768        11676 ns        11677 ns         2397
rotate_float/32768                11037 ns        11037 ns         2538
shift_float/32768                11192 ns        11192 ns         2502
$ (./build/bin/09_shuffle --benchmark_min_time=0.02s; ./build/bin/13_rotate --benchmark_min_time=0.02s; ./build/bin/14_shift --benchmark_min_time=0.02s) 2>/dev/null | grep -E "(rotate|shift|shuffle[12])_float/32768"
shuffle1_float/32768        20483 ns        20483 ns         1366
shuffle2_float/32768        11694 ns        11695 ns         2391
rotate_float/32768                11119 ns        11120 ns         2520
shift_float/32768                11195 ns        11196 ns         2500

# branch: opt-32bit-shuffle-avx2-i32x16
$ (./build/bin/09_shuffle --benchmark_min_time=0.02s; ./build/bin/13_rotate --benchmark_min_time=0.02s; ./build/bin/14_shift --benchmark_min_time=0.02s) 2>/dev/null | grep -E "(rotate|shift|shuffle[12])_float/32768"
shuffle1_float/32768         2960 ns         2960 ns         9452
shuffle2_float/32768         5463 ns         5463 ns         5147
rotate_float/32768                10917 ns        10918 ns         2562
shift_float/32768                11272 ns        11268 ns         2498
$ (./build/bin/09_shuffle --benchmark_min_time=0.02s; ./build/bin/13_rotate --benchmark_min_time=0.02s; ./build/bin/14_shift --benchmark_min_time=0.02s) 2>/dev/null | grep -E "(rotate|shift|shuffle[12])_float/32768"
shuffle1_float/32768         3046 ns         3047 ns         9176
shuffle2_float/32768         5689 ns         5689 ns         4943
rotate_float/32768                11007 ns        11007 ns         2540
shift_float/32768                11250 ns        11250 ns         2491
```

zephyr111:
@nurmukhametov

Please note that I also ran the shuffle test on a Zen3 CPU and I get an average speed-up of respectively ~5.5 for `shuffle1` and ~1.55 for `shuffle2`.

Regarding MCA, I use to check the generated code on Godbolt so far (only on Skylake, Zen3, Zen5), and then I tend to analyze results with uiCA (rather than MCA) since uiCA is significantly more accurate (though it is unfortunately often down and it does not support AMD CPUs yet as well as recent Intel CPUs). For example, the analysis report for gather instructions is completely wrong on many CPUs (I tested this on zen3/zen4/zen5 and it was considered as MUCH cheaper than it really is). The scheduling of several SIMD instructions on Zen CPUs looked suspiciously wrong IMHO. MCA is still pretty useful for AMD and ARM CPUs so to get a rough estimate of what could happend when we do not have the target CPU.

Regarding the tool, it is not clear to me what happens when we give both the target and the CPU. Like for example: `./compare_func.sh min_v avx2-i32x8 t.ispc "-O2 --cpu=skx"`. I tried this and I got a code using zmm register so it is not really an avx2 target but avx512 one, and in this case I guess we should rather specify the target `avx512skx-x8`. In such case, it looks like LLVM can replace the very inefficient code by a simple `vpermps zmm0, zmm1, zmm0` but I think this is not a normal use-case. Thus, I finally do not use `--cpu`. Here is my command-line:

```
OLD_ISPC=$HOME/Bureau/projects/ispc/main-build/bin/ispc NEW_ISPC=$HOME/Bureau/projects/ispc/build/bin/ispc PATH=$HOME/Bureau/projects/ispc/llvm-20.1.4-ubuntu22.04-Release-x86.arm.wasm/bin-20.1/bin:$PATH ~/Bureau/projects/ispc/scripts/compare_func.sh f32_shuffle1 avx2-i32x16 test.ispc "-O2" skylake
```

Based on the (maximum) CPU port pressure, I get the following speed-up with MCA:

```
f32_shuffle1:
    - skylake: 6.6x          (5.7x in practice in average)
    - icelake-client: 4.4x
    - pantherlake: 4.3x
    - znver3: 10x            (5.5x in practice in average)
    - znver5: 10x
f32_shuffle2:
    - skylake: 2.8x          (2.0x in practice in average)
    - icelake-client: 2.5x
    - pantherlake: 1.9x
    - znver3: 2.9x            (1.55x in practice in average)
    - znver5: 2.9x
```

By the way, with uiCA, I get 6.2x for `f32_shuffle1` and 2.5x for `f32_shuffle2`, which is a bit closer to the real-world results.

zephyr111:
@nurmukhametov

Regarding `rotate`, the assembly code is actually the same between the two branches. Its efficiency seems fine to me:

```
=== NEW ISPC (Intel(r) Implicit SPMD Program Compiler (Intel(r) ISPC), 1.31.0dev (build commit a6c207b2f72ea2f8 @ 20260410, LLVM 20.1.4)) ===
	push	rbp
	mov	rbp, rsp
	and	rsp, -64
	sub	rsp, 192
	vmovaps	ymmword ptr [rsp + 32], ymm1
	vmovaps	ymmword ptr [rsp], ymm0
	vmovaps	ymmword ptr [rsp + 96], ymm1
	vmovaps	ymmword ptr [rsp + 64], ymm0
	and	edi, 15
	vmovups	ymm0, ymmword ptr [rsp + 4*rdi]
	vmovups	ymm1, ymmword ptr [rsp + 4*rdi + 32]
	mov	rsp, rbp
	pop	rbp
	ret

=== OLD ISPC (Intel(r) Implicit SPMD Program Compiler (Intel(r) ISPC), 1.31.0dev (build commit fa6734f837c447fc @ 20260410, LLVM 20.1.4)) ===
	push	rbp
	mov	rbp, rsp
	and	rsp, -64
	sub	rsp, 192
	vmovaps	ymmword ptr [rsp + 32], ymm1
	vmovaps	ymmword ptr [rsp], ymm0
	vmovaps	ymmword ptr [rsp + 96], ymm1
	vmovaps	ymmword ptr [rsp + 64], ymm0
	and	edi, 15
	vmovups	ymm0, ymmword ptr [rsp + 4*rdi]
	vmovups	ymm1, ymmword ptr [rsp + 4*rdi + 32]
	mov	rsp, rbp
	pop	rbp
	ret

=== DETAILED DIFF (- OLD, + NEW) ===
No differences found
```

A similar method is certainly use for shifts.

nurmukhametov:
> @nurmukhametov
> 
> Regarding `rotate`, the assembly code is actually the same between the two branches. Its efficiency seems fine to me:

Right. You need to enable *_via_shuffle paths like 
```bash
--- a/stdlib/include/core.isph
+++ b/stdlib/include/core.isph
@@ -184,6 +184,11 @@ static const int32 programIndex = {ISPC_PROGRAM_INDEX_INITIALIZER};
 #else
 #define ISPC_TARGET_HAS_ROTATE_VIA_SHUFFLE_64_VAL 0
 #endif
+#elif defined(ISPC_TARGET_AVX2) && TARGET_WIDTH == 16
+#define ISPC_TARGET_HAS_ROTATE_VIA_SHUFFLE_8_VAL 0
+#define ISPC_TARGET_HAS_ROTATE_VIA_SHUFFLE_16_VAL 0
+#define ISPC_TARGET_HAS_ROTATE_VIA_SHUFFLE_32_VAL 1
+#define ISPC_TARGET_HAS_ROTATE_VIA_SHUFFLE_64_VAL 0
 #else
 #define ISPC_TARGET_HAS_ROTATE_VIA_SHUFFLE_8_VAL 0
 #define ISPC_TARGET_HAS_ROTATE_VIA_SHUFFLE_16_VAL 0
@@ -205,6 +210,11 @@ static const int32 programIndex = {ISPC_PROGRAM_INDEX_INITIALIZER};
 #else
 #define ISPC_TARGET_HAS_SHIFT_VIA_SHUFFLE_64_VAL 0
 #endif
+#elif defined(ISPC_TARGET_AVX2) && TARGET_WIDTH == 16
+#define ISPC_TARGET_HAS_SHIFT_VIA_SHUFFLE_8_VAL 0
+#define ISPC_TARGET_HAS_SHIFT_VIA_SHUFFLE_16_VAL 0
+#define ISPC_TARGET_HAS_SHIFT_VIA_SHUFFLE_32_VAL 1
+#define ISPC_TARGET_HAS_SHIFT_VIA_SHUFFLE_64_VAL 0
 #else
 #define ISPC_TARGET_HAS_SHIFT_VIA_SHUFFLE_8_VAL 0
 #define ISPC_TARGET_HAS_SHIFT_VIA_SHUFFLE_16_VAL 0
```
this to test these implementations:
https://github.com/ispc/ispc/blob/7af581e14b88af3ed69bf2a6f0a58afccdc74803/stdlib/stdlib.ispc#L100 and https://github.com/ispc/ispc/blob/7af581e14b88af3ed69bf2a6f0a58afccdc74803/stdlib/stdlib.ispc#L495

zephyr111:
Ok, indeed.

Here are short results (with the via-shuffle):

```
main:
  shuffle1_float/32768        20518 ns
  shuffle2_float/32768        11600 ns
  rotate_float/32768          20650 ns
  shift_float/32768           20243 ns

This PR:
  shuffle1_float/32768         2967 ns
  shuffle2_float/32768         5992 ns
  rotate_float/32768           2912 ns
  shift_float/32768            3400 ns
```

Here are complete results for 32-bit rotate/shifts (with the via-shuffle):

```
-----------------------------------------------------------------------------
Benchmark function           |   Main branch   |    This PR     |     ~iter
-----------------------------------------------------------------------------
rotate_int32/256             |       130 ns    |      15.4 ns   |   4533042
rotate_int32/4096            |      2339 ns    |       230 ns   |    303975
rotate_int32/32768           |     20467 ns    |      2747 ns   |     25579
rotate_int32/1048576         |    748723 ns    |    273943 ns   |       255
rotate_const_int32/256       |      13.9 ns    |      13.9 ns   |   4996202
rotate_const_int32/4096      |       226 ns    |       236 ns   |    293366
rotate_const_int32/32768     |      2958 ns    |      3341 ns   |     20916
rotate_const_int32/1048576   |    247729 ns    |    270897 ns   |       257
rotate_float/256             |       130 ns    |      15.4 ns   |   4533134
rotate_float/4096            |      2327 ns    |       230 ns   |    303774
rotate_float/32768           |     20468 ns    |      3066 ns   |     22795
rotate_float/1048576         |    683429 ns    |    174052 ns   |       398
rotate_const_float/256       |      20.9 ns    |      11.6 ns   |   5968625
rotate_const_float/4096      |       337 ns    |       232 ns   |    305425
rotate_const_float/32768     |      3225 ns    |      3294 ns   |     21196
rotate_const_float/1048576   |    174829 ns    |    174348 ns   |       398
-----------------------------------------------------------------------------
shift_int32/256              |       126 ns    |     16.8 ns    |   4177426
shift_int32/4096             |      2177 ns    |      293 ns    |    236611
shift_int32/32768            |     20374 ns    |     3290 ns    |     21236
shift_int32/1048576          |    787016 ns    |   279861 ns    |       249
shift_const_int32/256        |      10.7 ns    |     10.9 ns    |   6514935
shift_const_int32/4096       |       182 ns    |      182 ns    |    384126
shift_const_int32/32768      |      3305 ns    |     2818 ns    |     24878
shift_const_int32/1048576    |    260829 ns    |   265048 ns    |       260
shift_float/256              |       124 ns    |     16.7 ns    |   4177986
shift_float/4096             |      2139 ns    |      293 ns    |    238233
shift_float/32768            |     20109 ns    |     3136 ns    |     22181
shift_float/1048576          |    669637 ns    |   173328 ns    |       397
shift_const_float/256        |      14.0 ns    |     13.9 ns    |   5011649
shift_const_float/4096       |       227 ns    |      228 ns    |    306363
shift_const_float/32768      |      2989 ns    |     2855 ns    |     24471
shift_const_float/1048576    |    167961 ns    |   169152 ns    |       405
```

Interestingly, the const version is now almost as fast as the non-const version. In fact, we could enable the via-shuffle implementation back now since it is fast on this specific target:

```
----------------------------------------------------------------
                             |   Main branch   |    This PR     
Benchmark function           |  current implem |                
                             |    via memory   |  via shuffles  
----------------------------------------------------------------
rotate_int32/256             |      78.3 ns    |      15.4 ns   
rotate_int32/4096            |      1270 ns    |       230 ns   
rotate_int32/32768           |     11284 ns    |      2747 ns   
rotate_int32/1048576         |    472386 ns    |    273943 ns   
rotate_const_int32/256       |      13.9 ns    |      13.9 ns   
rotate_const_int32/4096      |       226 ns    |       236 ns   
rotate_const_int32/32768     |      2971 ns    |      3341 ns   
rotate_const_int32/1048576   |    250646 ns    |    270897 ns   
rotate_float/256             |      78.3 ns    |      15.4 ns   
rotate_float/4096            |      1270 ns    |       230 ns   
rotate_float/32768           |     11044 ns    |      3066 ns   
rotate_float/1048576         |    378700 ns    |    174052 ns   
rotate_const_float/256       |      21.8 ns    |      11.6 ns   
rotate_const_float/4096      |       336 ns    |       232 ns   
rotate_const_float/32768     |      3107 ns    |      3294 ns   
rotate_const_float/1048576   |    167276 ns    |    174348 ns   
-----------------------------------------------------------------
shift_int32/256              |      74.6 ns    |      16.8 ns     
shift_int32/4096             |      1206 ns    |       293 ns     
shift_int32/32768            |     11279 ns    |      3290 ns     
shift_int32/1048576          |    446056 ns    |    279861 ns     
shift_const_int32/256        |      10.7 ns    |      10.9 ns     
shift_const_int32/4096       |       182 ns    |       182 ns     
shift_const_int32/32768      |      3135 ns    |      2818 ns     
shift_const_int32/1048576    |    252950 ns    |    265048 ns     
shift_float/256              |      74.6 ns    |      16.7 ns     
shift_float/4096             |      1203 ns    |       293 ns     
shift_float/32768            |     11300 ns    |      3136 ns     
shift_float/1048576          |    383613 ns    |    173328 ns     
shift_const_float/256        |      13.9 ns    |      13.9 ns     
shift_const_float/4096       |       227 ns    |       228 ns     
shift_const_float/32768      |      3140 ns    |      2855 ns     
shift_const_float/1048576    |    169582 ns    |    169152 ns     
```

That being said, other targets should be considered, though I think `avx2-i16x16` is actually the only one affected too. Thus, we could copy-past this part (or maybe reuse it somehow) and it should be fine. Regarding the datatype size, setting only `ISPC_TARGET_HAS_ROTATE_VIA_SHUFFLE_32_VAL` to 1 already enable the use of shuffles in rotates only for 32-bit data-types, right? If so, can I include your stdlib change in this PR?

Btw, It is a bit weird the const version timing changes so much (~10%). I wonder whether the code changed a bit or if this is just the fluctuation coming from my machine (environment and frequency scaling).

aneshlya:
> Regarding the datatype size, setting only ISPC_TARGET_HAS_ROTATE_VIA_SHUFFLE_32_VAL to 1 already enable the use of shuffles in rotates only for 32-bit data-types, right? If so, can I include your stdlib change in this PR?

Yes, that's right. Please include this change to the PR.

nurmukhametov:
> The scheduling of several SIMD instructions on Zen CPUs looked suspiciously wrong IMHO. MCA is still pretty useful for AMD and ARM CPUs so to get a rough estimate of what could happend when we do not have the target CPU.

You are right, it could be so, but it is quite handy for quick estimation. 

> Regarding the tool, it is not clear to me what happens when we give both the target and the CPU. Like for example: `./compare_func.sh min_v avx2-i32x8 t.ispc "-O2 --cpu=skx"`. I tried this and I got a code using zmm register so it is not really an avx2 target but avx512 one, and in this case I guess we should rather specify the target `avx512skx-x8`. In such case, it looks like LLVM can replace the very inefficient code by a simple `vpermps zmm0, zmm1, zmm0` but I think this is not a normal use-case. Thus, I finally do not use `--cpu`. Here is my command-line:

It is probably reflects ISPC's behaviour, I don't remember the tool doing anything about this at all. It should just pass these flags to ISPCs.

> OLD_ISPC=$HOME/Bureau/projects/ispc/main-build/bin/ispc NEW_ISPC=$HOME/Bureau/projects/ispc/build/bin/ispc PATH=$HOME/Bureau/projects/ispc/llvm-20.1.4-ubuntu22.04-Release-x86.arm.wasm/bin-20.1/bin:$PATH ~/Bureau/projects/ispc/scripts/compare_func.sh f32_shuffle1 avx2-i32x16 test.ispc "-O2" skylake

I also don't pass `--cpu=` to flags here. 

> Regarding the datatype size, setting only ISPC_TARGET_HAS_ROTATE_VIA_SHUFFLE_32_VAL to 1 already enable the use of shuffles in rotates only for 32-bit data-types, right? If so, can I include your stdlib change in this PR?

Yes, sure. Please include it.

zephyr111:
I added the fix about rotate/shifts using shuffle and just rebased the commits on the latest `main` branch.

nurmukhametov (approved):
Please update copyright year in core.isph. LGTM!

PR #3794, opened 2026-04-17
Fix the stdlib default log function and make the fast-math one faster

### LINKED ISSUE #3801: Improve precision of `log` function

zephyr111:
Hello,

This PR fix several issues with the current log default-math-lib function. This has been previously mentioned in #2236 . It also make the fast-math-lib version faster. Details are in the commit description:

> The new default math-lib log implementation now correctly support special values (as opposed to the previous one):
>  - Subnormal numbers
>  - Infinities
>  - NaNs
> 
> This new default implementation also reduces the error from about ~3000 ULP (close to 1, due to a catastrophic cancellation) to <4 ULP.
> 
> The fast-math-lib implementation is now significantly faster (although less precise).

This PR is a better version of the previous PR #2937 . The default-math-lib version provided is faster than the previous PR while being also more precise. It also now support subnormal numbers. It is a bit slower than the current implementation in the main branch, but the later is bogus IMHO. The tradeoff between speed and precision of the versions provided by this PR seems optimal to me. Actually, I don't think a faster implementations can be implemented in AVX-2 for regular numbers. The main downside of this new (default) version is that it performs a branch so the implementation is significantly slower where there are special values like NaN, Inf, subnormal numbers or <=0 values. That being said, such input should be quite rare in real-world applications.

I found out that the current fast-math-lib version is actually significantly more precise (<2 ULP) than the default one (~3000 ULP close to 1) on regular numbers. Thus, I think the fast-math-lib version should have been the default one and vice-versa and this was a bug. This explains why the fast-math-lib version was *slower* than the default one while the opposite is expected.

Here are benchmark results on my machine (i5-9600KF CPU) on the avx2-i32x8 target. The reported values are the time in clocks. Please note that the Sleef implementation used in this benchmark achieves a precision similar to the new default-math-lib version (i.e. <=3.5 ULP, so let say <4 ULP).

```
Full range of positive 32-bit floats:
    - without subnormals:
        - Current (default):    3.591089e9
        - Old PR (default):     5.095525e9
        - This PR (default):    4.000402e9  <-----
        - Current (fast):       4.236597e9
        - This PR (fast):       2.555979e9  <-----
        - Sleef:                5.742247e9
        - Current (default):    3.591113e9
        - Old PR (default):     5.094209e9
        - This PR (default):    4.131417e9  <-----
        - Current (fast):       4.224303e9
        - This PR (fast):       2.555900e9  <-----
        - Sleef:                5.887635e9

Uniform random range [0;5]:
    - Current (default):    7.719988e9
    - Old PR (default):     8.333528e9
    - This PR (default):    7.266325e9  <-----
    - Current (fast):       8.304500e9
    - This PR (fast):       5.573054e9  <-----
    - Sleef:                9.087902e9
    - RNG alone (overhead): 2.199376e9
```

## Related Issue

## Checklist

### DIFF
--- a/stdlib/stdlib.ispc
+++ b/stdlib/stdlib.ispc
@@ -4588,8 +4588,8 @@ __declspec(safe) static inline float expm1(float x) {
 }
 
 // Range reduction for logarithms takes log(x) -> log(2^n * y) -> n
-// * log(2) + log(y) where y is the reduced range (usually in [1/2,
-// 1)).
+// * log(2) + log(y) where y is the reduced range (usually in [1/2,1)).
+// Warning: this function does not support subnormal numbers.
 __declspec(safe) static inline void __range_reduce_log(float input, varying float *uniform reduced,
                                                        varying int *uniform exponent) {
     int int_version = intbits(input);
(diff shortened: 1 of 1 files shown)


zephyr111:
By the way, I found out that the default log function in the stdlib is 5-6% slower than the same log implementation in my ISPC testing file. This is a bit sad since a 5-6% performance gain would make the new default implementation even more competitive (compared to the current default log function in the main branch). This is apparently due to the `inline` keyword (i.e. it is surprisingly faster *without* it). See the [slow (inlined) assembly](https://pastebin.com/4JU6d4B2) vs the [fast (not-inlined) assembly](https://pastebin.com/2Urr6CVh). I suspect this comes from a slightly-less efficient masking (more blend instructions) and register reloads (more broadcast instructions) caused by a different block layout. I don't think there is much more to do with that and maybe we should accept that the compiler can introduce some overhead (5-6%) due to that (depending on the implementation used) and so the benchmark is precise to 5-6%.
Please note that the reported benchmark timings are the one of the stdlib.

elena901:
@codex

nurmukhametov:
> By the way, I found out that the default log function in the stdlib is 5-6% slower than the same log implementation in my ISPC testing file. This is a bit sad since a 5-6% performance gain would make the new default implementation even more competitive (compared to the current default log function in the main branch). This is apparently due to the `inline` keyword (i.e. it is surprisingly faster _without_ it). See the [slow (inlined) assembly](https://pastebin.com/4JU6d4B2) vs the [fast (not-inlined) assembly](https://pastebin.com/2Urr6CVh). I suspect this comes from a slightly-less efficient masking (more blend instructions) and register reloads (more broadcast instructions) caused by a different block layout. I don't think there is much more to do with that and maybe we should accept that the compiler can introduce some overhead (5-6%) due to that (depending on the implementation used) and so the benchmark is precise to 5-6%. Please note that the reported benchmark timings are the one of the stdlib.

Can you create a separate issue about this with the ISPC source code?

nurmukhametov (approved):
LGTM!

dbabokin:
@codex review

PR #3796, opened 2026-04-18
Implement the log1p function

### LINKED ISSUE #3795: Implement the math function log1p
Similarly to `expm1`, the `log1p` function [available in C and C++](https://en.cppreference.com/w/c/numeric/math/log1p) (or in many other languages/math-library like [Numpy](https://numpy.org/doc/stable/reference/generated/numpy.log1p.html)/[Scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.log1p.html) in Python) is not provided yet in ISPC and it would be good to implement it. It is basically `log(1+x)` computed in a numerically-stable way (so to avoid a [catastrophic cancellation](https://en.wikipedia.org/wiki/Catastrophic_cancellation)). This function is sometimes pretty useful. It might be useful to implement `lgamma` accurately.

zephyr111:
This PR implement the function `log1p` mentioned in #3795. The precision of the default version is <5 ULP while the fast-math-lib one is <30 ULP (the maximum error is close to that bound in practice). The default implementation is much faster than Sleef (which is more precise though). The fast-math-lib version bring a significant performance boost.

Here are benchmark results on my machine (i5-9600KF CPU) on the avx2-i32x8 target. The reported values are the time in clocks. Please note that the Sleef implementation used in this benchmark achieves a precision relatively close to the default-math-lib version (i.e. <=3.5 ULP, so let say <4 ULP -- versus <5 ULP).

```
Full range of positive 32-bit floats:
    - without subnormals:
        - This PR (default):    4.004950e9
        - This PR (fast):       2.604127e9
        - Sleef:               10.076649e9
        - This PR (default):    4.829993e9
        - This PR (fast):       3.320860e9
        - Sleef:               20.559469e9
```

PS: Sleef does not correctly compute special values (e.g. negatives input values, inf, etc.) for this function. 

## Related Issue

## Checklist


dbabokin:
@codex review

tests/func-tests/log1p-double-uniform.ispc
```diff
@@ -0,0 +1,19 @@
...
+
+uniform bool ok(uniform double x, uniform double ref) {
+    // See the varying version
+    return (abs(x - ref) < 1e-15d) || abs((x - ref) / ref) < 1e-12d;
+}
+
+task void f_fu(uniform float RET[], uniform float aFOO[], uniform float b) {
+    for (uniform int i = 0; i != programCount; ++i) {
+        uniform double ref = log1p((uniform double)((aFOO[i] / programCount) * 2.5d - 0.5d));
+        uniform double res = log1p((aFOO[i] / programCount) * 2.5d - 0.5d);
```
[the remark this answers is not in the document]
zephyr111: This raise the question of what the uniform tests in float/double types should actually test. I think I initially meant to just test it is equal to the varying one and I wrongly thought `aFOO[i]` was a varying (because of the `for` loop confused with a `foreach` or the various kind of test function not very clear to me yet). 
This (sneaky) issue was present in the uniform tests of `erf`, `erfc`, `sinh`, `cosh`, `tanh`, `expm1`. I just blindly copy-pasted this file and changed the name of the target function as well as the precision. Interestingly, the AI bots did not found the issue previously and it looks like the minor changes (e.g. name of the function) can change the outcome of the review...

stdlib/stdlib.ispc
```diff
@@ -5039,6 +5039,183 @@ __declspec(safe) static inline float log(float x_full) {
     }
 }
 
+__declspec(safe) static inline uniform float log1p(uniform float x_full) {
```
dbabokin, line 5094: Does it make sense to have a uniform/varying template to reuse the same code implementation?

builtins/util.m4
```diff
@@ -6051,6 +6058,11 @@ define double @__stdlib_log(double) nounwind readnone alwaysinline {
   ret double %r
 }
 
+define double @__stdlib_log1p(double) nounwind readnone alwaysinline {
+  %r = call double @log1p(double %0)
```
dbabokin, line 6062: Where does this implementation come from?

zephyr111: I am not sure to understand the question. This should call the libm function like other function defined above/below. See for example https://www.man7.org/linux/man-pages/man3/log1p.3.html.

stdlib/stdlib.ispc
zephyr111: Yes. But this is not just for this function but for all functions of the stdlib. I plan to rework this. There are functions where templates works well and there are functions where it is a mess, typically because of mixed precision (or type deduction). Btw, this is not just an issue with the stdlib but also in my projects using ISPC: I wish to avoid code duplication but current ISPC templates are sometimes too limited. I used it for `cbrt` once. I remembered they was some issue with templates being in the stdlib at this time. There are open questions regarding this: does it impact the compilation time? Is is currently safe/OK? Could the language be extended to fix the issue with mixed precision (or ideally type deduction)?
It is certainly something to discuss after the fix of all math functions and possibly once the stdlib will be split into many files (so to better analyze its compilation time and possibly side effects on the generated code).

zephyr111:
I updated the tests so to take into account #3813 and rebased the commits onto the latest main branch.

stdlib/stdlib.ispc
elena901: > Is is currently safe/OK?

Of course not. C++ style templates are more about portability and user's involvement (don't need to learn new syntax/behavior) and not about safety.

Except non-type template arguments. But at this moment it's "untyped" too (no varying/uniform distinguish).

nurmukhametov (approved):
LGTM!

elena901:
> does it impact the compilation time? Is is currently safe/OK?



> Could the language be extended to fix the issue with mixed precision (or ideally type deduction)?

@zephyr111 You need to file an issue with practical examples!

*misposted previously*

PR #3798, opened 2026-04-19
Fix the acos function, make it faster and improve tests

### LINKED ISSUE #3797: `acos` is wongly computed in both single/double precision
Here is a C code calling `acos` with the constant `0.99999964237213134766` in both single and double precision (either from a cast or not).

```c
#include <stdlib.h>
#include <stdio.h>
#include <stdint.h>
#include <math.h>
#include <string.h>

int main()
{
    float xf;
    double xd;
    uint32_t reprf = 0x3F7FFFFAul;          // 0.99999964237213134766
    uint64_t reprd = 0x3FEFFFFF40000000ull; // 0.99999964237213134766
    memcpy(&xf, &reprf, sizeof(float));
    memcpy(&xd, &reprd, sizeof(double));
    double v1 = acosf(xf);
    double v2 = acos((double)xf);
    double v3 = acos(xd);

    // Result: "8.457279764e-4 8.457279586e-4 8.457279586e-4"
    printf("%.9lfe-4 %.9lfe-4 %.9lfe-4\n", v1*10000.0, v2*10000.0, v3*10000.0);
    return 0;
}
```

Here is its equivalent ISPC code:

```ispc
extern "C" uniform int main()
{
    uniform float xf;
    uniform double xd;
    uniform uint32 reprf = 0x3F7FFFFAul;          // 0.99999964237213134766
    uniform uint64 reprd = 0x3FEFFFFF40000000ull; // 0.99999964237213134766
    memcpy(&xf, &reprf, sizeof(uniform float));
    memcpy(&xd, &reprd, sizeof(uniform double));
    uniform double v1 = acos(xf);
    uniform double v2 = acos((uniform double)xf); // v2 = 8.456707000732421875e-4
    uniform double v3 = acos(xd);

    // Result: "8.456707e-4    8.457717e-4    8.457717e-4"
    print("%e-4    %e-4    %e-4\n", v1*10000.0d, v2*10000.0d, v3*10000.0d);

    // Result: 4560939158759014400 4560940090042009600 4560940090042009600
    print("% % %\n", intbits(v1), intbits(v2), intbits(v3));
    return 0;
}
```

And here is a build script (using the main branch of ISPC):

```bash
rm -f acos-c acos-ispc main.o
clang main.c -o acos-c -lm
~/Bureau/projects/ispc/build/bin/ispc main.ispc --pic --target=avx2-i32x8 -o main.o -g
clang main.o -fpie -o acos-ispc -lm
```

The results are different in both single and double precision. This is normal for the single precision since ISPC call the default stdlib implementation which is not as accurate as the (GNU) libm. However, the **double precision is very inaccurate**! It is 8.457717e-4 while it should be 8.4572795858789 (pretty-precise value from WolframAlpha) matching much better with the (GNU) libm implementation in C than the ISPC version. Sollya (which is very reliable) gives: 8.4572795858789498e-4 (using 256-bit FP numbers for the computation). This perfectly matches with the WolframAlpha version so we can safely say this is probably the right result. Here is the summary of the results:

```none
Sollya: 8.4572795858789498e-4
C:      8.457279764e-4 8.457279586e-4 8.457279586e-4
ISPC:   8.456707e-4    8.457717e-4    8.457717e-4
```

Overall, the relative precision of the single-precision C version calling the GNU libm is ~2.1e-8 so it is basically 0.5 ULP which is perfect. The double-precision one achieves a precision <=1.5e-11 and it is certainly limited to the limited number of digits printed, so this fine. The single-precision ISPC version supposed to call the GNU libm reach a precision of ~6.8e-05 which is quite bad (this is a known issue with the current stdlib default math implementation though). The double-precision version reach 5.168492960750737e-05 which is *very bad* for double-precision (actually, also for single precision)!

Here is the result of the linked files showing the same libm is linked:

```bash
$ ldd ./acos-c
	linux-vdso.so.1 (0x00007f0833591000)
	libm.so.6 => /usr/lib/x86_64-linux-gnu/libm.so.6 (0x00007f083346b000)
	libc.so.6 => /usr/lib/x86_64-linux-gnu/libc.so.6 (0x00007f0833275000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f0833593000)
$ ldd ./acos-ispc
	linux-vdso.so.1 (0x00007f21d205e000)
	libm.so.6 => /usr/lib/x86_64-linux-gnu/libm.so.6 (0x00007f21d1f35000)
	libc.so.6 => /usr/lib/x86_64-linux-gnu/libc.so.6 (0x00007f21d1d3f000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f21d2060000)
```

When we look the assembly code, we can see that ISPC generates a call to the libm as expected, but **it calls `asin` instead of `acos`** and it then seems to discard the result use a very-inaccurate *pre-computed constant* instead! I strongly suspect **this constant for a  double-precision call is computed in single-precision** which is clearly a bug (maybe not just one). Here is the assembly code:

```asm
Dump of assembler code for function ispc::main():
3		uniform float xf;
   0x0000555555555180 <+0>:	push   rbp
   0x0000555555555181 <+1>:	mov    rbp,rsp
   0x0000555555555184 <+4>:	and    rsp,0xffffffffffffffe0
   0x0000555555555188 <+8>:	sub    rsp,0xa0
   0x000055555555518f <+15>:	vmovsd xmm0,QWORD PTR [rip+0x2e71]        # 0x555555558008

4		uniform double xd;
5		uniform uint32 reprf = 0x3F7FFFFAul;          // 0.99999964237213134766
6		uniform uint64 reprd = 0x3FEFFFFF40000000ull; // 0.99999964237213134766
7		memcpy(&xf, &reprf, sizeof(uniform float));
8		memcpy(&xd, &reprd, sizeof(uniform double));
9		uniform double v1 = acos(xf);
10		uniform double v2 = acos((uniform double)xf);
   0x0000555555555197 <+23>:	call   0x555555555070 <asin@plt>
=> 0x000055555555519c <+28>:	movabs rax,0x4020e9d580000000

11		uniform double v3 = acos(xd);
12	
13	    // Result: "8.456707e-4    8.457717e-4    8.457717e-4"
14		print("%e-4    %e-4    %e-4\n", v1*10000.0d, v2*10000.0d, v3*10000.0d);
   0x00005555555551a6 <+38>:	mov    QWORD PTR [rsp+0x38],rax
   0x00005555555551ab <+43>:	lea    rax,[rsp+0x38]
   0x00005555555551b0 <+48>:	mov    QWORD PTR [rsp+0x60],rax
   0x00005555555551b5 <+53>:	movabs rax,0x4020ea59d7e0c492
   0x00005555555551bf <+63>:	mov    QWORD PTR [rsp+0x30],rax
   0x00005555555551c4 <+68>:	lea    rcx,[rsp+0x30]
   0x00005555555551c9 <+73>:	mov    QWORD PTR [rsp+0x68],rcx
   0x00005555555551ce <+78>:	mov    QWORD PTR [rsp+0x28],rax
   0x00005555555551d3 <+83>:	lea    rax,[rsp+0x28]
   0x00005555555551d8 <+88>:	mov    QWORD PTR [rsp+0x70],rax
   0x00005555555551dd <+93>:	lea    rdi,[rip+0x2e2c]        # 0x555555558010 <__str>
   0x00005555555551e4 <+100>:	lea    rsi,[rip+0x2e3b]        # 0x555555558026 <__str.1>
   0x00005555555551eb <+107>:	lea    rdx,[rsp+0x60]
   0x00005555555551f0 <+112>:	call   0x555555555260 <__do_print>
   0x00005555555551f5 <+117>:	movabs rax,0x3f4bb60000000000
```

-----

To prevent the compiler from pre-computing constants, I dynamically loaded the input values from main arguments and the issues remains present. Thus, we can conclude that this does not comes from pre-computing. Here is the ISPC code:

```ispc
extern "C" uniform int32 atol(const uniform int8 s[]);
extern "C" uniform int64 atoll(const uniform int8 s[]);

// TODO: call is with "./acos-ispc 1065353210 4607182415578791936"
extern "C" uniform int main(uniform int argc, uniform int8* uniform argv[])
{
    uniform float xf;
    uniform double xd;
    uniform uint32 reprf; // 0x3F7FFFFA         // 0.99999964237213134766
    uniform uint64 reprd; // 0x3FEFFFFF40000000 // 0.99999964237213134766
    reprf = atol(argv[1]);
    reprd = atoll(argv[2]);
    memcpy(&xf, &reprf, sizeof(uniform float));
    memcpy(&xd, &reprd, sizeof(uniform double));
    uniform double v1 = acos(xf);
    uniform double v2 = acos((uniform double)xf); // v2 = 8.456707000732421875e-4
    uniform double v3 = acos(xd);

    // Result: "8.456707e-4    8.457717e-4    8.457717e-4"
    print("%e-4    %e-4    %e-4\n", v1*10000.0d, v2*10000.0d, v3*10000.0d);

    // Result: 4560939158759014400 4560940090042009600 4560940090042009600
    print("% % %\n", intbits(v1), intbits(v2), intbits(v3));
    return 0;
}
```

If we look the assembly code, then we can see that the single-precision `acos` function is now properly dynamically performed in ISPC, but there is still calls to `asin` instead of `acos` for the two last calls.

```asm
   0x000055555555529d <+253>:	vcvtss2sd xmm0,xmm0,xmm0
   0x00005555555552a1 <+257>:	call   0x555555555090 <asin@plt>
   0x00005555555552a6 <+262>:	vmovsd xmm1,QWORD PTR [rip+0x2d92]        # 0x555555558040
   0x00005555555552ae <+270>:	vsubsd xmm0,xmm1,xmm0
   0x00005555555552b2 <+274>:	vmovsd QWORD PTR [rsp+0x20],xmm0
   0x00005555555552b8 <+280>:	vmovsd xmm0,QWORD PTR [rsp+0x18]
   0x00005555555552be <+286>:	call   0x555555555090 <asin@plt>

   0x00005555555552c3 <+291>:	vmovsd xmm1,QWORD PTR [rip+0x2d75]        # 0x555555558040
   0x00005555555552cb <+299>:	vsubsd xmm2,xmm1,xmm0
```

The call to `asin` followed by `vsubsd` seem to clearly indicate that **`acos` is computed using `pi/2 - asin(x)`**. This is not an accurate way of computing it because of a **catastrophic cancellation**.

-----

In the end, It looks like the problem lies in the implementation of `acos` in the stdlib code:

```ispc
__declspec(safe) static inline uniform double acos(const uniform double v) {
    if (__have_native_trigonometry)
        return __acos_uniform_double(v);
    else
        return 1.57079637050628662109375d - asin(v);    //  <=====  BUG HERE
}
```

`acos` should directly call the libm and never make such an operation (unless there is no direct mapping to the libm).

-----

Calling the `__stdlib_acosf` and `__stdlib_acos` (GNU libm) functions directly from ISPC fix the problem in both single and double precision (it gives 8.457280e-4). By the way, the double-precision result does not seems perfectly accurate (error of ~20 ULP), but I think this is a problem of the GNU libm and not ISPC. This is a hack though since we are not supposed to call such functions directly from ISPC. Here is the resulting code:

```ispc
extern "C" uniform float __stdlib_acosf(uniform float x);
extern "C" uniform double __stdlib_acos(uniform double x);

extern "C" uniform int main()
{
    uniform float xf;
    uniform double xd;
    uniform uint32 reprf = 0x3F7FFFFAul;          // 0.99999964237213134766
    uniform uint64 reprd = 0x3FEFFFFF40000000ull; // 0.99999964237213134766
    memcpy(&xf, &reprf, sizeof(uniform float));
    memcpy(&xd, &reprd, sizeof(uniform double));
    uniform double v1 = __stdlib_acosf(xf);
    uniform double v2 = __stdlib_acos((uniform double)xf); // v2 = 8.457279585878994e-4
    uniform double v3 = __stdlib_acos(xd);

    // Result: "8.457280e-4    8.457280e-4    8.457280e-4"
    print("%e-4    %e-4    %e-4\n", v1*10000.0d, v2*10000.0d, v3*10000.0d);

    // Result: 4560939687039991808 4560939686875597361 4560939686875597361
    print("% % %\n", intbits(v1), intbits(v2), intbits(v3));
    return 0;
}
```

zephyr111:
As the title/commit suggest, this PR fix precision issues in the `acos` function, make it faster and improve tests. Here is the commit description for more details:
> - acos now correctly calls the standard library in double precision
> - implement faster versions which are also much more accurate close to 1 (no catastrophic cancellation anymore)
> - improve the acos tests so to check more use-case and reduce the error bound

It fixes the issue #3797 about double-precision and solve the `acos` numerical stability issue mention in #2236 (which is also mentioned in the first issue).

The precision of the default math-lib function is now <5 ULP and it is <15 ULP for the fast one. Overall, the error is more balanced than before on the `[-1;+1]` range and now bounded. This means it can be a bit higher for x < 0.5 while being much lower for x > 0.95.

The default and fast implementation are both faster than previous ones. They are also much faster (about 3-4 times) than Sleef (having a relatively-close precision of <3.5 ULP). Here are benchmark results on my machine (i5-9600KF CPU) on the avx2-i32x8 target (the reported values are the time in clocks):

```
Full range of 32-bit floats (not very realistic use-case):
    - without subnormals:
        - Current main branch (default):    2.903069e9
        - Current main branch (fast):       1.650359e9
        - This PR (default):                1.404689e9  <-----
        - This PR (fast):                   1.372506e9  <-----
        - Sleef:                            5.071257e9
        - Current main branch (default):    4.063371e9
        - Current main branch (fast):       2.249455e9
        - This PR (default):                1.325650e9  <-----
        - This PR (fast):                   1.372743e9  <-----
        - Sleef:                           17.094187e9


Uniform random range [-1;+1] (more realistic use-case)
    - Current main branch (default):    7.446462e9
    - Current main branch (fast):       5.399786e9
    - This PR (default):                4.561175e9  <-----
    - This PR (fast):                   4.354596e9  <-----
    - Sleef:                            8.395003e9
    - RNG alone (overhead):             2.114988e9
```

## Related Issue

## Checklist

### DIFF
--- a/builtins/util-xe.m4
+++ b/builtins/util-xe.m4
@@ -4599,6 +4599,7 @@ define float @__stdlib_expm1f(float) nounwind readnone alwaysinline {
 declare double @sin(double) nounwind readnone
 declare double @sinh(double) nounwind readnone
 declare double @asin(double) nounwind readnone
+declare double @acos(double) nounwind readnone
 declare double @cos(double) nounwind readnone
 declare double @cosh(double) nounwind readnone
 declare void @sincos(double, double *, double *) nounwind
@@ -4626,6 +4627,11 @@ define double @__stdlib_sinh(double) nounwind readnone alwaysinline {
 
 define double @__stdlib_asin(double) nounwind readnone alwaysinline {
   %r = call double @asin(double %0)
+  ret double %r
+}
+
+define double @__stdlib_acos(double) nounwind readnone alwaysinline {
+  %r = call double @acos(double %0)
   ret double %r
 }
 
--- a/builtins/util.m4
+++ b/builtins/util.m4
@@ -5973,6 +5973,7 @@ define float @__stdlib_expm1f(float) nounwind readnone alwaysinline {
 declare double @sin(double) nounwind readnone
 declare double @sinh(double) nounwind readnone
 declare double @asin(double) nounwind readnone
+declare double @acos(double) nounwind readnone
 declare double @cos(double) nounwind readnone
 declare double @cosh(double) nounwind readnone
 declare void @sincos(double, double *, double *) nounwind
@@ -6000,6 +6001,11 @@ define double @__stdlib_sinh(double) nounwind readnone alwaysinline {
 
 define double @__stdlib_asin(double) nounwind readnone alwaysinline {
   %r = call double @asin(double %0)
+  ret double %r
+}
+
+define double @__stdlib_acos(double) nounwind readnone alwaysinline {
+  %r = call double @acos(double %0)
   ret double %r
 }
 
--- a/stdlib/include/builtins.isph
+++ b/stdlib/include/builtins.isph
@@ -590,6 +590,7 @@ EXT inline void __streaming_store_varying_i8(NOESCAPE uniform int8 *uniform, var
 EXT inline READNONE uniform float __stdlib_acosf(uniform float);
 EXT inline READNONE uniform float __stdlib_asinf(uniform float);
 EXT inline READNONE uniform double __stdlib_asin(uniform double);
+EXT inline READNONE uniform double __stdlib_acos(uniform double);
 EXT inline READNONE uniform float __stdlib_atan2f(uniform float, uniform float);
 EXT inline READNONE uniform double __stdlib_atan2(uniform double, uniform double);
 EXT inline READNONE uniform float __stdlib_atanf(uniform float);
--- a/stdlib/stdlib.ispc
+++ b/stdlib/stdlib.ispc
@@ -3740,21 +3740,124 @@ __declspec(safe) static inline float cosh(float x) {
     }
 }
 
-__declspec(safe) static inline float acos(float v) {
+__declspec(safe) static inline uniform float acos(uniform float x_full) {
     if (__have_native_trigonometry && !__is_xe_target) {
-        return __acos_varying_float(v);
-    } else if (__math_lib == __math_lib_svml) {
-        return __svml_acosf(v);
-    } else {
-        return 1.57079637050628662109375 - asin(v);
+        return __acos_uniform_float(x_full);
+    } else if (__math_lib == __math_lib_system || __math_lib == __math_lib_svml) {
+        return __stdlib_acos(x_full);
+    } else if (__math_lib == __math_lib_ispc_fast) {
+
+        // Precision: <15 ULP.
+        // Natively support special values (e.g. subnormal numbers and NaNs).
+
+        // Sollya:
+        // f = acos(1-x) / sqrt(x)
+        // fpminimax(f, [|0,...,5|], [|single...|], [1e-300;1.0]);
+        const uniform float c0 = +1.4142124653e+00;
+        const uniform float c1 = +1.1792322248e-01;
+        const uniform float c2 = +2.5741405785e-02;
+        const uniform float c3 = +1.0879410430e-02;
+        const uniform float c4 = -2.2721362770e-03;
+        const uniform float c5 = +4.3107867240e-03;
+
+        const uniform float x = abs(x_full);
+        const uniform float t = 1.0 - x;
+        const uniform float t2 = t * t;
+
+        // Estrin's scheme used to reduce the latency
+        uniform float poly = c5 * t + c4;
+        poly = poly * t2 + (c3 * t + c2);
+        poly = poly * t2 + (c1 * t + c0);
+        poly *= sqrt(t);
+
+        return select(x_full < 0.0, PI - poly, poly);
+    } else if (__math_lib == __math_lib_ispc) {
+
+        // Precision: <5 ULP
+
+        // Sollya:
+        // f = acos(1-x) / sqrt(x)
+        // fpminimax(f, [|0,...,6|], [|single...|], [1e-300;1.004]);
+        const uniform float c0 = +1.4142136574e+00;
+        const uniform float c1 = +1.1783879995e-01;
+        const uniform float c2 = +2.6711430400e-02;
+        const uniform float c3 = +6.7900302820e-03;
+        const uniform float c4 = +5.5398275140e-03;
+        const uniform float c5 = -2.5801914740e-03;
+        const uniform float c6 = +2.2826674390e-03;
+
+        const uniform float x = abs(x_full);
+        const uniform float t = 1.0 - x;
+        const uniform float t2 = t * t;
+
+        // Estrin's scheme used to reduce the latency
+        uniform float poly = c6;
+        poly = poly * t2 + (c5 * t + c4);
+        poly = poly * t2 + (c3 * t + c2);
+        poly = poly * t2 + (c1 * t + c0);
+        poly *= sqrt(t);
+
+        return select(x_full < 0.0, PI - poly, poly);
     }
 }
 
-__declspec(safe) static inline uniform float acos(uniform float v) {
-    if (__have_native_trigonometry && !__is_xe_target)
-        return __acos_uniform_float(v);
-    else
-        return 1.57079637050628662109375 - asin(v);
+__declspec(safe) static inline float acos(float x_full) {
+    if (__have_native_trigonometry && !__is_xe_target) {
+        return __acos_varying_float(x_full);
+    } else if (__math_lib == __math_lib_svml) {
+        return __svml_acosf(x_full);
+    } else if (__math_lib == __math_lib_system) {
+        float ret;
+        foreach_active(i) {
+            uniform float r = __stdlib_acosf(extract(x_full, i));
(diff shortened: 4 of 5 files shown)


dbabokin:
@codex review

nurmukhametov (approved):
LGTM!

dbabokin:
I'm wondering if acos is one off or we have similar problem with other similar functions? Asin? atan?

zephyr111:
@dbabokin When I checked `atan` few years ago, it looked fine, but I need to check this again (I am gonna do that soon). Regarding `asin`, there was issues in single precision, but I fixed them in #2948 . I did not check much double precision for most functions yet. I've just quickly checked this now and I did not found any obvious similar issues, so it might be the only one.
EDIT: note that `fmod` is inaccurate, but this is independent of the type: it is inaccurate for both single-precision and double-precision (I did not test for half-precision).

PR #3802, opened 2026-04-21
Fix AMX tests failing with SIGILL on newer Linux kernels

azwolski:
## Description
On Linux, applications must explicitly request permission from the kernel via arch_prctl to use Intel AMX tile data registers (XFEATURE_XTILEDATA) before executing any AMX instructions, otherwise the kernel kills the process with SIGILL. Add the syscall at test harness startup and warn on failure (e.g., when the CPU/kernel doesn't support AMX).

## Related Issue

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,6 +64,10 @@ set(PROJECT_NAME ispc)
 set(ISPC_BUILD TRUE)
 project(${PROJECT_NAME})
 
+# Create a job pool to serialize stdlib bitcode generation to avoid race conditions
+# where multiple jobs try to read/write the same .bc file simultaneously
+set_property(GLOBAL PROPERTY JOB_POOLS stdlib_generation=1)
+
 ## Check PIE support
 include(CheckPIESupported)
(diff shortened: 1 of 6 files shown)


ivankelarev (approved):
LGTM

dbabokin:
@codex review

nurmukhametov (changes requested):
Thanks for the fix!

A couple of suggestions:
1. I agree with the AI reviews. Please check the syscall return value, print a clear error and exit on failure. Otherwise the test will still SIGILL.
2. It is called for any target, but only needed for AMX-capable targets. I suggest gating it behind a compile-time flag (e.g. `-DTEST_REQUIRES_AMX`) that `run_tests.py` passes for AMX-capable targets only.

azwolski:
@dbabokin @nurmukhametov I’d appreciate approvals/reviews as soon as possible so I can proceed with the undef‑to‑poison migration.

nurmukhametov (approved):
LGTM!

Nit: I would suggest avoiding unrelated formatting changes.

PR #3808, opened 2026-04-25
Defer forced inlining of `inline`-qualified functions

### LINKED ISSUE #3804: Possible performance loss with inlined functions
As requested by @nurmukhametov [here](https://github.com/ispc/ispc/pull/3794#issuecomment-4284603518) (in PR #3794), this issue provides an example of case where function inlining can make the code slower. This can sometimes decrease the performance of stdlib math function which are all inlined. Here is the quote of the PR initially mentioning the issue:

> By the way, I found out that the default log function in the stdlib is 5-6% slower than the same log implementation in my ISPC testing file. This is a bit sad since a 5-6% performance gain would make the new default implementation even more competitive (compared to the current default log function in the main branch). This is apparently due to the `inline` keyword (i.e. it is surprisingly faster *without* it). See the [slow (inlined) assembly](https://pastebin.com/4JU6d4B2) vs the [fast (not-inlined) assembly](https://pastebin.com/2Urr6CVh). I suspect this comes from a slightly-less efficient masking (more blend instructions) and register reloads (more broadcast instructions) caused by a different block layout. I don't think there is much more to do with that and maybe we should accept that the compiler can introduce some overhead (5-6%) due to that (depending on the implementation used) and so the benchmark is precise to 5-6%.
Please note that the reported benchmark timings are the one of the stdlib.

Here is the actual test file:

```ispc
// Switch between 0 and 1 so to see the impact on the timings
#if 0
    #define OPT_INLINE
#else
    #define OPT_INLINE inline
#endif

// Implementation currently used by the stdlib on the main branch
static OPT_INLINE float my_log(float x_full) {
    const uint32 x_repr = intbits(x_full);
    const bool special = x_repr + 0x80800000 < 0x81000000;

    if (any(special)) {
        const float pinf = floatbits(0x7F800000);
        const float ninf = floatbits(0xFF800000);
        const float pnan = floatbits(0x7FC00000);

        if (isnan(x_full) || x_full == pinf)
            return x_full;

        if (x_full < 0.0)
            return pnan;

        if (x_full == 0.0)
            return ninf;

        if (x_repr < 0x00800000) {
            x_full *= 1e36;

            float reduced;
            int exponent;

            __range_reduce_log(x_full, &reduced, &exponent);

            const float ln1e36 = 8.2893063348e+01;
            const float ln2 = 6.931471825e-01;
            const float scale = (float)exponent * ln2 - ln1e36;

            const float c02 = -5.0002461670e-01;
            const float c03 = +3.3212903140e-01;
            const float c04 = -2.6966506240e-01;
            const float c05 = +5.0812456760e-02;
            const float c06 = -7.5164681670e-01;
            const float c07 = -1.0362093450e+00;
            const float c08 = -1.1606367826e+00;

            const float x = reduced - 1.0;
            const float x2 = x * x;

            float result = x * c08 + c07;
            result = x2 * result + (x * c06 + c05);
            result = x2 * result + (x * c04 + c03);
            result = x2 * result + (x * c02 + 1.0);
            result = x * result + scale;

            return result;
        }
    }

    const float unstable_range_size = 0.285;
    const uint32 start_repr = intbits(1.0);
    const uint32 end_repr = (uint32)intbits(1.0 + unstable_range_size) + 1ul;
    const uint32 unstable_range_ulp_size = end_repr - start_repr;
    const bool close_to_one = x_repr - start_repr < unstable_range_ulp_size;

    float reduced;
    int exponent;

    __range_reduce_log(x_full, &reduced, &exponent);
    reduced = select(close_to_one, x_full, reduced);

    const float ln2 = 6.931471825e-01;
    const float scale = select(close_to_one, 0.0, (float)exponent * ln2);

    const float c02 = -4.9999991060e-01;
    const float c03 = +3.3335432410e-01;
    const float c04 = -2.4996809660e-01;
    const float c05 = +1.9873061776e-01;
    const float c06 = -1.6905537250e-01;
    const float c07 = +1.6525325180e-01;
    const float c08 = -7.3399633170e-02;
    const float c09 = -4.0101176130e-03;
    const float c10 = -4.4349682330e-01;

    const float x = reduced - 1.0;
    const float x2 = x * x + 1e-30;

    float result = x * c10 + c09;
    result = x2 * result + (x * c08 + c07);
    result = x2 * result + (x * c06 + c05);
    result = x2 * result + (x * c04 + c03);
    result = x2 * result + (x * c02 + 1.0);
    result = x * result + scale;

    return result;
}

unmasked uniform float benchmark(uniform float cst)
{
    uniform int32 maxi = 0x7F800001;
    float checksum = 0.0;

    foreach(i=0...maxi)
        checksum += my_log(cst + floatbits(i));

    return reduce_add(checksum);
}

extern "C" uniform int32 main(uniform int32 argc, uniform int8* uniform argv[])
{
	// Using argc in the benchmark is meant to prevent ISPC & LLVM
	// to pre-compute everything.
    const uniform float res = benchmark((uniform float)argc);

    print("checksum: %\n", res);
}
```

Here is how to build and run the code on Linux:
```bash
ispc test.ispc -o test.o --pic --target=avx2-i32x8
clang test.o -o ispc_test -fpic
time ./ispc_test
```

Here are timings results on my i5-9600KF CPU:
```none
without inlining: 0m1,104s
with inlining: 0m1,254s
-----
with the log of the stdlib (inlined): 0m1,265s
```

nurmukhametov:
## Description

`inline` in ISPC was implemented by attaching LLVM's `alwaysinline` at parse time, which expanded the callee into the caller in the very first inliner pass, i.e.,  before some optimization passes could run on the callee in isolation. The over-large caller optimized worse: extra broadcasts/blends and more spills in hot loops (issue #3804).

This PR defers forced inlining until just before the final inliner pass, so each `inline`-qualified callee gets optimized standalone first.

## Changes

- Replace the parse-time `alwaysinline` with a custom string attribute `ispc-defer-alwaysinline`.
- New module pass `RestoreInlineAttrPass` strips the marker and adds `alwaysinline` (unless `noinline` is also present) right before the final `ModuleInlinerWrapperPass`.
- Add an cleanup after the final inliner so caller-context simplifications fold through the freshly inlined bodies.

## Results

- Issue reproducer should match now the non-inlined baseline.
- Side effect: a previously-XFAIL'd test (`const_shuffle_i8.ispc`, introduced in #3458) now passes. `IsCompileTimeConstant` can fold the `shuf2` const path now that  inlining + cleanup land at the right time.

## Related Issue
- [X] Linked to relevant issue: #3804 

## Checklist

### DIFF
--- a/src/func.cpp
+++ b/src/func.cpp
@@ -1,5 +1,5 @@
 /*
-  Copyright (c) 2011-2025, Intel Corporation
+  Copyright (c) 2011-2026, Intel Corporation
 
   SPDX-License-Identifier: BSD-3-Clause
 */
@@ -53,7 +53,12 @@ bool Function::IsInternal() const {
     bool isInline = false;
     llvm::Function *function = sym->function;
     if (function != nullptr) {
-        isInline = (function->getAttributes().getFnAttrs().hasAttribute(llvm::Attribute::AlwaysInline));
+        const llvm::AttributeSet &fnAttrs = function->getAttributes().getFnAttrs();
+        // ispc-defer-alwaysinline is the deferred form of AlwaysInline used for
+        // user-written `inline`; either marker means the source qualifier was
+        // present.
+        isInline =
+            fnAttrs.hasAttribute(llvm::Attribute::AlwaysInline) || fnAttrs.hasAttribute("ispc-defer-alwaysinline");
     }
     return sc == StorageClass::STATIC || isInline;
 }
@@ -500,10 +505,13 @@ void Function::emitCode(FunctionEmitContext *ctx, llvm::Function *function, Sour
         // entire thing inside code that tests to see if the mask is all
         // on, all off, or mixed.  If this is a simple function, then this
         // isn't worth the code bloat / overhead.
-        bool checkMask =
-            (!g->target->isXeTarget() && type->IsTask() == true) ||
-            ((function->getAttributes().getFnAttrs().hasAttribute(llvm::Attribute::AlwaysInline) == false) &&
-             costEstimate > CHECK_MASK_AT_FUNCTION_START_COST);
+        const llvm::AttributeSet &fnAttrs = function->getAttributes().getFnAttrs();
+        // ispc-defer-alwaysinline is the deferred form of alwaysinline used for
+        // user-written `inline`; treat both as "will be inlined".
+        const bool willBeInlined =
+            fnAttrs.hasAttribute(llvm::Attribute::AlwaysInline) || fnAttrs.hasAttribute("ispc-defer-alwaysinline");
+        bool checkMask = (!g->target->isXeTarget() && type->IsTask() == true) ||
+                         (!willBeInlined && costEstimate > CHECK_MASK_AT_FUNCTION_START_COST);
         checkMask &= (type->IsUnmasked() == false);
         checkMask &= (g->target->getMaskingIsFree() == false);
         checkMask &= (g->opt.disableCoherentControlFlow == false);
@@ -1326,7 +1334,9 @@ llvm::Function *TemplateInstantiation::createLLVMFunction(Symbol *functionSym) {
     g->target->markFuncWithTargetAttr(function);
 
     if (isInline) {
-        function->addFnAttr(llvm::Attribute::AlwaysInline);
+        // See AddFunctionDeclaration in module.cpp for why we use a custom
+        // attribute here instead of AlwaysInline.
+        function->addFnAttr("ispc-defer-alwaysinline");
     }
     if (isNoInline) {
         function->addFnAttr(llvm::Attribute::NoInline);
--- a/src/opt.cpp
+++ b/src/opt.cpp
@@ -605,7 +605,10 @@ void ispc::Optimize(llvm::Module *module, int optLevel) {
         optPM.addFunctionPass(llvm::SROAPass(llvm::SROAOptions::ModifyCFG));
         optPM.addFunctionPass(llvm::InferAlignmentPass());
         optPM.addFunctionPass(llvm::InstCombinePass());
-        optPM.addFunctionPass(IsCompileTimeConstantPass(true));
+        // Don't resolve __is_compile_time_constant_* in `inline` functions not
+        // yet inlined: the final inliner below may expose a constant argument.
+        // The post-inline cleanup resolves what is left (#3804).
+        optPM.addFunctionPass(IsCompileTimeConstantPass(true, /*skipDeferredInline=*/true));
         optPM.addFunctionPass(LowerAMXBuiltinsPass());
         optPM.addFunctionPass(IntrinsicsOpt());
         optPM.addFunctionPass(InstructionSimplifyPass());
@@ -673,6 +676,10 @@ void ispc::Optimize(llvm::Module *module, int optLevel) {
         // freshly inlined bodies, recovering the cross-procedural
         // simplifications we would otherwise miss by deferring inlining.
         optPM.initFunctionPassManager();
+        // Resolve __is_compile_time_constant_* left in the just-inlined bodies
+        // (the marker is gone, so nothing is skipped now). InstCombine/SimplifyCFG
+        // below fold the result.
+        optPM.addFunctionPass(IsCompileTimeConstantPass(true));
         optPM.addFunctionPass(llvm::InferAlignmentPass());
         optPM.addFunctionPass(llvm::InstCombinePass());
         optPM.addFunctionPass(InstructionSimplifyPass());
--- a/src/opt.cpp
+++ b/src/opt.cpp
@@ -343,6 +343,9 @@ void ispc::Optimize(llvm::Module *module, int optLevel) {
         optPM.addFunctionPass(IsCompileTimeConstantPass(true));
         optPM.commitFunctionToModulePassManager();
 
+        // Convert the deferred ispc-defer-alwaysinline marker to LLVM alwaysinline so
+        // the inliner that runs next will honor `inline`-qualified functions.
+        optPM.addModulePass(RestoreInlineAttrPass());
         optPM.addModulePass(llvm::ModuleInlinerWrapperPass(IP));
         optPM.addModulePass(RemovePersistentFuncsPass());
 
@@ -657,6 +660,12 @@ void ispc::Optimize(llvm::Module *module, int optLevel) {
         optPM.addFunctionPass(ScalarizePass());
         optPM.addFunctionPass(llvm::ADCEPass());
         optPM.commitFunctionToModulePassManager();
+        // Convert ispc-defer-alwaysinline (set on `inline`-qualified functions at
+        // parse time) to alwaysinline now that all function-level passes have
+        // run. The next inliner is the final one in the pipeline; deferring
+        // forced inlining until here lets each callee be optimized in
+        // isolation before being expanded into its callers (issue #3804).
+        optPM.addModulePass(RestoreInlineAttrPass());
         optPM.addModulePass(llvm::ModuleInlinerWrapperPass(IP));
         optPM.addModulePass(llvm::StripDeadPrototypesPass());
         optPM.addModulePass(RemovePersistentFuncsPass());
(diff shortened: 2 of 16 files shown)


src/opt.cpp
```diff
@@ -661,7 +664,25 @@ void ispc::Optimize(llvm::Module *module, int optLevel) {
...
+        // deferred by RestoreInlineAttrPass. Run a small cleanup so that
+        // caller-context constants and CFG simplifications fold through the
+        // freshly inlined bodies, recovering the cross-procedural
+        // simplifications we would otherwise miss by deferring inlining.
+        optPM.initFunctionPassManager();
+        optPM.addFunctionPass(llvm::InferAlignmentPass());
+        optPM.addFunctionPass(llvm::InstCombinePass());
+        optPM.addFunctionPass(InstructionSimplifyPass());
+        optPM.addFunctionPass(llvm::SimplifyCFGPass(simplifyCFGopt));
+        optPM.addFunctionPass(llvm::ADCEPass());
```
dbabokin, line 687: [P2] Deferred inlining runs after the final compile-time-constant pass

At optimized levels, `IsCompileTimeConstantPass(true)` still runs before this restores `alwaysinline` and runs the final inliner. That pass permanently lowers unresolved `__is_compile_time_constant_*` calls to false while deferred inline callees still see formal parameters; after final inlining exposes constant call-site arguments, the new cleanup cannot recover because it does not run `IsCompileTimeConstantPass` and the call is already gone. This can regress large `inline` wrappers around shuffle-style builtins. Keep the pre-inline pass non-final or rerun the last try after the final inliner before simplification.

tests/lit-tests/3804.ispc
```diff
@@ -0,0 +1,23 @@
...
+
+// RUN: %{ispc} %s --target=host --nowrap --emit-llvm-text -o - \
+// RUN:   | FileCheck %s --implicit-check-not='@helper'
+
+// Verify the callee is inlined and erased: only `caller` remains in the
+// final IR — no standalone `helper` definition and no call to `helper`
+// from inside `caller`. --implicit-check-not enforces this across the
+// whole input, including inside `caller`'s body.
+// CHECK-LABEL: define {{.*}} @caller
+// CHECK: ret
```
dbabokin: [P2] Regression test would pass before the fix

This only checks that `helper` is absent from final IR, which main already satisfied because parse-time `alwaysinline` forced early inlining. A regression that reintroduces early `AlwaysInline` would still pass. The test should assert the behavior that changed, for example with an intermediate `--debug-phase` check showing the deferred marker survives early inliners and is restored only before the final inliner, or with a codegen reproducer for issue #3804.

dbabokin (commented):
Posted two inline review findings from local inspection.

larsbohn:
Not sure whether this is related, but what we commonly do is pass uniform const args to our inlined or non-inlined functions which then do an if/switch on those passed args - the expectation is then that all this evaluation is done at compile time due to some constant propagation.

The idea is to have the code generating variants for all combinations of such args - much like a template does in C++.

Like this:
```
inline int Calculate(ValueType value, uniform const bool calculateWithMethodA)
{
	if (calculateWithMethodA)
	{
		return ...;
	}
	else
 	{
		return ...;
	}
}

void ProcessArray(ValuesArrayType values, uniform const bool calculateWithMethodA)
{
	for ("value in values")
	{
		value = Calculate(value, calculateWithMethodA);
	}
}

void main(ValuesArrayType values, uniform const bool calculateWithMethodA)
{
	if (calculateWithMethodA)
	{
		ProcessArray(values, true);
	}
	else
	{
		ProcessArray(values, false);
	}
}
```
I hope this will be properly optimized then.

aneshlya:
Would be good to add a test for larsbohn's uniform const argument specialization.
@nurmukhametov, are you planning finishing this PR?

nurmukhametov:
> Would be good to add a test for larsbohn's uniform const argument specialization. @nurmukhametov, are you planning finishing this PR?

Yes, I plan to address the comments and add this test by the end of the weekend.

src/opt.cpp
nurmukhametov: Good catch, thanks. I have added a skip parameter for the first run of `IsCompileTimeConstantPass` to skip substituting `__is_compile_time_constant_*` calls in inline deferred functions.

tests/lit-tests/3804.ispc
nurmukhametov: I have changed the test.

nurmukhametov:
> Would be good to add a test for larsbohn's uniform const argument specialization. @nurmukhametov, are you planning finishing this PR?

I have added the larsbohn's test and addressed review comments.

larsbohn:
> > Would be good to add a test for larsbohn's uniform const argument specialization. @nurmukhametov, are you planning finishing this PR?
> 
> I have added the larsbohn's test and addressed review comments.

Will this optimize also if the inner function is not inlined? (Per constant propagation or the like)?

src/opt/RestoreInlineAttrPass.cpp
```diff
@@ -0,0 +1,36 @@
...
+*/
+
+#include "RestoreInlineAttrPass.h"
+
+namespace ispc {
+
+llvm::PreservedAnalyses RestoreInlineAttrPass::run(llvm::Module &M, [[maybe_unused]] llvm::ModuleAnalysisManager &MAM) {
+    bool changed = false;
+    for (llvm::Function &F : M) {
+        if (F.hasFnAttribute("ispc-defer-alwaysinline")) {
```
aneshlya, line 14: Would be good to make a little clean up - the string "ispc-defer-alwaysinline" duplicated across 6 sites with description why it's needed. A typo in any one silently breaks forced inlining with no compile error. Please define it once, e.g. `inline constexpr const char *ISPCDeferAlwaysInlineAttr = "ispc-defer-alwaysinline"`; in `src/opt/ISPCPass.h` or `src/ispc.h`.

src/func.cpp
```diff
@@ -53,7 +53,12 @@ bool Function::IsInternal() const {
     bool isInline = false;
     llvm::Function *function = sym->function;
     if (function != nullptr) {
-        isInline = (function->getAttributes().getFnAttrs().hasAttribute(llvm::Attribute::AlwaysInline));
+        const llvm::AttributeSet &fnAttrs = function->getAttributes().getFnAttrs();
+        // ispc-defer-alwaysinline is the deferred form of AlwaysInline used for
+        // user-written `inline`; either marker means the source qualifier was
+        // present.
+        isInline =
```
aneshlya, line 60: From maintainability point of view I'd extract this condition to separate function e.g. `bool FunctionWillBeInlined(const llvm::Function *fn);   // checks AlwaysInline OR marker`

aneshlya:
There are several performance issues after this change:
## Issue 1 — optimization-barrier intrinsics survive inlining (≈ +20%)

An inlined SoA transpose helper left empty `llvm.stacksave` / `llvm.stackrestore` pairs
wrapped around pure `shufflevector` ops. They emit no instructions, but they act as
**optimization barriers** that block CSE and gather coalescing, roughly doubling the
shuffles and gathers in the hot loop.

## Issue 2 — un-cleaned caller-context redundancy (≈ +17%)

No barriers here. The merged body simply retained redundant loads/stores/GEPs that the
standalone callee couldn't have folded because they only become foldable once the call is
inlined into its caller's context. The thin cleanup left them in place.

The fix for both 1 and 2 - is in **ISPC's optimization-pipeline construction (`opt.cpp`)**, in the cleanup
sequence that runs **after the final/deferred inliner**. It should run SROA, SimplifyCFG, InstCombine.

## Issue 3 — NOT fixable by adding passes (gather index width, ≈ +13–17%)

A `VectorGather` over a 32-bit index should lower to a **dword-indexed gather** (32-bit
index, 8 lanes per instruction). ISPC's gather lowering (`ImproveMemoryOps`) picks that cheap form only when it
can *prove* the index is 32-bit and factor out the common base pointer. With deferred
inlining, the gather is lowered **after the index has already been widened into a 64-bit
pointer value**, so the lowering emits the **qword-indexed gather** (64-bit index, only 4
lanes per instruction → twice as many gather instructions). That choice is **baked into
the IR** and is unrecoverable — re-running the new cleanup, GVN/CSE, or even a full LLVM
`-O3` all keep the qword form.
[lit_examples.zip](https://github.com/user-attachments/files/29396194/lit_examples.zip)

PR #3813, opened 2026-04-26
Fix all tests of the recent stdlib-math-function PRs

zephyr111:
This PR update the tests operating on uniform types for all recent PRs merged since they was not actually computing something useful. It should now properly compare the uniform version with the varying one. This fix is initiated from [this (AI) review post](https://github.com/ispc/ispc/pull/3796#discussion_r3111571112) (from PR #3476).

**UPDATE:**

The PR also fix all tests to make them independent. Having the `float16` tests not dependent of the `float` one certainly fix the CI issues. Indeed, some values where converted to infinities due to overflow during `float16` casts. The thing is the fast-math version is not guaranteed to return a stable result in such a case and the `ok` function may results in internal NaN in pathological cases (which is an issue per se only if aggressive fast-math flags are enabled -- otherwise the function should just return `false` even though both results are `inf`). Thus, I fixed that so there is no overflows in input/output values. I also recomputed the `float` and `double` so the results should now be ULP-perfect and better formatted. All the tests have now the same structure (a bit easier to maintained although there are a lot of independent test files).

### DIFF
--- a/tests/func-tests/cosh-double-uniform.ispc
+++ b/tests/func-tests/cosh-double-uniform.ispc
@@ -7,11 +7,34 @@ uniform bool ok(uniform double x, uniform double ref) {
     return (abs(x - ref) < 1e-15d) || abs((x - ref) / ref) < 1e-12d;
 }
 
-task void f_fu(uniform float RET[], uniform float aFOO[], uniform float b) {
-    for (uniform int i = 0; i != programCount; ++i) {
-        uniform double ref = cosh((uniform double)((aFOO[i] / programCount) * 3.0d - 1.5d));
-        uniform double res = cosh((aFOO[i] / programCount) * 3.0d - 1.5d);
-        RET[i] = ok(res, ref) ? 0 : 1;
+task void f_v(uniform float RET[]) {
+    const uniform double x[] = {
+        0.27d, 0.99d,
+        2e-2d, 1e-12d,
+        5.0d, 7.0d,
+        -3.0d, -1e-2d,
+        -0.0d, +0.0d,
+        80.0d, 379.1d
+    };
+
+    const uniform double ref[] = {
+        +1.036671972535050390e+00d, +1.531405581685653994e+00d,
+        +1.000200006666755570e+00d, +1.000000000000000000e+00d,
+        +7.420994852478784765e+01d, +5.483170351552121247e+02d,
+        +1.006766199577776533e+01d, +1.000050000416668139e+00d,
+        +1.000000000000000000e+00d, +1.000000000000000000e+00d,
+        +2.770311192196754917e+34d, +2.187802397202284545e+164d
+    };
+
+    const uniform count = sizeof(x) / sizeof(x[0]);
+    assert(count == sizeof(ref) / sizeof(ref[0]));
+
+    RET[programIndex] = 0;
+
+    for (uniform int i = 0; i < count; ++i)
+    {
+        const uniform double res = cosh(x[i]);
+        RET[programIndex] += ok(res, ref[i]) ? 0 : 1;
     }
 }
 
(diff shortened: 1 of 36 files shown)


aneshlya (changes requested):
This PR uses `varying` as the reference, not independent `uniform` computation. The tests verify that uniform and varying implementations produce the same result, but that's circular if both implementations have the same bug, the test passes.
I suggest to reuse the same approach as you have in varying tests for uniform values (feel free to copy them, don't worry about code duplication).

zephyr111:
> This PR uses varying as the reference, not independent uniform computation. 

Yes.

> The tests verify that uniform and varying implementations produce the same result, but that's circular if both implementations have the same bug, the test passes.

This test will pass, but not the varying one. If both pass, then both version are fine. This works because equality is transitive (`(a = b ∧ b = c) ⇔ a = c` where `a` represent the reference results, `b` the `varying` implementation and `c` the `uniform` one).

> feel free to copy them, don't worry about code duplication

IMHO, It is quite a mess to maintain tests or to write new ones, so I wish to make this worse. For example if new tests are added independently of the precision, then we then need to modify 6 files per function instead of 3. This is bug bug-prone when there are many files. It would be OK if the code of the uniform and varying test would be very similar. Maybe we can improve this a bit by defining helper function calling the uniform math function for each item of the gang (similar to what is done for libm functions in double-precision varying functions in the stdlib) so to mostly keep the same main function we can then copy-past from the varying test to the uniform one. Another approach is to iterate over arrays in both version, but this means to remove test-cases section which is less debug-friendly (and I have to say this modification is quite tedious to do since there are 12 tests to update).

aneshlya:
I agree with you about maintenance cost (and we can cover it by new issue and implement consistently for all tests). But the current codebase prioritizes independent verification in tests. The established pattern is that each test independently validates one functionality with its own reference values:

  - ceil-float-uniform.ispc / ceil-float-varying.ispc - each has its own test cases, not referencing each other
  - clampfloat16_uniform.ispc / clampfloat16_varying.ispc - each defines independent result() with hardcoded expected
  values
  - sinh-float-varying.ispc - uses hardcoded reference values: ref = float4(0.27329248, 1.1598289, ...)
  - transcendentals-*-*.ispc - same pattern with hardcoded input/output pairs

  For the uniform tests, I suggest copying the structure from the corresponding varying tests and adapting for uniform.
  For example, sinh-float-uniform.ispc could look like:
```

  task void f_fu(uniform float RET[], uniform float aFOO[], uniform float b) {
      uniform int error = 0;

      // Case 1: Basic values
      error += ok(sinh(0.27f), 0.27329248f) ? 0 : 1;
      error += ok(sinh(0.99f), 1.1598289f) ? 0 : 1;
      error += ok(sinh(1.17f), 1.4558128f) ? 0 : 1;
      error += ok(sinh(2.67f), 7.185359f) ? 0 : 1;

      // Case 2: Small values
      error += ok(sinh(4e-2f), 4.001067e-02f) ? 0 : 1;
      // ... etc, same reference values as varying test

      RET[0] = error;
  }

  task void result(uniform float RET[]) { RET[programIndex] = 0; }
```
  This uses the same reference values as the varying test, but tests the uniform implementation independently. The
  duplication is intentional.

zephyr111:
@aneshlya Ok, I understand. I applied the second approach on the `tanh` test so you can see the result. The uniform and varying tests are now very similar so they can be more easily be modified/maintained. This ease the creation of new tests too. The tests cases are not so important compared to that. If you think it is better with them, then we can add them as comments at the end of each line. If `tanh` is now fine to you, I will generalize this to all other tests.

aneshlya:
@zephyr111, looks good to me!

zephyr111:
@aneshlya I fixed all tests, including `float16` ones. I updated the description of the PR so to explain the new changes. I also rebased this on `main`. Everything should be fine now.

aneshlya (approved):
Looks good! Thank you!

PR #3816, opened 2026-04-30
Fix foreach macro range preprocessing

### LINKED ISSUE #3809: Parsing error with defines in foreach ranges without spaces
The following ISPC code fails to compile. This is because of a missing space between `...` and `MAXI`. I don't expect this to happen and consider this as a parser bug. Do you agree?
```ispc
#define MAXI 5
void test() {
    foreach(i = 0...MAXI)
    {}
}
```

azwolski:
## Description
Fixes #3809.

Clang tokenizes cases like `0...MAXI` as a single preprocessing-number token, so `MAXI` is not expanded as a macro. This breaks `foreach` ranges when there is no space between the numeric start value, the `...` operator, and the macro upper bound.

This patch normalizes the main input file before running Clang's preprocessor. It uses Clang's raw lexer to find numeric tokens containing `...` followed by an identifier-start character, then inserts a space before the ellipsis:

```ispc
0...MAXI -> 0 ...MAXI
```

## Related Issue

## Checklist

### DIFF
--- a/src/preprocessor.cpp
+++ b/src/preprocessor.cpp
@@ -1,5 +1,5 @@
 /*
-  Copyright (c) 2025, Intel Corporation
+  Copyright (c) 2026, Intel Corporation
 
   SPDX-License-Identifier: BSD-3-Clause
 */
@@ -16,6 +16,7 @@
 #include "util.h"
 #include "version.h"
 
+#include <cstring>
 #include <ctype.h>
 #include <fcntl.h>
 #include <memory>
@@ -44,6 +45,7 @@
 #include <clang/Frontend/Utils.h>
 #include <clang/Lex/HeaderSearch.h>
 #include <clang/Lex/HeaderSearchOptions.h>
+#include <clang/Lex/Lexer.h>
 #include <clang/Lex/ModuleLoader.h>
 #include <clang/Lex/Preprocessor.h>
 #include <clang/Lex/PreprocessorOptions.h>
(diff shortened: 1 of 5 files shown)


src/preprocessor.cpp
```diff
@@ -369,12 +371,79 @@ static void lSetPreprocessorOptions(const std::shared_ptr<clang::PreprocessorOpt
...
 
+// Normalize cases like "0...MAXI" before running Clang's preprocessor.
+// Clang treats this as one preprocessing-number token, so MAXI is not
+// expanded as a macro. Insert a space before "..." so the range operator
+// and the macro are tokenized separately.
+//
+// Clang's raw lexer is used to avoid changing comments, strings, or
+// character literals.
+//
+// TODO: This currently only handles the main input file. Included files need separate support.
```
zephyr111: Do you plan to add this in another PR or later in this one?

src/preprocessor.cpp
```diff
@@ -369,12 +371,79 @@ static void lSetPreprocessorOptions(const std::shared_ptr<clang::PreprocessorOpt
...
+    clang::LangOptions langOpts;
+    lSetLangOptions(&langOpts);
+
+    // Collect offsets of "..." that need a preceding space
+    std::vector<size_t> insertions;
+
+    clang::Lexer lexer(clang::SourceLocation(), langOpts, source.data(), source.data(), source.data() + source.size());
+    clang::Token tok;
+
+    while (!lexer.LexFromRawLexer(tok)) {
```
zephyr111, line 393: I was a bit worried about the possible performance impact of using the lexer pass one more time, but it turns out that the lexer is pretty fast and this does not introduce any significant performance loss when computing the `stdlib.ispc` (with the `-dD` flag so to measure the preprocessing step which is actually only a fraction of the compilation time already). The performance difference is tiny (<5%) even when the stdlib is modified so 800 replacements are needed, especially considering the preprocessing step is <10% of the whole compilation time.

zephyr111 (approved):
Looks good to me (besides the TODO note)!
I did not found any issues while performing some additional tests.

dbabokin:
@codex review

src/preprocessor.cpp
azwolski: I would like to try

azwolski: Would like me to remove the TOODs?

src/preprocessor.cpp
```diff
@@ -422,17 +491,66 @@ static int lExecPreprocessor(llvm::Module *module, const char *infilename, llvm:
...
+            llvm::errs() << "ISPC: error reading stdin\n";
+            return 1;
+        }
+
+        std::string sourceContent = fileBufferOrError.get()->getBuffer().str();
+        std::string fixedContent = lFixDotDotDotSpacing(sourceContent);
+
+        stdinBuffer = llvm::MemoryBuffer::getMemBufferCopy(fixedContent, "<stdin>");
+        inputFile = clang::FrontendInputFile(stdinBuffer->getMemBufferRef(), clang::InputKind());
+    } else {
```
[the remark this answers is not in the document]
azwolski: @copilot apply changes based on [this feedback](https://github.com/ispc/ispc/pull/3816#discussion_r3168905018)

src/preprocessor.cpp
```diff
@@ -1,5 +1,5 @@
 /*
-  Copyright (c) 2025, Intel Corporation
+  Copyright (c) 2026, Intel Corporation
```
aneshlya: Please update copyright to 2025-2026

tests/lit-tests/3809-include-limitation.ispc
```diff
@@ -0,0 +1,5 @@
+// XFAIL: *
+// RUN: %{ispc} %s --target=host --nostdlib --nowrap -o %t.o
```
aneshlya: Consider reordering XFAIL test to put // RUN: before // XFAIL: (more common pattern)

src/preprocessor.cpp
aneshlya: Please remove TODO and create an issues instead. Thanks for creating lit test for this.

aneshlya (changes requested):

elena901:
A little bit late with it, but i like this [option](https://github.com/ispc/ispc/issues/3809#issuecomment-4322484569).
Richard Jerome:
`Fix the pre-processor in the clang code itself so for it not to recognize this as a single token`.

aneshlya:
> A little bit late with it, but i like this https://github.com/ispc/ispc/issues/3809#issuecomment-4322484569.

@elena901, this PR already implements the option 1 and it's not relevant place for such comment, it should go to the linked issue.

aneshlya (approved):
Looks good!

PR #3830, opened 2026-05-09
Fix #3828: track header deps in builtins build

### LINKED ISSUE #3828: Modifying builtins-c-common.hpp does not trigger a rebuild
When `builtins-c-common.hpp` is modified, I expect the depending files to be recompiled. For example, `builtins-c-cpu.hpp` should be compiled again. However, this is not the case: the modification is silently ignored. This results in sneaky issues.

**EDIT:**
There is an even bigger issue: when the cpp files cannot be compiled due to errors in it, errors are shown (and ignored) but it does not break the build the first time. In subsequent builds, the file is not even re-compiled! It looks like an empty llvm file is generated in this case and this is used to generate a bc file which actually contains nothing... This certainly results in weird bug later (possibly even at runtime).

nurmukhametov:
## Description

Modifications to headers included by builtins/builtins-c-cpu.cpp (builtins-c-common.hpp, array.hpp, src/builtins-info.h) did not trigger a rebuild because add_custom_command in builtin_to_cpp and builtin_wasm_to_cpp only listed the .cpp file in DEPENDS.
    
Add BUILTINS_C_CPU_HEADER_DEPS, mirroring the existing M4_IMPLICIT_DEPENDENCIES pattern, and pass it to DEPENDS in both functions so a header edit correctly rebuilds the four bitcode targets.

## Related Issue
- [X] Linked to relevant issue: #3828

## Checklist

### DIFF
--- a/cmake/GenerateBuiltins.cmake
+++ b/cmake/GenerateBuiltins.cmake
@@ -63,6 +63,15 @@ if (${LLVM_VERSION_NUMBER} VERSION_GREATER_EQUAL "20.1.2")
         builtins/target-avx10_2-x64-common.ll)
 endif()
 
+# Headers included by builtins/builtins-c-cpu.cpp. Listed explicitly so that
+# add_custom_command rebuilds the bitcode when these headers change. DEPFILE
+# support in add_custom_command is generator-specific for CMake < 3.20, so we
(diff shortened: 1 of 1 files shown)


aneshlya (approved):
LGTM!

PR #3831, opened 2026-05-10
Fix the build (i.e. bc file generation issues)

zephyr111:
This fix the build issue reference in the issue #3828 .
The fix is meant to be more generic than #3830 (though IDK if this is overkill or dangerous).
Besides the missing dependencies, it replace the pipes with two dependent commands. Indeed, when a command fails in a pipe, the encompassing command (i.e. the one including the pipe) does not fail. Dependent commands do.
However, this causes a side effect : other build issues (referenced in the issue by @nurmukhametov) which was already there before (just ignored / silent) and not fixed yet in this PR (hence the draft). 

## Related Issue

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -493,6 +493,10 @@ endif()
 message(STATUS "BITCODE_FOLDER is ${BITCODE_FOLDER}")
 file(MAKE_DIRECTORY ${BITCODE_FOLDER})
 
+set(LLVMIR_FOLDER ${CMAKE_CURRENT_BINARY_DIR}/llvmir)
+message(STATUS "LLVMIR_FOLDER is ${LLVMIR_FOLDER}")
+file(MAKE_DIRECTORY ${LLVMIR_FOLDER})
+
 # Generate stdlib width families from X86_TARGETS and ARM_TARGETS
 # and C++ code for stdlib target mapping
 define_stdlib_families()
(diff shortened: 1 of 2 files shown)


nurmukhametov:
As I mentioned in https://github.com/ispc/ispc/issues/3828#issuecomment-4413864378, I'm not sure we should fix it this way. It's a correct fix, but it introduces build failures that didn't occur before.

How about a different approach: during the configuration step, detect when the build environment can't compile a specific configuration and disable its build accordingly?

zephyr111:
I just made this draft PR so to share the modifications I made locally (before your changes) on my side -- just not to do the same thing twice. IDK if it could help somehow. Feel free to use it so to improve your PR, use it as a new basis so to fix the issue, or possibly discard it if this is not really useful.

Please note that it tracks the dependencies automatically.

> How about a different approach: during the configuration step, detect when the build environment can't compile a specific configuration and disable its build accordingly?

This seems quite good to me, especially compared to the current state. Aren't some configurations required for the build to be actually correct?

zephyr111:
I am closing this draft since the current solution looks good enough for my needs (and also since the automatic detection is not as simple as it seems -- at least for someone not very familiar with CMake). I also do not want to pollute the list of open PRs.

@nurmukhametov Do you think we should to open an (enhancement) issue on the matter?

aneshlya:
I created https://github.com/ispc/ispc/issues/3845 as follow up.

zephyr111:
@aneshlya Thank you!

PR #3846, opened 2026-05-23
Fix reduce_add functions

### LINKED ISSUE #3834: Several discrepancies in the implementation of reduction functions
In the documentation, we can read that the reduction function returns a type bigger than the input (certainly to avoid overflow):
```ispc
uniform int16 reduce_add(int8 x)
uniform unsigned int16 reduce_add(unsigned int8 x)
uniform int32 reduce_add(int16 x)
uniform unsigned int32 reduce_add(unsigned int16 x)
uniform int64 reduce_add(int32 x)
uniform unsigned int64 reduce_add(unsigned int32 x)
uniform int64 reduce_add(int64 x)
uniform unsigned int64 reduce_add(unsigned int64 x)
```
Note that this is not the case for the 64-bit type since it is the biggest supported type so far.

In the stdlib implementation, we can see that functions `__reduce_add_intxx` are called internally. For example, here is 3 of them:
```ispc
__declspec(safe) static inline uniform unsigned int16 reduce_add(unsigned int8 x) {
    return __reduce_add_int8(__mask ? x : (int8)0);
}
__declspec(safe) static inline uniform int32 reduce_add(int16 x) { return __reduce_add_int16(__mask ? x : (int16)0); }
__declspec(safe) static inline uniform int64 reduce_add(int32 x) {
    // Zero out the values for lanes that aren't running
    return __reduce_add_int32(__mask ? x : 0);
}
```

The thing is the underlying functions does not actually use the matching proper types in all cases. See `builtins.isph`:
```ispc
EXT inline READNONE uniform int16 __reduce_add_int8(varying int8);
EXT inline READNONE uniform int16 __reduce_add_int16(varying int16);
EXT inline READNONE uniform int32 __reduce_add_int32(varying int32);
EXT inline READNONE uniform int64 __reduce_add_int64(varying int64);
```
We can see that only the `int8` type is correctly supported here.

Fixing this will certainly impact performance (not that much overall since reductions are generally performed once after loops), but the current code is clearly bogus IMHO (at least, this is neither documented nor expected regarding the output types). Do you agree?

-----

There is another issue in the implementation: the unsigned version just call the signed version so big unsigned integers are converted to negative value and I highly doubt the result is correct in this case. See `stdlib.ispc`:

```ispc
__declspec(safe) static inline uniform unsigned int64 reduce_add(unsigned int32 x) {
    // Set values for non-running lanes to zero so they don't affect the
    // result.
    return __reduce_add_int32(__mask ? x : 0);
}
```

IMHO, unsigned functions (like `__reduce_add_uint32` for the above implementation) are missing.

zephyr111:
## Description

This PR fixes #3834 . More specifically it:
- Fix reduce_add(int8) for signed (negative) values;
- Implement unsigned versions;
- Change the output type so it is now the same than the input type (no widening) -- most target was not computing this correctly anyway;
- Keep using psadbw in `__reduce_add_uint8` though the output type is now `uint8` instead of `int16` for sake of consistency with other functions.
- Fix the tests and add new ones, especially to catch the issue in `__reduce_add_int8` and check the unsigned versions.

I think it would be great to add more stdlib functions with widened output type later. Indeed, it would be more efficient on x86 CPU for `uint8` when the user want a `int16`/`uint16` result (no `uint8` wrap-around). On x86, I think we could also use the instruction `pmaddwd` for computing `reduce_add(int16) -> int32` (AFAIK not yet done previously). AFAIK, RISC and SVE (SVE2?) targets could also provide better implementations than naive ones (though SVE is not supported yet and RISC is certainly used by very few people). It is not critical since SIMD reductions are known to be a bit expensive and they rarely are on a hot path anyway.

Please note that I also reorganized the order of some function for sake of clarity. I also added a missing `undef` (minor).

PS: updating the IR of the 19 targets is pretty tedious and bug prone.

## Related Issue

## Checklist

### DIFF
--- a/docs/ispc.rst
+++ b/docs/ispc.rst
@@ -281,10 +281,16 @@ New Features:
 
 Standard Library Changes:
 
-* **Breaking change**: the return type of the ``reduce_add()`` functions is now
-  the same than the input type. Users should now cast the inpout to a wider type
-  themselves so to prevent overflows. Any code relying on the widened return
-  type needs to be updated!
+* **Breaking change**: ``reduce_add()`` now returns the same type as its
+    input. Previously the signatures widened the result, but the
+    implementation was inconsistent and partially incorrect. The new
+    same-type convention also matches the prevailing convention in other
+    SIMD/SPMD ecosystems and is consistent with ``reduce_min()`` /
+    ``reduce_max()`` in ISPC. For overflow-safe accumulation, cast the
+    input to a wider type explicitly, e.g. ``reduce_add((int32)x)`` for an
+    ``int16`` ``x``. Any code that relied on the previous widened return
+    type must be updated.
+
 * The Gauss error function ``erf`` and the complementary error function ``erfc``
   have been added to the standard library.
 * The ``expm1`` function (meant to compute ``exp(x)-1`` accurately), and the
--- a/docs/ispc.rst
+++ b/docs/ispc.rst
@@ -269,12 +269,18 @@ New Architecture Support:
   included in official ISPC binaries. To use it, build ISPC from source with
   the `PPC64_ENABLED=ON` CMake option. Usage: `ispc --arch=ppc64le
   --target=generic-i32x4 foo.ispc -o foo.o`.
+
+New Features:
+
 * Implementation of two new fast-math modes: ``balanced`` and ``aggressive``.
   These modes bring more aggressive optimizations compared to the default
   ``legacy`` mode. The latter was the mode provided so far in previous ISPC
   versions. It is still available and left unchanged. The fast-math mode can be
   specified with the option ``--opt=fast-math:<mode>``. See the optimization
   settings section for more information about the new modes.
+
+Standard Library Changes:
+
 * The Gauss error function ``erf`` and the complementary error function ``erfc``
   have been added to the standard library.
 * The ``expm1`` function (meant to compute ``exp(x)-1`` accurately), and the
@@ -282,6 +288,9 @@ New Architecture Support:
   added to the standard library.
 * The hyperbolic functions ``sinh``, ``cosh`` and ``tanh`` have also been added
   to the standard library.
+* The return type of the ``reduce_add()`` functions is now the same than the
+  input type. Users should now cast the inpout to a wider type themselves so to
+  prevent overflows.
 
 ISPC Targets:
 
@@ -6726,12 +6735,12 @@ instances are added together by the ``reduce_add()`` function.
 
 ::
 
-    uniform int16 reduce_add(int8 x)
-    uniform unsigned int16 reduce_add(unsigned int8 x)
-    uniform int32 reduce_add(int16 x)
-    uniform unsigned int32 reduce_add(unsigned int16 x)
-    uniform int64 reduce_add(int32 x)
-    uniform unsigned int64 reduce_add(unsigned int32 x)
+    uniform int8 reduce_add(int8 x)
+    uniform unsigned int8 reduce_add(unsigned int8 x)
+    uniform int16 reduce_add(int16 x)
+    uniform unsigned int16 reduce_add(unsigned int16 x)
+    uniform int32 reduce_add(int32 x)
+    uniform unsigned int32 reduce_add(unsigned int32 x)
     uniform int64 reduce_add(int64 x)
     uniform unsigned int64 reduce_add(unsigned int64 x)
 
--- a/docs/ispc.rst
+++ b/docs/ispc.rst
@@ -281,6 +281,10 @@ New Features:
 
 Standard Library Changes:
 
+* **Breaking change**: the return type of the ``reduce_add()`` functions is now
+  the same than the input type. Users should now cast the inpout to a wider type
+  themselves so to prevent overflows. Any code relying on the widened return
+  type needs to be updated!
 * The Gauss error function ``erf`` and the complementary error function ``erfc``
   have been added to the standard library.
 * The ``expm1`` function (meant to compute ``exp(x)-1`` accurately), and the
@@ -288,9 +292,6 @@ Standard Library Changes:
   added to the standard library.
 * The hyperbolic functions ``sinh``, ``cosh`` and ``tanh`` have also been added
   to the standard library.
-* The return type of the ``reduce_add()`` functions is now the same than the
-  input type. Users should now cast the inpout to a wider type themselves so to
-  prevent overflows.
 
 ISPC Targets:
 
--- a/builtins/generic.ispc
+++ b/builtins/generic.ispc
@@ -758,65 +758,62 @@ EXT READNONE varying float16 __pow_varying_half(varying float16 x, varying float
 EXT READNONE uniform float16 __sqrt_uniform_half(uniform float16 x) { return @llvm.sqrt(x); }
 EXT READNONE varying float16 __sqrt_varying_half(varying float16 x) { return @llvm.sqrt(x); }
 
-// 16-bit float reduction
+// 16-bit float add/min/max reductions
 EXT READNONE uniform float16 __reduce_add_half(varying float16 x) { return @llvm.vector.reduce.fadd(-0.0f16, x); }
 EXT READNONE uniform float16 __reduce_max_half(varying float16 x) { return @llvm.vector.reduce.fmax(x); }
 EXT READNONE uniform float16 __reduce_min_half(varying float16 x) { return @llvm.vector.reduce.fmin(x); }
 
-// 16-bit float min and max functions
-EXT READNONE uniform float16 __min_uniform_half(uniform float16 x, uniform float16 y) { return x < y ? x : y; }
-EXT READNONE varying float16 __min_varying_half(varying float16 x, varying float16 y) { return x < y ? x : y; }
-EXT READNONE uniform float16 __max_uniform_half(uniform float16 x, uniform float16 y) { return x > y ? x : y; }
-EXT READNONE varying float16 __max_varying_half(varying float16 x, varying float16 y) { return x > y ? x : y; }
-
-// horizontal int8 ops
-EXT READNONE uniform int16 __reduce_add_int8(varying int8 x) { return @llvm.vector.reduce.add((int16)x); }
-EXT READNONE uniform int8 __reduce_min_int8(varying int8 x) { return @llvm.vector.reduce.smin(x); }
-EXT READNONE uniform int8 __reduce_max_int8(varying int8 x) { return @llvm.vector.reduce.smax(x); }
-
-// horizontal int16 ops
-EXT READNONE uniform int16 __reduce_add_int16(varying int16 x) { return @llvm.vector.reduce.add((int32)x); }
-EXT READNONE uniform int16 __reduce_min_int16(varying int16 x) { return @llvm.vector.reduce.smin(x); }
-EXT READNONE uniform int16 __reduce_max_int16(varying int16 x) { return @llvm.vector.reduce.smax(x); }
-
-// horizontal float ops
+// float add/min/max reductions
 EXT READNONE uniform float __reduce_add_float(varying float x) { return @llvm.vector.reduce.fadd(-0.0f, x); }
 EXT READNONE uniform float __reduce_max_float(varying float x) { return @llvm.vector.reduce.fmax(x); }
 EXT READNONE uniform float __reduce_min_float(varying float x) { return @llvm.vector.reduce.fmin(x); }
 
-// horizontal int32 ops
-// TODO!: don't we lose result with reduce_add having int32 return type?
-EXT READNONE uniform int32 __reduce_add_int32(varying int32 x) { return @llvm.vector.reduce.add((int64)x); }
-EXT READNONE uniform int32 __reduce_min_int32(varying int32 x) { return @llvm.vector.reduce.smin(x); }
-EXT READNONE uniform int32 __reduce_max_int32(varying int32 x) { return @llvm.vector.reduce.smax(x); }
-
-// horizontal uint8 ops
-EXT READNONE uniform uint8 __reduce_min_uint8(varying uint8 x) { return @llvm.vector.reduce.umin(x); }
-EXT READNONE uniform uint8 __reduce_max_uint8(varying uint8 x) { return @llvm.vector.reduce.umax(x); }
-
-// horizontal uint16 ops
-EXT READNONE uniform uint16 __reduce_min_uint16(varying uint16 x) { return @llvm.vector.reduce.umin(x); }
-EXT READNONE uniform uint16 __reduce_max_uint16(varying uint16 x) { return @llvm.vector.reduce.umax(x); }
-
-// horizontal uint32 ops
-EXT READNONE uniform uint32 __reduce_min_uint32(varying uint32 x) { return @llvm.vector.reduce.umin(x); }
-EXT READNONE uniform uint32 __reduce_max_uint32(varying uint32 x) { return @llvm.vector.reduce.umax(x); }
-
-// horizontal double ops
+// double add/min/max reductions
 EXT READNONE uniform double __reduce_add_double(varying double x) { return @llvm.vector.reduce.fadd(-0.0d, x); }
 EXT READNONE uniform double __reduce_min_double(varying double x) { return @llvm.vector.reduce.fmin(x); }
 EXT READNONE uniform double __reduce_max_double(varying double x) { return @llvm.vector.reduce.fmax(x); }
 
-// horizontal int64 ops
+// int8 add/min/max reductions
+EXT READNONE uniform int8 __reduce_add_int8(varying int8 x) { return @llvm.vector.reduce.add(x); }
+EXT READNONE uniform int8 __reduce_min_int8(varying int8 x) { return @llvm.vector.reduce.smin(x); }
+EXT READNONE uniform int8 __reduce_max_int8(varying int8 x) { return @llvm.vector.reduce.smax(x); }
+
+// int16 add/min/max reductions
+EXT READNONE uniform int16 __reduce_add_int16(varying int16 x) { return @llvm.vector.reduce.add(x); }
+EXT READNONE uniform int16 __reduce_min_int16(varying int16 x) { return @llvm.vector.reduce.smin(x); }
+EXT READNONE uniform int16 __reduce_max_int16(varying int16 x) { return @llvm.vector.reduce.smax(x); }
+
+// int32 add/min/max reductions
+EXT READNONE uniform int32 __reduce_add_int32(varying int32 x) { return @llvm.vector.reduce.add(x); }
+EXT READNONE uniform int32 __reduce_min_int32(varying int32 x) { return @llvm.vector.reduce.smin(x); }
+EXT READNONE uniform int32 __reduce_max_int32(varying int32 x) { return @llvm.vector.reduce.smax(x); }
+
+// int64 add/min/max reductions
 EXT READNONE uniform int64 __reduce_add_int64(varying int64 x) { return @llvm.vector.reduce.add(x); }
 EXT READNONE uniform int64 __reduce_min_int64(varying int64 x) { return @llvm.vector.reduce.smin(x); }
 EXT READNONE uniform int64 __reduce_max_int64(varying int64 x) { return @llvm.vector.reduce.smax(x); }
 
-// horizontal uint64 ops
+// uint8 add/min/max reductions
+EXT READNONE uniform uint8 __reduce_add_uint8(varying uint8 x) { return @llvm.vector.reduce.add(x); }
+EXT READNONE uniform uint8 __reduce_min_uint8(varying uint8 x) { return @llvm.vector.reduce.umin(x); }
+EXT READNONE uniform uint8 __reduce_max_uint8(varying uint8 x) { return @llvm.vector.reduce.umax(x); }
+
+// uint16 add/min/max reductions
+EXT READNONE uniform uint16 __reduce_add_uint16(varying uint16 x) { return @llvm.vector.reduce.add(x); }
+EXT READNONE uniform uint16 __reduce_min_uint16(varying uint16 x) { return @llvm.vector.reduce.umin(x); }
+EXT READNONE uniform uint16 __reduce_max_uint16(varying uint16 x) { return @llvm.vector.reduce.umax(x); }
+
+// uint32 add/min/max reductions
+EXT READNONE uniform uint32 __reduce_add_uint32(varying uint32 x) { return @llvm.vector.reduce.add(x); }
+EXT READNONE uniform uint32 __reduce_min_uint32(varying uint32 x) { return @llvm.vector.reduce.umin(x); }
+EXT READNONE uniform uint32 __reduce_max_uint32(varying uint32 x) { return @llvm.vector.reduce.umax(x); }
+
+// uint64 add/min/max reductions
+EXT READNONE uniform uint64 __reduce_add_uint64(varying uint64 x) { return @llvm.vector.reduce.add(x); }
 EXT READNONE uniform uint64 __reduce_min_uint64(varying uint64 x) { return @llvm.vector.reduce.umin(x); }
 EXT READNONE uniform uint64 __reduce_max_uint64(varying uint64 x) { return @llvm.vector.reduce.umax(x); }
 
-// horizontal equality ops
+// equality reductions
 template <typename T> unmasked uniform bool __reduce_equal(varying T x, uniform T *uniform ret, UIntMaskType mask) {
     uniform int first_active = __count_trailing_zeros_uniform_i64(__movmsk(mask));
     uniform T first_val = __extract(x, first_active);
@@ -829,6 +826,15 @@ template <typename T> unmasked uniform bool __reduce_equal(varying T x, uniform 
         return true;
     }
     return false;
+}
+EXT uniform bool __reduce_equal_half(varying float16 val, uniform float16 *uniform ret, UIntMaskType mask) {
+    return __reduce_equal<float16>(val, ret, mask);
+}
+EXT uniform bool __reduce_equal_float(varying float val, uniform float *uniform ret, UIntMaskType mask) {
+    return __reduce_equal<float>(val, ret, mask);
+}
+EXT uniform bool __reduce_equal_double(varying double val, uniform double *uniform ret, UIntMaskType mask) {
+    return __reduce_equal<double>(val, ret, mask);
 }
 EXT uniform bool __reduce_equal_int8(varying int8 val, uniform int8 *uniform ret, UIntMaskType mask) {
     return __reduce_equal<int8>(val, ret, mask);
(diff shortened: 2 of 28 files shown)


dbabokin:
@codex review

aneshlya:
> updating the IR of the 19 targets is pretty tedious and bug prone.

Does this implementation produce the same code as generic.ispc?
```
define i8 @__reduce_add_int8(<N x i8>) nounwind readnone alwaysinline {
  reduceN(i8, @__add_varying_i8, @__add_uniform_i8)
}
```
If yes, there is no need to add it to builtin files, the generic implementation will be used if target-specific one is not found.

zephyr111:
> the generic implementation will be used if target-specific one is not found

Initial rational: 

The main point is that generic implementation does not call `reduceN`. It calls the following one:
```ispc
EXT READNONE uniform int8 __reduce_add_int8(varying int8 x) { return @llvm.vector.reduce.add(x); }
```
So I blindly called called `reduceN` in order to avoid any sneaky performance issue (I expected `reduceN` to be pretty efficient). Indeed, the LLVM intrinsic might not be optimal as it might perform a reduction on widened types depending on the output type. In fact, [here](https://godbolt.org/z/9bETsWdnd) is the difference between the generic `i32x8` implementation (on AVX2 CPUs) and the `avx2-i32x8` target where we can see that the generic signed 8-bit implementation performs 3 `vpaddw` (16-bit instruction) while the new 8-bit implementation should perform 3 `vpaddb` (8-bit instruction) instead. We can see that the current implementation wrongly calls `vpsadbw` btw. Please keep in mind that this is the current version (with type widening & bugs).

Here is the practical assembly code of respectively `out-avx2-i32x8.s` and `out-generic-i32x8-skylake.s` in this PR branch:

```asm
	.file	"test.ispc"
	.text
	.globl	test_s8___UM_vyt                # -- Begin function test_s8___UM_vyt
	.p2align	4
	.type	test_s8___UM_vyt,@function
test_s8___UM_vyt:                       # @test_s8___UM_vyt
# %bb.0:
	vpshufd	$85, %xmm0, %xmm1               # xmm1 = xmm0[1,1,1,1]
	vpaddb	%xmm0, %xmm1, %xmm0
	vpsrld	$16, %xmm0, %xmm1
	vpaddb	%xmm0, %xmm1, %xmm0
	vpextrb	$1, %xmm0, %eax
	vmovd	%xmm0, %ecx
	addb	%cl, %al
	vmovd	%eax, %xmm0
	vpbroadcastb	%xmm0, %xmm0
	retq
.Lfunc_end0:
	.size	test_s8___UM_vyt, .Lfunc_end0-test_s8___UM_vyt
                                        # -- End function
	.globl	test_u8___UM_vyT                # -- Begin function test_u8___UM_vyT
	.p2align	4
	.type	test_u8___UM_vyT,@function
test_u8___UM_vyT:                       # @test_u8___UM_vyT
# %bb.0:
	vmovq	%xmm0, %xmm0                    # xmm0 = xmm0[0],zero
	vpxor	%xmm1, %xmm1, %xmm1
	vpsadbw	%xmm1, %xmm0, %xmm0
	vpextrq	$1, %xmm0, %rax
	vmovq	%xmm0, %rcx
	orl	%eax, %ecx
	vmovd	%ecx, %xmm0
	vpbroadcastb	%xmm0, %xmm0
	retq
.Lfunc_end1:
	.size	test_u8___UM_vyT, .Lfunc_end1-test_u8___UM_vyT
                                        # -- End function
	.globl	test_s16___UM_vys               # -- Begin function test_s16___UM_vys
	.p2align	4
	.type	test_s16___UM_vys,@function
test_s16___UM_vys:                      # @test_s16___UM_vys
# %bb.0:
	vpshufd	$238, %xmm0, %xmm1              # xmm1 = xmm0[2,3,2,3]
	vpaddw	%xmm0, %xmm1, %xmm0
	vpshufd	$85, %xmm0, %xmm1               # xmm1 = xmm0[1,1,1,1]
	vpaddw	%xmm0, %xmm1, %xmm0
	vpextrw	$1, %xmm0, %eax
	vmovd	%xmm0, %ecx
	addl	%ecx, %eax
	vmovd	%eax, %xmm0
	vpbroadcastw	%xmm0, %xmm0
	retq
.Lfunc_end2:
	.size	test_s16___UM_vys, .Lfunc_end2-test_s16___UM_vys
                                        # -- End function
	.globl	test_u16___UM_vyS               # -- Begin function test_u16___UM_vyS
	.p2align	4
	.type	test_u16___UM_vyS,@function
test_u16___UM_vyS:                      # @test_u16___UM_vyS
# %bb.0:
	vpshufd	$238, %xmm0, %xmm1              # xmm1 = xmm0[2,3,2,3]
	vpaddw	%xmm0, %xmm1, %xmm0
	vpshufd	$85, %xmm0, %xmm1               # xmm1 = xmm0[1,1,1,1]
	vpaddw	%xmm0, %xmm1, %xmm0
	vpextrw	$1, %xmm0, %eax
	vmovd	%xmm0, %ecx
	addl	%ecx, %eax
	vmovd	%eax, %xmm0
	vpbroadcastw	%xmm0, %xmm0
	retq
.Lfunc_end3:
	.size	test_u16___UM_vyS, .Lfunc_end3-test_u16___UM_vyS
                                        # -- End function
	.globl	test_s32___UM_vys               # -- Begin function test_s32___UM_vys
	.p2align	4
	.type	test_s32___UM_vys,@function
test_s32___UM_vys:                      # @test_s32___UM_vys
# %bb.0:
	vpshufd	$238, %xmm0, %xmm1              # xmm1 = xmm0[2,3,2,3]
	vpaddw	%xmm0, %xmm1, %xmm0
	vpshufd	$85, %xmm0, %xmm1               # xmm1 = xmm0[1,1,1,1]
	vpaddw	%xmm0, %xmm1, %xmm0
	vmovd	%xmm0, %eax
	vpextrw	$1, %xmm0, %ecx
	addl	%eax, %ecx
	movswl	%cx, %eax
	vmovd	%eax, %xmm0
	vpbroadcastd	%xmm0, %ymm0
	retq
.Lfunc_end4:
	.size	test_s32___UM_vys, .Lfunc_end4-test_s32___UM_vys
                                        # -- End function
	.globl	test_u32___UM_vyS               # -- Begin function test_u32___UM_vyS
	.p2align	4
	.type	test_u32___UM_vyS,@function
test_u32___UM_vyS:                      # @test_u32___UM_vyS
# %bb.0:
	vpshufd	$238, %xmm0, %xmm1              # xmm1 = xmm0[2,3,2,3]
	vpaddw	%xmm0, %xmm1, %xmm0
	vpshufd	$85, %xmm0, %xmm1               # xmm1 = xmm0[1,1,1,1]
	vpaddw	%xmm0, %xmm1, %xmm0
	vmovd	%xmm0, %eax
	vpextrw	$1, %xmm0, %ecx
	addl	%eax, %ecx
	movzwl	%cx, %eax
	vmovd	%eax, %xmm0
	vpbroadcastd	%xmm0, %ymm0
	retq
.Lfunc_end5:
	.size	test_u32___UM_vyS, .Lfunc_end5-test_u32___UM_vyS
                                        # -- End function
	.globl	test_s64___UM_vyI               # -- Begin function test_s64___UM_vyI
	.p2align	4
	.type	test_s64___UM_vyI,@function
test_s64___UM_vyI:                      # @test_s64___UM_vyI
# %bb.0:
	vpaddq	%ymm0, %ymm1, %ymm0
	vextracti128	$1, %ymm0, %xmm1
	vpaddq	%xmm0, %xmm1, %xmm0
	vpextrq	$1, %xmm0, %rax
	vmovq	%xmm0, %rcx
	addq	%rcx, %rax
	vmovq	%rax, %xmm0
	vpbroadcastq	%xmm0, %ymm0
	vmovdqa	%ymm0, %ymm1
	retq
.Lfunc_end6:
	.size	test_s64___UM_vyI, .Lfunc_end6-test_s64___UM_vyI
                                        # -- End function
	.globl	test_u64___UM_vyU               # -- Begin function test_u64___UM_vyU
	.p2align	4
	.type	test_u64___UM_vyU,@function
test_u64___UM_vyU:                      # @test_u64___UM_vyU
# %bb.0:
	vpaddq	%ymm0, %ymm1, %ymm0
	vextracti128	$1, %ymm0, %xmm1
	vpaddq	%xmm0, %xmm1, %xmm0
	vpextrq	$1, %xmm0, %rax
	vmovq	%xmm0, %rcx
	addq	%rcx, %rax
	vmovq	%rax, %xmm0
	vpbroadcastq	%xmm0, %ymm0
	vmovdqa	%ymm0, %ymm1
	retq
.Lfunc_end7:
	.size	test_u64___UM_vyU, .Lfunc_end7-test_u64___UM_vyU
                                        # -- End function
	.ident	"Intel(r) Implicit SPMD Program Compiler (Intel(r) ISPC), 1.31.0dev (build commit ef22cd37b171dace @ 20260526, LLVM 21.1.8)"
	.ident	"LLVM version 21.1.8 (https://github.com/llvm/llvm-project.git 2078da43e25a4623cab2d0d60decddf709aaea28)"
	.section	".note.GNU-stack","",@progbits
```

```asm
	.file	"test.ispc"
	.text
	.globl	test_s8___UM_vyt                # -- Begin function test_s8___UM_vyt
	.p2align	4
	.type	test_s8___UM_vyt,@function
test_s8___UM_vyt:                       # @test_s8___UM_vyt
# %bb.0:
	vpshufd	$85, %xmm0, %xmm1               # xmm1 = xmm0[1,1,1,1]
	vpaddb	%xmm1, %xmm0, %xmm0
	vpsrld	$16, %xmm0, %xmm1
	vpaddb	%xmm1, %xmm0, %xmm0
	vpsrlw	$8, %xmm0, %xmm1
	vpaddb	%xmm1, %xmm0, %xmm0
	vpbroadcastb	%xmm0, %xmm0
	retq
.Lfunc_end0:
	.size	test_s8___UM_vyt, .Lfunc_end0-test_s8___UM_vyt
                                        # -- End function
	.globl	test_u8___UM_vyT                # -- Begin function test_u8___UM_vyT
	.p2align	4
	.type	test_u8___UM_vyT,@function
test_u8___UM_vyT:                       # @test_u8___UM_vyT
# %bb.0:
	vpshufd	$85, %xmm0, %xmm1               # xmm1 = xmm0[1,1,1,1]
	vpaddb	%xmm1, %xmm0, %xmm0
	vpsrld	$16, %xmm0, %xmm1
	vpaddb	%xmm1, %xmm0, %xmm0
	vpsrlw	$8, %xmm0, %xmm1
	vpaddb	%xmm1, %xmm0, %xmm0
	vpbroadcastb	%xmm0, %xmm0
	retq
.Lfunc_end1:
	.size	test_u8___UM_vyT, .Lfunc_end1-test_u8___UM_vyT
                                        # -- End function
	.globl	test_s16___UM_vys               # -- Begin function test_s16___UM_vys
	.p2align	4
	.type	test_s16___UM_vys,@function
test_s16___UM_vys:                      # @test_s16___UM_vys
# %bb.0:
	vpshufd	$238, %xmm0, %xmm1              # xmm1 = xmm0[2,3,2,3]
	vpaddw	%xmm1, %xmm0, %xmm0
	vpshufd	$85, %xmm0, %xmm1               # xmm1 = xmm0[1,1,1,1]
	vpaddw	%xmm1, %xmm0, %xmm0
	vpsrld	$16, %xmm0, %xmm1
	vpaddw	%xmm1, %xmm0, %xmm0
	vpbroadcastw	%xmm0, %xmm0
	retq
.Lfunc_end2:
	.size	test_s16___UM_vys, .Lfunc_end2-test_s16___UM_vys
                                        # -- End function
	.globl	test_u16___UM_vyS               # -- Begin function test_u16___UM_vyS
	.p2align	4
	.type	test_u16___UM_vyS,@function
test_u16___UM_vyS:                      # @test_u16___UM_vyS
# %bb.0:
	vpshufd	$238, %xmm0, %xmm1              # xmm1 = xmm0[2,3,2,3]
	vpaddw	%xmm1, %xmm0, %xmm0
	vpshufd	$85, %xmm0, %xmm1               # xmm1 = xmm0[1,1,1,1]
	vpaddw	%xmm1, %xmm0, %xmm0
	vpsrld	$16, %xmm0, %xmm1
	vpaddw	%xmm1, %xmm0, %xmm0
	vpbroadcastw	%xmm0, %xmm0
	retq
.Lfunc_end3:
	.size	test_u16___UM_vyS, .Lfunc_end3-test_u16___UM_vyS
                                        # -- End function
	.globl	test_s32___UM_vys               # -- Begin function test_s32___UM_vys
	.p2align	4
	.type	test_s32___UM_vys,@function
test_s32___UM_vys:                      # @test_s32___UM_vys
# %bb.0:
	vpshufd	$238, %xmm0, %xmm1              # xmm1 = xmm0[2,3,2,3]
	vpaddw	%xmm1, %xmm0, %xmm0
	vpshufd	$85, %xmm0, %xmm1               # xmm1 = xmm0[1,1,1,1]
	vpaddw	%xmm1, %xmm0, %xmm0
	vpsrld	$16, %xmm0, %xmm1
	vpaddw	%xmm1, %xmm0, %xmm0
	vmovd	%xmm0, %eax
	cwtl
	vmovd	%eax, %xmm0
	vpbroadcastd	%xmm0, %ymm0
	retq
.Lfunc_end4:
	.size	test_s32___UM_vys, .Lfunc_end4-test_s32___UM_vys
                                        # -- End function
	.globl	test_u32___UM_vyS               # -- Begin function test_u32___UM_vyS
	.p2align	4
	.type	test_u32___UM_vyS,@function
test_u32___UM_vyS:                      # @test_u32___UM_vyS
# %bb.0:
	vpshufd	$238, %xmm0, %xmm1              # xmm1 = xmm0[2,3,2,3]
	vpaddw	%xmm1, %xmm0, %xmm0
	vpshufd	$85, %xmm0, %xmm1               # xmm1 = xmm0[1,1,1,1]
	vpaddw	%xmm1, %xmm0, %xmm0
	vpsrld	$16, %xmm0, %xmm1
	vpaddw	%xmm1, %xmm0, %xmm0
	vpextrw	$0, %xmm0, %eax
	vmovd	%eax, %xmm0
	vpbroadcastd	%xmm0, %ymm0
	retq
.Lfunc_end5:
	.size	test_u32___UM_vyS, .Lfunc_end5-test_u32___UM_vyS
                                        # -- End function
	.globl	test_s64___UM_vyI               # -- Begin function test_s64___UM_vyI
	.p2align	4
	.type	test_s64___UM_vyI,@function
test_s64___UM_vyI:                      # @test_s64___UM_vyI
# %bb.0:
	vpaddq	%ymm1, %ymm0, %ymm0
	vextracti128	$1, %ymm0, %xmm1
	vpaddq	%xmm1, %xmm0, %xmm0
	vpshufd	$238, %xmm0, %xmm1              # xmm1 = xmm0[2,3,2,3]
	vpaddq	%xmm1, %xmm0, %xmm0
	vpbroadcastq	%xmm0, %ymm0
	vmovdqa	%ymm0, %ymm1
	retq
.Lfunc_end6:
	.size	test_s64___UM_vyI, .Lfunc_end6-test_s64___UM_vyI
                                        # -- End function
	.globl	test_u64___UM_vyU               # -- Begin function test_u64___UM_vyU
	.p2align	4
	.type	test_u64___UM_vyU,@function
test_u64___UM_vyU:                      # @test_u64___UM_vyU
# %bb.0:
	vpaddq	%ymm1, %ymm0, %ymm0
	vextracti128	$1, %ymm0, %xmm1
	vpaddq	%xmm1, %xmm0, %xmm0
	vpshufd	$238, %xmm0, %xmm1              # xmm1 = xmm0[2,3,2,3]
	vpaddq	%xmm1, %xmm0, %xmm0
	vpbroadcastq	%xmm0, %ymm0
	vmovdqa	%ymm0, %ymm1
	retq
.Lfunc_end7:
	.size	test_u64___UM_vyU, .Lfunc_end7-test_u64___UM_vyU
                                        # -- End function
	.ident	"Intel(r) Implicit SPMD Program Compiler (Intel(r) ISPC), 1.31.0dev (build commit ef22cd37b171dace @ 20260526, LLVM 21.1.8)"
	.ident	"LLVM version 21.1.8 (https://github.com/llvm/llvm-project.git 2078da43e25a4623cab2d0d60decddf709aaea28)"
	.section	".note.GNU-stack","",@progbits
```

We can see:
- `vpsadbw` is not that good (without widened types) because it needs some kind of preparation, but it should be still faster with it on int8=>int8 reduction.
- The end of the reduction (using instructions like `vmovd`, `addl`, etc.) seems less efficient here when using `reduceN` than generic LLVM intrinsics, but IDK if this is still true in real-world codes where a scalar register is needed to do some computations (instead of keeping everything in SIMD registers). IDK if masking can have some impact on the matter. IDK if this is also true on SSEx targets.

I also found that LLVM can produce a `vpsadbw` instruction in the following cases involving overflow:
```ispc
unmasked int8 new_test(int8 v) {
	// vpxor	%xmm1, %xmm1, %xmm1
	// vpsadbw	%xmm1, %xmm0, %xmm0
	// vmovd	%xmm0, %eax
	// movl	%eax, %ecx
	// shll	$4, %ecx
	// addl	%eax, %ecx
	// vmovd	%ecx, %xmm0
	// vpbroadcastb	%xmm0, %xmm0
    return reduce_add(v) * (int8)17;
}
```
It is somehow "smarter" than I though first. It looks like generic code can benefit from optimizations while non-generic code don't (or at least not as much). I testested few other cases and it looks like the generic implementation is generally surprisingly more efficient for signed values on avx2 x8 targets. There are cases where the generic implementation is as fast but I did not found cases where it looks much smaller during a quick investigation (on avx2 targets).

So put it shortly, generic implementations might be better (faster and simpler) in the end though there are certainly cases where it might cause some small regressions!

aneshlya:
Thanks for your analysis! So it seems that generic target is good enough and we can remove `__reduce_add_int8/int16` and helpers from all target .ll files. Keep only `__reduce_add_uint8` with `psadbw`.
This would significantly simplify the PR.

Also please add regression lit test checking that `psadbw` is produced for `__reduce_add_uint8` and is not produced for `__reduce_add_int8`.

zephyr111:
Ok. Should we also remove this from the WASM target? I cannot test it easily on my machine. That being said, I expect LLVM to produce relatively good assembly for this target similarly to others and it is certainly good for maintainability too.

Btw, I did few check on AVX-512 and I did not found any significantly bigger code using generic instructions too. I found that the 32-bit reduce_add of sse2/sse4 is just a bit bigger with the generic code (1 more instruction), but in practice UICA reports that both are about equally fast overall (on Sandy Bridge, Skylake and Icelake) so this is eventually not an issue.

aneshlya:
For WASM target generic implementation is enough.
Regarding AVX 512  - if generic target produces good enough code for all types/widths, we can use generic target there as well.

elena901:
>Change the output type so it is now the same than the input type (no widening)

A little bit late with it (again), but i think we need to fix functions with widening output first to not break source compatibility one more time.

>M [builtins/generic.ispc](https://github.com/ispc/ispc/pull/3846/files#diff-4fef5ac8cd7204f7e2942c83580efb5d0e9fd79f237cae7723062012b4f96924) (123)
M [builtins/target-avx1-i32x16.ll](https://github.com/ispc/ispc/pull/3846/files#diff-7703cc3a492c3ef26365618b3dbad8ae4a58cfbd77531d15ea1b2a5b23e9d12b) (47)
M [builtins/target-avx1-i32x8.ll](https://github.com/ispc/ispc/pull/3846/files#diff-a50f5c342c48e8afe2cc23feb4d85b8e00a99ac2ff7c286035f43994a047ae29) (38)
M [builtins/target-avx1-i64x4.ll](https://github.com/ispc/ispc/pull/3846/files#diff-16618bffcdfa13c58fc2b7d30fc9efe50ec0519091ac8e299abc23213cad8da9) (51)
M [builtins/target-avx2-i16x16.ll](https://github.com/ispc/ispc/pull/3846/files#diff-993f722571f4d95efbad91987081ce921bfbbe72e238cf91bc9ea40511376ff4) (36)
M [builtins/target-avx2-i8x32.ll](https://github.com/ispc/ispc/pull/3846/files#diff-168470856d62e8a28b584b4950dbf025b42fc84b7b544188b789f25ed6314694) (38)
M [builtins/target-avx512skx-x16-nozmm.ll](https://github.com/ispc/ispc/pull/3846/files#diff-4268a700835f50a9631894edde8ed8f78d973d088ef1643ab8e59ee4827d2094) (36)
M [builtins/target-avx512skx-x16.ll](https://github.com/ispc/ispc/pull/3846/files#diff-463b3bd282b35b50fb81257190898b6a4ffc444268030d4a89e61bfd74f4177a) (36)
M [builtins/target-avx512skx-x32.ll](https://github.com/ispc/ispc/pull/3846/files#diff-76742acb767b4424b1ec9a16cbc37c436ae654afdbe2f9e24411f52d8e6fe61a) (36)
M [builtins/target-avx512skx-x4.ll](https://github.com/ispc/ispc/pull/3846/files#diff-0f0058a3b1a749363cef2fd865d0315e6c486f180f70c98bfeb19d6da54a31d8) (43)
M [builtins/target-avx512skx-x64.ll](https://github.com/ispc/ispc/pull/3846/files#diff-17eca34eca82f3b8c177794a5e36f7a053eedaf533b0f221ae7450168ef502bb) (39)
M [builtins/target-avx512skx-x8.ll](https://github.com/ispc/ispc/pull/3846/files#diff-a0de5d39893b2e28337702170c24434887f3b7189ad9b69e96d51076b64af35b) (36)
M [builtins/target-sse2-i32x4.ll](https://github.com/ispc/ispc/pull/3846/files#diff-d0ae01324bcb47e0304fd966751c67a129d759b79e779fec1d38a5da2e8e8509) (45)
M [builtins/target-sse2-i32x8.ll](https://github.com/ispc/ispc/pull/3846/files#diff-838120aac6cefa1641d02eb60d0ada9f7014a8757d8b65cee4551dda50bff18c) (36)
M [builtins/target-sse4-i16x8.ll](https://github.com/ispc/ispc/pull/3846/files#diff-d1ca5dc4619634c5900056d23e143780fad6872eaeabd765321471217f61f558) (36)
M [builtins/target-sse4-i32x4.ll](https://github.com/ispc/ispc/pull/3846/files#diff-3f5dc9b0401df62fe6415ff66bd1c8ba12371bc35a89d4c52494dca9f978a935) (45)
M [builtins/target-sse4-i32x8.ll](https://github.com/ispc/ispc/pull/3846/files#diff-4b8e429f1031c5506018bbf6e9bb2ec5cdce7a4f1f95cc57f7ddd02d8f71f341) (36)
M [builtins/target-sse4-i8x16.ll](https://github.com/ispc/ispc/pull/3846/files#diff-b1084712675c01ab33921f05fbe163092c9562d58579d634feeae71ba9e45891) (36)
M [builtins/target-wasm-i32x4.ll](https://github.com/ispc/ispc/pull/3846/files#diff-1bacb7641d8d439bbcac61c7aaa87a66656b162c0e56a0d5c8dd346acc1fc16b) (47)
M [builtins/target-xe.ll](https://github.com/ispc/ispc/pull/3846/files#diff-e1a095f74bf1dc89b511fb28c050e2faf68440c0945eefa0da20c6c25bd67d85) (38)

RISC-V/SVE/SVE2(?)/~Gen9~ ? I can handle it anyway (a bit later).

zephyr111:
@elena901 

> A little bit late with it (again), but i think we need to fix functions with widening output first to not break source compatibility one more time

As mentioned before, only fixing while letting wider types would make the code slower, mainly on x86/x86-64 which is the most widely used platform (and mainly the one where it was not actually supported under the hood). Similarly, I expect using smaller types will certainly make reduction (at least a bit) faster on most other platform. I understand that changing the type breaks compatibility is clearly not great, but I think this is probably the best time to do it. Later, there will be less reasons to change this. On top of that, currently, no users apparently reported the missing/broken widened operations on x86 so I expect people not to actually use much this feature. Personally, I never noticed so far the output types were wider. Ideally, it would be great to detect cases were it is dangerous but I do not see any simple (and reliable) ways to do that so far. Maybe we can make the transition smoother somehow?

> RISC-V/SVE/SVE2(?)/~~Gen9~~ ? I can handle it anyway (a bit later).

SVE and SVE2 are not yet supported in ISPC. I am not sure RISC-V is very well supported yet. The very large majority (>95%) of applications are running on either x86-64 CPU or ARM these days (and x86-64 lead so far on the desktop/server market -- though ARM lead on the mobile/low-power market). Thus, I think the x86-86 and ARM support matter more so far in ISPC than other target -- with a main focus on x86-64 (not to mention ISPC is an Intel software in the first place).

I am going to make this more clear that this can break compatibility in the doc. Maybe we could also increase the major version number so to be more clear about this too (especially if there is other similar PR).

PS: there is also another argument: when ISPC will support 128-bit integers, the 64-bit reduction having widened return types will become severely slower (all certainly all targets). Meanwhile, very few people actually need/want 64-bit to 128-bit reductions. 64-bit to 64-bit reduction will be unaffected.

aneshlya:
@zephyr111 ,

Yes, please add a clear note to the "updating the programs" section in the documentation explaining that `reduce_add` for 8- and 16-bit types now returns the same type as the input rather than a widened type, and that any code relying on the widened return type will need to be updated.

On the versioning side, we don't plan to bump the major version for this change - a clearly documented breaking-change note in the release notes, combined with the usual minor version increment, is sufficient to communicate this to users. The change is scoped and well-motivated (correctness + performance), and we've seen no evidence of real-world code depending on the old widened return types.

zephyr111:
@aneshlya Ok. I made the modification of all targets except XE one. I expect it not to use generic functions as fallback one. Is this true? If no, it might be better not to use generic functions on XE anyway, isn't it? I also updated the doc regarding the update / breaking change. Is this OK now?

> Also please add regression lit test checking that psadbw is produced for __reduce_add_uint8 and is not produced for __reduce_add_int8.

~~There is a corner case: the `psadbw` is not generated for `__reduce_add_uint8` on target having a gand size of only 4 items. Is there an easy way to test all targets while discarding such target (while avoiding to list all x86 targets since new ones can be added later)?~~ (see below)

-----

**EDIT:** I was confused because I found that the generic LLVM code generated the `vpsadbw` for signed integers in my (local) lit-test although func-tests where passing. It turns out that this seems OK for 8-bit integers as long as there is no widening. Moreover, here again, the generic code looks quite good for signed functions. The code generated for the current custom unsigned builtin seems less efficient in some cases. Thus, I think we can even use the generic code for unsigned int now. I am not sure this is always the case. I need to check the code of all target to be sure.

Here are results for some target (`test_uint8___UM_vyT` is the custom intrinsic for `uint8` while `test_int8___UM_vyt` is now the generic one for `int8` -- both work on both uint8 and int8 in practice). Btw, please note I currently use LLVM 21.

```asm
     ; target avx2-i8x32

test_uint8___UM_vyT:
	vpxor	xmm1, xmm1, xmm1
	vpsadbw	ymm0, ymm0, ymm1
	vpextrq	rcx, xmm0, 1
	vmovq	rdx, xmm0
	vextracti128	xmm0, ymm0, 1
	vpextrq	rsi, xmm0, 1
	vmovq	rax, xmm0
	add	edx, ecx
	add	eax, edx
	add	eax, esi
                                        # kill: def $al killed $al killed $rax
	vzeroupper
	ret

test_int8___UM_vyt:
	vextracti128	xmm1, ymm0, 1
	vpaddb	xmm0, xmm0, xmm1
	vpshufd	xmm1, xmm0, 238                 # xmm1 = xmm0[2,3,2,3]
	vpaddb	xmm0, xmm0, xmm1
	vpxor	xmm1, xmm1, xmm1
	vpsadbw	xmm0, xmm0, xmm1
	vmovd	eax, xmm0
                                        # kill: def $al killed $al killed $eax
	vzeroupper
	ret


     ; target avx2-i32x8

test_uint8___UM_vyT:
	vmovq	xmm0, xmm0                      # xmm0 = xmm0[0],zero
	vpxor	xmm1, xmm1, xmm1
	vpsadbw	xmm0, xmm0, xmm1
	vmovq	rax, xmm0
	vpextrq	rcx, xmm0, 1
	or	eax, ecx
                                        # kill: def $al killed $al killed $rax
	ret

test_int8___UM_vyt:
	vpxor	xmm1, xmm1, xmm1
	vpsadbw	xmm0, xmm0, xmm1
	vmovd	eax, xmm0
                                        # kill: def $al killed $al killed $eax
	ret


     ; target avx2-i32x16

test_uint8___UM_vyT:
	vpxor	xmm1, xmm1, xmm1
	vpsadbw	xmm0, xmm0, xmm1
	vmovq	rax, xmm0
	vpextrq	rcx, xmm0, 1
	add	eax, ecx
                                        # kill: def $al killed $al killed $rax
	ret

test_int8___UM_vyt:
	vpshufd	xmm1, xmm0, 238                 # xmm1 = xmm0[2,3,2,3]
	vpaddb	xmm0, xmm0, xmm1
	vpxor	xmm1, xmm1, xmm1
	vpsadbw	xmm0, xmm0, xmm1
	vmovd	eax, xmm0
                                        # kill: def $al killed $al killed $eax
	ret


     ; target sse4-i8x16

test_uint8___UM_vyT:
	pxor	xmm1, xmm1
	psadbw	xmm1, xmm0
	movq	rax, xmm1
	pextrq	rcx, xmm1, 1
	add	eax, ecx
                                        # kill: def $al killed $al killed $rax
	ret

test_int8___UM_vyt:
# %bb.0:
	pshufd	xmm1, xmm0, 238                 # xmm1 = xmm0[2,3,2,3]
	paddb	xmm1, xmm0
	pxor	xmm0, xmm0
	psadbw	xmm0, xmm1
	movd	eax, xmm0
                                        # kill: def $al killed $al killed $eax
	ret


     ; target avx512spr-x64

test_uint8___UM_vyT:
	vpxor	xmm1, xmm1, xmm1
	vpsadbw	zmm0, zmm0, zmm1
	vmovq	rcx, xmm0
	vpextrq	rdx, xmm0, 1
	vextracti128	xmm1, ymm0, 1
	vmovq	rsi, xmm1
	vpextrq	rdi, xmm1, 1
	vextracti32x4	xmm1, zmm0, 2
	vpextrq	r8, xmm1, 1
	vmovq	r9, xmm1
	vextracti32x4	xmm0, zmm0, 3
	vmovq	rax, xmm0
	vpextrq	r10, xmm0, 1
	add	ecx, edx
	add	esi, ecx
	add	esi, edi
	add	r9d, r8d
	add	r9d, esi
	add	eax, r10d
	add	eax, r9d
                                        # kill: def $al killed $al killed $rax
	vzeroupper
	ret

test_int8___UM_vyt:
	vextracti64x4	ymm1, zmm0, 1
	vpaddb	ymm0, ymm0, ymm1
	vextracti128	xmm1, ymm0, 1
	vpaddb	xmm0, xmm0, xmm1
	vpshufd	xmm1, xmm0, 238                 # xmm1 = xmm0[2,3,2,3]
	vpaddb	xmm0, xmm0, xmm1
	vpxor	xmm1, xmm1, xmm1
	vpsadbw	xmm0, xmm0, xmm1
	vmovd	eax, xmm0
                                        # kill: def $al killed $al killed $eax
	vzeroupper
	ret
```

As you can see, the generic code is often actually better, especially on `avx512spr-x64`. Thus, I propose to use the generic implementation on all x86 CPU target for `reduce_add` as long as is no widening.

docs/ispc.rst
```diff
@@ -268,12 +268,22 @@ New Architecture Support:
...
   versions. It is still available and left unchanged. The fast-math mode can be
   specified with the option ``--opt=fast-math:<mode>``. See the optimization
   settings section for more information about the new modes.
+
+Standard Library Changes:
+
+* **Breaking change**: the return type of the ``reduce_add()`` functions is now
+  the same than the input type. Users should now cast the inpout to a wider type
+  themselves so to prevent overflows. Any code relying on the widened return
+  type needs to be updated!
```
aneshlya: Let's explain why this change was made: 
```suggestion
* **Breaking change**: ``reduce_add()`` now returns the same type as its
    input. Previously the signatures widened the result, but the
    implementation was inconsistent and partially incorrect. The new
    same-type convention also matches the prevailing convention in other
    SIMD/SPMD ecosystems and is consistent with ``reduce_min()`` /
    ``reduce_max()`` in ISPC. For overflow-safe accumulation, cast the
    input to a wider type explicitly, e.g. ``reduce_add((int32)x)`` for an
    ``int16`` ``x``. Any code that relied on the previous widened return
    type must be updated.

```

aneshlya:
Thanks for the detailed asm dumps. I agree: let's drop the custom `__reduce_add_uint8` from all x86 targets too and rely on `@llvm.vector.reduce.add` everywhere except XE (it doesn't fall back to generic). 

Also please add regression lit test checking that `psadbw` is produced for `__reduce_add_uint8` and is not produced for `__reduce_add_int8`.

elena901:
There is a [research](https://github.com/ispc/ispc/issues/3834) for now.(Just for the reminder).

nurmukhametov:
I haven't reviewed the PR, but replacing LLVM IR implementations with generic ones is a good move.

aneshlya:
@zephyr111, this PR is almost ready for merging. Would you be able to finish it before 06/22? I'd love to include it into upcoming release.

zephyr111:
@aneshlya Ok. It should be fine for a merge now: I dropped all the custom `__reduce_add_uint8` so there is only the generic+Xe ones now, committed your requested change on the doc, and added a test for the presence of the fast instruction in the case of `__reduce_add_uint8`. The signed case is not tested since the instruction could be used too (`(v)psadbw` can be used on both unsigned and signed 8-bit integer as long as the result is not widened -- otherwise, it only works for unsigned ones). In fact, the above asm code shows LLVM generates it for the signed case too. I also rebased the branch on `main`.

aneshlya (approved):
LGTM! Thanks a lot for your work!

PR #3866, opened 2026-06-18
Fix #3863: Correct _mm_shuffle_pd immediate decoding and add tests

### LINKED ISSUE #3863: SSE2 _mm_shuffle_pd uses incorrect immediate decoding and lacks LIT/runtime coverage
## Problem

While reviewing constexpr-related stdlib changes, we found that the ISPC implementation of the SSE2 intrinsic `_mm_shuffle_pd()` appears to decode the immediate operand like `_mm_shuffle_ps()`.

The current/old logic was effectively:

```ispc
_R[0] = _A.m128d_f64[_I & 0x3];
_R[1] = _B.m128d_f64[(_I >> 2) & 0x3];
```

That treats the immediate as two 2-bit selectors. For `_mm_shuffle_pd`, the mask is only two bits total, normally produced by `_MM_SHUFFLE2(x, y) == ((x) << 1) | (y)`. Each result lane uses one selector bit:

```ispc
_R[0] = _A.m128d_f64[_I & 0x1];
_R[1] = _B.m128d_f64[(_I >> 1) & 0x1];
```

Concrete example:

```ispc
uniform __m128d a = _mm_setr_pd(1.0, 2.0);
uniform __m128d b = _mm_setr_pd(3.0, 4.0);
uniform __m128d r = _mm_shuffle_pd(a, b, _MM_SHUFFLE2(1, 0));
```

Expected result is `{1.0, 4.0}`. With the old 2-bit-field decoding, `_MM_SHUFFLE2(1, 0)` is `2`, so the first lane indexes `_A[2]`, which is out of range for `double<2>`.

## Test coverage gap

We also found that this was not covered by correctness tests:

- No existing LIT test appears to verify `_mm_shuffle_pd(a, b, _MM_SHUFFLE2(...))` semantics for the legacy 3-argument form.
- No runtime `tests/func-tests` coverage appears to exercise `intrinsics/emmintrin.isph` / `_mm_shuffle_pd` behavior.
- Existing benchmark coverage is not sufficient as correctness coverage, and the benchmark only uses `_mm_shuffle_pd(S, S, 0x1)`, which does not expose this bug.

## Suggested coverage

Add both LIT and runtime tests:

1. LIT: verify `_mm_shuffle_pd(a, b, _MM_SHUFFLE2(...))` and any constexpr/template overload variant produce the expected folded result for all four masks: `(0,0)`, `(0,1)`, `(1,0)`, `(1,1)`.
2. Runtime: add a `tests/func-tests` case that includes `intrinsics/emmintrin.isph`, runs `_mm_shuffle_pd` on known `{1,2}` and `{3,4}` inputs, and checks all four `_MM_SHUFFLE2` masks.

This should prevent regressions in both the existing compatibility form and any new constexpr-friendly wrapper forms.

aneshlya:
- Fixed _mm_shuffle_pd to use 1-bit selector fields instead of 2-bit fields
- The intrinsic now correctly interprets the immediate as two 1-bit selectors (one for each output lane) instead of treating it like _mm_shuffle_ps
- Added LIT test covering all four shuffle masks: (0,0), (0,1), (1,0), (1,1)
- Added functional test to verify runtime correctness
- Updated copyright year to 2026

## Related Issue
- [x] Fixes #3863 

## Checklist

### DIFF
--- a/stdlib/include/intrinsics/emmintrin.isph
+++ b/stdlib/include/intrinsics/emmintrin.isph
@@ -1,5 +1,5 @@
 /*
-  Copyright (c) 2010-2025, Intel Corporation
+  Copyright (c) 2010-2026, Intel Corporation
 
   SPDX-License-Identifier: BSD-3-Clause
 */
@@ -569,8 +569,8 @@ static inline unmasked uniform int _mm_movemask_pd(uniform __m128d _A) {
 
 static inline unmasked uniform __m128d _mm_shuffle_pd(uniform __m128d _A, uniform __m128d _B, uniform unsigned int _I) {
     uniform double<2> _R;
-    _R[0] = _A.m128d_f64[_I & 0x3];
-    _R[1] = _B.m128d_f64[(_I >> 2) & 0x3];
+    _R[0] = _A.m128d_f64[_I & 0x1];
+    _R[1] = _B.m128d_f64[(_I >> 1) & 0x1];
     return *((uniform __m128d * uniform) & _R);
 }
 
--- a/tests/func-tests/3863.ispc
+++ b/tests/func-tests/3863.ispc
@@ -1,0 +1,58 @@
+// Copyright (c) 2026, Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+
+// Test for _mm_shuffle_pd SSE2 intrinsic correctness (#3863)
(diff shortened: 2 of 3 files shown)


dbabokin (approved):
Looks good!

PR #3878, opened 2026-07-01
Fix missing reference to roundeven + fix bad round call on double-precision

zephyr111:
Everything is in the name. This PR fix the issue detected #3818 (in tests filing on Windows).

I did not test this thoroughly yet. Please note that the IR code has been modified manually and has not been automatically  generated (it matter for the review).
As discussed in #3818, this probably impact performance of rounding operations on targets using generic targets like RISC-V and POWER.

## Related Issue

## Checklist

### DIFF
--- a/builtins/generic.ispc
+++ b/builtins/generic.ispc
@@ -447,10 +447,66 @@ EXT READNONE uniform double __floor_uniform_double(uniform double x) { return @l
 EXT READNONE uniform float __floor_uniform_float(uniform float x) { return @llvm.floor(x); }
 EXT READNONE varying double __floor_varying_double(varying double x) { return @llvm.floor(x); }
 EXT READNONE varying float __floor_varying_float(varying float x) { return @llvm.floor(x); }
-EXT READNONE uniform double __round_uniform_double(uniform double x) { return @llvm.roundeven(x); }
-EXT READNONE uniform float __round_uniform_float(uniform float x) { return @llvm.roundeven(x); }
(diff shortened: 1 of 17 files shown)


azwolski:
It is needed to fix failing tests

zephyr111:
I did not expect the AVX-512 target to fallback on the generic implementation for ``round``. It is now fixed.
I also fixed the neon implementation for the same reason (fallback on the generic implementation).

Technically, PPC should be fixed too for sake of performance, but I do not have a PPC machine so to test it.
We could assume that this is OK if tests running on PPC pass, but one test fails on the matter and I think it might hide something more critical. Please let me know if you want me to apply the same fix than on Neon or AVX-512.
I expect the problem to happen on all generic code, but to be detected only on PPC so far.
That being said, ``stdlib-short-vec-funcs-3.ispc`` already reports an issue with froundeven. 
I do not understand why the test fail yet... I would be glad if you could shed some light on the matter.

Please note that I re-enabled the generic tests as requested in #3818 (and I added few more tests so to be safer). It is not clear to me why the ``round-float-varying.ispc`` test has been disabled on ``ppc64le`` though... 

I also fixed double-precision issues related to ``round``.

I think the issue mentioned in ``round-float16-varying.ispc`` can be fixed in this PR. Indeed, there is no reason not to emulate half-precision ``round`` instructions since both single-precision and double-precision ``round`` instructions are now emulated on generic targets. I can add this later if you wish (or possibly in another PR if you think it make more sense). What do you think about it?

dbabokin:
@codex review

PR #3883, opened 2026-07-07
Fix crashing gather index

### LINKED ISSUE #3881: Assertion failed (llvmutil.cpp:893): "baseValue > 0 && (baseValue & (baseValue - 1)) == 0" — gather index derived from 64-bit hash math (1.30 & 1.31, all targets)
Compiling a kernel where a gather index is derived from 64-bit hash math (splitmix64-style mixer followed by a fixed-point range mapping) crashes the compiler:

```
llvmutil.cpp:893: Assertion failed: "baseValue > 0 && (baseValue & (baseValue - 1)) == 0".
***
...
main.cpp(40): FATAL ERROR: Unhandled signal sent to process; terminating.
```

## Minimal reproducer (10 lines)

```c
export void repro(uniform int n, uniform int colorCount,
                  const uniform int* uniform lut, uniform int* uniform result) {
    foreach (index = 0 ... n) {
        unsigned int64 z = (unsigned int64)index * 0xD1B54A32D192ED03;
        z = (z ^ (z >> 30)) * 0xBF58476D1CE4E5B9;
        z = z ^ (z >> 31);
        unsigned int32 r = (unsigned int32)(z >> 32);
        int lutIdx = (int)(((unsigned int64)r * (unsigned int32)colorCount) >> 32);
        result[index] = lut[lutIdx];
    }
}
```

```
ispc -O3 --target=avx2-i32x16 --arch=x86_64 repro.ispc -o repro.o
```

## Scope

Reproduces on:

- **Versions**: 1.30.0 (LLVM 21.1.8) and 1.31.0 (LLVM 23.0.0), official release binaries
- **Targets**: sse2-i32x8, sse4.2-i32x8, avx1-i32x16, avx2-i32x16, avx512skx-x16 (x86_64) and neon-i32x8 (aarch64)
- **Target OS**: `--target-os=macos` and `--target-os=linux`
- **Opt levels**: `-O1`, `-O2`, `-O3` crash; `-O0` compiles OK
- Host: macOS 26 arm64 (universal binary); the assert is in target-independent gather-index analysis so the host likely doesn't matter

## Sensitivity of the trigger

The crash needs the full chain — each of these one-line changes makes it compile cleanly:

- removing the final `z = z ^ (z >> 31);`
- removing the xorshift-multiply round (keeping only the initial multiply)
- replacing the fixed-point mapping `(int)(((unsigned int64)r * (unsigned int32)colorCount) >> 32)` with a plain shift of the hash (e.g. `(int)(z >> 60)`)

The gather element type is irrelevant (`int`, `uint8<4>` both crash).

## Workaround

Inserting a semantically no-op clamp between the hash math and the gather avoids the crash:

```c
        lutIdx = min(lutIdx, colorCount - 1);   // no-op: (r * n) >> 32 < n
        result[index] = lut[lutIdx];
```

## Context

Hit while porting per-pixel effects in [xLights](https://github.com/xLightsSequencer/xLights) to ISPC: a stateless hash RNG (for reproducible random effects) picks a palette index per pixel, which then gathers from a color LUT — a fairly natural pattern for anything using hash-based procedural randomness with table lookups.

azwolski:
## Description
When a gather index came from 64-bit hash math, the compiler crashed. The problem was in lVectorShiftRightAllEqual: it built the shift divisor as a 32-bit int, so a shift by 31 or 32 overflowed and turned negative, which tripped an assertion. The divisor is now int64_t, with a check on the shift amount. There's also a new guard in lAllDivBaseEqual for large baseValue values that weren't possible before. Both places just give up when unsure, so the worst case is that a gather stays a gather, the result is never wrong. Added a regression test with the 64-bit and 32-bit cases that used to crash.

Fixes #3881 

## Related Issue

## Checklist

### DIFF
--- a/src/llvmutil.cpp
+++ b/src/llvmutil.cpp
@@ -11,6 +11,7 @@
 #include "llvmutil.h"
 #include "type.h"
 
+#include <climits>
 #include <cstdlib>
 #include <map>
 #include <set>
@@ -995,6 +996,14 @@ static bool lAllDivBaseEqual(llvm::Value *val, int64_t baseValue, int vectorLeng
         // the addConstants[], mod baseValue.  If we round that up to the
         // next power of 2, we'll have a value that will be no greater than
         // baseValue and sometimes less.
+        //
+        // The requiredAlignment computation below is done in int; if baseValue
+        // is large enough that (addConstants[i] % baseValue) could overflow
+        // int, the alignment could be corrupted and produce a false-positive
+        // equality, so conservatively give up.
+        if (baseValue > INT_MAX) {
+            return false;
+        }
         int maxMod = int(addConstants[0] % baseValue);
         for (int i = 1; i < vectorLength; ++i) {
             maxMod = std::max(maxMod, int(addConstants[i] % baseValue));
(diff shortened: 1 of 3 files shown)


dbabokin:
@codex review

aneshlya (approved):
LGTM!

aneshlya:
@azwolski , why a lit test failing on ppc target?

PR #3885, opened 2026-07-14
Fix #3882: Avoid UB in float-to-unsigned constant conversion causing …

### LINKED ISSUE #3882: Silent miscompile: `--` on a varying unsigned int (any width) becomes a no-op (step constant lowered to 0)
`--` (pre- or post-decrement) applied to a **varying unsigned** integer is silently miscompiled into a **no-op** — no diagnostic, wrong result at runtime. The front-end lowers the decrement's step constant to `0` instead of `-1`, so the expression becomes `v + 0`.

Increment (`++`) on the same types is fine, decrement on **signed** types is fine, and `v -= 1` / `v = v - 1` are fine — so this is easy to miss until a result is silently wrong.

## Reproducer

```c
// dec_repro.ispc — each function should subtract 1 from every element of buf.
export void u8_predec(uniform unsigned int8  buf[], uniform int n)  { foreach (i = 0 ... n) { unsigned int8  v = buf[i]; v--;    buf[i] = v; } }
export void u16_predec(uniform unsigned int16 buf[], uniform int n) { foreach (i = 0 ... n) { unsigned int16 v = buf[i]; v--;    buf[i] = v; } }
export void u32_predec(uniform unsigned int32 buf[], uniform int n) { foreach (i = 0 ... n) { unsigned int32 v = buf[i]; v--;    buf[i] = v; } }
export void s8_predec(uniform int8  buf[], uniform int n)           { foreach (i = 0 ... n) { int8  v = buf[i]; v--;             buf[i] = v; } }  // control: works
export void u8_minuseq(uniform unsigned int8  buf[], uniform int n) { foreach (i = 0 ... n) { unsigned int8  v = buf[i]; v -= 1; buf[i] = v; } }  // control: works
export void u8_inc(uniform unsigned int8  buf[], uniform int n)     { foreach (i = 0 ... n) { unsigned int8  v = buf[i]; v++;    buf[i] = v; } }  // control: works
```

Compiling and calling each with a buffer of `100` (u8) / `1000` (u16/u32):

```
unsigned int8   v--        100 -> 100   (want 99)   *** WRONG ***
unsigned int16  v--        1000 -> 1000 (want 999)  *** WRONG ***
unsigned int32  v--        1000 -> 1000 (want 999)  *** WRONG ***
signed   int8   v--        100 -> 99    (want 99)   ok   (control)
unsigned int8   v -= 1     100 -> 99    (want 99)   ok   (control)
unsigned int8   v++        100 -> 101   (want 101)  ok   (control)
```

## Root cause (from `--emit-llvm-text -O0`)

The decrement's step constant is emitted as `zeroinitializer` for the unsigned type. The temp is even named `%v_load_minus1`, showing the intended decrement:

| kernel | emitted `add <8 x i8>` step constant |
|---|---|
| `u8_dec` (unsigned `--`) | **`zeroinitializer`**  ← bug (`v + 0`) |
| `u8_inc` (unsigned `++`) | `splat (i8 1)`  ✓ |
| `s8_dec` (signed `--`)   | `splat (i8 -1)`  ✓ |

So `--` on an unsigned varying computes its constant as `0` rather than the unsigned representation of `-1`.

## Scope

- **Types**: varying `unsigned int8`, `unsigned int16`, `unsigned int32` all affected. Signed types unaffected.
- **Operators**: both `v--` and `--v` affected; `++` unaffected; `v -= 1` and `v = v - 1` unaffected.
- **Versions**: 1.30.0 (LLVM 21.1.8) and 1.31.0 (LLVM 23.0.0), official release binaries.
- **Targets / opt**: reproduces on `neon-i32x8`/`aarch64` and `avx2-i32x16`/`x86_64`, and at `-O0` through `-O2` — consistent with a front-end lowering bug, not an LLVM optimization.

## Note

Separately, `unsigned int64 v--` (the 64-bit case) instead trips a compiler assertion — `expr.cpp:1102: Assertion failed: "value == (int64_t)i"` — rather than the silent miscompile. Likely related front-end handling of the unsigned decrement constant; happy to file separately if you'd prefer.

Found in real code ([xLights](https://github.com/xLightsSequencer/xLights)) where a `unsigned int8` counter `wind--` in a per-pixel kernel silently never decremented, diverging the ISPC path from the scalar/GPU paths. Workaround is to use `v -= 1`.

azwolski:
…decrement miscompile

lLLVMConstantValue() cast a negative double directly to an unsigned integer type (e.g. (unsigned int)(-1.0)) when emitting the step constant for varying unsigned decrement. This conversion is undefined behavior in C++ for out-of-range values; on aarch64 hosts it saturates to 0, silently turning `--` into a no-op, and for UINT64 it tripped an Assert. Route the conversion through a signed intermediate type first, which is well-defined for -1, then narrow to the unsigned type.

## Description
Brief description of changes

## Related Issue

## Checklist

### DIFF
--- a/src/expr.cpp
+++ b/src/expr.cpp
@@ -1071,8 +1071,9 @@ static llvm::Constant *lLLVMConstantValue(const Type *type, llvm::LLVMContext *c
             return isUniform ? LLVMInt8(i) : LLVMInt8Vector(i);
         }
         case AtomicType::TYPE_UINT8: {
-            unsigned int i = (unsigned int)value;
-            return isUniform ? LLVMUInt8(i) : LLVMUInt8Vector(i);
+            int i = (int)value;
+            Assert((double)i == value);
+            return isUniform ? LLVMUInt8((uint8_t)i) : LLVMUInt8Vector((uint8_t)i);
         }
         case AtomicType::TYPE_INT16: {
             int i = (int)value;
@@ -1080,8 +1081,9 @@ static llvm::Constant *lLLVMConstantValue(const Type *type, llvm::LLVMContext *c
             return isUniform ? LLVMInt16(i) : LLVMInt16Vector(i);
         }
         case AtomicType::TYPE_UINT16: {
-            unsigned int i = (unsigned int)value;
-            return isUniform ? LLVMUInt16(i) : LLVMUInt16Vector(i);
+            int i = (int)value;
+            Assert((double)i == value);
+            return isUniform ? LLVMUInt16((uint16_t)i) : LLVMUInt16Vector((uint16_t)i);
         }
         case AtomicType::TYPE_INT32: {
             int i = (int)value;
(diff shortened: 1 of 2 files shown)


dbabokin:
@codex review

aneshlya (approved):
Looks good to me! Thanks for the fix!

PR #3890, opened 2026-08-03
Add support for Zen6 AMD CPU

### LINKED ISSUE #3889: Add AMD Zen6 CPU definition
Support znver6 option for --cpu= flag.

nurmukhametov:
## Description
LLVM gained znver6 in 23, so add it as a --cpu=znver6 value guarded by ISPC_LLVM_23_0.

Zen6 is Zen5 plus AVX512-FP16, AVX-VNNI-INT8, AVX-IFMA, AVX-NE-CONVERT and AVX512-BMM. It maps to avx512icl-x16.

No new ISPC target is introduced, but passing the CPU through can be useful: float16 arithmetic on avx512icl-x16 now lowers to native instructions instead of the round-trip through fp32.

## Related Issue
- [X] Linked to relevant issue: #3889 

## Checklist

### DIFF
--- a/src/ispc.cpp
+++ b/src/ispc.cpp
@@ -375,12 +375,15 @@ typedef enum {
 #if ISPC_LLVM_VERSION >= ISPC_LLVM_22_0
     CPU_NVL,
 #endif
-    // Zen1 to Zen5
+    // Zen1 to Zen6
     CPU_ZNVER1,
     CPU_ZNVER2,
     CPU_ZNVER3,
     CPU_ZNVER4,
     CPU_ZNVER5,
+#if ISPC_LLVM_VERSION >= ISPC_LLVM_23_0
+    CPU_ZNVER6,
+#endif
 
 // FIXME: LLVM supports a ton of different ARM CPU variants--not just
 // listed below.  We should be able to handle any of them that also
@@ -503,6 +506,9 @@ std::map<DeviceType, std::set<std::string>> CPUFeatures = {
     {CPU_ZNVER3, {"mmx", "sse", "sse2", "ssse3", "sse41", "sse42", "avx", "avx2"}},
     {CPU_ZNVER4, {"mmx", "sse", "sse2", "ssse3", "sse41", "sse42", "avx", "avx2", "avx512", "avx512_vnni"}},
     {CPU_ZNVER5, {"mmx", "sse", "sse2", "ssse3", "sse41", "sse42", "avx", "avx2", "avx512", "avx_vnni", "avx512_vnni"}},
+#if ISPC_LLVM_VERSION >= ISPC_LLVM_23_0
+    {CPU_ZNVER6, {"mmx", "sse", "sse2", "ssse3", "sse41", "sse42", "avx", "avx2", "avx512", "avx_vnni", "avx512_vnni"}},
+#endif
 // TODO: Add features for remaining CPUs if valid.
 #ifdef ISPC_ARM_ENABLED
     {CPU_CortexA35, {}},
@@ -632,6 +638,9 @@ class AllCPUs {
         names[CPU_ZNVER3].push_back("znver3");
         names[CPU_ZNVER4].push_back("znver4");
         names[CPU_ZNVER5].push_back("znver5");
+#if ISPC_LLVM_VERSION >= ISPC_LLVM_23_0
+        names[CPU_ZNVER6].push_back("znver6");
+#endif
 
 #ifdef ISPC_ARM_ENABLED
         names[CPU_CortexA35].push_back("cortex-a35");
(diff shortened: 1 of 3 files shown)


aneshlya (approved):
Thanks! Looks good to me!

dbabokin:
@codex review

zephyr111:
Good!

> Zen6 is Zen5 plus AVX512-FP16, AVX-VNNI-INT8, AVX-IFMA, AVX-NE-CONVERT and AVX512-BMM. 

Considering https://en.wikipedia.org/wiki/AVX-512 , Zen5 should already support AVX-IFMA (and it is not clear for AVX-VNNI-INT8, but AFAIR it is not fully supported indeed). Other sources (starting from https://en.wikipedia.org/wiki/Zen_6) tend to agree with you. Do you know if this is an error in this page?

The GCC code appears to also agree with your statement: https://github.com/gcc-mirror/gcc/blob/df494ca28d75f80b4d0b4e0b93b5d23d2092131a/gcc/config/i386/i386.h#L2542 . It looks like the same for LLVM: https://github.com/llvm/llvm-project/commit/a362593e0d02ff5677620b10ad5adeff0a8196a5 .

The same thing applies for IceLake: did it support IFMA already? If so, Zen5 should not have been marked compatible with ICL/ICX. On one hand, AFAIK, Ice Lake is built on the [Sunny Cove](https://en.wikipedia.org/wiki/Sunny_Cove_(microarchitecture)) microarchitecture and has the exact same supported instruction regarding AVX-512. If so, then, IFMA was not supported yet on IceLake, assuming https://en.wikipedia.org/wiki/Sunny_Cove_(microarchitecture) is correct. On the other hand, the [Intel intrinsics-guide](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=ifma&ig_expand=4209) reports IFMA instructions are supported on both ICL and ICX... Thus, I am confused. Can you shed some light on the matter?

I think we should be very careful about this since it could end up with sneaky crash if instructions are lowered to unsupported ones on the target platform.

PS: the same thing applies to BF16 instructions (AFAIK they are supported by both ICL/ICX and Zen5/Zen6)

src/ispc.cpp
```diff
@@ -722,6 +731,12 @@ class AllCPUs {
         compat[CPU_ICL] =
             Set(CPU_ICL, CPU_x86_64, CPU_Bonnell, CPU_Penryn, CPU_Core2, CPU_Nehalem, CPU_Silvermont, CPU_SandyBridge,
                 CPU_IvyBridge, CPU_Haswell, CPU_Broadwell, CPU_Skylake, CPU_SKX, CPU_None);
+#if ISPC_LLVM_VERSION >= ISPC_LLVM_23_0
+        compat[CPU_ZNVER6] =
```
zephyr111, line 735: Minor remark: IMHO, it is better for readability to move this block after ZNVER5 (especially since CPU_ZNVER5 is mentioned in the end of this block).

zephyr111 (approved):
Looks good to me besides the above comment regarding IFMA/VNNI instruction support/compatibility (and the minor one in the review).

src/ispc.cpp
```diff
@@ -503,6 +506,9 @@ std::map<DeviceType, std::set<std::string>> CPUFeatures = {
     {CPU_ZNVER3, {"mmx", "sse", "sse2", "ssse3", "sse41", "sse42", "avx", "avx2"}},
     {CPU_ZNVER4, {"mmx", "sse", "sse2", "ssse3", "sse41", "sse42", "avx", "avx2", "avx512", "avx512_vnni"}},
     {CPU_ZNVER5, {"mmx", "sse", "sse2", "ssse3", "sse41", "sse42", "avx", "avx2", "avx512", "avx_vnni", "avx512_vnni"}},
+#if ISPC_LLVM_VERSION >= ISPC_LLVM_23_0
+    {CPU_ZNVER6, {"mmx", "sse", "sse2", "ssse3", "sse41", "sse42", "avx", "avx2", "avx512", "avx_vnni", "avx512_vnni"}},
```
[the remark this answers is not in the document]
dbabokin: tests/lit-tests/znver6-fp16.ispc tests exactly this and seems to work. But the reasoning here sounds right as well. I'm wondering why.

Otherwise the changes look good to me.

src/ispc.cpp
```diff
@@ -722,6 +731,12 @@ class AllCPUs {
         compat[CPU_ICL] =
             Set(CPU_ICL, CPU_x86_64, CPU_Bonnell, CPU_Penryn, CPU_Core2, CPU_Nehalem, CPU_Silvermont, CPU_SandyBridge,
                 CPU_IvyBridge, CPU_Haswell, CPU_Broadwell, CPU_Skylake, CPU_SKX, CPU_None);
+#if ISPC_LLVM_VERSION >= ISPC_LLVM_23_0
+        compat[CPU_ZNVER6] =
+            Set(CPU_x86_64, CPU_Bonnell, CPU_Penryn, CPU_Core2, CPU_Nehalem, CPU_Silvermont, CPU_SandyBridge,
```
azwolski: `CPU_ADL` seems to be missing from this compatibility set. The `avx2vnni-*` targets use CPU_ADL as CPUfromISA, and BackwardCompatible() only performs a direct lookup in this set (there is no transitive compatibility). Since znver6 inherits AVX-VNNI support from znver5, --cpu=znver6 --target=avx2vnni-i32x8 should be valid but is currently rejected. CPU_ZNVER5 appears to have the same pre-existing omission.

Issue #3893, opened 2026-09-14
Crash passing reference to template arg


larsbohn:
Hi!

This crashes ispc, also reproducible on latest:

```
int<3> convert(const bool<3> v) {
    int<3> result;

    result.x = v.x ? 1 : 0;
    result.y = v.y ? 1 : 0;
    result.z = v.z ? 1 : 0;

    return result;
}

void f(const bool<3>& v)
{
    int<3> i = convert(v);
}
```

Error message:

```
Call parameter type does not match function signature!
  %15 = load [3 x <8 x i8>], ptr %12, align 8, !dbg !137
 [3 x <8 x i32>]  %18 = call [3 x <8 x i32>] @convert___Cvyb_3C_3_3E_([3 x <8 x i8>] %15, <8 x i32> %17), !dbg !138
```

[Godbolt](https://godbolt.org/z/MEzqKnP6q)

Seems to be related to passing references onto templates.

PR #3896, opened 2026-09-17
Fix bool short-vector reference conversion

azwolski:
## Summary
- convert bool short-vector reference loads from storage bytes to LLVM boolean vectors
- add a regression test for passing `const bool<3>&` to a conversion function

## Root cause
The uniform load path converted scalar bools from stored `i8` values to internal `i1` values, but did not apply the conversion when the loaded type was a short vector with bool elements. This produced a call argument of `[3 x <W x i8>]` for a parameter expecting `[3 x <W x i1>]`.

## Related Issue
- [X] Linked to relevant issue: #3893 

## Checklist


Lycan7hropus:
@codex review

